Issue #3, opened 2018-04-19
voicechat bug


Alf21:
If a player forces another role, the voice chat vars won't reset.

Issue #25, opened 2018-10-03
Traitorvoicechat bug


V3kta:
The traitor voicechat can be heard by everyone sometimes, everyone who knows about it can reveal him as a traitor.

Alf21:
I'm aware of this. The bugfix will be included into the next update. Thanks 4 the report !:)

Issue #19, opened 2018-09-23
TeamChat BUG


Alf21:
Currently, the TeamChat doesn‘t support the custom roles.

Issue #12, opened 2018-07-29
Role‘s index


Alf21:
Change role‘s initialization of data from dynamic (with syncing) to static (init shared on start). There, use a GetIndex function to get the index of a role (like equipment addons do).
Last but not least, remove ROLES.team and change it to subroles. Traitor, Hitman and Vampire will have ROLE = 1 while T has SUBROLE 0, H SUBROLE 1,... This will support the compatibility with other addons to check eg if someone is a Traitor.

PR #26, opened 2018-10-04
Complete revision of TTT2

Alf21:


Alf21:
this should fix #3 #4 #12 #19 and #25

Issue #58, opened 2018-11-11
Damagelogs additional roles.


dok441:
Can you make "Damagelog.ULX_Autoslay_ForceRole" function in Damagelogs to work properly? It doesn't work properly with new roles. It is very sad to see when some active role dies because of RDM. Currently if this function is enabled, all new roles will disappear or will not work properly.

iViscosity:
I've created a solution to this.

As for 

> Currently if this function is enabled, all new roles will disappear or will not work properly.

This is due to how Tommy implements the ForceRole feature. Basically, he overwrites the SelectRoles() function to prevent people from being selected as Special roles (Traitor/Detective) if they have autoslays. This is done in an Initialize hook, meaning after the Gamemode (TTT2) is done loading. In other words, when SelectRoles() is called, it's calling [Tommy's version](https://github.com/Tommy228/tttdamagelogs/blob/289535fe540b1aaa7fe346b012d3e201f55a2775/lua/damagelogs/server/autoslay.lua#L441) which, obviously, doesn't include custom roles.

My solution requires you removing from [this line](https://github.com/Tommy228/tttdamagelogs/blob/289535fe540b1aaa7fe346b012d3e201f55a2775/lua/damagelogs/server/autoslay.lua#L420) down.

dok441:
![22](https://user-images.githubusercontent.com/7480655/48966774-78f4bb80-eff9-11e8-92d3-2361b2b9ccbf.png)
Не работает.

dok441:
Так же это отключает возрождения шута и неизвестного. Если их убьют.

SchokoShampoo:
Will be fixed with #116

PR #112, opened 2018-11-24
Solution to #58

iViscosity:
Checks both if Damagelog exists, and if it does it will then check for the ForceRole config boolean.

If it's true, will ignore people with Autoslays.

This is not an "end-all be-all" solution, as it requires small editing of Damagelogs (mainly removing from [this line](https://github.com/Tommy228/tttdamagelogs/blob/289535fe540b1aaa7fe346b012d3e201f55a2775/lua/damagelogs/server/autoslay.lua#L420) down).

### DIFF
--- a/gamemodes/terrortown/gamemode/init.lua
+++ b/gamemodes/terrortown/gamemode/init.lua
@@ -1164,7 +1164,7 @@ function SelectRoles(plys, max_plys)
 	for _, v in ipairs(player.GetAll()) do
 
 		-- everyone on the spec team is in specmode
-		if IsValid(v) and not v:GetForceSpec() and (not plys or table.HasValue(plys, v)) then
+		if IsValid(v) and not v:GetForceSpec() and (not plys or table.HasValue(plys, v)) and not (v.AutoslaysLeft and tonumber(v.AutoslaysLeft) > 0) then
 			if not plys then
 				tmp[#tmp + 1] = v
 			end
--- a/gamemodes/terrortown/gamemode/init.lua
+++ b/gamemodes/terrortown/gamemode/init.lua
@@ -1164,7 +1164,7 @@ function SelectRoles(plys, max_plys)
 	for _, v in ipairs(player.GetAll()) do
 
 		-- everyone on the spec team is in specmode
-		if IsValid(v) and not v:GetForceSpec() and (not plys or table.HasValue(plys, v)) and not (v.AutoslaysLeft and tonumber(v.AutoslaysLeft) > 0) then
+		if IsValid(v) and not v:GetForceSpec() and (not plys or table.HasValue(plys, v)) and (Damagelog and Damagelog.ULX_Autoslay_ForceRole and not (v.AutoslaysLeft and tonumber(v.AutoslaysLeft) > 0)) then
 			if not plys then
 				tmp[#tmp + 1] = v
 			end


Alf21:
Maybe it would be a better solution to add it to damagelogs directly (I did a compatible fork already)? Some commits ago, i‘ve changed the selection system to easily add roles for players that should be forced or marked (if possible). What about using this system instead?

Btw. if we doing it in this way, there should also be a check "or not Damagelogs" to support roleselection for players that don‘t have it :)

Thanks for this suggestion / direct commit, keep going on! :)

iViscosity:
I did add support for whether or not they have damagelogs.

`(Damagelog and Damagelog.ULX_Autoslay_ForceRole --...`

It **would** be easier to directly push the solution to Damagelogs. However, (forgive me, just found this addon recently), is there some global table or way to detect that TTT2 is running?

Or, perhaps, along the lines of what you said, use the [ULibCommandCalled](https://ulyssesmod.net/docs/files/lua/ulib/shared/defines-lua.html#ULibCommandCalled) hook to capture when "ulx aslay" is called, and mark a player accordingly? Then, the SelectRoles() here can simply check if the player is marked.

Alf21:
I like your solution with ulib. But up to now, i never worked with this code directly. Modifying damagelog directly is the way we should go. Maybe, this would be the best solution, because damagelog needs to get modified, otherwise, it would break TTT2 or vice versa. There is still an artifact "ROLES" in that code ("ROLES" have been the old check for TTT2). Therefor, i already created a PR weeks ago. Feel free to modify the TTT2 damagelog fork, if you want too :)

iViscosity:
> I like your solution with ulib. But up to now, i never worked with this code directly, because you need to add Ulib as dependency). Modifying damagelog directly is the way we should go. Maybe, this would be the best solution, because damagelog needs to get modified, otherwise, it would break TTT2 or vice versa. There is still an artifact "ROLES" in that code ("ROLES" have been the old check for TTT2). Therefor, i already created a PR weeks ago. Feel free to modify the TTT2 damagelog fork, if you want too :)
> 
> (not Damagelog or Damagelog.ULX_Autoslay_ForceRole --...)
> 
> Btw gj, I am amazed how quickly you get involved

I don't think you'd have to add ULib as a dependency. Because all I'm adding is a hook, I could just do this:

```lua
hook.Add("ULibCommandCalled", "Mark Slayed Player", function(ply, cmd, args)
    if cmd == "ulx aslay" then
        if #args >= 2 then -- Need at least a player and number
            args[1].marked = tonumber(args[2]) -- Marks them for this many turns.
            -- SelectRoles() can then check if a player is marked and not select them.
            -- It could also be responsible for decrementing "marked". Or, it could be done in a "TTTBeginRound" hook.
        end
    end
end)
```

Alf21:
Ah I understand, just trying to add an ulib hook. That could be the best solution. So we just need to mark them for x amounts and decrease this var in SelectRoles directly.  :)

iViscosity:
Yes, just add the hook.

I was thinking though, this would **still** require editing of Damagelogs (due to how Damagelogs implements ForceRole by overwriting SelectRoles).

SchokoShampoo:
Will be solved by this #116

SchokoShampoo:
I saw your PR after i fixed it ^^

PR #154, opened 2018-12-05
Update player_main.lua

TheNickSkater:
Fix for "ttt_credits_award_pct"


Alf21:
What does "Fix" mean? You are changing default functions back to old deprecated ones: https://wiki.garrysmod.com/page/Global/GetConVarNumber

PR #242, opened 2019-01-30
Add possibility to change the role color at TargetID + Corpse Icon again

lebroomer:
See https://github.com/Alf21/ttt2-role_siki/pull/3


Alf21:
Instead of adding new hooks, i suggest to just modify the var that is returned by ply:GetColor()

gamemodes/terrortown/gamemode/cl_targetid_main.lua
```diff
@@ -363,6 +363,7 @@ function GM:HUDDrawTargetID()
 
 		text = L["target_" .. rd.name]
 		c = IsValid(ent) and ent:GetRoleColor() or rd.color
+		c = hook.Run("TTT2ModifyTargetIDColor", ent) or c
```
saibotk: no new line needed, just append this with an or

Alf21:
Could you use a hook that fits the sense? eg instead of "TTT2ModifyBodyFoundRoleIcon" -> "TTT2ModifyBodyFoundRoleColor"

gamemodes/terrortown/gamemode/server/sv_corpse.lua
```diff
@@ -211,6 +211,7 @@ function CORPSE.ShowSearch(ply, rag, covert, long_range)
 	-- init a heap of data we'll be sending
 	local nick = CORPSE.GetPlayerNick(rag)
 	local subrole = rag.was_role
+	local role_color = rag.role_color or GetRoleByIndex(subrole).color
```
Alf21: The second color (after "or") isn‘t needed :)

Alf21:
thanks!

PR #246, opened 2019-01-30
Fixed the auto converter

lebroomer:
Error was caused by a falsely assumed filename in the fallback shop init


lebroomer:
So first its "why" not "warum" :P and yeah you are right, I forgot about this possibility.

PR #252, opened 2019-02-05
Rework kind + slot system

lebroomer:
First working solution!

However I would like to expand the rework to the slot system, in detail to use the weapon kinds/inventory places to generate slot places, e.g., Slot 1 for crowbar, Slot 2 for Primary, Slot 3 for Secondary and so on.


lebroomer:
The new slot system works nice and intuitive together with the inventory system.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -287,22 +287,75 @@ function plymeta:CanCarryWeapon(wep)
 	if not wep or not wep.Kind then
 		return false
 	end
-
+	
+	--appeareantly TTT can't handle two times the same weapon
+	for k, v in pairs(self:GetWeapons()) do
+		if IsValid(wep) and IsValid(v) and v:GetClass() == wep:GetClass() then
+			return false
+		elseif istable(wep) and IsValid(v) and v:GetClass() == wep.ClassName then
```
Alf21: This elseif should never get called...

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -287,22 +287,75 @@ function plymeta:CanCarryWeapon(wep)
...
 	end
+	
+	return InventorySlotFree(self, t)
+end
 
-	for _, w in pairs(self:GetWeapons()) do
-		if w.Kind and w.Kind == t then
-			return false
+function plymeta:GetInventory()
+	CleanupInventoryIfDirty(self)
```
Alf21: I suggest to clean direclty after modifying the data, not on trying to access it. This could cause issues.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -287,22 +287,75 @@ function plymeta:CanCarryWeapon(wep)
...
+end
+
+function plymeta:GetWeaponsOnSlot(slot)
+	if slot > WEAPON_EXTRA then
+		return
+	end
+	
+	CleanupInventoryIfDirty(self)
+	
+	if istable(self.inventory) then
```
Alf21: This sould always be a tablt. Just use the faster check: if self.inventory then... end

gamemodes/terrortown/gamemode/client/cl_wepswitch.lua
```diff
@@ -158,29 +174,28 @@ function WSWITCH:Draw(client)
...
-	--	self.WeaponCache = LocalPlayer():GetWeapons()
-	-- So copy over the weapon refs
 	self.WeaponCache = {}
-
-	CopyVals(LocalPlayer():GetWeapons(), self.WeaponCache)
-
-	table.sort(self.WeaponCache, SlotSort)
+	
+	local inventory = LocalPlayer():GetInventory()
+	InsertIfValid(self.WeaponCache, inventory[WEAPON_MELEE], SlotnumberFromKind(WEAPON_MELEE))
```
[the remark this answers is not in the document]
lebroomer: Mmh I don't know. I am setting the order of the weapons in WSWITCH, which should be always the same no matter how their variables are defined (e.g. WEAPON_MELEE = 9 for some reason)

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
lebroomer: Actually it is called when a player is getting a new weapon (e.g. via Shop/Console)

gamemodes/terrortown/gamemode/shared/sh_inventory.lua
```diff
@@ -0,0 +1,82 @@
...
+		return WEAPON_EXTRA
+	else
+		return kind
+	end	
+end
+
+--we appeareantly need to have that, because added and removed entities are sometimes broken on client side and need to get updated later
+--additionally spawn can be called when a player is alive in which case he doesn't loose weapons beforehand
+function CleanupInventoryIfDirty(ply)
+	if !ply.inventory || ply.refresh_inventory_cache then
```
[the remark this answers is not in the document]
lebroomer: The idea behind that is a kind of caching functionality, which only updates the inventory if needed and dirty (e.g. GetInventory is called). Updating weapons work great on Server side, but doesn't on Client, because Entities are updated late/are invalid for some reason.

A solution to beautify the code would be to always use caching (not only when something is invalid). However, this would probably be less runtime efficient.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
lebroomer: The idea behind that is a kind of caching functionality, which only updates the inventory if needed and dirty (e.g. GetInventory is called). Updating weapons work great on Server side, but doesn't on Client, because Entities are updated late/are invalid for some reason.

A solution to beautify the code would be to always use caching (not only when something is invalid). However, this would probably be less runtime efficient.

(the same comment but now in the right place ^^)

lebroomer: So I noticed, that when a player buys a weapon from Shop the WeaponEquip hook is called on SERVER. There the entity is valid. If i send a message from here to the Client with the same entity, it is a NULL Entity. 

Yeah of cause that would be perfect, but leads to weird behaviour on client as described above. But the cache is only updated once (if weapons were changed), because after that the cache isn't dirty anymore. This "Dirty-Flag" is an often used technique.

Alf21:
Ouh, really nice changes! Let's discuss the idea in the Discord :)

Alf21:
Nice, thanks! :)

Alf21:
Good job Dude!!

PR #253, opened 2019-02-05
Add alive check for corpse so player only get revealed when they are dead

lebroomer:
- e.g. through Fake Body/Dead Ringer etc.

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_corpse.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_corpse.lua
@@ -87,12 +87,10 @@ local function IdentifyBody(ply, rag)
 	-- Register find
 	if not CORPSE.GetFound(rag, false) then -- will return either false or a valid ply
 		local deadply = player.GetBySteamID64(rag.sid64)
-		if deadply then
+		if deadply and not deadply:Alive() then
 			deadply:SetNWBool("body_found", true)
 
-			if not deadply:Alive() then
-				SendPlayerToEveryone(deadply) -- confirm player for everyone
-			end
+			SendPlayerToEveryone(deadply) -- confirm player for everyone
 
 			SCORE:HandleBodyFound(ply, deadply)
 		end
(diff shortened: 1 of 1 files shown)


gamemodes/terrortown/gamemode/server/sv_corpse.lua
```diff
@@ -90,7 +90,9 @@ local function IdentifyBody(ply, rag)
 		if deadply then
```
Alf21: Why not checking directly in this if-check?

Alf21:
So can you confirm a fake body in the scoreboard? Or what will happen if trying to confirm it?

lebroomer:
What do you mean with confirm in scoreboard? The problem was that HUDTargetID shows the role of dead players, because their true role was sent to the client before. Now this won't happen with fake corpses. Confirming the bodies can still make them appear on "Confirmed Dead" if the "body_found" Network Variable is set in the addon.

Alf21:
Hmm that‘s right!

lebroomer:
Good point, then the player gets revealed, but I think that should be fine, because first this have to happen exactly in the time, when the player i dead and then if this happens, its not so unintuitive.

lebroomer:
The only way to fix this problem would be to either modify TTT2 in a way that its compatible with the Fake Body (very bad style) or I could rewrite the complete IdentifyBody method in the addon (which will be a problem, if the original method is changed for example). So please let me know, if you have a different idea :)

TheNickSkater:
> The only way to fix this problem would be to either modify TTT2 in a way that its compatible with the Fake Body (very bad style) or I could rewrite the complete IdentifyBody method in the addon (which will be a problem, if the original method is changed for example). So please let me know, if you have a different idea :)

Well, like I said. I fixed it myself. But I also had to rewerite the the Death Faker and Death Ringer. And also edited some of the identify function.

Only problem which I had (may not occur with this here):
When somebody was confirmed, and got shot as sidekick as example, he was revealed as sidekick.

PR #261, opened 2019-02-17
Fixing the WSwitch Error

lebroomer:
Forgot to commit this change in my PR before...


lebroomer:
See https://github.com/Alf21/TTT2/issues/260

PR #273, opened 2019-03-02
Settings Binding Categories

lebroomer:
A liitle QOL change which adds binding categories. Its still possible to not set a category - these bindings will be added to Other Bindings.


Alf21:
nice!

PR #326, opened 2019-05-06
Improved Random Shop

lebroomer:
* added "reroll after every buy feature"
* improved "Shopeditor Option" design
* implemented Team rerolling (very good together with reroll after buy :D)


Alf21:
Ouh very nice options :3 I really wanna test it :D

Issue #331, opened 2019-05-14
attempt to perform arithmetic on field 'TipIndex' (a nil value)


Perwex:
**Describe the bug**
Some players (every 5/6th player) receive an error as soon as they join the server.
They can't see hud and they can't use f1(settings), scoreboard.
In addition, they see a white window on the right.

**To Reproduce**
Steps to reproduce the behaviour:
1. Connect to the server.
2. Just it.

**Expected behaviour**
1. HUD should work - it doesn't
2. F1 should open settings - it opens empty 
3. Tab should open scoreboard - it doesn't

**Screenshots**
Below. (cl_error)

**Desktop (please complete the following information):**
 - Collection-Link of your Server: Below (workshop.txt)
- Server and Client Errors are the same
`[ERROR] addons/trouble_in_terrorist_town2(advancedupdate)-_v0.5.2b_1357204556/gamemodes/terrortown/gamemode/client/cl_lang.lua:75: bad argument #1 to 'rawget' (table expected, got nil)
  1. rawget - [C]:-1
   2. GetRaw - addons/trouble_in_terrorist_town2(advancedupdate)-_v0.5.2b_1357204556/gamemodes/terrortown/gamemode/client/cl_lang.lua:75
    3. SafeTranslate - addons/trouble_in_terrorist_town2(advancedupdate)-_v0.5.2b_1357204556/gamemodes/terrortown/gamemode/client/cl_lang.lua:82
     4. GetEquipmentTranslation - addons/trouble_in_terrorist_town2(advancedupdate)-_v0.5.2b_1357204556/gamemodes/terrortown/gamemode/shared/sh_init.lua:244
      5. unknown - addons/trouble_in_terrorist_town2(advancedupdate)-_v0.5.2b_1357204556/gamemodes/terrortown/gamemode/client/cl_hudpickup.lua:20
[ERROR] addons/trouble_in_terrorist_town2(advancedupdate)-_v0.5.2b_1357204556/gamemodes/terrortown/gamemode/client/cl_tips.lua:213: bad argument #1 to 'pairs' (table expected, got nil)
  1. pairs - [C]:-1
   2. unknown - addons/trouble_in_terrorist_town2(advancedupdate)-_v0.5.2b_1357204556/gamemodes/terrortown/gamemode/client/cl_tips.lua:213
[...]

PR #332, opened 2019-05-17
fixed hud bug (#331)

Dzwiedziasty:
this fixes a bug which caused the hud to not render for some people when damagelogs were installed


Alf21:
#331

PR #346, opened 2019-06-30
introduced a simple workaround that allows HUD creators to set a default design

TimGoll:
This does not break anything that is currentyl working but fixed the problem that currently exists. Right now you can disabledUnlessForced HUDs, but since it always gets the first one, nil is returned when it is disabledUnlessForced.

I know @saibotk hates this but I did not want to throw it away and hear the opinion of the other devs. :D

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/huds/base_huds/hud_base/cl_init.lua
+++ b/gamemodes/terrortown/gamemode/shared/huds/base_huds/hud_base/cl_init.lua
@@ -130,7 +130,8 @@ function HUD:GetElementByType(elementType)
 	if table.HasValue(self.disabledTypes, elementType) then return nil end
 
 	local element = self.forcedElements[elementType]
-	local elementTbl = not element and hudelements.GetTypeElement(elementType) or hudelements.GetStored(element)
+
+	local elementTbl = not element and self:GetEnabledTypeElement(elementType) or hudelements.GetStored(element)
 
 	-- return nil if the elements table is not found
 	if not elementTbl then return nil end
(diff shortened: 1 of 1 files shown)


saibotk:
I still hate this workaround, since having this "fixed" will eliminate the urgency of the issue thus might lead to forgetting about the other issues, which then may be forgotten. 
And just because it does not break more than it is now, does not mean it is acceptable. 
The only option is to just get this whole issue fixed instead of just what you need at the moment.

TimGoll:
But someone has to explain the whole issue since I'm not aware of any other problems than fixed in this commit

Alf21:
Hmm maybe we should think about fixing it finally. I don‘t like buggy or inconsistent code too. If there is a better way, we should take it. TTT2 stands for quality, not quantity

TimGoll:
I think both of you are right. But there is this bug right now and this fixes it. I have no clue about the other problem and how to fix it. Therefore I have to rely on you. An since Tobi has no time in the next few months I propose this change as a work around.

TimGoll:
Without this change hud elements are missing if the player chooses a theme that has this element not implememted. And therefore it is not playable.

saibotk:
Since this is no high security issue, where a quick fix or workaround is needed, it should be done as a whole and correctly, instead of introducing workarounds. I can/will do this when i get some spare time in the next few days, so that it will fix all remaining issues.

TimGoll:
if saibotk implements it in this week or so it would be nice. Otherwise I would have to implement a workaround inside the octagonal HUD and I would not be able to put the marker-element-skin inside the octagonal addon.

PR #361, opened 2019-07-23
Changed TellTraitorsAboutTraitors function

TheNickSkater:
You can now decide (with a hook) if the Message should even be shown for certain players and what players should be shown.

First of all this brings more possibilities into the game (for me for my BodyGuard :) ) and I think it's better if the hook is changeable PER player, and not only for all at once.

Please look over it and tell me what needs to be changed for you to accept it.

Oh, also the Spy needs to be updated if you accept this (I may do it if needed)

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -844,11 +844,10 @@ function TellTraitorsAboutTraitors()
 	for _, v in ipairs(player.GetAll()) do
 		if v:HasTeam(TEAM_TRAITOR) then
 
-			local tmp, shouldShow = table.Copy(traitornicks), true
+			local tmp, shouldShow = table.Copy(traitornicks)
 
-			tmp, shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
+			shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
 
-			if shouldShow == nil then shouldShow = true end
 			if shouldShow == false then continue end
 
 			if tmp == nil then tmp = table.Copy(traitornicks) end
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -848,16 +848,14 @@ function TellTraitorsAboutTraitors()
 
 			local shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
 
-			if shouldShow == false or tmp == nil then continue end
+			if shouldShow == false or tmp == nil or #tmp == 0 then continue end
 
-			table.RemoveByValue(tmp, v:Nick())
-
-			if #tmp == 0 then
+			if #tmp == 1 then
 				LANG.Msg(v, "round_traitors_one")
 				return
 			end
 
-			if #tmp >= 2 then
+			if #tmp >= 3 then
 				table.sort(tmp, function(a, b)
 					return a and b and a:upper() < b:upper()
 				end)
@@ -866,7 +864,9 @@ function TellTraitorsAboutTraitors()
 			local names = ""
 
 			for _, name in ipairs(tmp) do
-				names = names .. name .. ", "
+				if name ~= v:Nick() then
+					names = names .. name .. ", "
+				end
 			end
 
 			names = string.sub(names, 1, -3)
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -846,9 +846,7 @@ function TellTraitorsAboutTraitors()
 
 			local tmp = table.Copy(traitornicks)
 
-			local shouldShow
-
-			shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
+			local shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
 
 			if shouldShow == false or tmp == nil then continue end
 
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -844,13 +844,13 @@ function TellTraitorsAboutTraitors()
 	for _, v in ipairs(player.GetAll()) do
 		if v:HasTeam(TEAM_TRAITOR) then
 
-			local tmp, shouldShow = table.Copy(traitornicks)
+			local tmp = table.Copy(traitornicks)
+
+			local shouldShow
 
 			shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
 
-			if shouldShow == false then continue end
-
-			if tmp == nil then tmp = table.Copy(traitornicks) end
+			if shouldShow == false or tmp == nil then continue end
 
 			table.RemoveByValue(tmp, v:Nick())
 
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -840,14 +840,16 @@ function TellTraitorsAboutTraitors()
 			traitornicks[#traitornicks + 1] = v:Nick()
 		end
 	end
-	
+
 	for _, v in ipairs(player.GetAll()) do
 		if v:HasTeam(TEAM_TRAITOR) then
 
-			local tmp, shouldShow = hook.Run("TTT2TellTraitors", v, table.Copy(traitornicks))
+			local tmp, shouldShow = table.Copy(traitornicks), true
+
+			tmp, shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
 
 			if shouldShow == nil then shouldShow = true end
-			if not shouldShow then continue end
+			if shouldShow == false then continue end
 
 			if tmp == nil then tmp = table.Copy(traitornicks) end
 
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -840,38 +840,40 @@ function TellTraitorsAboutTraitors()
 			traitornicks[#traitornicks + 1] = v:Nick()
 		end
 	end
-
-	hook.Run("TTT2TellTraitors", traitornicks)
-
-	-- This is ugly as hell, but it's kinda nice to filter out the names of the
-	-- traitors themselves in the messages to them
-	local traitornicks_min = #traitornicks < 2
-
-	if not traitornicks_min then
-		table.sort(traitornicks, function(a, b)
-			return a and b and a:upper() < b:upper()
-		end)
-	end
-
+	
 	for _, v in ipairs(player.GetAll()) do
 		if v:HasTeam(TEAM_TRAITOR) then
-			if traitornicks_min then
+
+			local tmp, shouldShow = hook.Run("TTT2TellTraitors", v, table.Copy(traitornicks))
+
+			if shouldShow == nil then shouldShow = true end
+			if not shouldShow then continue end
+
+			if tmp == nil then tmp = table.Copy(traitornicks) end
+
+			table.RemoveByValue(tmp, v:Nick())
+
+			if #tmp == 0 then
 				LANG.Msg(v, "round_traitors_one")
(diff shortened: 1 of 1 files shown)


TimGoll:
I talked with nick about this and while I can't say if his implementation is good and if this will break anything, I can say that the change is needed.

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -840,38 +840,40 @@ function TellTraitorsAboutTraitors()
...
+	for _, v in ipairs(player.GetAll()) do
+		if v:HasTeam(TEAM_TRAITOR) then
 
-	hook.Run("TTT2TellTraitors", traitornicks)
+			local tmp, shouldShow = hook.Run("TTT2TellTraitors", v, table.Copy(traitornicks))
 
-	-- This is ugly as hell, but it's kinda nice to filter out the names of the
-	-- traitors themselves in the messages to them
-	local traitornicks_min = #traitornicks < 2
+			if shouldShow == nil then shouldShow = true end
```
Alf21: simplify with `shouldShow = shouldShow or true`

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -840,38 +840,40 @@ function TellTraitorsAboutTraitors()
...
+			if not shouldShow then continue end
```
Alf21: or just use the check `if shouldShow == false then ...`

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -840,38 +840,40 @@ function TellTraitorsAboutTraitors()
 			traitornicks[#traitornicks + 1] = v:Nick()
 		end
 	end
+	
+	for _, v in ipairs(player.GetAll()) do
+		if v:HasTeam(TEAM_TRAITOR) then
 
-	hook.Run("TTT2TellTraitors", traitornicks)
+			local tmp, shouldShow = hook.Run("TTT2TellTraitors", v, table.Copy(traitornicks))
```
Alf21: I suggest to not breaking any other add-ons. To do so, you should change this into
```lua
local shouldShow = hook.Run("TTT2TellTraitors", traitornicks, v)
```

traitornicks should be a copy for each player that is created before calling this hook. Maybe the structure needs to change, but this will support the integrity of already existing add-ons.

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -840,38 +840,40 @@ function TellTraitorsAboutTraitors()
...
-			return a and b and a:upper() < b:upper()
-		end)
-	end
+			if tmp == nil then tmp = table.Copy(traitornicks) end
 
-	for _, v in ipairs(player.GetAll()) do
-		if v:HasTeam(TEAM_TRAITOR) then
-			if traitornicks_min then
-				LANG.Msg(v, "round_traitors_one")
+			table.RemoveByValue(tmp, v:Nick())
```
Alf21: I never used this function :D

gamemodes/terrortown/gamemode/server/sv_main.lua
TheNickSkater: So should I just move the creation of the traitornicks table into the for loop of the players?

Alf21: Just creating a new table as a copy of traitornicks

gamemodes/terrortown/gamemode/server/sv_main.lua
TheNickSkater: when I do that, it somehow always returns true.

Alf21: what do you mean? ^^

TheNickSkater: If I use `shouldShow = shouldShow or true` instead of my thing above, then It will always get back to true, even if I return false in the hook.

Alf21: Sure, I did too much C#, or just checks for "not nil or not false"... Just focus on the following comment ^^

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -841,37 +841,41 @@ function TellTraitorsAboutTraitors()
...
-			return a and b and a:upper() < b:upper()
-		end)
-	end
+			tmp, shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
 
-	for _, v in ipairs(player.GetAll()) do
-		if v:HasTeam(TEAM_TRAITOR) then
-			if traitornicks_min then
-				LANG.Msg(v, "round_traitors_one")
+			if shouldShow == nil then shouldShow = true end
```
Alf21: now this line is useless :D

gamemodes/terrortown/gamemode/server/sv_main.lua
TheNickSkater: Made changes, please look at them, is that what you meant ?
https://github.com/TTT-2/TTT2/pull/361/commits/0008e2a4c86f48eb4a1780de7946edeb6e8feb0e

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -841,37 +841,41 @@ function TellTraitorsAboutTraitors()
...
 	end
 
-	hook.Run("TTT2TellTraitors", traitornicks)
+	for _, v in ipairs(player.GetAll()) do
+		if v:HasTeam(TEAM_TRAITOR) then
 
-	-- This is ugly as hell, but it's kinda nice to filter out the names of the
-	-- traitors themselves in the messages to them
-	local traitornicks_min = #traitornicks < 2
+			local tmp, shouldShow = table.Copy(traitornicks), true
```
Alf21: you don't need to init shouldShow, the `if shouldShow == false` doesn't trigger if it's nil or true

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -841,37 +841,41 @@ function TellTraitorsAboutTraitors()
...
+			tmp, shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
 
-	for _, v in ipairs(player.GetAll()) do
-		if v:HasTeam(TEAM_TRAITOR) then
-			if traitornicks_min then
-				LANG.Msg(v, "round_traitors_one")
+			if shouldShow == nil then shouldShow = true end
+			if shouldShow == false then continue end
+
+			if tmp == nil then tmp = table.Copy(traitornicks) end
```
Alf21: we don't need a return

Alf21: tmp is modified if the hook is called

TheNickSkater: https://github.com/TTT-2/TTT2/pull/361/commits/ad2d57d01c1b959d461464993c872474c46828fd
I mean, I will get there eventually

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -844,11 +844,10 @@ function TellTraitorsAboutTraitors()
 	for _, v in ipairs(player.GetAll()) do
 		if v:HasTeam(TEAM_TRAITOR) then
 
-			local tmp, shouldShow = table.Copy(traitornicks), true
+			local tmp, shouldShow = table.Copy(traitornicks)
```
Alf21: shouldShow should be defined (locally) the line below ;D

gamemodes/terrortown/gamemode/server/sv_main.lua
Alf21: if the table is nil, this table was set to nil manually. This shouldn't be reverted, there should be a continue if you wanna check this :)

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -844,13 +844,13 @@ function TellTraitorsAboutTraitors()
 	for _, v in ipairs(player.GetAll()) do
 		if v:HasTeam(TEAM_TRAITOR) then
 
-			local tmp, shouldShow = table.Copy(traitornicks)
+			local tmp = table.Copy(traitornicks)
 
-			shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
+			local shouldShow
```
Alf21: nearly xD
Just put the lines together, don't lose track ;)
```lua
local shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
```

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -841,37 +841,40 @@ function TellTraitorsAboutTraitors()
...
-	end
+			local shouldShow
 
-	for _, v in ipairs(player.GetAll()) do
-		if v:HasTeam(TEAM_TRAITOR) then
-			if traitornicks_min then
-				LANG.Msg(v, "round_traitors_one")
+			shouldShow = hook.Run("TTT2TellTraitors", tmp, v)
+
+			if shouldShow == false or tmp == nil then continue end
```
Alf21: add `or #tmp == 0`. Wouldn't make sense to keep process if the table is empty

gamemodes/terrortown/gamemode/server/sv_main.lua
Alf21: Would be nice to handle it in another way as I did, this functions is looping through the table. This means a duplicated amount of processes

gamemodes/terrortown/gamemode/server/sv_main.lua
TheNickSkater: Well sure, if it's empty, it will show the "Traitor you stand alone" message.

TheNickSkater: Also about `table.RemoveByValue(tmp, v:Nick())`, how are you doing it? Then I can change it

Alf21: If it is empty, you aren't in the list too. So this shouldn't be displayed for you, right? Focused on the roles which manipulate roles like the spy. For sure, the shouldShow method should be used instead, but if you aren't on the list, this shouldn't be shown too...

TheNickSkater: > If it is empty, you aren't in the list too. So this shouldn't be displayed for you, right? Focused on the roles which manipulate roles like the spy. For sure, the shouldShow method should be used instead, but if you aren't on the list, this shouldn't be shown too...

Yeah I understand that. I also had a little "miss-tought". Will add the check.

TheNickSkater: https://github.com/TTT-2/TTT2/pull/361/commits/b8119c88b2a95c9a3a4603c49d4e39dc698233fd
maybe now

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -841,37 +841,38 @@ function TellTraitorsAboutTraitors()
...
-			else
-				local names = ""
+			end
 
-				for _, name in ipairs(traitornicks) do
-					if name ~= v:Nick() then
-						names = names .. name .. ", "
-					end
-				end
+			if #tmp >= 3 then
```
Alf21: last little thing: `if #tmp > 1 then` :D

TheNickSkater: why already sort it if only 2 names (where one is my own) is in?

Alf21: ah, you are right, forgot the own one that is removed later

Alf21:
Needs to be tested (and maybe the spy needs to get modified, because this hook is called multiple times now)

TheNickSkater:
> Needs to be tested (and maybe the spy needs to get modified, because this hook is called multiple times now)

Well without the spy I already tested it trough many situations and all worked.
I will also test the Spy now and if it needs a change, I will open a PR on the Spy too.

Alf21:
gj!

PR #362, opened 2019-07-24
Add StaminaMult Hook to sh_sprint.lua

TheOnly8Z:
A shared hook that modifies sprint drain (and recovery).

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/sh_sprint.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_sprint.lua
@@ -87,12 +87,13 @@ hook.Add("Think", "TTT2PlayerSprinting", function()
 		if not ply.sprintTS then continue end
 
 		local timeElapsed = CurTime() - ply.sprintTS
-		local staminaMult = hook.Run("TTT2StaminaMult", ply)
+		local drainMult = hook.Run("TTT2StaminaDrain", ply)
+		local regenMult = hook.Run("TTT2StaminaRegen", ply)
 			
 		if not ply.isSprinting then
-			ply.sprintProgress = math.min((ply.oldSprintProgress or 0) + timeElapsed * (staminaMult or 1) * GetGlobalFloat("ttt2_sprint_stamina_regeneration"), 1)
+			ply.sprintProgress = math.min((ply.oldSprintProgress or 0) + timeElapsed * (regenMult or 1) * GetGlobalFloat("ttt2_sprint_stamina_regeneration"), 1)
 		else
-			ply.sprintProgress = math.max((ply.oldSprintProgress or 0) - timeElapsed * (staminaMult or 1) * GetGlobalFloat("ttt2_sprint_stamina_consumption"), 0)
+			ply.sprintProgress = math.max((ply.oldSprintProgress or 0) - timeElapsed * (drainMult or 1) * GetGlobalFloat("ttt2_sprint_stamina_consumption"), 0)
 		end
 
 		if ply.sprintProgress == 1 then
--- a/gamemodes/terrortown/gamemode/shared/sh_sprint.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_sprint.lua
@@ -87,14 +87,12 @@ hook.Add("Think", "TTT2PlayerSprinting", function()
 		if not ply.sprintTS then continue end
 
 		local timeElapsed = CurTime() - ply.sprintTS
-		local isSprinting = (ply.sprintMultiplier ~= nil)
-
-		local drainMult, regenMult = hook.Run("TTT2StaminaMult", ply, isSprinting)
+		local staminaMult = hook.Run("TTT2StaminaMult", ply)
 			
-		if not isSprinting then
-			ply.sprintProgress = math.min((ply.oldSprintProgress or 0) + timeElapsed * (regenMult or 1) * GetGlobalFloat("ttt2_sprint_stamina_regeneration"), 1)
+		if not ply.isSprinting then
+			ply.sprintProgress = math.min((ply.oldSprintProgress or 0) + timeElapsed * (staminaMult or 1) * GetGlobalFloat("ttt2_sprint_stamina_regeneration"), 1)
 		else
-			ply.sprintProgress = math.max((ply.oldSprintProgress or 0) - timeElapsed * (drainMult or 1) * GetGlobalFloat("ttt2_sprint_stamina_consumption"), 0)
+			ply.sprintProgress = math.max((ply.oldSprintProgress or 0) - timeElapsed * (staminaMult or 1) * GetGlobalFloat("ttt2_sprint_stamina_consumption"), 0)
 		end
 
 		if ply.sprintProgress == 1 then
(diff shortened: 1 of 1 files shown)


TimGoll:
I'm with @saibotk 
While you're at it, what about adding a speed multiplier hook?

TheOnly8Z:
> I'm with @saibotk
> While you're at it, what about adding a speed multiplier hook?

I was planning on doing that separately, since I would need to modify the TTTPlayerSpeedModifier hook for that.
Technically you can already accomplish this by checking whether the player is sprinting in this hook, but it's ugly and I'll see what I can do later.

TimGoll:
That's fine too! Your code looks good to me. Have you tested it on your server?

TheOnly8Z:
Tested and works as intended (shared hook). Ready to merge whenever you are.

TimGoll:
I'd give my ok too

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -87,11 +87,16 @@ hook.Add("Think", "TTT2PlayerSprinting", function()
 		if not ply.sprintTS then continue end
 
 		local timeElapsed = CurTime() - ply.sprintTS
-
-		if not ply.sprintMultiplier then
-			ply.sprintProgress = math.min((ply.oldSprintProgress or 0) + timeElapsed * GetGlobalFloat("ttt2_sprint_stamina_regeneration"), 1)
+		local isSprinting = (ply.sprintMultiplier ~= nil)
```
Alf21: theoretically, there already exists a client var: https://github.com/TTT-2/TTT2/blob/d387992086fd9c4fe4669afaa2670ee9bd6aa23a/gamemodes/terrortown/gamemode/shared/sh_sprint.lua#L57

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -87,11 +87,16 @@ hook.Add("Think", "TTT2PlayerSprinting", function()
...
+
+		local drainMult, regenMult = hook.Run("TTT2StaminaMult", ply, isSprinting, ply.sprintProgress)
```
Alf21: it is right to send data with this hook, but should think about consistency. isSprinting and ply.sprintProgess is accessible on the submitted Player "object" "ply" itself.

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -87,11 +87,16 @@ hook.Add("Think", "TTT2PlayerSprinting", function()
...
+		drainMult = drainMult or 1
```
Alf21: this could be inlined, depending on the code style :D

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -87,11 +87,16 @@ hook.Add("Think", "TTT2PlayerSprinting", function()
...
+		regenMult = regenMult or 1
```
Alf21: same

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -87,11 +87,16 @@ hook.Add("Think", "TTT2PlayerSprinting", function()
...
+			
+		if !isSprinting then
```
Alf21: I suggest using "not" instead of the C-Syntax

Alf21:
I've created some comments show you some possible improvements, just little things :D

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
TheOnly8Z: Does this exist on the server, however? doing something like `local isSprinting = CLIENT and LocalPlayer().isSprinting or (ply.sprintMultiplier ~= nil)` seems awkward.

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
TheOnly8Z: isSprinting should probably be kept - I think it's not intuitive to potential addon developers, and providing the information makes sense.

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
Alf21: https://github.com/TTT-2/TTT2/blob/d387992086fd9c4fe4669afaa2670ee9bd6aa23a/gamemodes/terrortown/gamemode/shared/sh_sprint.lua#L46 it's serverside too, same as ply.sprintMultiplier

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
Alf21: Hmm... It's okay if we handle it this way...

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -87,11 +87,14 @@ hook.Add("Think", "TTT2PlayerSprinting", function()
 		if not ply.sprintTS then continue end
 
 		local timeElapsed = CurTime() - ply.sprintTS
+		local isSprinting = (ply.sprintMultiplier ~= nil)
 
-		if not ply.sprintMultiplier then
-			ply.sprintProgress = math.min((ply.oldSprintProgress or 0) + timeElapsed * GetGlobalFloat("ttt2_sprint_stamina_regeneration"), 1)
+		local drainMult, regenMult = hook.Run("TTT2StaminaMult", ply, isSprinting)
```
Alf21: @saibotk I really don't like this solution, because this seems to be inconsistent. Every time, one var is not used and all the time nil. Using this hook seems to be tricky too. I think one var should be enough. The only thing that makes sense would be to call a "TTT2StaminaDrainMult" and a "TTT2StaminaRegenMult" hook some lines below. Then we can remove the isSprinting value and everything should be more consistent with better integrity.

TheOnly8Z: The separation of vars was suggested by @saibotk earlier in Discord, and I did initially plan to only keep one variable. Would calling it staminaMult be a good fit?

saibotk: No, thats not what @Alf21 means, they should be separate but he dislikes the combined hook, which can be split up in 2 hooks, i just suggested it with the thought of not adding too many useless hooks, but i suppose this should be okay here.

Alf21:
We should think about a referenced value too (to support multiple hooking, see https://github.com/TTT-2/TTT2/blob/da4b418dfad7111eb0526b95c353bc497c594f36/gamemodes/terrortown/gamemode/shared/sh_main.lua#L56). There are two ways: using a table as a param like in this link as a reference (noLag param) or using a player object stored var (e.g. ply.staminaRegenMult, see https://github.com/Alf21/ttt2-item_speedrun/blob/master/lua/terrortown/entities/items/item_ttt_speedrun.lua#L21 => ply.speedrun_mul)

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -87,11 +87,13 @@ hook.Add("Think", "TTT2PlayerSprinting", function()
 		if not ply.sprintTS then continue end
 
 		local timeElapsed = CurTime() - ply.sprintTS
-
-		if not ply.sprintMultiplier then
-			ply.sprintProgress = math.min((ply.oldSprintProgress or 0) + timeElapsed * GetGlobalFloat("ttt2_sprint_stamina_regeneration"), 1)
+		local drainMult = hook.Run("TTT2StaminaDrain", ply)
```
Alf21: These hooks should be placed in the right if-clause to avoid the call if it's not necessary :)

PR #366, opened 2019-07-26
Added documentation

Alf21:
and removed the old docs file


Alf21:
found some bugs while creating the docs, have fun to review LOL

TheOnly8Z:
Would it be better if function names and parameters are provided in the wiki for easier access?

Alf21:
What do you mean exactly? Currently, I‘ve added each param with the type to the documentation for each shared function. With the help of our own documentation tool, a website should be generated.

TheOnly8Z:
I meant the Github wiki in particular. If auto-generation is possible, that would help greatly.

TimGoll:
The GitHub wiki will consist of some getting started guides, like the one for the HUD that is currently available

Alf21:
Yea it wouldn't make any sense to put these function into the wiki. There is no nice search function and this would lead to inconsistency.

Histalek:
i can't find it right now, but the docs were hosted somewhere right? Just put the link in the README and it should be quite accessible (just if you weren't planning to do so anyway)

TheOnly8Z:
Could possibly use Github Pages, or maybe a separate page in the wiki? Unfortunately I've never used either so I can't vouch for them.

If all the functions are in one page, I think searching won't be too big of a hassle (with CTRL+F).

TimGoll:
Our docu page will be basically like the GMOD docu. Way better than a classic github wiki

TheOnly8Z:
That sounds inspiring. I'll leave the web development up to you guys!

Alf21:
Now I need to modify the doc tool again for some special cases...

PR #370, opened 2019-08-11
Armor system

TimGoll:
I made a simple implemenatation of a new armor system based on Histaleks idea. The aim of the new system is to make a visually easy to understand armor system that is not confusing when multiple armor effects stack. The percentual system right now is confusing too many players, especially when you have for example the classic armor in combination with a super soda.

- armor is a secondary health information on top of the normal HP
- it blocks only bulletdamage
- 20% of  the damage are sucked into the armor and reduce the armor value, 10% vanish and 70% get reduced from the health (values open for discussion)
- the classic armor item is replaced with a multibuy armor item, that instantly grants 30 armor to the player
- other roles can set specific values if the roledev plans on giving roles armor on spawn
- heroes roles and supersodas can use this system too instead of the classic armor
- classic armor is included as a nonbuyable fallback option
- right now only pureskin support is included and the armor bar is hidden if a player has no armor


TimGoll:
**Why this armorsystem?**
The armor system in TTT is always a bit confusing, especially if you have multiple addons that reduce the damage taken. This approach changes this.

**Are there any drawbacks?**
No. If players do not like the new system, they can disable it with a convar.

**How does it work?**
Additional to the HP points, each player now has armor points. When he receives damage, the damage gets split between the armor and the HP (20% armor, 70% HP, 10% vanish, values tweakable). If a player has more that 50 armor, the values increase by about 10%. This allows for stronger armor if a player buys the armor item twice, or finds multiple armor sodas for example.

Alf21:
I like the idea, but I think there is a better solution than defining a fixed constant like "50" armor.
In my opinion, the currently stored armor should be used for the calculation relatively. Defining 50 as a fixed constant and maybe 100 as a maximum value seems to be inconsistent because there isn't a maximum value for the health. 100 is the default health amount, but there should be the possibility to have a higher value always. As a second argument, restricting the armor amount to 100 doesn't make sense thinking about the issue that you are able to protect yourself with new technologies in real life better (more than the currently defined "100%").
In conclusion, there could never be a fixed or defined 100%. As a result, there can't (or shouldn't) be a fixed value like "50".

gamemodes/terrortown/entities/weapons/weapon_ttt_armor.lua
```diff
@@ -0,0 +1,43 @@
+SWEP.Base = "weapon_tttbase"
```
Alf21: Why is there a SWEP weapon?

gamemodes/terrortown/gamemode/client/cl_status.lua
```diff
@@ -8,14 +8,19 @@ STATUS.active = {}
 
 ---
 -- Registers a @{STATUS}
--- @param number id The index of the new status
+-- @param number id the index of the new status
```
Alf21: btw. the previous one was right ;) everything after the param name is the string arg for the docu ^^

gamemodes/terrortown/gamemode/client/cl_status.lua
```diff
@@ -8,14 +8,19 @@ STATUS.active = {}
...
 -- @return boolean whether the creation was successfully
 -- @realm client
 function STATUS:RegisterStatus(id, data)
 	if STATUS.registered[id] ~= nil then  -- name is not unique
 		return false
 	end
+	
+	-- support single and multible icons per status effect
+	if data.hud.GetTexture then
+		data.hud = {data.hud}
```
Alf21: lol, why not using `istable(data.hud)` check?

gamemodes/terrortown/gamemode/client/cl_status.lua
```diff
@@ -25,23 +30,26 @@ end
...
 -- Adds a status to the currently active ones
 -- @param number id The id of the registered @{STATUS}
+-- @param number active_icon
 -- @realm client
-function STATUS:AddStatus(id)
+function STATUS:AddStatus(id, active_icon)
 	if STATUS.registered[id] == nil then return end
 
 	STATUS.active[id] = table.Copy(STATUS.registered[id])
+	self:SetActiveIcon(id, active_icon or 1)
```
Alf21: missing linebreak in front

gamemodes/terrortown/gamemode/client/cl_status.lua
```diff
@@ -56,6 +64,41 @@ function STATUS:AddTimedStatus(id, duration, showDuration)
...
 
+---
+-- Changes the active icon for a specifiv active effect for a given @{Player}
+-- @param number id The id of the registered @{STATUS}
+-- @param number active_icon
+-- @realm client
+function STATUS:SetActiveIcon(id, active_icon)
+	if STATUS.active[id] == nil then return end
+
+	local max_amount = (self.registered[id].hud.GetTexture) and 1 or #STATUS.registered[id].hud
```
Alf21: why brackets around the check?

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -78,22 +80,25 @@ function GM:PostDrawTranslucentRenderables(bDrawingDepth, bDrawingSkybox)
...
-			local pos = ply:GetPos()
-			pos.z = pos.z + 80
-
+			
+			local pos = ply:GetBonePosition(6)
+			pos.z = pos.z + 20
+			
+			local ea = ply:EyeAngles()
+			ea.pitch = 0
+			local shift = Vector(2,0,0)
```
Alf21: missing linebreak in front

Alf21: missing spaces in Vector instruction

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -78,22 +80,25 @@ function GM:PostDrawTranslucentRenderables(bDrawingDepth, bDrawingSkybox)
...
+			shift:Rotate(ea)
+			pos:Add(shift)
```
Alf21: missing linebreak in front

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -78,22 +80,25 @@ function GM:PostDrawTranslucentRenderables(bDrawingDepth, bDrawingSkybox)
...
+			
 			if ply ~= client
 			and ply:IsActive()
 			and ply:IsSpecial()
 			and (not client:IsActive() or ply:IsInTeam(client))
 			and not ply:GetSubRoleData().avoidTeamIcons
 			then
-				local base = Material("vgui/ttt/dynamic/sprite_base")
-				local base_overlay = Material("vgui/ttt/dynamic/sprite_base_overlay.png")
 				local icon = Material("vgui/ttt/dynamic/roles/icon_" .. rd.abbr)
```
Alf21: should be cached too

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,64 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+	-- DISABLED for now since the hook gets called twice, once with 0 damage, but with a hitgroup and once
+	-- with damage but no hitgroup. I have to dig deeper into this
+
+	-- some entities cause this hook to be triggered, ignore 0 damage events
+	if dmginfo:GetDamage() == 0 then return end
+
+	-- fallback for players who prefer the vanilla armor
+	if GetConVar("ttt_armor_classic"):GetBool() and ent:Armor() > 0 then
+		dmginfo:ScaleDamage(0.7)
+		return
```
Alf21: missing linebreak in front

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,64 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+	end
+
+	print("-------------------------------------------------------")
+
+	-- handle different damage type factors
+	local armor_factor, body_factor
+	if dmginfo:IsDamageType(DMG_BULLET) or dmginfo:IsDamageType(DMG_CLUB) then -- bullet or crowbar damage
```
Alf21: missing linebreak in front

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,64 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+	-- some entities cause this hook to be triggered, ignore 0 damage events
+	if dmginfo:GetDamage() == 0 then return end
+
+	-- fallback for players who prefer the vanilla armor
+	if GetConVar("ttt_armor_classic"):GetBool() and ent:Armor() > 0 then
+		dmginfo:ScaleDamage(0.7)
+		return
+	end
+
+	print("-------------------------------------------------------")
```
Alf21: print should be removed or used in a debugging function...

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,64 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+
+	-- handle different damage type factors
+	local armor_factor, body_factor
+	if dmginfo:IsDamageType(DMG_BULLET) or dmginfo:IsDamageType(DMG_CLUB) then -- bullet or crowbar damage
+		armor_factor = 1.0
+		body_factor = 1.0
+		print("damage type	: bullet or crowbar")
```
Alf21: ...

Alf21: missing linebreak in front

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,64 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+	elseif dmginfo:IsDamageType(DMG_BURN) or dmginfo:IsDamageType(DMG_BLAST) then -- fire or explosion damage
+		armor_factor = 1.5
+		body_factor = 0.75
+		print("damage type	: fire or explosion")
```
Alf21: ...

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,64 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+	else
+		return
+	end
+
+	-- calculate damage
+	local damage = dmginfo:GetDamage()
+	local armor = ent:Armor()
+	
+	print("current HP	 : " .. tostring(ent:Health()))
```
Alf21: ...

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,64 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+	print("current Armor  : " .. tostring(armor))
+	print("damage to take : " .. tostring(damage))
+
+	-- normal damage handling when no armor is available
+	if armor == 0 then return end
+
+	local cv_armor_factor = ent:ArmorIsReinforced() and GetConVar("ttt_armor_rei_damage_block_pct"):GetFloat() or GetConVar("ttt_armor_damage_block_pct"):GetFloat()
+	local cv_body_factor = ent:ArmorIsReinforced() and GetConVar("ttt_armor_rei_damage_health_pct"):GetFloat() or GetConVar("ttt_armor_damage_health_pct"):GetFloat()
+
+	print("armor factor   : " .. tostring(cv_armor_factor))
```
Alf21: ...

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,64 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+	print("body factor	: " .. tostring(cv_body_factor))
+
+	armor = armor - cv_armor_factor * armor_factor * damage
+	print("armor internal : " .. tostring(armor))
```
Alf21: ...

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,64 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+	ent:SetArmor(math.max(armor, 0))
+	print("new armor	  : " .. tostring(ent:Armor()))
```
Alf21: ...

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,64 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+	
+	local new_damage = cv_body_factor * body_factor * damage - math.min(armor, 0)
+	print("calced dmg	 : " .. tostring(new_damage))
```
Alf21: ...
missing linebreak in front

gamemodes/terrortown/gamemode/shared/hud_elements/base_elements/pure_skin_element.lua
```diff
@@ -52,10 +52,10 @@ if CLIENT then
 		surface.SetDrawColor(clr(c))
 		surface.DrawRect(x, y, w, h)
 
-		local w2 = math.Round(w * (p or 1))
+		local w_ = math.Round(w * (p or 1))
```
Alf21: this var name is horrible :D

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/old_ttt_info.lua
```diff
@@ -63,8 +63,13 @@ if CLIENT then
 			local health = math.max(0, client:Health())
 			local health_y = y + margin
 
+			local health_string = tostring(health)
+			if not GetGlobalBool("ttt_armor_classic", false) and client:Armor() > 0 then
```
Alf21: missing linebreak in front

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
```diff
@@ -206,8 +209,36 @@ if CLIENT then
...
+			local armor = math.max(0, client:Armor())
+			local armor_print = tostring(armor)
+
+			self:DrawBar(nx, ty, bw, bh, Color(234, 41, 41), health / client:GetMaxHealth(), self.scale, "HEALTH: " .. health_print)
 
-			self:DrawBar(nx, ty, bw, bh, Color(234, 41, 41), health / client:GetMaxHealth(), self.scale, "HEALTH: " .. health)
+			-- draw armor information
+			if not GetGlobalBool("ttt_armor_classic", false) and armor > 0 then
+				local icon_mat
+				if client:ArmorIsReinforced() then
```
Alf21: missing linebreak in front

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
```diff
@@ -206,8 +209,36 @@ if CLIENT then
...
+					icon_mat = icon_armor_rei
+				else
+					icon_mat = icon_armor
+					--icon_mat = watching_icon
+				end
+				local a_size = bh - math.Round(11 * self.scale)
```
Alf21: missing linebreak in front

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
```diff
@@ -206,8 +209,36 @@ if CLIENT then
...
+				local a_pos_y = ty + math.Round(5 * self.scale)
+				local a_pos_x = nx + bw - math.Round(65 * self.scale)
+
+				local at_pos_y = ty + 1
+				local at_pos_x = a_pos_x + a_size + a_pad
+
+				util.DrawFilteredTexturedRect(a_pos_x + math.Round(2*self.scale), a_pos_y + math.Round(2*self.scale), a_size, a_size, icon_mat, 200, {r=0,g=0,b=0})
+				util.DrawFilteredTexturedRect(a_pos_x + math.Round(self.scale), a_pos_y + math.Round(self.scale), a_size, a_size, icon_mat, 255, {r=0,g=0,b=0})
+				util.DrawFilteredTexturedRect(a_pos_x, a_pos_y, a_size, a_size, icon_mat, 255, {r=255,g=255,b=255})
+				draw.AdvancedText(armor_print, "PureSkinBar", at_pos_x, at_pos_y, self:GetDefaultFontColor(Color(234, 41, 41)), TEXT_ALIGN_LEFT, TEXT_ALIGN_LEFT, true, self.scale)
```
Alf21: missing linebreak in front

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
```diff
@@ -206,8 +209,36 @@ if CLIENT then
...
+				local a_size = bh - math.Round(11 * self.scale)
+				local a_pad = math.Round(5 * self.scale)
+
+				local a_pos_y = ty + math.Round(5 * self.scale)
+				local a_pos_x = nx + bw - math.Round(65 * self.scale)
+
+				local at_pos_y = ty + 1
+				local at_pos_x = a_pos_x + a_size + a_pad
+
+				util.DrawFilteredTexturedRect(a_pos_x + math.Round(2*self.scale), a_pos_y + math.Round(2*self.scale), a_size, a_size, icon_mat, 200, {r=0,g=0,b=0})
```
Alf21: missing spaces between `r=0,g=0,b=0`

gamemodes/terrortown/gamemode/shared/hud_elements/tttsidebar/pure_skin_sidebar.lua
```diff
@@ -81,7 +81,13 @@ if CLIENT then
 		surface.SetDrawColor(item.hud_color.r, item.hud_color.g, item.hud_color.b, math.Round(factor * 255))
 		surface.DrawRect(pos.x, curY, size.w, size.w)
 
-		util.DrawFilteredTexturedRect(pos.x, curY, size.w, size.w, item.hud, iconAlpha, fontColor)
+
+		local hud_icon = item.hud
+		if not hud_icon.GetTexture then
```
Alf21: missing linebreak in front

Alf21: maybe using `istable(hud_icon)` check too?

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,122 @@ function plymeta:GetRole()
...
+function plymeta:GetMaxArmor()
+	return self.armor_max or 100
+end
+
+---
+-- Returns wether the armor is reinforced
+-- @return boolean is armor reinforced
+-- @realm shared
+function plymeta:ArmorIsReinforced()
+	return self:Armor() > 50
```
Alf21: 50 as a defined constant?

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,122 @@ function plymeta:GetRole()
...
+	end)
+
+	---
+	-- Sets the armor to a specific value
+	-- @param number armor the new armor to be set
+	-- @realm server
+	function plymeta:SetArmor(armor)
+		self.armor = armor
+
+		print("Setting armor to: " .. tostring(self.armor))
```
Alf21: ...

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,122 @@ function plymeta:GetRole()
...
+
+	hook.Add("PlayerSpawn", "ttt2_player_armor_spawn_reset", function(ply)
+		ply:SetArmor(GetConVar("ttt_armor_on_spawn"):GetInt())
+	end)
+else
+	net.Receive("ttt2_sync_armor", function()
+		local client = LocalPlayer()
+
+		-- prevent error from netmessage prior to the client beeing ready
+		if not client or not IsValid(client) then return end
```
Alf21: first check is useless

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,122 @@ function plymeta:GetRole()
...
+		client.armor = net.ReadUInt(16)
+
+		-- UPDATE STATUS ICONS
+		-- removed armor
+		if STATUS:Active("ttt_weapon_armor") and client.armor == 0 then
+			STATUS:RemoveStatus("ttt_weapon_armor")
+		end
+
+		-- check if reinforced
+		local icon_id = 1
```
Alf21: missing linebreak

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,122 @@ function plymeta:GetRole()
...
+		if not STATUS:Active("ttt_weapon_armor") and client.armor > 0 then
+			STATUS:AddStatus("ttt_weapon_armor", icon_id)
+		end
+
+		-- normal armor level change
+		if STATUS:Active("ttt_weapon_armor") then
+			STATUS:SetActiveIcon("ttt_weapon_armor", icon_id)
+		end
+	end)
+	net.Receive("ttt2_sync_armor_max", function()
```
Alf21: missing linebreak in front

lua/terrortown/entities/items/item_ttt_armor.lua
```diff
@@ -9,7 +9,7 @@ ITEM.EquipMenuData = {
 	desc = "item_armor_desc"
 }
 ITEM.material = "vgui/ttt/icon_armor"
-ITEM.CanBuy = {ROLE_TRAITOR, ROLE_DETECTIVE}
+ITEM.CanBuy = nil
```
Alf21: why?

TimGoll:
100 is not the max. It's the same as with the health. Max gets used if you want to draw a bar, but it is not capped to 100

Should I add a convar for the reinforced armor? Or do you want function that calculates the protection based on the value? I considered the latter, but ditched it because I think it will confuse players

TimGoll:
Additionally I'd like to add that armor is no percentual value

lebroomer:
I think a percental value could be ambigous, since stacking armor is a possibility. With a percentual value the new reinforced border would be complicated, where as an additive procedure with "50" always being the reinforced border would be simple and understandable.

Alf21:
I don't think that the current solution is understandable and reasonable. You need to use a formula to calculate the damage. Same as using the armor as a relative factor. The reinforced armor thing is a nice idea too, but maybe a thing that makes it more complicated. Displaying the current armor and using a constant formula is (in my opinion) the easiest way in this armor system.

lebroomer:
Yes the current solution is quite complex too, but since the concept is broadly used in other games (mainly BRs) users might understand it fairly quick.

TimGoll:
I strongy disagree with alf. This system was designed with the community to stop using a formula, since a formula is confusing and a constant armor value is more reasonable

Right now, if a player has armor and an armor soday he receives 0.7*0.7 damage. There is no way to display this in a good way and it is nowhere near easy to understand. It's a reason nobody ever bothered to buy armor. 

A formula having a non constant corelation between the armor value and the damage blocked  would make it more complicated and the armor would be pretty op for high armor values when it is not capped.

Additionally you can disable the system with a convar

gamemodes/terrortown/entities/weapons/weapon_ttt_armor.lua
TimGoll: Is there a better way for an armor item that can be bought multiple times?

gamemodes/terrortown/gamemode/client/cl_status.lua
TimGoll: You're right. My bad

gamemodes/terrortown/gamemode/client/cl_status.lua
TimGoll: Since materials are tables too, I have to check if it is a material. The table check returns true for both.

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: I left it there on puurpose until the whole thing gets mergen. Just in case somebody notices a bug

gamemodes/terrortown/gamemode/shared/hud_elements/tttsidebar/pure_skin_sidebar.lua
TimGoll: Same problem here

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
TimGoll: Might change this to a convar

lua/terrortown/entities/items/item_ttt_armor.lua
TimGoll: Because I wanted to keep the old item as a fallback but it should no be buyable. Might add the new functionality to the old item, but it has to be rebuyable

TimGoll:
I talked with Nick for a bit about this and he had some good points. He kinda agrees with you, Alf.
Basically he said that the new system isn't any more confusing than the old one, but it is different than the old one. Additionally he mentioned that displaying the value might confuse players.

So now I'm not 100% sure how to progress from this point. Do we want to change the system at all? Because I really don't like the original system. But it seems like the playerbase is a bit split on this. Other opinions are very welcome.

These are the options right now:
- scrap this idea
- change it in a way that it makes it easier for people to understand: my first idea was to add an armor value on top of the HP that gets consumed first. Like in GTA. Basically additional HP. 
- add it as a disabled feature that can be enabled if people want to use it

LunexKun:
I am in favor of Tim's current idea, and system overall. I'd really like a redisign of the current armor, since I never got the point of armor in the first place. I know it does something, but I honestly never bothered.
Making armor more interesting in general would open up windows for posible ideas and advancements in the future, for example when designing new roles, items and other features such as events. 
Overall I believe this to be a great addition, but ofcourse since it is something new, some people might not like it, and I think they shouldn't be forced to use it.
(though addons using this system in the future would require them to do so)
So yes, I'd appreciate this armor system to be added. A balancing sided issue could be resolved through convars to change the values that armor covers.

TimGoll:
@histalek
The classic system always sets the damage to 70%, no matter how high your armor value is

Or do you talk abiut stacking in the vanilla system?

Histalek:
> @Histalek
> The classic system always sets the damage to 70%, no matter how high your armor value is
> 
> Or do you talk abiut stacking in the vanilla system?

yes i was talking about stacking the vanilla system ^^

gamemodes/terrortown/entities/weapons/weapon_ttt_armor.lua
Alf21: this is a trashy way, then, multiple item bought should be implemented

gamemodes/terrortown/gamemode/client/cl_status.lua
Alf21: lol ok, nm

gamemodes/terrortown/entities/weapons/weapon_ttt_armor.lua
TimGoll: Oof. This would be the best solution. I'll look into the item code.

TimGoll:
yup, the weapon-> item will be done after holidays, comments will be removed after testing. Anything else missing?

TimGoll:
multibuy for items is now finished

TimGoll:
Armor system is now finished. I'll remove the debug prints once it is ready to merge into the master.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -101,6 +101,7 @@ if SERVER then
 	CreateConVar('ttt_armor_damage_health_pct', 0.7, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
 	CreateConVar('ttt_armor_on_spawn', 0, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
 	CreateConVar('ttt_armor_for_reinforced', 50, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+	CreateConVar('ttt_armor_buy_value', 30, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
```
Histalek: IMO the conVar should be named something like 'ttt_item_armor_value'.
For me personally 'ttt_armor_...' means it is a conVar directly related to the armorsystem. Throwing an item convar into it seems not a good idea, right?

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -190,17 +190,18 @@ end
 -- @param string id
 -- @realm server
 function plymeta:AddEquipmentItem(id)
-	if not self:HasEquipmentItem(id) then
-		self.equipmentItems = self.equipmentItems or {}
-		self.equipmentItems[#self.equipmentItems + 1] = id
+	local item = items.GetStored(id)
+	
+	if item.limited and self:HasEquipmentItem(id) then return end
```
Alf21: Bug. Check `if item and item.kimited and ... then`

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -190,17 +190,18 @@ end
...
 
-		local item = items.GetStored(id)
-		if item and isfunction(item.Equip) then
-			item:Equip(self)
-		end
+	self.equipmentItems = self.equipmentItems or {}
+	self.equipmentItems[#self.equipmentItems + 1] = id
 
-		self:SendEquipment()
+	if item and isfunction(item.Equip) then
```
Alf21: remove item check

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -460,13 +461,17 @@ end
...
-	if self:HasEquipmentItem(id) then
-		return false
-	elseif id then
-		self:AddEquipmentItem(id)
+	if not id then return end
 
-		return true
+	local item = items.GetStored(id)
+
+	if item.limited and self:HasEquipmentItem(id) then
```
Alf21: add if item and ... check

gamemodes/terrortown/gamemode/server/sv_player.lua
Alf21: ...

gamemodes/terrortown/gamemode/server/sv_player.lua
Alf21: ...

gamemodes/terrortown/gamemode/server/sv_player.lua
Alf21: ...

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,57 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+	print("current HP	 : " .. tostring(ent:Health()))
+	print("current Armor  : " .. tostring(armor))
+	print("damage to take : " .. tostring(damage))
+
+	-- normal damage handling when no armor is available
+	if armor == 0 then return end
+
+	local cv_armor_factor = GetConVar("ttt_armor_damage_block_pct"):GetFloat()
+	local cv_body_factor = GetConVar("ttt_armor_damage_health_pct"):GetFloat()
+	if ent:ArmorIsReinforced() then
```
Alf21: missing linebreak in front

gamemodes/terrortown/gamemode/server/sv_player.lua
Alf21: ...

gamemodes/terrortown/gamemode/server/sv_player.lua
Alf21: ...

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/old_ttt_info.lua
```diff
@@ -62,9 +62,14 @@ if CLIENT then
 			-- Draw health
 			local health = math.max(0, client:Health())
 			local health_y = y + margin
+			local health_string = tostring(health)
+			
+			if not GetGlobalBool("ttt_armor_classic", false) and client:Armor() > 0 then
+				health_string = health_string .. " + " .. tostring(client:Armor())
```
Alf21: sure u need to call the "tostring" function?

gamemodes/terrortown/gamemode/shared/sh_init.lua
```diff
@@ -47,6 +47,7 @@ SPECIAL_EQUIPMENT = {
 	"weapon_ttt_stungun",
 	"weapon_ttt_cse",
 	"weapon_ttt_teleport",
+	"weapon_ttt_armor",
```
Alf21: still exists?

gamemodes/terrortown/gamemode/shared/sh_init.lua
```diff
@@ -65,6 +66,7 @@ TRAITOR_EQUIPMENT = {
 	"weapon_ttt_teleport",
 	"weapon_ttt_decoy",
 	"item_ttt_armor",
+	"weapon_ttt_armor",
```
Alf21: still exists?

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,131 @@ function plymeta:GetRole()
...
+	return GetGlobalBool("ttt_armor_reinforced_enabled") and self:Armor() > GetGlobalInt("ttt_armor_for_reinforced")
+end
+
+if SERVER then
+	util.AddNetworkString("ttt2_sync_armor")
+	util.AddNetworkString("ttt2_sync_armor_max")
+
+	CreateConVar('ttt_armor_damage_block_pct', 0.2, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+	CreateConVar('ttt_armor_damage_health_pct', 0.7, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+	CreateConVar('ttt_armor_on_spawn', 0, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
```
Alf21: put this above and store the var

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,131 @@ function plymeta:GetRole()
...
+function plymeta:GetMaxArmor()
+	return self.armor_max or 100
+end
+
+---
+-- Returns wether the armor is reinforced
+-- @return boolean is armor reinforced
+-- @realm shared
+function plymeta:ArmorIsReinforced()
+	return GetGlobalBool("ttt_armor_reinforced_enabled") and self:Armor() > GetGlobalInt("ttt_armor_for_reinforced")
```
Alf21: use stored vars instead

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,131 @@ function plymeta:GetRole()
...
+	end)
+
+	---
+	-- Sets the armor to a specific value
+	-- @param number armor the new armor to be set
+	-- @realm server
+	function plymeta:SetArmor(armor)
+		self.armor = armor
+
+		print("Setting armor to: " .. tostring(self.armor))
```
Alf21: ...

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,131 @@ function plymeta:GetRole()
...
+		-- normal armor level change
+		if STATUS:Active("ttt_weapon_armor") then
+			STATUS:SetActiveIcon("ttt_weapon_armor", icon_id)
+		end
+	end)
+
+	net.Receive("ttt2_sync_armor_max", function()
+		local client = LocalPlayer()
+
+		client.armor_max = net.ReadUInt(16)
```
Alf21: if IsValid(player) then check missed

Alf21: or negate it. Look above

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: I even said in the comment above that I remove the prints once it is proper tested

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/old_ttt_info.lua
TimGoll: yes

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
TimGoll: what do you mean by this?

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/old_ttt_info.lua
Alf21: why?

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
Alf21: var localization.
```lua
local armorOnSpawn = CreateConVar('ttt_armor_on_spawn', 0, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
```

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/old_ttt_info.lua
TimGoll: without it it will throw an error because it can't concat an integer to a string

Alf21: Could you send this error?

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
Alf21: same in line 125

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -124,6 +124,9 @@ function GM:PlayerSpawn(ply)
 	ply:SetupHands()
 
 	SCORE:HandleSpawn(ply)
+
+	-- reset armor of player on spawn
+	ply:SetArmor(0)
```
Alf21: Very ugly to paste it anywhere. Should be used in functions that are created for reset. E.g. plymeta:ResetRoundFlags()

lua/terrortown/entities/items/item_ttt_armor.lua
```diff
@@ -19,6 +19,12 @@ function ITEM:Equip(buyer)
 	end
 end
 
+function ITEM:Reset(buyer)
+	if SERVER then
+		buyer:DecreaseArmor(GetConVar("ttt_item_armor_value"):GetInt())
```
Alf21: Hmmm... It's tricky. Is there a possibility to get a negative amount? What, if I bought another armor and this item removes (but I already got damage and lost armor)...

TimGoll: Therefore the function "DecreaseArmor", it handles the capping

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: I removed it at the other place, it is only in this one position. I could use a function for this if you really prefer it.

TimGoll:
So are there any more things requested to change except removing the debug prints?

I kind of lost the oversight, but I sthink everything is taken care of.

gamemodes/terrortown/gamemode/server/sv_player.lua
Alf21: yea, you should use the requested or any similar function :)

gamemodes/terrortown/gamemode/client/cl_targetid.lua
Alf21: what is with the icon caching?

Alf21:
The prints, the armor function move and the material caching

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: okay, no problem

TimGoll:
material caching: since this was the same problem prior to this PR, I'd say this is no part of this PR. I can do it though, but not today. But one might merge this into the master for a general audience (Nick etc) to test the new system on their server.

The other two problems will be taken care of in a moment.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -460,13 +463,17 @@ end
 -- @return boolean success?
 -- @realm server
 function plymeta:GiveEquipmentItem(id)
```
Alf21: What's the difference between ply:GiveEquipmentItem and ply:AddEquipmentItem?

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/old_ttt_info.lua
Alf21: !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!?

lua/terrortown/entities/items/item_ttt_armor.lua
```diff
@@ -2,20 +2,38 @@ if SERVER then
...
+	hook.Add("Initialize", "ttt2_base_register_armor_status", function() 
+		STATUS:RegisterStatus("ttt_weapon_armor", {
+			hud = {
+				Material("vgui/ttt/perks/hud_armor.png"),
+				Material("vgui/ttt/perks/hud_armor_reinforced.png")
+			},
+			type = "good"
+		})
+	end)
+end
```
Alf21: eol missing

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/old_ttt_info.lua
```diff
@@ -62,9 +62,14 @@ if CLIENT then
 			-- Draw health
 			local health = math.max(0, client:Health())
 			local health_y = y + margin
+			local health_string = tostring(health)
```
Alf21: tostring again?

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,49 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
 	end
 end
 
+function HandlePlayerArmorSystem(ent, infl, att, amount, dmginfo)
+	-- armor does not help when the head receives damage
+	--if ent:LastHitGroup() == HITGROUP_HEAD then return end
```
saibotk: Please remove uncommented code

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1430,14 +1430,21 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
 		-- scale everything to karma damage factor except the knife, because it
 		-- assumes a kill
 		if not dmginfo:IsDamageType(DMG_SLASH) then
 			dmginfo:ScaleDamage(att:GetDamageFactor())
 		end
+	end
+
+	-- before the karma is calculated, but after all other damage hooks / damage change is processed,
+	-- the armor system should come into place (GM functions are called last)
+	HandlePlayerArmorSystem(ent, infl, att, amount, dmginfo)
```
saibotk: Could you explain, why this is excluded from the previous if statement?

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1446,6 +1453,49 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
+	if armor == 0 then return end
+
+	local cv_armor_factor = GetConVar("ttt_armor_damage_block_pct"):GetFloat()
+	local cv_body_factor = GetConVar("ttt_armor_damage_health_pct"):GetFloat()
+	
+	if ent:ArmorIsReinforced() then
+		cv_body_factor = cv_body_factor - 0.15
+	end
+
+	armor = armor - cv_armor_factor * damage
```
saibotk: This would add armor for negative damage values.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,155 @@ function plymeta:GetRole()
...
+
+---
+-- Returns wether the armor is reinforced
+-- @return boolean is armor reinforced
+-- @realm shared
+function plymeta:ArmorIsReinforced()
+	return self.armor_reinforced or false
+end
+
+if SERVER then
```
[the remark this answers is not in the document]
Alf21: or in it's own sv_armor.lua file like the sprint system

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,155 @@ function plymeta:GetRole()
...
+	CreateConVar('ttt_armor_damage_block_pct', 0.2, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+	CreateConVar('ttt_armor_damage_health_pct', 0.7, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+	CreateConVar('ttt_item_armor_value', 30, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+	CreateConVar('ttt_armor_classic', 0, {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+
+	cvars.AddChangeCallback("ttt_armor_classic", function(cv, old, new)
+		SetGlobalBool("ttt_armor_classic", tobool(tonumber(new)))
+	end)
+	cvars.AddChangeCallback("ttt_armor_reinforced_enabled", function(cv, old, new)
+		SetGlobalBool("ttt_armor_reinforced_enabled", tobool(tonumber(new)))
```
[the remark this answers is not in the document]
Alf21: I don't know, times ago I had a bug and needed to use something similar. Maybe patched in this nearly 2 years

lua/autorun/ttt2_force_download.lua
```diff
@@ -18,6 +18,8 @@ if SERVER then
 	resource.AddFile("materials/vgui/ttt/dynamic/roles/icon_traitor.vmt")
 	resource.AddFile("materials/vgui/ttt/dynamic/roles/icon_det.vmt")
 	resource.AddFile("materials/vgui/ttt/dynamic/roles/icon_no_team.vmt")
+	
+	resource.AddFile("materials/vgui/ttt/equip/reroll.png")
```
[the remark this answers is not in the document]
TimGoll: I just added this to the downloadlist, it has nothing todo with the armor system, but I commited it to the wrong branch. I can change it though

lua/autorun/ttt2_force_download.lua
```diff
@@ -73,10 +75,14 @@ if SERVER then
 
 	-- ITEMs
 	resource.AddFile("materials/vgui/ttt/icon_armor.vmt") -- armor
-	resource.AddFile("materials/vgui/ttt/perks/hud_armor.png") -- armor HUD
 	resource.AddFile("materials/vgui/ttt/perks/hud_disguiser.png") -- disguiser
 	resource.AddFile("materials/vgui/ttt/perks/hud_radar.png") -- radar
 
+	resource.AddFile("materials/vgui/ttt/perks/hud_armor.png") -- armor HUD
+	resource.AddFile("materials/vgui/ttt/perks/hud_armor_reinforced.png") -- armor reinforced HUD
```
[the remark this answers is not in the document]
TimGoll: no. All items and status effects use pngs. Alf or you decided this.

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -78,22 +80,26 @@ function GM:PostDrawTranslucentRenderables(bDrawingDepth, bDrawingSkybox)
...
+			
+			local pos = ply:GetBonePosition(6)
+			pos.z = pos.z + 20
+			
+			local ea = ply:EyeAngles()
+			ea.pitch = 0
+			
+			local shift = Vector(2, 0, 0)
+			shift:Rotate(ea)
+			pos:Add(shift)
```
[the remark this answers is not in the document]
TimGoll: It has nothing todo with the armor system. I know, bad practive. I got a bit carried away. I improved the overheadicon a bit, nothing worth mentioning, nobody will notie a difference

lua/terrortown/entities/items/item_ttt_armor.lua
```diff
@@ -2,20 +2,38 @@ if SERVER then
...
+
+function ITEM:Reset(buyer)
+	if SERVER then
+		buyer:DecreaseArmor(GetConVar("ttt_item_armor_value"):GetInt())
+	end
+end
+
+-- REGISTER STATUS ICONS
+if CLIENT then
+	hook.Add("Initialize", "ttt2_base_register_armor_status", function()
```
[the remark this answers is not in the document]
TimGoll: like explained on the discord, it does belong to the armor system. It could have been split into two PRs, but since they depend on each other, I thought this might be a valid solution.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -190,17 +190,17 @@ end
...
-	if not self:HasEquipmentItem(id) then
-		self.equipmentItems = self.equipmentItems or {}
-		self.equipmentItems[#self.equipmentItems + 1] = id
+	local item = items.GetStored(id)
 
-		local item = items.GetStored(id)
-		if item and isfunction(item.Equip) then
-			item:Equip(self)
-		end
+	if not item then return end
```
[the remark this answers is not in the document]
TimGoll: I tested it with the new armor, the default TTT items and "a second chance". Without this check, buying an item that is not installed on the server, throws an error. This was never notived earlier, since IsItem prevented this.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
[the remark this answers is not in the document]
TimGoll: If I move this to sv_armor, I have to add sh_armor too. Or it will be inconsistent

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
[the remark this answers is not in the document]
TimGoll: I used the same code that was used in the TTT2 code by you and Alf

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,155 @@ function plymeta:GetRole()
 	return self.role or ROLE_INNOCENT
 end
 
+-- ARMOR SYSTEM
+---
+-- Returns the current armor
+-- @return[default=0] number armor
+-- @realm shared
+function plymeta:Armor()
```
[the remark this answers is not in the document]
TimGoll: yes and no. While I'm 100% with you, I copied the style of the health system. One way or another, there will be an inconsitency. I though using the style of the current systems is a good idea

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -68,6 +68,155 @@ function plymeta:GetRole()
...
+		net.WriteUInt(armor_max, 16)
+		net.Send(self)
+	end
+
+	---
+	-- Increases the armor about a specific value
+	-- @param number increaseby the amount to be increased
+	-- @realm server
+	function plymeta:IncreaseArmor(increaseby)
+		self:SetArmor(self:Armor() + increaseby)
```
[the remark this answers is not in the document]
TimGoll: Good point. And yes, maxarmor is used in the same way as maxhealth. It is only intended for HUD purposes, not to limit it

gamemodes/terrortown/gamemode/shared/hud_elements/tttsidebar/pure_skin_sidebar.lua
```diff
@@ -81,7 +81,9 @@ if CLIENT then
 		surface.SetDrawColor(item.hud_color.r, item.hud_color.g, item.hud_color.b, math.Round(factor * 255))
 		surface.DrawRect(pos.x, curY, size.w, size.w)
 
-		util.DrawFilteredTexturedRect(pos.x, curY, size.w, size.w, item.hud, iconAlpha, fontColor)
+		local hud_icon = item.hud.GetTexture and item.hud or item.hud[item.active_icon]
```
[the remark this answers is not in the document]
TimGoll: I see ne reason why it shouldn't be here?

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: Because I wanted to keep it a bit seperated to not clutter the code above. Additionally I like to work with `if bla then return end` instead of nested ifs

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: I have not thought about negative damage, thanks!

TimGoll:
@saibotk Thanks for the feedback. Except one or two comments I agree with your points.

> Language strings are not selected based on which mode (classic armor system) is enabled.
yep, I'll change that

> Is Item.limited is checked properly on serverside?
I implemented and tested it client and serverside. It worked for me. I'll ask Nick to test this on his server.

> Changes to the Status system do not belong here.
yes and no. You're right, but it would have been a nightmare to make this in two different PRs.

> Removal of hud_radar.png also does not belong to the Armor system.
I just changed the icon texture while I made the armor icon texture. Doesn't belong to the armor system, yes. I should have changed the branch, will do it next time.

lua/autorun/ttt2_force_download.lua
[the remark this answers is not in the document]
TimGoll: Hmm, I changed it to a vmt/vtf, now it is the only icon of the shop tabs that is a vmt/vtf. All the others are pngs. Therefore I think it should stay a png

gamemodes/terrortown/gamemode/shared/hud_elements/tttsidebar/pure_skin_sidebar.lua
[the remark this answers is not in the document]
TimGoll: It has to be this way since the same function is used for items and statusicons. Therefore this check has to be made

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: Very good question. I was confused too, since they are basically the same with the same checks. But they are not done by me and maybe something relies on them. Therefore I did not want to change anything inside them

gamemodes/terrortown/gamemode/server/sv_armor.lua
```diff
@@ -0,0 +1,146 @@
...
+-- @param number amount Amount of damage
+-- @param CTakeDamageInfo dmginfo Damage info
+-- @realm server
+-- @internal
+function ARMOR:HandlePlayerTakeDamage(ply, infl, att, amount, dmginfo)
+	-- some entities cause this hook to be triggered, ignore 0 damage events
+	if dmginfo:GetDamage() <= 0 then return end
+
+	-- fallback for players who prefer the vanilla armor
+    if self.cv.armor_classic:GetBool() and ply:Armor() > 0 then
```
Alf21: pls give a look to this indent

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -460,13 +462,17 @@ end
...
-	if self:HasEquipmentItem(id) then
-		return false
-	elseif id then
-		self:AddEquipmentItem(id)
+	if not id then return end
 
-		return true
+	local item = items.GetStored(id)
+
+	if item and item.limited and self:HasEquipmentItem(id) then
```
Alf21: what if item is nil?

gamemodes/terrortown/gamemode/shared/sh_armor.lua
```diff
@@ -0,0 +1,32 @@
...
+	return self.armor_max or 100
+end
+
+---
+-- Returns wether the armor is reinforced
+-- @return boolean is armor reinforced
+-- @realm shared
+function plymeta:ArmorIsReinforced()
+	return self.armor_is_reinforced or false
+end
```
Alf21: missing EOL

gamemodes/terrortown/gamemode/server/sv_armor.lua
```diff
@@ -0,0 +1,146 @@
...
+-- Sets the initial @{Player} Armor, called at @{GM:TTTBeginRound}
+-- @realm server
+-- @internal
+function ARMOR:InitPlayerArmor() 
+    for _, p in ipairs(player.GetAll()) do
+        if p:IsTerror() then 
+            p:SetArmor(self.cv.armor_on_spawn:GetInt())
+        end
+    end
+end
```
Alf21: missing EOL

gamemodes/terrortown/gamemode/client/cl_armor.lua
```diff
@@ -0,0 +1,59 @@
...
+end)
+
+net.Receive("ttt2_sync_armor_max", function()
+    local client = LocalPlayer()
+
+    -- prevent error from netmessage prior to the client beeing ready
+    if not IsValid(client) then return end
+
+    client.armor_max = net.ReadUInt(16)
+end)
```
Alf21: missing EOL

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: you're right. It is not checked here. However `self:AddEquipmentItem()` checks if the item is nil

lua/autorun/ttt2_force_download.lua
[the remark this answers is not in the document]
saibotk: Wether or not, this should be moved out of this PR

lua/autorun/ttt2_force_download.lua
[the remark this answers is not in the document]
saibotk: then we should change this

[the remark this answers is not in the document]
TimGoll: why though? there's only a performance difference on initial material creation and we have already about 30 icons in TTT2 and heroes.

gamemodes/terrortown/gamemode/shared/hud_elements/tttsidebar/pure_skin_sidebar.lua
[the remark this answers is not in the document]
saibotk: This also belongs to your status system etc this should be done in a separate PR

lua/autorun/ttt2_force_download.lua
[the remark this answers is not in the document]
TimGoll: Since I did not change anything besides adding a line to the forcedownload list, I see no point in removing it here and adding it manually to the master.

Yes, it does not belong in here, but adding two commits for one line of code seems pointless to me

gamemodes/terrortown/gamemode/server/sv_armor.lua
```diff
@@ -0,0 +1,146 @@
...
+	if armor == 0 then return end
+
+	self.cv.armor_factor = self.cv.armor_damage_block_pct:GetFloat()
+	self.cv.health_factor = self.cv.armor_damage_health_pct:GetFloat()
+	
+	if ply:ArmorIsReinforced() then
+		self.cv.health_factor = self.cv.health_factor - 0.15
+	end
+
+	armor = armor - self.cv.armor_factor * damage
```
saibotk: This will still allow negative damage to increase the current armor.

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1430,14 +1430,21 @@ function GM:PlayerTakeDamage(ent, infl, att, amount, dmginfo)
...
 		-- scale everything to karma damage factor except the knife, because it
 		-- assumes a kill
 		if not dmginfo:IsDamageType(DMG_SLASH) then
 			dmginfo:ScaleDamage(att:GetDamageFactor())
 		end
+	end
+
+	-- before the karma is calculated, but after all other damage hooks / damage change is processed,
+	-- the armor system should come into place (GM functions are called last)
+	ARMOR:HandlePlayerTakeDamage(ent, infl, att, amount, dmginfo)
```
saibotk: This should not be excluded from the upper if, instead it would be good, to include it, since the checks are useful for the armor handling too and it would remove the duplicate if.

lua/terrortown/entities/items/item_ttt_armor.lua
saibotk: That is not a valid solution, since it is possible to have many sources, where additional armor comes from, this would remove armor points, without considering the armor that is left, from this item. (eg. get 2 armor items +100, get shot -10, remove one -100 and you got less than you should

gamemodes/terrortown/gamemode/server/sv_armor.lua
```diff
@@ -0,0 +1,146 @@
...
+    net.WriteUInt(math.Round(self.armor), 16)
+    net.WriteBool(self.armor_is_reinforced)
+    net.Send(self)
+end
+
+---
+-- Sets the max armor to a specific value
+-- @param number armor_max the new max armor to be set
+-- @realm server
+function plymeta:SetMaxArmor(armor_max)
```
saibotk: This will not change the current armor accordingly

gamemodes/terrortown/gamemode/server/sv_armor.lua
```diff
@@ -0,0 +1,146 @@
...
+end)
+
+
+-- SERVERSIDE ARMOR FUNCTIONS
+---
+-- Sets the armor to a specific value
+-- @param number armor the new armor to be set
+-- @realm server
+function plymeta:SetArmor(armor)
+    self.armor = armor
```
saibotk: This does not consider max armor

gamemodes/terrortown/gamemode/shared/hud_elements/tttsidebar/pure_skin_sidebar.lua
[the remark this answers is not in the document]
TimGoll: no, this was done to make the armor system work. Without this, the armor system does not work

gamemodes/terrortown/gamemode/server/sv_armor.lua
TimGoll: like I said previously: This system is build the same as the health system, therefore MaxArmor is only used for HUDs

gamemodes/terrortown/gamemode/server/sv_armor.lua
TimGoll: MaxArmor is only used for HUDs

TimGoll:
From my point of view, the AS is finished

gamemodes/terrortown/gamemode/server/sv_armor.lua
```diff
@@ -15,77 +15,120 @@ util.AddNetworkString("ttt2_sync_armor_max")
...
 ---
--- Sets the armor to a specific value
+-- Sets the @{ARMOR} to a specific value that is capped by the max armor value
 -- @param number armor the new armor to be set
 -- @realm server
 function plymeta:SetArmor(armor)
-    self.armor = armor
-    self.armor_is_reinforced = ARMOR.cv.armor_is_reinforced_enabled:GetBool() and self:Armor() > ARMOR.cv.armor_threshold_for_reinforced:GetInt()
+	self.armor = math.Clamp(armor, 0, self:GetMaxArmor())
+	self.armor_is_reinforced = ARMOR.cv.armor_is_reinforced_enabled:GetBool() and self:GetArmor() > ARMOR.cv.armor_threshold_for_reinforced:GetInt()
```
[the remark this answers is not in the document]
TimGoll: I had this in a function. But Alf or you requested a change to this^^

gamemodes/terrortown/gamemode/server/sv_armor.lua
```diff
@@ -15,77 +15,120 @@ util.AddNetworkString("ttt2_sync_armor_max")
...
 
 ---
--- Increases the armor about a specific value
--- @param number increaseby the amount to be increased
+-- Sets the @{ARMOR} to a specific value while also increasing (if necessary) the max value.
+-- @param number armor the new armor to be set
 -- @realm server
-function plymeta:IncreaseArmor(increaseby)
-    self:SetArmor(self:Armor() + math.max(increaseby, 0))
+function plymeta:SetArmorUnbound(armor)
```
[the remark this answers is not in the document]
TimGoll: This is basically an internal function. It shouldn't be called from outside. I could call it `SetRawArmor` though

gamemodes/terrortown/gamemode/client/cl_armor.lua
```diff
@@ -0,0 +1,60 @@
...
+		icon_id = ply:ArmorIsReinforced() and 2 or 1
+	end
+
+	-- added armor
+	if not STATUS:Active("ttt_weapon_armor") and ply.armor > 0 then
+		STATUS:AddStatus("ttt_weapon_armor", icon_id)
+	end
+
+	-- normal armor level change
+	if STATUS:Active("ttt_weapon_armor") then
```
Alf21: Why this needs to be checked again?

gamemodes/terrortown/gamemode/server/sv_armor.lua
```diff
@@ -0,0 +1,184 @@
...
+---
+-- Sets the initial @{Player} Armor, called in @{GM:TTTBeginRound} after (dumb) players
+-- are respawned. Therefore no armor has to be reset and it is safe that the player
+-- receives their armor.
+-- @realm server
+-- @internal
+function ARMOR:InitPlayerArmor()
+	for _, p in ipairs(player.GetAll()) do
+		if p:IsTerror() then
+			p:GiveArmor(self.cv.armor_on_spawn:GetInt())
```
Alf21: GiveArmor, not SetArmor?

gamemodes/terrortown/gamemode/server/sv_armor.lua
```diff
@@ -0,0 +1,184 @@
...
+	self.cv.armor_factor = self.cv.armor_damage_block_pct:GetFloat()
+	self.cv.health_factor = self.cv.armor_damage_health_pct:GetFloat()
+
+	if ply:ArmorIsReinforced() then
+		self.cv.health_factor = self.cv.health_factor - 0.15
+	end
+
+	ply:DecreaseArmorValue(self.cv.armor_factor * damage)
+
+	local new_damage = self.cv.health_factor * damage - math.min(armor, 0)
```
Alf21: `- math.min(armor, 0)` is 0 I think...

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
Alf21: We should give a look again to this

gamemodes/terrortown/gamemode/server/sv_armor.lua
TimGoll: yes, since GiveArmor also handles the increase of MaxArmor, SetArmor does not increase the bound and is therefore capped at maxarmor

gamemodes/terrortown/gamemode/client/cl_armor.lua
TimGoll: It is checked to make sure the icon is only changed when there should be a change. Without this, the function for the sidebar is called on every armorchange. This should be better performancewise

Alf21: better performance would be to break in the first lines if this is the intention instead of doing useless checks...

gamemodes/terrortown/gamemode/server/sv_armor.lua
Alf21: ?

gamemodes/terrortown/gamemode/client/cl_armor.lua
TimGoll: One could add a return there, yes. But SetStatus is still heavier than two checks. Additionally you can see, that I have three scenarios, Add, Remove and Change.
One could always call Remove and Add, but this is bad in my opinion. Therefore these checks, basically as "edge detection" of a signal change.

gamemodes/terrortown/gamemode/server/sv_armor.lua
TimGoll: see the latest commit, I addressed this issue and added a short explanation in a comment.

PR #372, opened 2019-08-15
GetAvoidDetective compatibility

Histalek:
Damagelog depends on this function, other ttt addons might as well. Therefore i thought fixing the problem with damagelogs in the ttt2 project is the better approach.
Also feel free to change my documentation :)


Alf21:
thanks, this would be the best solution

PR #374, opened 2019-08-31
introduced new hooks

TimGoll:
Right now there exists one problem in TTT2. If you want to give a player a rolespecific loadout, you have to rely on the RoleChange hook. But if you want to give the same loadout on respawn (e.g if the player was revived) you have to use the PlayerSpawn Hook. There exists the problem that this hook is called on every spawn and a player would get the loadout of the role he was in the previous round on some ocasions. There are workarounds to this, but it is not intuitive.

Right now there exist quite a few loadout specific hooks, but none of them is particular useful in this scenario. Therefore let me introduce these two new hooks: `TTT2PlayerRoleInit` and `TTT2PlayerRoleDeinit`. Why two of them? At first I had this functionality in one hook. But this caused the problem that two roles, that give the same item, could create raceconditions if the hook of the old role is called after the hook of the new role.

The new hook is called on roleChange, teamChange, death and spawn and can be easily used like this:

```lua

hook.Add('TTT2PlayerRoleInit', 'TTT2MarkerInit', function(ply, new_role, new_team)
	if new_role ~= ROLE_MARKER then return end
	
	InitRoleMarker(ply)
end)

hook.Add('TTT2PlayerRoleDeinit', 'TTT2MarkerDeinit', function(ply, old_role, old_team)
	if old_role ~= ROLE_MARKER then return end

	DeinitRoleMarker(ply)
	MARKER_DATA:MarkerDied()
end)

```

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_player.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player.lua
@@ -128,7 +128,7 @@ function GM:PlayerSpawn(ply)
 	-- a hook to handle the rolespecific stuff that should be done on
 	-- rolechange and respawn (while a round is active)
 	if ply:IsActive() then -- round is active and player is terror player
-		hook.Run("TTT2PlayerInitRole", ply, false, ply:GetSubRole(), ply:GetTeam())
+		hook.Run("TTT2GiveRoleLoadout", ply, false, ply:GetSubRole(), ply:GetTeam())
 	end
 end
 
@@ -909,7 +909,7 @@ function GM:PlayerDeath(victim, infl, attacker)
 			-- a hook to handle the rolespecific stuff that should be done on death or rolechange
 			-- this hook is called prior to setting the player team to spectator!
 			if victim:IsActive() then -- round is active and player was terror player
-				hook.Run("TTT2PlayerDeinitRole", victim, victim:GetSubRole(), victim:GetTeam())
+				hook.Run("TTT2RemoveRoleLoadout", victim, false, victim:GetSubRole(), victim:GetTeam())
 			end
 			
 			victim:SetTeam(TEAM_SPEC)
--- a/gamemodes/terrortown/gamemode/server/sv_player.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player.lua
@@ -128,7 +128,7 @@ function GM:PlayerSpawn(ply)
 	-- a hook to handle the rolespecific stuff that should be done on
 	-- rolechange and respawn (while a round is active)
 	if ply:IsActive() then -- round is active and player is terror player
-		hook.Run("TTT2PlayerRoleInit", ply, ply:GetSubRole(), ply:GetTeam())
+		hook.Run("TTT2PlayerInitRole", ply, false, ply:GetSubRole(), ply:GetTeam())
 	end
 end
 
@@ -906,6 +906,12 @@ function GM:PlayerDeath(victim, infl, attacker)
 
 	timer.Simple(0, function()
 		if IsValid(victim) then
+			-- a hook to handle the rolespecific stuff that should be done on death or rolechange
+			-- this hook is called prior to setting the player team to spectator!
+			if victim:IsActive() then -- round is active and player was terror player
+				hook.Run("TTT2PlayerDeinitRole", victim, victim:GetSubRole(), victim:GetTeam())
+			end
+			
 			victim:SetTeam(TEAM_SPEC)
 			victim:Freeze(false)
 			victim:SetRagdollSpec(true)
@@ -929,12 +935,6 @@ function GM:PlayerDeath(victim, infl, attacker)
 			end
 
 			hook.Run("TTT2PostPlayerDeath", victim, infl, attacker)
-		end
-
-		-- a hook to handle the rolespecific stuff that should be done on
-		-- rolechange and respawn (while a round is active)
-		if GetRoundState() == ROUND_ACTIVE then -- round is active and player is terror player
-			hook.Run("TTT2PlayerRoleDeinit", victim, victim:GetSubRole(), victim:GetTeam())
 		end
 	end)
 end
--- a/gamemodes/terrortown/gamemode/server/sv_player.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player.lua
@@ -124,6 +124,12 @@ function GM:PlayerSpawn(ply)
 	ply:SetupHands()
 
 	SCORE:HandleSpawn(ply)
+
+	-- a hook to handle the rolespecific stuff that should be done on
+	-- rolechange and respawn (while a round is active)
+	if ply:IsActive() then -- round is active and player is terror player
+		hook.Run("TTT2PlayerRoleInit", ply, ply:GetSubRole(), ply:GetTeam())
+	end
 end
 
 ---
@@ -923,6 +929,12 @@ function GM:PlayerDeath(victim, infl, attacker)
 			end
 
 			hook.Run("TTT2PostPlayerDeath", victim, infl, attacker)
+		end
+
+		-- a hook to handle the rolespecific stuff that should be done on
+		-- rolechange and respawn (while a round is active)
+		if GetRoundState() == ROUND_ACTIVE then -- round is active and player is terror player
+			hook.Run("TTT2PlayerRoleDeinit", victim, victim:GetSubRole(), victim:GetTeam())
 		end
 	end)
 end
--- a/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
@@ -140,9 +140,9 @@ function plymeta:SetRole(subrole, team, forceHooks)
 
 	-- a hook to handle the rolespecific stuff that should be done on
 	-- rolechange and respawn (while a round is active)
-	if forceHooks or oldSubrole ~= newSubrole or oldTeam ~= newTeam then
-		hook.Run("TTT2PlayerRoleDeinit", self, oldSubrole, oldTeam)
-		hook.Run("TTT2PlayerRoleInit", self, newSubrole, newTeam)
+	if SERVER then
+		hook.Run("TTT2PlayerDeinitRole", self, oldSubrole, oldTeam)
+		hook.Run("TTT2PlayerInitRole", self, true, newSubrole, newTeam)
 	end
 end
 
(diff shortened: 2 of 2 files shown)


Alf21:
This is too inconsistent. Maybe think about using another additional hook for other cases. Dying and changing the role isn‘t the same. Maybe InitialRoleInit, but InitRole and Deinit Role already can be handled with existing RoleChanged hooks

lebroomer:
Maybe we should think about two different types of Loadout, which is the main use case of these hooks: 
1. Loadout, which is only handed out on RoleChange (Sidekick Deagle)
2. Loadout, which is handed out on every Spawn.

Maybe this will tackle the problem more precisely.

TimGoll:
Or, building on top of LeBroomers idea: What about a variable that is passed in the hook, like `isRespawned` to check if the player changed its role or respawned. Otherwise we would need 4 hooks instead of 2

TimGoll:
But @Alf21 I get where you come from. I have only thought about roles that get their equipment always (like the marker)

Alf21:
This doesn't solve the inconsistency. You should think about using different hooks for different issues. This does not touch the performance in any way (just a registration in the hooks table). And this would support the style of already existing hooks like PlayerInitialSpawn. Using vars mixes things up, makes it more complicated and seems to be a duplication of already existing RoleChanged hooks (more or less)

TimGoll:
But using multiple hooks makes it more complicated than it needs to be. Since - at least in my opinion - it makes no difference if you receive your loadout on rolechange or respawn.

If you really prefer splitting things up, I propose something like this, which makes the implementation in the roles file a bit more bloated:

```lua
hook.Add('TTT2PlayerRoleInitiaSpawn', 'TTT2MarkerInit', function(ply, new_role, new_team)
	if new_role ~= ROLE_MARKER then return end
	
	InitRoleMarker(ply)
end)

hook.Add('TTT2PlayerRoleRespawn', 'TTT2MarkerInit_respawn', function(ply, new_role, new_team)
	if new_role ~= ROLE_MARKER then return end
	
	InitRoleMarker(ply)
end)

hook.Add('TTT2PlayerRoleDeath', 'TTT2MarkerDeinit', function(ply, old_role, old_team)
	if old_role ~= ROLE_MARKER then return end

	DeinitRoleMarker(ply)
	MARKER_DATA:MarkerDied()
end)

hook.Add('TTT2PlayerRoleChange', 'TTT2MarkerDeinit_rc', function(ply, old_role, old_team)
	if old_role ~= ROLE_MARKER then return end

	DeinitRoleMarker(ply)
	MARKER_DATA:MarkerDied()
end)
```

And when doing it like this, I see no benefit over the current solution since I have to handle multiple hooks to solve one simple problem: Giving a rolespecific player loadout.

Alf21:
As you can see, it's currently possible to create a specific implementation without new inconsistent hooks. Maybe using a ROLE table var (like in TTTC) can solve this problem easily.

TimGoll:
Yes it is possible. Introducing a roletable will add only more checks and makes it even more complicated. It was also possible to create new roles for TTT without the role module. Also why is this inconsistent? The old way is inconsistent too. Additionally it creates code duplication.

But if you insist on keeping it the old way, I'll close this.

TimGoll:
Since Alf and I seem to find no solution on this problem, I'll open this to others. What do you think about this change?

I'm confused that the item race condition seems to bother noone. An other example: Right now in 50% of the cases this happens when changing from a marker to a jackal:
if the marker spawns with 60 armor and looses 15 and becomes a jackal with 30 armor, he spawns with only 15. This happens because the current hooks are not ordered by give/remove, but by role. So sometimes you add your new armor before removing the old (45 + 30 = 75, 75 - 60 = 15 instead of 45 - 60 = 0, 0 + 30 = 30).
This can happen with other items and weapons too. Right now it happens quite often (in 50% of the cases) that an error is printed in the chat, when a player switches from a role with radar to another role with radar, since it tries to add a second radar to the already existing one. You then end up with no radar at all. This is right now a synthetic error, since there are no rolechanges between two roles with radar, but it is an existing bug that I fixed with my pullrequest.

Alf21:
wait, how did you fix this bug with adding just some hooks? o.O

TimGoll:
Because the whole point of this is to have one hook to remove stuff and one hook to give stuff, while the remove hook is called first. Not one hook that does both.

If it does both, role A could remove their equipment after role B added theirs. This can lead to race conditions. Therefore this is no hook duplication.

Alf21:
Then it's possible to add a post hook. That's a general problem of the gmod hooking lib. It's already discussed on the gmod fp repo (possibility of prioritizing function and modify the hook order), but that seems to be tricky. I understand your point, but that doesn't change something related to the already discussed inconsistency. The solution should be something related to @lebroomer arguments and the use of already existing hooks. Maybe adding a pre and post hook. Or maybe using an additional var in these hooks as you already mentioned.

TimGoll:
but then we'd need a PostPlayerSpawn, a PostRoleChange and a PostTeamChange hook. This makes six hooks for such a simple thing as giving loadout. I'd be fine with this if you really prefer it. But I think it is way more convenient to have two hooks with a var that defines if it is a rolechange or respawn.

And yes I know about the hook order thing, but I don't think it will be implemented... ever.

Alf21:
What about creating just a PlayerInitRole hook (+ initialSpawn argument) for adding stuff and using the default PlayerChangeRole and PlayerDeath hooks for removing stuff?

TimGoll:
While I still prefer my solution, this would solve this as well

TimGoll:
After thinking about this: No, I still not like your proposal because with your proposal, we'd still have the item racecondition and therefore gained nothing at all. I can now use the playerspawn hook and can check if the round is active to achieve your proposed hook. This helps nothing.

TimGoll:
Another thing I noticed why another hook is needed:
You can't give items in the player spawn hook without using a timer, since the player is not yet completely spawned.

Right now you have to do it like this:
```lua
-- make sure that the supervillain always receives his radar
hook.Add("PlayerSpawn", "RespawnSupervillain", function(ply) -- called on player (re-)spawn
	-- this is an ugly workaround, since on calling of the player respawn hook, the player can not yet receive items
	timer.Simple(0.1, function()
		if GetRoundState() ~= ROUND_ACTIVE then return end
		if ply:GetSubRole() ~= ROLE_SUPERVILLAIN then return end

		ply:GiveEquipmentWeapon("weapon_ttt_crystalknife")
		ply:GiveEquipmentItem("item_ttt_radar")
	end)
end)
```

Alf21:
You can remove the items / weapons of a role in PlayerDeath or PlayerChangeRole hook and add new weapons in PlayerInitRole(respawned) hook as I already mentioned. This solved item stacking and removing issues and maybe could be a workaround for what you mentioned. But afaik, an item can be given all the time because it doesn't require a spawn of a player. Giving a weapon shouldn't be a problem too, because if the weapon is not given, it will be placed into a loop that retries to give a player the weapon until it works...

Alf21:
and placing the PlayerInitRole hook into PlayerSpawn (delaying PlayerSpawn hook) is a bad idea too, because at this time, the role isn't initialized (it's too late and doesn't match the sense of the hook's name). This should be related to a loadout called hook

TimGoll:
It doesn't work witout a delay. Weapons do work, items do not work. The hook is called prior to the player being ready. The loop seems to work with weapons, but not with items. Since I did not receive any items on player respawn withot this timer. You can try it yourself if you don't believe me.

But we're running in circles here. Your proposal still does not fix the item racecondition. If you do not like my solution, please let me know about a different way of giving a loadout without having this racecondition.

The hook can be renamed. And this edge case can be detected too.

Alf21:
Then we should fix the items bug instead of using timers in this ugly way...

TimGoll:
I know that timers are an ugly solution, but they are the only way to get this to work (without hunting through the item code). Hence the comment "this is an ugly workaround"

Alf21:
They aren‘t the only way, they are the easiest way.

TimGoll:
Of course. But that's not the point. I added this as a temporary fix since dok has reported multiple issues with the loadout that I wanted to fix for weeks.

I skipped through the code and acutally ... I do not understand why the player doesn't receive the loadout. 
My only guess is that the playerloadout is removed on playerspawn, not on death and therefore not added, since the player already has this item.

TimGoll:
The hook is now finished and looks like this:

```lua
hook.Add('TTT2GiveRoleLoadout', 'ttt2_marker_loadout_give', function(ply, isRoleChange, role, team)
	if role ~= ROLE_MARKER then return end

	InitRoleMarker(ply)
end)

hook.Add('TTT2RemoveRoleLoadout', 'ttt2_marker_loadout_remove', function(ply, isRoleChange, role, team)
	if role ~= ROLE_MARKER then return end

	DeinitRoleMarker(ply)
end)
```

This hook is called on rolechange, death and respawn and prevents item raceconditions. There currently still exists the problem, that this hook will not be called when the round begins and the player is innocent, because all players receive the role innocent on initial spawn. To fix this, one would have to change stuff in the roleselect code. This should not be part of this PR.

Alf21:
I think `TTT2GiveRoleLoadout` is called two times if a player died and respawns with a new role?

TimGoll:
yes it is. is this a problem?

TimGoll:
This might be possible with functions, I don't know for sure.

There are two main usecases which I explained at least twice in this thread so far, but here you go:
- introducing a simple way of giving a player loadout without the use of multiple hooks and checks (as it is right now), with support for loadout on respawn and/or rolechange
- eliminating the item racecondition by always calling remove first and give afterwards

Point one would be somewhat solvable with the already existing loadout hook, but you'd still have to listen to at least the rolechange hook, if not even more in some cases. I already wanted to close this PR because I found this hook. But Alf then reminded me that this hook still not solves the item racecondition. Therefore I modified the hooks slightly to his liking yesterday. As shown in the following example, you can now give equipment really easy on rolechange and respawn two specific roles without using multiple hooks, while also making sure that you receive the loadout and do not end up with no radar at all, since the remove function was called after the give function (this happens with the rolechange hook, another solution would be a PostRoleChange hook)

```lua
hook.Add('TTT2GiveRoleLoadout', 'ttt2_marker_loadout_give', function(ply, isRoleChange, role, team)
	if role ~= ROLE_MARKER then return end

	InitRoleMarker(ply)
end)

hook.Add('TTT2RemoveRoleLoadout', 'ttt2_marker_loadout_remove', function(ply, isRoleChange, role, team)
	if role ~= ROLE_MARKER then return end

	DeinitRoleMarker(ply)
end)
```

TimGoll:
closing this in favor of #378

PR #375, opened 2019-09-01
sprint modifier hook

TimGoll:
A really simple change to remove the static player sprint multiplier in favor of a hook. Right now the sprint multiplier ist set with  `ply.sprintMultiplierModifier`, this hook allows to do the same (but stackable) with a hook:

```lua
hook.Add("TTT2PlayerSprintMultiplier", "sprintchange", function(ply, multi)
	multi[1] = 10 * multi[1]
end)
```

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/sh_main.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_main.lua
@@ -73,10 +73,11 @@ function GM:Move(ply, mv)
 		local mul = hook.Call("TTTPlayerSpeedModifier", GAMEMODE, ply, slowed, mv, noLag) or 1
 		mul = basemul * mul * noLag[1]
 
-		local sprintMultiplierModifier = {1}
-		hook.Run("TTT2PlayerSprintMultiplier", ply, sprintMultiplierModifier)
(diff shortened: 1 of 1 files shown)


Alf21:
You can stack the ply.sprintMultiplierModifier too with multiplication. We could include this hook for consistency, but I do not like to have for each thing a new hook. Hooks in the Move hook are expensive to call. It would be better to use the sprintMultiplierModifier var...

TimGoll:
This is a fair thought that I had not thought of. I just looked at the heroes code (switft) and did not like the current implementation too much.
The new one has its drawbacks too. Hmm

Alf21:
We should use this way for consistency (ref.: https://github.com/TTT-2/TTT2/blob/badf11a98348e45f777d9dcee08e5c2b0b6e49d0/gamemodes/terrortown/gamemode/shared/sh_sprint.lua).

TimGoll:
You mean the two hooks in the bottom? I tried to use the same style

TimGoll:
Ah yes, nice. That's a simple and good performance improvements. Haven't thought of this!

PR #380, opened 2019-10-07
Role loading

saibotk:
This is a first approach at cleaning up the ROLE system.

The change is not that big, but should make it more clear for other developers, on how to use the 
`ROLE:PreInitialize()` and `ROLE:Initialize()` functions.  
They are used to clean up the messy hook usage that has since been used to eg. set the `BaseRole` or set the `LANG` entries.

But this mainly addresses the issue with roles that are loaded before other roles, which they reference with their global variables to eg. set the fallback Shop or the `defaultTeam` variable.
Now the loading is ordered in a way, that the global variables of all roles are initialized, then the `ROLE:PreInitialize()` function is called on each role and after setting up the internal default ConVars, etc for each role (see SetupData(v) in the roles module) the `ROLE:Initialize()` function is called on each role.

@Alf21 i did not know how to use the documentation symbols correctly, so i will need some input of you on how to make this approvable.

The code is tested and is compatible with the existing roles (non-breaking).

General feedback is welcome.

### DIFF
--- a/lua/includes/modules/roles.lua
+++ b/lua/includes/modules/roles.lua
@@ -484,3 +484,28 @@ function GetSortedRoles()
 
 	return rls
 end
+
+---
+-- Connects a SubRole with its BaseRole.
+-- @notice This will also set the defaultTeam variable to the BaseRole's team.
+-- @param ROLE roleTable the role table (of the SubRole)
+-- @param ROLE baserole the BaseRole
+-- @realm shared
+function SetBaseRole(roleTable, baserole)
+	if roleTable.baserole then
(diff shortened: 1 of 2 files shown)


Alf21:
Nice :)

@deprecated comment can be used to mark deprecated functions
@hook can be used to mark functions that can should get overwritten, e.g. ROLE:Initialize

Alf21:
Do you know whether it fixes the faulty shop init bug ?

TimGoll:
@Alf21 afaik this PR was born on the fix of said bug, so yes

saibotk:
Yes it should fix the shop init bugs i tested it and the roles got their shops after the fix

PR #381, opened 2019-10-09
add weapon switch without dropping

TimGoll:
With this pullrequest players are able to switch their weapons without dropping them. By default a player has to press E to switch a highlighted weaon.

This code is based on code by @TheNickSkater 

image


Alf21:
There are some lines that are not a part of this PR theoretically. As small changes, they are okay. But think about the last PR, this should be avoided.

I'm going to improve this code tomorrow and fix few bugs. Currently, spectators should be able to access these weapons theoretically. There are some performance leaks too.

Alf21:
Is it tested with custom settings of the weaponslot system. Afaik there is the possibility of increasing the weapons per slot. Maybe there is a bug if there are some more slots available and you wanna pickup a weapon manually (or need to do so).

TimGoll:
uhm no I haven't tested this scenario. Will do

TimGoll:
If you have two primaries, the primary selected will be dropped

TimGoll:
Good changes. What about using the clients role color for the halo?

Alf21:
I think that would be a nice solution because it's done the same way with the crosshair

Alf21:
Garry's Mod 15 10 2019 21_29_40

gamemodes/terrortown/gamemode/server/sv_weapon_pickup.lua
```diff
@@ -0,0 +1,32 @@
+util.AddNetworkString("ttt2_switch_weapon")
+
+net.Receive("ttt2_switch_weapon", function(_, ply)
+	-- player and wepaon must be valid
+	if not IsValid(ply) or not ply:IsTerror() or not ply:Alive() then return end
+
+	-- do not allow weapon switch when the weapon switch was started in the previous round
+	local roundstate_client = net.ReadString()
+	if roundstate_client == ROUND_POST and roundstate_client ~= GetRoundState() then return end
```
saibotk: Is there a reason for this being sent by the client? Why do we not use the serverside roundstate?

gamemodes/terrortown/gamemode/client/cl_weapon_pickup.lua
```diff
@@ -0,0 +1,33 @@
+local cv_draw_halo = CreateClientConVar("ttt_weapon_switch_draw_halo", "1", true, false)
+local defaultCol = Color(250, 210, 210)
+
+bind.Register("ttt2_weaponswitch", function()
+	net.Start("ttt2_switch_weapon")
+	net.WriteString(GetRoundState())
+	net.SendToServer()
```
lebroomer: This should be encapsulated in a local function for better readibiity

Alf21:
> 
> 
> I think there should be a hint in the HUD, what key to press to switch weapons together with maybe the weapon name (just like the weapon switch in CS:GO). Otherwise older TTT players might not get this new swap interaction.
> And another thing: Shouldn't swapping be possible too if you don't have the correct weapon in hand but in inventory? Maybe for umabigous cases only (primary, secondary, grenade)

Nice idea. And yea, I thought the same :)

TimGoll:
This system was originally planned together with my draft for the overlay update. This overlay updated shows which button to press.
We can delay it until we've come around doing it. Or add a small note on the display for now as a placeholder.

TimGoll:
@lebroomer "I just thought a little bit more about the binding with our TTT2 bind system and I don't really like it. I think you should query the currently bound use-key with input.LookupBinding( "+use" ) and then always use this key instead of another binding possibility."

Are you talking about the default key here? Because I don't think we should force users to use the use key. Using the use-key as default key seems like a good idea

TimGoll:
I think everything should be addressed by now. Could you please rereview? Thanks!

Alf21:
round state syncing didn't change...

TimGoll:
If you propose a solution without the roundstate sync, I'll be happy to implement it

Alf21:
@lebroomer do you have a possibility to throw the last picked-up weapon based on your inventory system?

lebroomer:
> @lebroomer do you have a possibility to throw the last picked-up weapon based on your inventory system?

No the last picked-up weapon is not saved anywhere at the moment. What do you want achieve with that?

Alf21:
> 
> 
> > @lebroomer do you have a possibility to throw the last picked-up weapon based on your inventory system?
> 
> No the last picked-up weapon is not saved anywhere at the moment. What do you want achieve with that?

instead of needing to switch to a certain weapon in order to be able to switch with the aimed weapon, we wanna switch with the weapon that is in the current slot of the aimed weapon. The issue occurs if using your inventory system because a loop through the normal weapons will lead to just drop the first weapon in this loop (of this type)

lebroomer:
> > > @lebroomer do you have a possibility to throw the last picked-up weapon based on your inventory system?
> > 
> > 
> > No the last picked-up weapon is not saved anywhere at the moment. What do you want achieve with that?
> 
> instead of needing to switch to a certain weapon in order to be able to switch with the aimed weapon, we wanna switch with the weapon that is in the current slot of the aimed weapon. The issue occurs if using your inventory system because a loop through the normal weapons will lead to just drop the first weapon in this loop (of this type)

But then you rather need the last picked-up weapon per slot right? Another possibility for that case would be to disable switching together with a hint on the HUDOverlay. I'm thinking of a situation where you want to pick up a weapon and accidentally drop a T item revealing yourself to bystanders.

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -69,7 +69,7 @@ else
 	end,
 	function()
 		PlayerSprint(false)
-	end, "TTT2 Bindings", "TTT2 Sprint", KEY_LSHIFT)
+	end, "TTT2 Bindings", "f1_bind_sprint", KEY_LSHIFT)
```
lebroomer: Why is that part of the PR?

gamemodes/terrortown/gamemode/shared/lang/german.lua
```diff
@@ -1164,6 +1165,9 @@ L.f1_bind_description = "Klicke und drücke einen Knopf um ihn zu setzen."
 L.f1_bind_reset_default_description = "Auf Standardwert zurücksetzen."
 L.f1_bind_disable_description = "Deaktiviere diese Bindung."
 
+L.f1_bind_weaponswitch = "Waffe Wechseln"
+L.f1_bind_sprint = "Sprinten"
```
lebroomer: See above

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
Alf21: already clarified

gamemodes/terrortown/gamemode/shared/lang/german.lua
Alf21: "There are some lines that are not a part of this PR theoretically. As small changes, they are okay. But think about the last PR, this should be avoided." (first comment)

Alf21:
Yea but there is no other possibility. If you wanna switch with the old system, you need to switch to the T weapon too.

lebroomer:
> Yea but there is no other possibility. If you wanna switch with the old system, you need to switch to the T weapon too.

Since the weapons on the ground doesn't always tell you directly about their slot (talking about players perspective) its kind of hard to let them drop a T weapon when they spam "E" on it to just pick it up. This could be "fixed" again with a hint in the HUD overlay telling them which weapon they will drop to pick up the new weapon. With that hint i would be fine.

Alf21:
> Since the weapons on the ground doesn't always tell you directly about their slot (talking about players perspective) its kind of hard to let them drop a T weapon when they spam "E" on it to just pick it up. This could be "fixed" again with a hint in the HUD overlay telling them which weapon they will drop to pick up the new weapon. With that hint i would be fine.

Alright, but I won't implement this feature.

TimGoll:
The Hud stuff is planned. One of my next PRs actually. But I don't think that this PR has to be perfect right now
One could add a temporary print on the display though

TimGoll:
About the hud stuff: One could Display the slot, the droped weapon, etc

Alf21:
Feel free to modify, I'm happy with the current state :)

Alf21:
@lebroomer the hud thing should be a thing of another PR or commit imo...

PR #386, opened 2019-10-26
added a native event popup to TTT2

TimGoll:
This is a really simple system that introduces `EPOP` - an event popup. It does not feature a stack, but always displays the most recent message. It could be used for role changes, randomat events and other things.

It features a title, a text body and icon support. Only the first one is needed to create a popup, the other two are optional. Additionally there is a display time, which is 4 seconds by default.

image
An older picture of a popup with title, text and icons, the red border is now fixed though.

How to use:

```lua
EPOP:AddMessage("test", "test text", 10)
```
Creating a simple popup is done by calling this function. It is client only.

This is all that has to be done to create a popup. However popups can be removed by calling `EPOP:RemoveMessage()`


gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
```diff
@@ -0,0 +1,40 @@
...
+
+	return true
+end
+
+---
+-- Instantly removed the currently displayed @{EPOP}
+-- @realm client
+function EPOP:RemoveMessage()
+	self.msg = nil
+end
```
Alf21: missing EOL

gamemodes/terrortown/gamemode/client/cl_fonts.lua
```diff
@@ -146,3 +149,68 @@ function draw.AdvancedText(text, font, x, y, color, xalign, yalign, shadow, scal
...
 end
+
+---
+-- Returns a list of lines to wrap the text matching the given width
+-- @param string text
+-- @param number width
+-- @param string font
+-- @return table
+-- @realm client
+function WrapText(text, width, font)
```
Alf21: needs a performance rework

gamemodes/terrortown/gamemode/shared/hud_elements/ttteventpopup/pure_skin_eventpopup.lua
```diff
@@ -0,0 +1,165 @@
+local base = "pure_skin_element"
```
Alf21: needs a performance rework

PR #389, opened 2019-10-27
Reworked TargetID

TimGoll:
This is a complete rework of the TTT targetID. It features a few things:

- Codestructure reworked. It is now actually readble and not a single block with 500 lines

- New player targetID:
  image
  Shows roleicon (when role is known), shows basic TTT information, supports scroreboard hints

- Player targetID without any extra information:
  image

- New ragdoll targetID:
  image
  Features a new ragdoll icon, shows basic TTT information

- Supports weapons:
  image
  Weapons also feature the ourline which now got moved into targetID

- Supports external addons:
  image
  External Addons can use it too without any problems, Here's an example of super soda using the system

- Expandable:
  - External Addons can register the hook and add their information to existing targetIDs (player, ...) or add new ones for their own entities
  - the description are can contain as many elements as the addon dev wants to add

**How to use:**
```lua
-- handle looking at sodas
hook.Add('TTTRenderEntityInfo', 'ttt2_supersoda_highlight_sodas', function(data, params)
    local client = LocalPlayer()
    local soda_class = data.ent:GetClass()

    if data.distance > 100 then return end
    if not table.HasValue(SUPERSODA.sodas, soda_class) then return end -- no valid soda
    if not IsValid(client) or not client:IsTerror() or not client:Alive() then return end

    params.drawInfo = true
    params.displayInfo.key = input.GetKeyCode(input.LookupBinding('+use'))
    params.displayInfo.title.text = LANG.GetTranslation(soda_class)
    params.displayInfo.subtitle.text = LANG.GetTranslation('ttt_pickup_soda')

    params.displayInfo.desc = {
        {text = LANG.GetTranslation('ttt_pickup_' .. soda_class), color = COLOR_WHITE}
    }

    params.drawOutline = true
    params.outlineColor = client:GetRoleColor()
end)
```
This simple example of the SuperSoda addon shows how to use the new targetID system.

In my tests no performance difference was noticed. Even disabling the complete targetID produces the same framerate for me.


TimGoll:
I changed the rendering of the text a bit, it looks now better

gamemodes/terrortown/gamemode/client/cl_fonts.lua
```diff
@@ -119,30 +120,43 @@ function draw.AdvancedText(text, font, x, y, color, xalign, yalign, shadow, scal
...
+-- @param string text The text that the length should be calculated
+-- @param [default="DefaultBold"] string font The font ID
+-- @return number, number w, h The size of the given text
+-- @2D
+-- @realm client
+function draw.GetTextSize(text, font)
+	surface.SetFont(font or "DefaultBold")
+
+	return surface.GetTextSize(text)
+end
```
Alf21: missing EOL

PR #392, opened 2019-10-29
fixes confirmation spam and delayed nwvar syncing (needed for msb)

Alf21:


Alf21:
Any thoughts? Maybe using a function `ply:GetNetworkingData(...)` instead of using `self.networking. ...` directly...

TimGoll:
Looks good to me. We have to update ULX though, since the ULX commands also set the `role_found` NWvar.

To your question: I think it is better if it is playerindependant. If it is independant the networking system can be extended in the future without relying on players.

Alf21:
Currently there is the issue that a newly connected player doesn't see the confirmed players. This needs to get fixed

Alf21:
> question: I think it is better if it is playerindependant. If it is independant the networking system can be extended in the future without relying on players.

Yes, but then we need to have this system. This could take a long time. But we need to fix the Miniscoreboard ^^

TimGoll:
> Currently there is the issue that a newly connected player doesn't see the confirmed players. This needs to get fixed

I had the same problem with advanced spectator. I solved this with a clientside hook that is called once the client is ready. It then sends a request to the server. See: https://github.com/TTT-2/ttt2-advanced_spectator/blob/master/lua/autorun/client/cl_spec_handler.lua#L30

PR #393, opened 2019-10-30
cleaned up roles

TimGoll:
- fixed var-typo from `buildin` to `builtin`
- added ColorLighten and ColorDarken functions and used them in the roles module. This prevents the problem that roles have to set three different colors. Also it removes code duplication, since the sidekick etc already add these color functions
- used new Init and PreInit functions for builtin roles as well


TimGoll:
Additionally one should think about changing `togglable` to `toggleable` prior to the 1.0 release

TimGoll:
> > Additionally one should think about changing `togglable` to `toggleable` prior to the 1.0 release
> 
> Why? This is not even a defined word, and when i researched it, both variants exist and are used. None is considered better / correct.

Okay. You're right. I've never heard of `togglable` and google corrected it to `toggleable`. Therefore I assumed, the latter is the only correct way.

PR #394, opened 2019-11-01
Pickup Additions

TimGoll:
So I just wanted to do something simple and it got kind of complicated. But after hours of testing, it works now!

What's changed?
- players can use the weapon switch functionality to pick up weapons
- the pickup system works with items like C4 and decoy that have their custom pickup system
- prevents weapon pickup when `PlayerCanPickupWeapon` returns false --> see TTTC
- prevents the radio from vanishing when there is no space in the inventory
- introduces a convar to disable the auto-pickup of weapons (does not affect ammo), autopickup is disabled in this PR since I like it better. But I'm open for a change if you'd like it the other way
- unifies the pickupsystem of equip and normal weapons

All in all this is a pretty simple change. Why did I modify the `ply:Give()` function? And why did I use the hook? Blame Gmod. The `PlayerCanPickupWeapon` is always called when `ply:Give()` is called. The same hook is used for autopickup. Therefore I introduced a flag to differentiate between these two cases.

The only downside right now is that no two weapons of the same class can be switched. This is because the `PlayerCanPickupWeapon` is used for everything and there are only two possibilities:
- use the hook and allow the pickup if it is a weapon_switch --> the new weapon takes the ammo of the old weapon (not nice)
- ignore the hook if it is a weapon change and the weaponclasses are identical --> could lead to problems (I don't know about a realy scenario, but it could lead to problems if a player already has weapons, but switching those weapons should be prohibited because then it is no longer prohibited if a player wants to change a deagle with a deagle) (not nice, but maybe okay)

The two interesting files to review are:
1. https://github.com/TTT-2/TTT2/pull/394/files#diff-bb7d466d24e307b3292a4722e8ade165
2. https://github.com/TTT-2/TTT2/pull/394/files#diff-e88781d331ac3dd3fddad6d1a4582688


TimGoll:
Update: The previously explained problem had another root. It was caused by ignoring the stored ammo in the dropped weapon. 

Now it works: A player can switch weapons, even when they have the same class without any of the previously mentioned shortcomings. Please review and merge, thanks! :)

PR #400, opened 2019-11-20
Add a hook for the next map load

menof36go:
Add and call hook in CheckForMapSwitch instead of setting a possibly unwanted timer. This is based on a suggestion / request from the Partly Adequate Mapvote (PAM) creator. Instead of having to override the entire CheckForMapSwitch function, addons can simply hook into the map switch process. This change is compatible with old TTT addons, since they override the affected function in its entirety anyways.
Working example with PAM:
```lua
hook.Add("Initialize", "PAMAutoStart", function()
	if GAMEMODE_NAME == "terrortown" then
		function CheckForMapSwitch()
			local rounds_left = math.max(0, GetGlobalInt("ttt_rounds_left", 6) - 1)

			SetGlobalInt("ttt_rounds_left", rounds_left)

			local time_left = math.max(0, (GetConVar("ttt_time_limit_minutes"):GetInt() * 60) - CurTime())

			if rounds_left <= 0 or time_left <= 0 then
				timer.Stop("end2prep")

				PAM.UpdateRecentMaps()
				PAM.Start(nil, false)
			end
		end
	end
end)
```
Can now easily be reduced to:
```lua
hook.Add("TTT2LoadNextMap", "PAMAutoStart", function() 
	PAM.UpdateRecentMaps()
	PAM.Start(nil, false)
	return true
end)
```
> [...] important things such as a gamemode checking for a map switch or events that you want external scripts to edit should have hooks.

Quote from the [Wiki](https://wiki.garrysmod.com/page/Hook_Library_Usage)


gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -701,6 +701,15 @@ function GM:PostCleanupMap()
 	ents.TTT.FixParentedPostCleanup()
 end
 
+---
+-- Called if CheckMapSwitch has determined that a map change should happen
```
Reispfannenfresser: Some obligatory nitpicking:

The function mentioned here (line 705) is called "CheckForMapSwitch" and not "CheckMapSwitch"

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -1162,7 +1171,7 @@ function CheckForMapSwitch()
 
 	if switchmap then
 		timer.Stop("end2prep")
-		timer.Simple(15, game.LoadNextMap)
+		hook.Run("TTT2LoadNextMap")
```
Alf21: you should add `nextmap` as param and an empty line in front of this line to seperate hooks from other code blocks :)

menof36go: Done

gamemodes/terrortown/gamemode/server/sv_main.lua
menof36go: Done ._.

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -701,6 +703,15 @@ function GM:PostCleanupMap()
 	ents.TTT.FixParentedPostCleanup()
 end
 
+---
+-- Called if CheckForMapSwitch has determined that a map change should happen
+-- @note Can be used for custom map voting system. Just hook this and return true to override.
+-- @hook
```
Alf21: `@param string nextmap`

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -1162,7 +1173,8 @@ function CheckForMapSwitch()
 
 	if switchmap then
 		timer.Stop("end2prep")
-		timer.Simple(15, game.LoadNextMap)
+
```
Alf21: https://ci.neoxult.de/job/TTT2/job/PR-400/2/

Alf21:
I don't know whether the ConVar makes sense, but anyway nice to have one :)

Reispfannenfresser:
There are some messages being sent when the conditions for a map change apply.
(lines 1163 and 1167)
The content of these messages won't always be true now that the hook is there.

gamemodes/terrortown/gamemode/server/sv_main.lua
Alf21: `gamemodes/terrortown/gamemode/server/sv_main.lua: [Warning] line 1176, column 1 - line 1176, column 2: Trailing whitespace`

Reispfannenfresser:
You technically don't even need nextmap as a parameter now, since it is only used from within the hook.
I don't know if there are some hidden functionalities in game.GetMapNext though.

Alf21:
> You technically don't even need nextmap as a parameter now, since it is only used from within the hook.
> I don't know if there are some hidden functionalities in game.GetMapNext though.

IMO we should provide this var as a param. Even if there are no hidden functionalities, it is better to use the exact var that leads to this result. And it is used in this function too (and maybe just made for functions like this).

PR #408, opened 2019-12-04
Changes in the English localization file

Satton2:
- Line 1194: every saved changes (positions and color) getting reset.
- Line 1173: I think «editor» would be a better name for that tab.
- Lines 1215, 1216, 1217: questions? Ahem... I think player, admin, or server owner expects to receive a direct answer, not questions to themselves. Also, admin can be without needed permission.
- Lines 1213, 1214, 1225, 1226, 1228, 1229: improved.
- Line 1211 moved to line 1207.
- Line 1181: I really want to change it to «Open the shop by pressing {menukey} instead of the score menu during preparing / at the end of a round?» but this {menukey} doesn't work for some reason. (probably I should make an Issue for this one later)


TimGoll:
I'm not quite sure about Hud Editor / Switcher. Hmmm.

And 1194 is wrong. It only resets positions. Not the color.

Satton2:
> I'm not quite sure about Hud Editor / Switcher. Hmmm.

It's already an editor in this line: https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/shared/lang/english.lua#L1200
Also, in this editor, you can **edit** the hud or switch to another one and **edit** it if it supports the editor.

> And 1194 is wrong. It only resets positions. Not the color.

When I change the HUD scale, the color got reset. Perhaps this another personal bug that I sometimes get for some reason, but I don't think so. I will test this again a bit later.

Satton2:
Yes, settings that player set stays untouched, but it fact color gets reset.
I downloaded TTT2 on Nov 25, 2019 (±1 day) from GitHub, so my version of TTT2 is new enough.
Also, I didn't touch any files except localization files, so it's not an issue that I made by myself.

saibotk:
Reworking all menus is on our roadmap for version 1.0, but we'll look into this, thank you for reporting this problem, and thank you very much for your translation work 👍🏻

gamemodes/terrortown/gamemode/shared/lang/english.lua
```diff
@@ -1169,7 +1169,7 @@ L.shop_selected = "Selected {role}"
 L.shop_settings_desc = "Change the values to adapt Random Shop ConVars. Don't forget to save your changes!"
 
 L.f1_settings_changes_title = "Changes"
-L.f1_settings_hudswitcher_title = "HUD Switcher"
```
saibotk: This should still be called `HUD Switcher`, as the code refers to the HUD Editor as the position editor.
Also this will be reworked in version 1.0.

Satton2: Hmmm... look's like I got the point now.
HUD Editor is a part of the HUD Switcher, while the HUD Switcher is the left panel when you can choose what HUD you want to use or edit if it supports HUD Editor.

I will change it back to `HUD Switcher`.

PR #416, opened 2019-12-21
Fix addondev color

menof36go:
This has bugged me for a while now so I finally got of my ass and made a PR


TimGoll:
I'm bot saying your PR is bad. But how do you define a false color? :D

menof36go:
> I'm bot saying your PR is bad. But how do you define a false color? :D

Last time I checked rgb values were supposed to be in the range 0 to 255. So I figured someone accidently put in 300 instead of 30 and it went unnoticed since Color just caps the value at 255

TimGoll:
Oof. Valid point. Thanks!

TimGoll:
I'm gonna merge this now, there is no point in waiting for other reviews

PR #417, opened 2019-12-21
Fix sprint consuming stamina when there is no move input

menof36go:
You will no longer get exhausted by standing still lol
No move input -> No stamina used
I also reworked the progress to rely on the FrameTime instead of elapsed time,
which uses 2 variables instead of 3 and is cleaner in my opinion


menof36go:
It wasn't subject of this pull request but you can now also double tap the forward key to sprint while we're at it

TimGoll:
Like wtf. You'll finish all our features from the 1.0 ToDo list before we get around to release 0.6

Thanks!

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -44,31 +44,54 @@ if SERVER then
...
 	end)
 else
-	local function PlayerSprint(bool)
+	local lastPress = 0
+
+	local function PlayerSprint(trySprinting, bind)
 		local client = LocalPlayer()
 
-		if bool and not GetGlobalBool("ttt2_sprint_enabled", true) or not bool and not client.isSprinting then return end
+		if trySprinting and not GetGlobalBool("ttt2_sprint_enabled", true) then return end
```
Alf21: @TimGoll could you clean this up if you find some time? :) I'm unable to do so currently :/

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -44,31 +44,54 @@ if SERVER then
...
+		if not trySprinting and not client.isSprinting or trySprinting and client.isSprinting then return end
+		if client.isSprinting and (client.useSprintBind and not bind or not client.useSprintBind and bind) then return end
 
 		client.oldSprintProgress = client.sprintProgress
-		client.sprintMultiplier = bool and (1 + GetGlobalFloat("ttt2_sprint_max", 0)) or nil
-		client.isSprinting = bool
-		client.sprintTS = CurTime()
```
Alf21: Doesn't it break something or were these the only places this var was used?

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -44,31 +44,54 @@ if SERVER then
...
 		net.Start("TTT2SprintToggle")
-		net.WriteBool(bool)
+		net.WriteBool(trySprinting)
 		net.SendToServer()
 	end
 
+	hook.Add("KeyPress", "TTT2DoublePressSprint", function(ply, key)
+		if not IsFirstTimePredicted() then return end
+		if key == IN_FORWARD and not ply.isSprinting then
+			local time = CurTime()
```
Alf21: I don't gave a deep look, I think CurTime is correct because this hook should get triggered as soon as I press a key (not any frame / tick). But just to think about, it's correct not using FrameTime, right?

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -44,31 +44,54 @@ if SERVER then
...
+				PlayerSprint(true, false)
+			end
+
+			lastPress = time
+		end
+	end)
+
+	hook.Add("KeyRelease", "TTT2EndDoublePressSprint", function(ply, key)
+		if not IsFirstTimePredicted() then return end
+		if key == IN_FORWARD and ply.isSprinting then
```
Alf21: so double pressing backwards / sidewards won't trigger the sprint, right?

Alf21: WIth the bindings, you are able doing it.

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -85,24 +108,23 @@ hook.Add("Think", "TTT2PlayerSprinting", function()
 
 	for i = 1, #plys do
 		local ply = plys[i]
+		local wantsToMove = ply:KeyDown(IN_FORWARD) or ply:KeyDown(IN_BACK) or ply:KeyDown(IN_MOVERIGHT) or ply:KeyDown(IN_MOVELEFT)
 
-		if not ply.sprintTS then continue end
+		if ply.sprintProgress == 1 and (not ply.isSprinting or not wantsToMove) then continue end
+		if ply.sprintProgress == 0 and ply.isSprinting and wantsToMove then continue end
```
Alf21: So I need to release the key manually to be able to regenerate? Or will it start automatically? How about having just a slower sprint like in the current version?

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -85,24 +108,23 @@ hook.Add("Think", "TTT2PlayerSprinting", function()
...
 
-		local timeElapsed = CurTime() - ply.sprintTS
 		local modifier = {1} -- Multiple hooking support
 
-		if not ply.isSprinting then
+		if not ply.isSprinting or not wantsToMove then
```
Alf21: In which case the second condition is needed? :o

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
Alf21: And will the player start regenerating even if he is moving? The stamina should just be consumed if the player sprints.

Alf21:
Thanks for this hard work. I just made some comments to avoid some inconsistent behaviour. Sorry for asking something that should be clear viewing the code, but giving a look on my smartphone doesn't makes fun :D

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
menof36go: sprintTS isn't used anywhere else or in any TTT2 related repository

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
menof36go: Yes, the hook is triggered on key down events. I use CurTime to check if the button was pressed twice in the last 0.4 seconds. In this case CurTime makes most sense, since the button presses could be multiple frames apart and still count as a double tap :)

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
menof36go: Nope, only the forward key does. This particular hook is about making sure we stop sprinting if the player started sprinting by double tapping w and now lets go of it

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
menof36go: 1. You will regenerate stamina while standing still, regardless if you press the sprint bind or not.
2. The stamina is only ever consumed if the player is moving.
3. If you keep trying to sprint while moving, but have no stamina, the stamina will not start regenerating (I tested it and it would just feel kind of off if it does)
4. If you just walk around normally without trying to sprint, the stamina will regenerate, just like before.

Number 1 is the only thing that's different from before, except you can now also start sprinting by double tapping w.

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
menof36go: If the player isSprinting (pressing the bind for example) but standing still, we still want to regenerate stamina

Issue #401, opened 2019-11-20
reroll_no_credits translation bug


Alf21:
https://github.com/TTT-2/TTT2/blob/2d24e0b181a84d1da2784296f1ff9a7b523c39cc/gamemodes/terrortown/gamemode/shared/lang/english.lua#L140

This should be
`You need {amount} credits to reroll!`

Alf21:
https://github.com/TTT-2/TTT2/blob/e6fba6e555bb5711b1f443657519abf25596c028/gamemodes/terrortown/gamemode/client/cl_reroll.lua#L25

this should be something like

`client:GetCredits() < cost`

Alf21:
@lebroomer : @Satton2 told me the following:

> I set the needed credits amount to 100 but... «Use {amount} credits to get new items in your shop!» appears. When I have no credits at all, the Reroll tab disappears.

PR #419, opened 2019-12-21
Russian translation update

Satton2:
- Fixed **many** grammar and translation mistakes made by other translators.
- Translated strings that weren't translated in the past.
- Updated old strings.


Satton2:
I will translate a few strings after the [issue](https://github.com/TTT-2/TTT2/issues/401) will be resolved.

PR #421, opened 2019-12-23
Change default loading screen

menof36go:
This PR changes the default loading screen to show the TTT2 logo instead. Ofcourse all design is subject to change, I am not a graphic designer after all.
There are two things I kind of dislike, but see no way of resolving due to gmod just being gmod:
1. If the loading url is changed while the server is running, there seems to be no way to react to it, since the callback only works for lua created convars, which means that if you change the loading url while the server is still running to "" (no custom loading screen), the ttt2 default loading screen will not show up until the next restart / map change. This is not really an issue however, since that is incredibly rare (loading url is set in the config in the vast majority of cases) and is only a minor inconvenience if it does happen.
2. If you do not have TTT2 installed and are joining a TTT2 server for the very first time, the loading screen will not show up. Having a "fallback" loading screen in this case is also kind of impossible due to the nature of garrysmod. It should happen only once, so I consider this to be a minor inconvenience as well.
The new default TTT2 loading screen would not affect servers with custom loading screens in any way.
[What it looks like as of now](https://i.imgur.com/CB3qEny.jpg) (ofcourse very much subject to change). I like the minimalism since it is close to the default gmod loading screen, but there is ofcourse no limitation since it is just a loading screen like any other.


materials/vgui/ttt/loadingscreen/loading.html
```diff
@@ -0,0 +1,11 @@
+<html>
+	<head>
+		<title>Loading..</title>
+		<link rel="stylesheet" type="text/css" href="loading.css?3" />
+		<script type="text/javascript" src="js/thirdparty/jquery.js"></script>
```
TimGoll: Why do you need jQuery here?

TimGoll: okay, after talking about this, I think this can stay in here since vanilla TTT also has this included

saibotk:
Wow, nice thank you!
This looks awesome ;)

materials/vgui/ttt/loadingscreen/loading.html
```diff
@@ -0,0 +1,11 @@
...
+	</head>
+	<body>
+		<img src='img/bg.jpg' class="background"/>
+		<img src='img/TTT2Icon.png' class="center"/>
+	</body>
+</html>
```
Alf21: missing EOF

materials/vgui/ttt/loadingscreen/loading.css
```diff
@@ -0,0 +1,61 @@
...
+	margin: 0 auto;
+	text-align: right;
+	font-size: 12px;
+	font-family: helvetica;
+	color: rgba( 0, 0, 0, 0.4);
+	letter-spacing: -1px;
+	text-shadow: 0px 0px 3px white;
+	overflow: hidden;
+	white-space: nowrap;
+}
```
Alf21: missing EOF

gamemodes/terrortown/gamemode/shared/sh_include.lua
```diff
@@ -99,6 +99,7 @@ if SERVER then
 		sv_weapon_pickup = {file = "sv_weapon_pickup.lua", on = "server"},
 		sv_weaponry = {file = "sv_weaponry.lua", on = "server"},
 		sv_addonchecker = {file = "sv_addonchecker.lua", on = "server"},
+		sv_loadingscreen = {file = "sv_loadingscreen.lua", on = "server"},
```
Alf21: wrong position as well as the line above (@TimGoll ). Should be ordered alphabetically

gamemodes/terrortown/gamemode/server/sv_loadingscreen.lua
```diff
@@ -0,0 +1,4 @@
+local loadingUrl = GetConVar("sv_loadingurl")
+if loadingUrl and loadingUrl:GetString() == "" then
+    game.ConsoleCommand("sv_loadingurl \"asset://garrysmod/materials/vgui/ttt/loadingscreen/loading.html\"\n")
+end
```
Alf21: missing EOF

PR #424, opened 2019-12-30
Fix weapon switch and minor german localization changes

menof36go:
I've tested it and this seems to fix the weapon switch bug that occurs when weapons are switched too rapidly. I've also corrected some grammar and spelling issues in the german localization.


TimGoll:
Okay. You removed the ResetWeapon call when the pickup process is stopped by starting a new one. This is strange since without this call the old weapon should be still without physics

TimGoll:
I tested it, the problem still exists. I used 4 weapons in the corner of a map. One vanished through the ground (because of disabled physics?) and one stopped working like earlier.

But it seems like I'm unable to switch weapons as fast as I could earlier. Is this a change you've made?  Or is this a change by me? :D

TimGoll:
ignore the push, I removed the last commit since I was still on the wrong branch

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
```diff
@@ -639,6 +639,11 @@ concommand.Add("ttt_dropammo", DropActiveAmmo)
 -- @ref https://wiki.garrysmod.com/page/GM/WeaponEquip
 -- @local
 function GM:WeaponEquip(wep, ply)
+	if IsValid(ply) then
+		-- allow the player to pick up another weapon
+		ply.wpickup_waitequip = false
```
Alf21: Theoretically this could be simplified because there are multiple checks of `IsValid(ply)` (nothing of that except the weapon removing should happen if ply is invalid)

gamemodes/terrortown/gamemode/shared/lang/german.lua
```diff
@@ -84,11 +84,11 @@ L.equip_tabtitle = "Ausrüstung bestellen"
 L.equip_cost = "Du hast {num} Credit(s) übrig."
 L.equip_help_cost = "Jedes Ausrüstungsteil, das du kaufst, kostet 1 Credit."
 
-L.equip_help_carry = "Du kannst nur das kaufen, für das du auch Platz hast."
+L.equip_help_carry = "Du kannst nur das Kaufen, für das du auch Platz hast."
```
Alf21: That's wrong. the verb isn't made to a substantive...

gamemodes/terrortown/gamemode/shared/lang/german.lua
```diff
@@ -38,8 +38,8 @@ L.win_innocents = "Die Innos haben gewonnen!"
 L.win_bees = "Die Bienen haben gewonnen! (Es ist ein Unentschieden)"
 L.win_showreport = "Schauen wir uns den Rundenbericht die nächste(n) {num} Sekunde(n) an."
 
-L.limit_round = "Rundenlimit erreicht. {mapname} wird bald geladen."
-L.limit_time = "Zeitlimit erreicht. {mapname} wird bald geladen."
+L.limit_round = "Rundenlimit erreicht. Die nächste Map wird bald geladen."
```
Alf21: Why doing this? Fixing this here seems to be inconsistent. If mapname is displayed cryptically, this is a fault on the lua side.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1106,6 +1106,12 @@ end
 -- @returns Weapon if successful, nil if not
 -- @realm server
 function plymeta:PickupWeapon(wep, dropBlockingWeapon, shouldAutoSelect)
+	-- If the player has picked up a weapon, but not yet received it, ignore this request
+	-- to prevent GMod from making the weapon unusable and behaving weird
+	if self.wpickup_waitequip then
+		return nil
```
Alf21: `return nil` is the same as `return`

gamemodes/terrortown/gamemode/shared/lang/english.lua
```diff
@@ -38,8 +38,8 @@ L.win_innocents = "The Innos have won!"
 L.win_bees = "The Bees has won! (Its a draw)"
 L.win_showreport = "Let's look at the round report for {num} seconds."
 
-L.limit_round = "Round limit reached. {mapname} will load soon."
-L.limit_time = "Time limit reached. {mapname} will load soon."
+L.limit_round = "Round limit reached. The next map will load soon."
```
Alf21: Inconsistent. Give a look into the comment in the german lang section

gamemodes/terrortown/gamemode/shared/lang/german.lua
```diff
@@ -612,11 +612,12 @@ L.corpse_too_far_away = "Leiche zu weit weg zum Untersuchen."
 L.corpse_binoculars = "Drücke [{key}] um Leiche mit Fernglas zu untersuchen."
 L.corpse_searched_by_detective = "Diese Leiche wurde von einem Detektiv untersucht"
 
-L.target_disg = "gestarnt"
+L.target_disg = "Getarnt"
 L.target_unid = "Unidentifizierter Körper"
 
 L.target_credits = "Durchsuche, um ungenutzte Credits zu erhalten."
 
+L.target_switch_weapon = "Drücke [Knopf] um mit aktueller Waffe zu tauschen"
```
Alf21: "Knopf" shouldn't be statically haha :D But that's nothing on your side.

gamemodes/terrortown/gamemode/shared/lang/german.lua
```diff
@@ -729,21 +730,21 @@ L.tip10 = "Halte als Verräter auf rote oder als Detektiv auf blaue Nachrichten
...
+L.tip13 = "Wenn du in der Nähe von jemandem standest, den du getötet hast, hinterlässt du deine DNA auf der Leiche. Diese DNA kann ein Detektiv mit seinem DNA-Scanner untersuchen, um deine momentane Position herauszufinden. Es wäre besser, wenn du die Leiche versteckst, nachdem du jemanden mit dem Messer getötet hast!"
 
-L.tip14 = "Je weiter du dich von der Leiche entfernst, an der deine DNA hängt, dest schneller verschwindet die DNA Spur."
+L.tip14 = "Je weiter du dich von der Leiche entfernst, an der deine DNA hängt, desto schneller verschwindet die DNA-Spur."
 
-L.tip15 = "Du bist Verräter und gehst snipen? Du solltest in Betracht ziehen eine Tarnung zu kaufen. Wenn du verfehlst, renn' an einen sicheren Ort und deaktiviere deine Tarnung. Niemand wird wissen, dass du es warst, der auf sie schießen wollte."
+L.tip15 = "Du bist ein böser Heckenschütze? Dann solltest du in Betracht ziehen eine Tarnung zu kaufen. Wenn du verfehlst, renn an einen sicheren Ort und deaktiviere deine Tarnung. Niemand wird wissen, dass du der Heckenschütze warst."
```
Alf21: "Heckenschütze" isn't used often, but if you think it's better, I'm okay with it :D

gamemodes/terrortown/gamemode/shared/lang/german.lua
```diff
@@ -1217,4 +1218,4 @@ L.disable_spectatorsoutline = "Deaktiviere Zuschauerumrandungen"
 L.disable_spectatorsoutline_tip = "Deaktiviert die Umrandungen um ein Objekt, in dem sich ein Zuschauer befindet (+Performance)"
 
 L.disable_overheadicons = "Deaktiviere Rollenicons"
-L.disable_overheadicons_tip = "Deaktiviert die Rollenicons über den Köpfen der Spieler (+Performance)"
+L.disable_overheadicons_tip = "Deaktiviert die Rollenicons über den Köpfen der Spieler (+Performance)"
```
Alf21: Missing EOF

gamemodes/terrortown/gamemode/shared/lang/german.lua
menof36go: I saw that too and wasn't sure whether or not Tim intended it to be that way haha

gamemodes/terrortown/gamemode/shared/lang/german.lua
menof36go: It is often impossible to tell the next mapname beforehand. It's more consistent, since as of now it will always be wrong if you are using a mapvote addon instead of the default map cycling, which is atleast in my perception the case on the vast majority of servers. See this [PR](https://github.com/TTT-2/TTT2/pull/400).

menof36go:
I've tested it and it seems to work. The commit still requires plenty of cleanup and is for testing purposes only. Someone should please test it and tell me whether or not it works (@TimGoll ), since it does for me, but results may vary. If it does work, I'll clean it up a lot ofcourse..

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -396,7 +397,7 @@ function HUDDrawTargetIDWeapons(data, params)
 	local client = LocalPlayer()
 
 	if not IsValid(client) or not client:IsTerror() or not client:Alive()
-	or data.distance > 100 or not data.ent:IsWeapon() then
+	or data.distance > 1000 or not data.ent:IsWeapon() then
```
TimGoll: is this a testing change? because the distance "100" is our default distance

TimGoll:
okay, yes. I spam switched for about two minutes without any problems. While this still not confirms that it is working, it seems to be good (and better than it was previously). At some point I tried adding a flag that waits until the weapon is equipped too, but it seems like I've done a bad job because my implementation did not work at all.
A big thank you for your work, it's great!

gamemodes/terrortown/gamemode/shared/lang/german.lua
TimGoll: yeah, we should talk about this before the 0.6 release. Currently there is no mention of the key in the text, it only references the keysymbol next to it. We could change it however

saibotk:
A big THANK YOU <3 good job, looks like this could be cleaned up and then should be good to be merged. You were a huge help ;)

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1106,19 +1106,31 @@ end
...
-	if wep.wpickup_player then return end
+	if wep.wpickup_player then
+		self.wpickup_waitequip = false
+		return
+	end
 
-	-- if the player tries to pickup another weapon while the pickup process of a previous
-	-- weapon is still running, the old flags have to be reset
-	ResetWeapon(self.wpickup_weapon)
+	print("Pickup", wep)
```
saibotk: Remove this debug msg

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1131,20 +1143,28 @@ function plymeta:PickupWeapon(wep, dropBlockingWeapon, shouldAutoSelect)
 	wep.wp__WeaponSwitchFlag = dropBlockingWeapon
 
 	local canPickupWeapon = self:CanPickupWeapon(wep)
+	--local canPickupWeapon = true
```
saibotk: remove commented out code

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1131,20 +1143,28 @@ function plymeta:PickupWeapon(wep, dropBlockingWeapon, shouldAutoSelect)
...
 	if dropBlockingWeapon then
 		local dropWeapon, isActiveWeapon, switchMode = GetBlockingWeapon(self, wep)
 
 		if switchMode == SWITCHMODE_FULLINV then return end
 
+		-- Very very rarely happens but definitely breaks the weapon and should be avoided at all costs
+		if dropWeapon == wep then return end
+
 		self:SafeDropWeapon(dropWeapon, true)
+		--self:DropWeapon(dropWeapon)
```
saibotk: remove commented out code

gamemodes/terrortown/gamemode/shared/lang/german.lua
```diff
@@ -664,19 +667,19 @@ L.spec_help2 = "Zum Verlassen des Zuschauer-Modus öffne das Menü mit {helpkey}
 -- quotes. That's a Lua thing. Every line break (enter) will show up in-game.
 L.info_popup_innocent = [[Du bist ein unschuldiger Terrorist!
 Aber es gibt Verräter… Wem kannst du trauen, und wem füllst du den Körper mit Blei?
-Gib Acht und arbeite mit anderen zusammen um zu überleben!]]
+Gib acht und arbeite mit anderen zusammen, um zu überleben!]]
```
Alf21: That's wrong. "Acht" is a substantive in that case (it isn't written together). Ref.: https://www.duden.de/suchen/dudenonline/Gib%20acht

gamemodes/terrortown/gamemode/shared/lang/german.lua
```diff
@@ -996,7 +999,7 @@ L.aw_mac1_title = "Pray and Slay"
 L.aw_mac1_text = "tötete {num} Leute mit der MAC10, sagte allerdings nicht wie viel Munition er dafür gebraucht hat."
 
 L.aw_mac2_title = "Mac and Cheese"
-L.aw_mac2_text = "fragt sich, was passieren würde, wenn man zwei MAC10s gleichzeitig benutzen würde. Zwei mal {num}?"
+L.aw_mac2_text = "fragt sich, was passieren würde, wenn man zwei MAC10s gleichzeitig benutzen würde. Zwei Mal {num}?"
```
Alf21: Maybe "Zweimal"? ^^

gamemodes/terrortown/gamemode/shared/lang/german.lua
Alf21: @saibotk @TimGoll okay with it too?

gamemodes/terrortown/gamemode/shared/lang/german.lua
Alf21: @TimGoll resolve it if you created an issue, pls :)

gamemodes/terrortown/gamemode/shared/lang/german.lua
saibotk: yes

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
Alf21: Just a small request. You could let it stay as it is, but just in case to keep the code simple :)

gamemodes/terrortown/gamemode/client/cl_targetid.lua
Alf21: @TimGoll

TimGoll: already resolved. Maybe your mobile website did not refresh it

gamemodes/terrortown/gamemode/shared/lang/german.lua
menof36go: Most of the spelling and grammar changes were proposed by word, Ill change it

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
menof36go: It's checked twice because it happens at two different points in time. The player could be valid and suddenly become invalid during the next frame, which is why I would keep it

gamemodes/terrortown/gamemode/shared/lang/german.lua
menof36go: Damnit word! :D Fixed

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1155,57 +1178,62 @@ function plymeta:PickupWeapon(wep, dropBlockingWeapon, shouldAutoSelect)
...
-		-- of the player
-		-- IMPORTANT: If the weapon gets teleported into other entities, it gets stuck. Therefore
-		-- the weapon is teleported to half player height
-		local pWepPos = self:EyePos()
-		pWepPos.z = pWepPos.z - 20 -- -20 to move it outside the viewing area
+		-- Since we all love GMOD we do some really funny things here. Sometimes the weapon is in
+		-- a position where a player is unable to pick it up, even if there is nothing that hinders
+		-- it from being picked up. Therefore we randomise the position a bit.
+		local function SetWeaponPos()
+			if not IsValid(self) or not IsValid(wep) or not self.wpickup_waitequip[kind] then return end
```
Alf21: self should be localized for this localized function...

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1155,57 +1178,62 @@ function plymeta:PickupWeapon(wep, dropBlockingWeapon, shouldAutoSelect)
...
 	end
 
-	-- this flag is set to the player to make sure he only picks up this weapon
-	self.wpickup_weapon = wep
+	-- prevent race conditions
+	self.wpickup_waitequip[kind] = true
 
-	-- the flag is set to the weapon to stop other players from auto-picking up this weapon
-	wep.wpickup_player = self
+	timer.Create(kind .. "_WeaponPickup_" .. self:SteamID64(), 0, 1, function()
```
Alf21: functions like this should be localized too to save performance

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
Alf21: Sure, but the timer shouldn't start if the player is invalid at the first time too. Do you understand what I mean? Theoretically, the weapon should get removed if it is incompatible (return ?!). After this, if the player is valid, the other stuff should run (equipping, ...). Currently it seems to be inconsistent :o

menof36go: Ah, I see now, I must have been looking at outdated files. I get your point now, Ill fix it

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
menof36go: Sure (It's an artifact), but I am curious though. Isn't self just like wep in this context? In my understanding wep is just the second argument to a function while self is the implicit first argument. Since my understanding of the lua stack is very limited I am curious as to how they are different.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
menof36go: Localized as in create a local function outside of PickupWeapon or as in create a local function inside of PickupWeapon (similar to SetWeaponPos)? I would assume the first thing, since in my understanding the later wouldn't improve performance. If so, I should probably also move SetWeaponPos outside of WeaponEquip

Alf21: Both is - regarding to the lua docu - a better performance than doing function declarations just inline. Don't ask me why, I didn't gave a deep look into luajit or stuff like this. Maybe because it isn't optimized and need to do several other actions localizing the inlined function. Maybe not. But answering your question, defining a - maybe abstract / general - function outside this function (as a first instance local function) would lead to the best performance.

Alf21: additionally, there need to be a valid check for self (or the localized thing of self :) )

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
Alf21: yea self is a - more or less - localized var in this case (the player's var). But it is just recommended to use self in exact this instance - where it was meant to be defined. Using it outside this situation or inside another function, this is an indicator that there exists inconsistency or something that can be done in a better way. Sometimes, jenkins doesn't like those situation too, so often we use "self" in a enclosed / included other localized function, we declare a var too avoid issues (smth like `local ply = self`).

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
```diff
@@ -639,16 +639,17 @@ concommand.Add("ttt_dropammo", DropActiveAmmo)
 -- @ref https://wiki.garrysmod.com/page/GM/WeaponEquip
 -- @local
 function GM:WeaponEquip(wep, ply)
+	if not IsValid(ply) or not IsValid(wep) then return end
+
 	if IsValid(wep) and not wep.Kind then
```
Alf21: IsValid(wep) is useless now :D

PR #431, opened 2020-01-03
add fourth parameter to LANG.Msg function

TimGoll:
This allows the call of `LANG.MSG(ply, msg, params, mode)`, with `mode` being the new parameter. This removes the strange table in cl_lang that defines different modes for specific strings and opens them up to be used in any context.

This keeps compatibilitry with old addons that define their own style with `LANG.SetStyle` (a rather strange construct that allows the use of callback functions for specific lang strings, but who knows). While compatibility is kept, sex new and easy to use styles are added as default styles with four of them beeing eyisting and renamed ones and two more added for consistency.

```
MSG_MSTACK_ROLE
MSG_MSTACK_WARN
MSG_MSTACK_PLAIN

MSG_CHAT_ROLE
MSG_CHAT_WARN
MSG_CHAT_PLAIN
```

What does this change mean in reality?

Previously a lang string looked like this:
```lua
LANG.AddToLanguage("English", "my_lang_elem", "TEXT")
LANG.SetStyle("my_lang_elem", "the_style")

LANG.MSG(ply, "my_lang_elem")
```
This is rather stupid in my opinion because this is not straighforward and only allows the use of a string for one message environment. Now it looks like this:
```lua
LANG.AddToLanguage("English", "my_lang_elem", "TEXT")

LANG.MSG(ply, "my_lang_elem", nil, MSG_TYPE)
```

Edit: I want to add though that the old way still works in case someone does use this. But the new way is way more convenient.


gamemodes/terrortown/gamemode/shared/sh_lang.lua
```diff
@@ -59,16 +68,20 @@ if SERVER then
...
 
+		-- make mode valid, use MSG_MSTACK_PLAIN as default since it was always this way
+		mode = mode or MSG_MSTACK_PLAIN
+
 		-- number of keyval param pairs to send
 		local c = params and count(params) or 0
 
 		net.Start("TTT_LangMsg")
 		net.WriteString(name)
+		net.WriteUInt(mode, 8)
```
Alf21: 8 bits for the mode is a bit too much ;D

gamemodes/terrortown/gamemode/shared/sh_lang.lua
```diff
@@ -108,9 +121,12 @@ if SERVER then
 	end
 
 	concommand.Add("_ttt_request_serverlang", ServerLangRequest)
-else -- CLIENT
+end
+
+if CLIENT then
 	local function RecvMsg()
 		local name = net.ReadString()
+		local mode = net.ReadUInt(8)
```
Alf21: 8 bits

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -341,18 +358,36 @@ LANG.MsgStyle = {}
 -- @param string name style name
 -- @return function style table
 -- @realm client
-function LANG.GetStyle(name)
-	return LANG.MsgStyle[name] or LANG.Styles.default
+function LANG.GetStyle(name, mode)
+	print("name: " .. tostring(name))
```
Alf21: too many prints...

gamemodes/terrortown/gamemode/shared/sh_lang.lua
```diff
@@ -9,13 +9,15 @@ if LANG then return end
...
+MSG_MSTACK_PLAIN = 2
 
-MSG_CHAT_ROLE = 20
-MSG_CHAT_WARN = 21
-MSG_CHAT_PLAIN = 22
+MSG_CHAT_ROLE = 3
+MSG_CHAT_WARN = 4
+MSG_CHAT_PLAIN = 5
+
+MSG_MODE_BITS = 4
```
Alf21: 3 should be enough too (up to uint 7).

TimGoll: yes, I'm stupid :D

PR #441, opened 2020-01-13
Rework shop funcs

menof36go:
Replace RunConsoleCommand with net, but keep concommand for compatibility

Add RemoveEquipmentWeapon alias in order to be consistent with RemoveEquipmentItem

Add RemoveWeapon in order to be consistent with RemoveItem

Add callback function to GiveEquipmentWeapon, which is useful due to the delayed nature of the function and allows you to do stuff like this:
```lua
ply:GiveEquipmentWeapon("some_weapon_class", function(ply, cls, wep)
	wep.UserID = ply:UserID()
	wep.SomeProperty = x
end)
```
At first glance this seems like something the Equip function is for, but often times you aren't necessarily the creator of a weapon you are giving to a player and have no access to it.

GiveEquipmentItem and AddEquipmentItem now return the actual item instead of a boolean. Due to the nature of lua this should not break any existing checks and provides an easy and fast way to access the item you just gave someone.

Add is_item to TTTCanOrderEquipment, since it is required for compatibility (I don't know how it was lost since this is actually part of the vanilla hook)

Change/Fix id parameter to support the shitty old vanilla CanOrder Hook with id and cls ambiguity. The old system sucks, but always sending the class (args[1] in the last version always is a class) is not compatible with old items and probably an oversight.

Change/Fix the is_item parameter of OrderedEquipment Hook

Add TTT2 equivalents of the CanOrder and OrderedEquipment Hook. Since the old hooks are shitty it is best just to leave them as they are and make our own hooks. It is less confusing to use and easier to expand this way. Also it cannot break anything if done this way. I think this kind of approach should actually be taken for any old hook that needs to be reworked for whatever reason.

Move some buyable checks into the proper function EquipmentIsBuyable
Move RandomShop check into a localized function
Also some refactoring and minor optimization

Thoughts?


TimGoll:
I will look at the code at a later point. But since we already talked about these changes I can say that I approve them.

gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -72,6 +72,7 @@ end
 
 local function PreqLabels(parent, x, y)
 	local tbl = {}
+	local ply = LocalPlayer()
```
TimGoll: maybe rename `ply` to `client` to keep it consitent with our code and make it clear which player is meant.

gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -155,7 +156,7 @@ local function PreqLabels(parent, x, y)
 	tbl.info.img:SetImage("vgui/ttt/equip/icon_info")
 
 	tbl.info.Check = function(s, sel)
-		return EquipmentIsBuyable(sel, LocalPlayer():GetTeam())
+		return EquipmentIsBuyable(sel, ply)
```
TimGoll: is this intedened? Shouldn't it be `ply:GetTeam()`?

gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -385,7 +386,7 @@ local function CreateEquipmentList(t)
 						or items.IsItem(item.id) and item.limited and ply:HasEquipmentItem(item.id)
 						-- already carrying a weapon for this slot
 						or ItemIsWeapon(item) and not CanCarryWeapon(item)
-						or not EquipmentIsBuyable(item, ply:GetTeam())
+						or not EquipmentIsBuyable(item, ply)
```
TimGoll: same thing here

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -195,19 +195,21 @@ end
...
 -- Gives a specific @{ITEM} (if possible)
 -- @param string cls
+-- @return @{ITEM} or false
 -- @realm server
 -- @internal
 function plymeta:AddEquipmentItem(cls)
 	local item = items.GetStored(cls)
 
-	if not item or item.limited and self:HasEquipmentItem(cls) then return end
+	if not item or item.limited and self:HasEquipmentItem(cls) then return false end
```
TimGoll: why return false here? if `nil` is returned, a simple `IsValid` check is enough

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -195,19 +195,21 @@ end
...
 
 	self.equipmentItems = self.equipmentItems or {}
 	self.equipmentItems[#self.equipmentItems + 1] = cls
 
 	item:Equip(self)
 
 	self:SendEquipment()
+	return item
```
TimGoll: add an empty line in front of a return

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -476,20 +497,18 @@ end
...
 	local item = items.GetStored(cls)
 
 	if not item or item.limited and self:HasEquipmentItem(cls) then
 		return false
 	end
 
-	self:AddEquipmentItem(cls)
-
-	return true
+	return self:AddEquipmentItem(cls)
```
TimGoll: we have to talk about this. Why return true/false and not the item? Maybe some addon relies on the returned item? A simple IsValid check solves the problem and I see no need in switching to booleans here

gamemodes/terrortown/gamemode/shared/sh_init.lua
```diff
@@ -628,23 +628,30 @@ end)
...
 -- @param table tbl equipment table
--- @param string team @{ROLE}'s team
+-- @param @{Player} player
 -- @return boolean
 -- @return string text as an icon
 -- @return string result or error
 -- @realm shared
-function EquipmentIsBuyable(tbl, team)
-	if not tbl then
+function EquipmentIsBuyable(tbl, ply)
```
TimGoll: okay I see why you changed it above. Isn't this a breaking change? I haven't used this function and I suspect nobody else does. But who knows?

menof36go: It is potentially breaking, I don't think anyone uses it though (It isn't used in any TTT2 repo). I can make it compatible, if necessary

TimGoll: It's okay by me

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
menof36go: I think you got it backwards. Before the change it was always a boolean value, now it is the item on success. The function never returns true and instead returns the item.

gamemodes/terrortown/gamemode/client/cl_equip.lua
menof36go: I think you saw it already, but if not have a look at the EquipmentIsBuyable function

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
menof36go: Sure. I am personally not a fan of returning nil (I think it's bad code style), but Ill change it. IsValid(false) should return false, so it makes no difference whatsoever, but return nil is probably more consistent with other functions.

TimGoll: I see where you're coming from. But in lua or other untyped languages I think it is not a bad style. But maybe someone else has a different opinion

gamemodes/terrortown/gamemode/server/sv_shop.lua
```diff
@@ -16,138 +16,147 @@ end
...
+local function NetOrderEquipment(len, ply)
+	local cls = net.ReadString()
+
+	if #cls == 0 then return end
+
+	OrderEquipment(ply, cls)
 end
-concommand.Add("ttt_order_equipment", OrderEquipment)
+
+local function ConOrderEquipment(ply, cmd, args)
```
lebroomer: The function name could be read misleadingly as "CanOrderEquipment" (at least I stuck a moment here). Maybe change it to ConcommandOrderEquipment or something similar.

menof36go: I agree. Ill change it to ConCommandOrderEquipment

gamemodes/terrortown/gamemode/client/cl_equip.lua
Alf21: No, he changed the params of this func

gamemodes/terrortown/gamemode/server/sv_shop.lua
```diff
@@ -16,138 +16,147 @@ end
...
+	local allow, ignoreCost, message = hook.Run("TTT2CanOrderEquipment", ply, cls, is_item, credits)
 
-				key = true
+	if message then
+		LANG.Msg(ply, message, nil, MSG_MSTACK_ROLE)
+	end
 
-				break
-			end
+	if allow == false then
```
Alf21: this can be simplified into one line (because there isn't any necessary information / no special return like true / false).

gamemodes/terrortown/gamemode/server/sv_shop.lua
```diff
@@ -16,138 +16,147 @@ end
...
+	hook.Run("TTTOrderedEquipment", ply, idOrCls, is_item)
+
+	-- add our own hook with more consistent class parameter
+	hook.Run("TTT2OrderedEquipment", ply, cls, is_item, credits, ignoreCost or false)
+end
+
+local function NetOrderEquipment(len, ply)
+	local cls = net.ReadString()
+
+	if #cls == 0 then return end
```
Alf21: this should never happen, never send an empty string ^^

gamemodes/terrortown/gamemode/server/sv_shop.lua
```diff
@@ -16,138 +16,147 @@ end
...
 end
-concommand.Add("ttt_order_equipment", OrderEquipment)
+
+local function ConCommandOrderEquipment(ply, cmd, args)
+	if #args ~= 1 then return end
+
+	OrderEquipment(ply, args[1])
+end
+
+net.Receive("TTT2OrderEquipment", NetOrderEquipment)
```
Alf21: Put this right under NetOrderEquipment func

gamemodes/terrortown/gamemode/server/sv_shop.lua
```diff
@@ -16,138 +16,147 @@ end
...
+
+concommand.Add("ttt_order_equipment", ConCommandOrderEquipment)
```
Alf21: Put this right under ConCommandOrderEquipment func :)

menof36go:
All requested changes have been implemented. 
While implementing some of them I saw that the code currently relies on a lot of table.HasValue calls to handle the CanBuy table, which is pretty bad from a performance point of view and only gets worse with increasing numbers of items and weapons. I have removed all those calls and replaced them with key checks. To do so I created a function that will automatically convert old CanBuy tables at startup. 
Tables that looked like 
`{ ROLE_TRAITOR, ROLE_JACKAL }` 
will now look like 
`{ [ROLE_TRAITOR] = ROLE_TRAITOR, [ROLE_JACKAL] = ROLE_JACKAL}`. 
I kept the role as a value to keep it compatible with any addon that still relies on table.HasValue. This change definitely requires some more testing, which I'll do on my server as soon as possible (My players are my forced test subjects).
I also moved AddBought out of the GiveEquipmentWeapon function, because it isn't consistent with GiveEquipmentItem and doesn't fit to the name and description of the function. Maybe wrapper functions like BuyEquipmentWeapon and BuyEquipmentItem should be added for that specific purpose.

gamemodes/terrortown/gamemode/shared/sh_item_module.lua
```diff
@@ -17,6 +17,15 @@ for i = 1, #itemsFiles do
 
 	local cls = string.sub(fl, 0, #fl - 4)
 
+	ITEM.CanBuy = ITEM.CanBuy or {}
```
Alf21: Please add a comment, because without knowing why you doing this, it seems weird :D

gamemodes/terrortown/gamemode/shared/sh_item_module.lua
```diff
@@ -51,6 +60,15 @@ for i = 1, #itemsFolders do
 		end
 	end
 
+	ITEM.CanBuy = ITEM.CanBuy or {}
```
Alf21: Comment to :)

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -794,10 +785,8 @@ function AddToShopFallback(fallback, subrole, eq)
 		if not equip then return end
 
 		equip.CanBuy = equip.CanBuy or {}
-
-		if table.HasValue(equip.CanBuy, subrole) then return end
-
-		equip.CanBuy[#equip.CanBuy + 1] = subrole
+		print("AddToShopFallBack", fallback, subrole, eq)
```
Alf21: Shall this be printed all the time?

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -836,11 +838,44 @@ local function InitDefaultEquipmentForRole(roleData)
...
 end
 
+local function CleanUpDefaultCanBuyIndices()
+	local itms = items.GetList()
+
+	-- load items
+	for i = 1, #itms do
+		local itm = itms[i]
+		local CanBuy = {}
+		itm.CanBuy = itm.CanBuy or {}
```
Alf21: Maybe a small converting function would be useful. If you wanna save performance, then pls add a comment too :)

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
menof36go: Oops, forgot to remove it, thanks!

Alf21:
@TimGoll to keep the overview, I suggest that you close the resolved requests ;)

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
menof36go: Done

gamemodes/terrortown/gamemode/shared/sh_item_module.lua
menof36go: Actually, your comment made me realize that it isn't even necessary, since the automatic cleanup always happens later on anyways. I removed it

gamemodes/terrortown/gamemode/shared/sh_item_module.lua
menof36go: This has been removed now, see my other comment :)

gamemodes/terrortown/gamemode/server/sv_shop.lua
menof36go: Done fb4628c

gamemodes/terrortown/gamemode/server/sv_shop.lua
menof36go: Done 06aa3db

PR #442, opened 2020-01-14
Stamina autoregen on exhaustion

menof36go:
You will automatically start regenerating stamina if you have zero stamina for some time (Currently 1.5 seconds) regardless if you are still pressing sprint / moving. It feels a lot more natural that way and prevents you from having to stop moving if the sprint was initiated by double tapping. I also removed the hooks to use the corresponding GM functions instead.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_keys.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_keys.lua
@@ -4,6 +4,7 @@
 
 local timer = timer
 local IsValid = IsValid
+local UpdateInputSprintFunc
 local cv_sv_cheats = GetConVar("sv_cheats")
 
 local function SendWeaponDrop()
@@ -174,6 +175,13 @@ function GM:KeyPress(ply, key)
 		end
 
 		timer.Simple(0.05, _func)
+	elseif key == IN_FORWARD or key == IN_BACK or key == IN_MOVERIGHT or key == IN_MOVELEFT then
+		UpdateInputSprintFunc = UpdateInputSprintFunc or UpdateInputSprint
+
+		-- Might not be loaded yet due to load order
+		if UpdateInputSprintFunc then
+			UpdateInputSprintFunc(ply, key, true)
+		end
 	end
 end
 
@@ -199,6 +207,13 @@ function GM:KeyRelease(ply, key)
 		end
 
 		timer.Simple(0.05, _func)
+	elseif key == IN_FORWARD or key == IN_BACK or key == IN_MOVERIGHT or key == IN_MOVELEFT then
+		UpdateInputSprintFunc = UpdateInputSprintFunc or UpdateInputSprint
+
+		-- Might not be loaded yet due to load order
+		if UpdateInputSprintFunc then
+			UpdateInputSprintFunc(ply, key, false)
+		end
 	end
 end
 
--- a/gamemodes/terrortown/gamemode/client/cl_keys.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_keys.lua
@@ -4,7 +4,7 @@
 
 local timer = timer
 local IsValid = IsValid
-local UpdateInputSprintFunc
+local UpdateInputSprint = UpdateInputSprint
 local cv_sv_cheats = GetConVar("sv_cheats")
 
 local function SendWeaponDrop()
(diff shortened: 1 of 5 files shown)


gamemodes/terrortown/gamemode/client/cl_keys.lua
```diff
@@ -174,6 +175,13 @@ function GM:KeyPress(ply, key)
 		end
 
 		timer.Simple(0.05, _func)
+	elseif key == IN_FORWARD or key == IN_BACK or key == IN_MOVERIGHT or key == IN_MOVELEFT then
+		UpdateInputSprintFunc = UpdateInputSprintFunc or UpdateInputSprint
+
+		-- Might not be loaded yet due to load order
```
saibotk: What load order? isn't this deterministic as we define the order?

saibotk: Move this line up to the other shared includes, and it should work:
https://github.com/TTT-2/TTT2/blob/5c51c24077282f9b898f93ecff3125fcfb99f2a2/gamemodes/terrortown/gamemode/client/cl_main.lua#L74

menof36go: The cl_main load order looks pretty messy to me and looks like it will eventually be cleaned up at which point it would immediately break again. I will move it up nonetheless, but should I keep the check, just in case?

saibotk: I dont think it will be cleaned up soon, but yes please move it up and remove the checks, as they are useless and we want to let it fail quickly, when the loading order is incorrect. Otherwise we might not notice the bug as fast as possible.

gamemodes/terrortown/gamemode/shared/sh_main.lua
```diff
@@ -135,6 +135,15 @@ function GM:TTTPlayerColor(model)
 	return COLOR_WHITE
 end
 
+local UpdateSprint = UpdateSprint
```
saibotk: Please put variable definitions to the top of the file

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -1,3 +1,22 @@
+local function PlayerSprint(trySprinting, moveKey)
+	if SERVER then return end
```
saibotk: If this is a clientside function only, why is it even defined on the server?

gamemodes/terrortown/gamemode/shared/sh_main.lua
menof36go: I put the variable definition there, because it's consistent with your code just a few lines further down:
https://github.com/TTT-2/TTT2/blob/77c5dbcee7ecc620d3ab7af7073106b1911c55b6/gamemodes/terrortown/gamemode/shared/sh_main.lua#L138-L145
I'll change it

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
menof36go: It used to be a local clientside function that's used in a shared hook. That leaves us with four options:
Define a local var in shared realm and populate it with the function in the client realm
Make PlayerSprint global
Split the shared hook into a clientside and a serverside hook
Or do it the way I did it
All have benefits and disadvantages. Ill change it according to your preference.

saibotk: I thought about only defining it on the clientside, as it is secured by another `if CLIENT ...` below, and the function is never called serverside. I think it should be moved back into the previous position, where it was only defined on the client (the big `if SERVER then ... else .... here ... end`. This will also ensure that it is not callable on the server and fail when used otherwise.

gamemodes/terrortown/gamemode/shared/sh_main.lua
saibotk: Jeah thats a thing where we would need a good linter for :D but we should put variables at the top, so that they can be found easily. Thanks :+1:

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
menof36go: I think you misunderstood me. It's not only a matter of realm, but also a matter of variable scope. It's a local function and therefore cannot be referenced outside of the block it was defined in. Therefore, putting it back into the else block leads to a nil error, since the shared hook is outside of the else block.

saibotk: I see, then its ok as it is now. 👍🏻

gamemodes/terrortown/gamemode/shared/sh_main.lua
```diff
@@ -135,6 +136,13 @@ function GM:TTTPlayerColor(model)
 	return COLOR_WHITE
 end
 
+---
+-- @hook
+-- @realm shared
+function GM:Think()
+	UpdateSprint()
```
Alf21: The old sprint system was modular designed, to move it easilyy out. If you (and the other) agree connecting it together, I'm okay with it. Idk why we/I decided creating it modular, but we/I did :D (maybe it got already connected, I just woke up when I saw this line). @saibotk @TimGoll @lebroomer any reasons? (I thought about the option to use the ingame sprinting system of gmod or if there is another system, but this maybe doesn't make much sense)

PR #453, opened 2020-01-27
Issue template update

Histalek:
Updated the Issue template to better reflect what information is absolutely necessary.
Also changed the formatting and updated some information.


Alf21:
Hmm i‘m thinking about the need of git hashes... shouldn‘t it be enough to get information about the exact version, and how it was installed / put into folder / modified? I don‘t want that users keep away from reporting bugs being afraid of git commands

Histalek:
i also thought about just asking for the day of the clone/pull but that is still kinda vague. 
I agree that it might be off-putting using git commands if you don't know your way around them, but i would think that using the github version instead of the workshop would at least come with some knowledge of git.

Maybe we could think of more alternatives of providing the versioning. I will definitely think about it

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,34 +1,54 @@
...
+- "Your Version of TTT2"
+- "Describe the Bug"
+
+Also please make sure to read the [FAQ](https://github.com/TTT-2/TTT2/wiki/faq) first!
+
+## Your Version of TTT2 (mandatory)
+
+Do you use the Workshop Version of TTT2 or did you pull a Version from GitHub?  
+If the latter is the case please also provide the (shortened) commithash.  
+(executing `git log -n 1 --pretty=format:"%h"` in your Servers TTT2 repository will provide it to you)
```
saibotk: This is not always the case, as most people do not clone the repository and rather download the zip file. I think adding a hint, that this command is only possible when the repository was cloned would be enough. Otherwise a commit hash can only be given, if the user still knows when and what commit he downloaded.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,34 +1,54 @@
...
 A clear and concise description of what you expected to happen.
 
-**Screenshots**
-If applicable, add screenshots to help explain your problem.
+## Context (please provide as much as you can):
+
+- Collection-Link of your Server: [e.g. https://steamcommunity.com/sharedfiles/filedetails/?id=1888516630]
+- **All** Server-Errors in the server console
+- Client-Errors in the developer console
+- other Addons installed in your Servers `addons`-folder
```
saibotk: The first word in this and the following line should be uppercase

.github/ISSUE_TEMPLATE/bug_report.md
Histalek: I actually have not thought about people getting the zip ... 
We definitely need some clever and intuitive way to provide the versioning.

I also thought about asking to update to the latest version first before submitting an issue but then i would be worried that some bugs will just not be reported as fast (or not at all) and hence will be missed.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,34 +1,56 @@
...
+
+Also please make sure to read the [FAQ](https://github.com/TTT-2/TTT2/wiki/faq) first!
+
+## Your Version of TTT2 (mandatory)
+
+Leave one of the following:
+
+- Workshopversion from Steam
+- Latest GitHub Version
+- Non-latest GitHub Version (please provide what you can to specify)
```
Histalek: what do you think of this approach @saibotk @Alf21

saibotk: Thats better in my opinion

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,34 +1,56 @@
 ---
 name: Bug report
 about: Create a report to help us improve
-
 ---
 
-**Please make sure to read the [FAQ](https://github.com/Alf21/TTT2/wiki/faq) first!**
+## Disclaimer
+
+If you are missing the following information your Issue will be closed!
```
TimGoll: issue should be lower case here

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,34 +1,56 @@
...
-**Expected behaviour**
+## Expected behaviour
+
 A clear and concise description of what you expected to happen.
 
-**Screenshots**
-If applicable, add screenshots to help explain your problem.
+## Context (please provide as much as you can):
+
+- Collection-Link of your Server: [e.g. https://steamcommunity.com/sharedfiles/filedetails/?id=1888516630]
```
TimGoll: "Collection link"

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,34 +1,56 @@
...
+- **All** Server-Errors in the server console
```
TimGoll: "Server errors"

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,34 +1,56 @@
...
+- Client-Errors in the developer console
```
TimGoll: "Client errors"

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,34 +1,56 @@
...
+- Other Addons installed in your Servers `addons`-folder
```
TimGoll: "addons" "server's" "`addons` folder"

PR #454, opened 2020-01-27
reworked audio hooks

TimGoll:
I've reworked the audio hooks and cleaned up some elements

The hook `TTT2AvoidTeamVoiceChat` is now `TTT2CanSeeChat` since this hook had nothing todo with voice

Added the hook `TTT2CanHearVoiceChat` and modified `TTT2CanUseVoiceChat`

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_voice.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_voice.lua
@@ -28,7 +28,7 @@ g_VoicePanelList = nil
 local function VoiceTryEnable()
 	local client = LocalPlayer()
 
-	if not hook.Run("TTT2CanUseVoiceChat", client, false) then return false end
+	if hook.Run("TTT2CanUseVoiceChat", client, false) == false then return false end
 
 	if not VOICE.IsSpeaking() and VOICE.CanSpeak() then
 		VOICE.isTeam = false
@@ -53,7 +53,7 @@ end
 local function VoiceTeamTryEnable()
 	local client = LocalPlayer()
 
-	if not hook.Run("TTT2CanUseVoiceChat", client, true) then return false end
+	if hook.Run("TTT2CanUseVoiceChat", client, true) == false then return false end
 
 	if not IsValid(client) then return false end
 
@@ -104,17 +104,6 @@ local function VoiceNotifyThink(pnl)
 	local d = client:GetPos():Distance(pnl.ply:GetPos())
 
 	pnl:SetAlpha(math.max(-0.1 * d + 255, 15))
-end
-
----
--- Whether or not the @{Player} use the voice chat.
--- @param Player ply @{Player} who wants to use the voice chat
--- @param boolean isTeam Are they trying to use the team voice chat
--- @return boolean Whether or not the @{Player} can use the voice chat
--- @hook
--- @realm client
-function GM:TTT2CanUseVoiceChat(ply, isTeam)
-	return true
 end
 
 local PlayerVoicePanels = {}
--- a/gamemodes/terrortown/gamemode/client/cl_voice.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_voice.lua
@@ -28,7 +28,9 @@ g_VoicePanelList = nil
 local function VoiceTryEnable()
 	local client = LocalPlayer()
 
-	if hook.Run("TTT2CanUseVoiceChat", client, false) == false then return false end
+	if hook.Run("TTT2CanUseVoiceChat", client, false) == false then
+		return false
+	end
 
 	if not VOICE.IsSpeaking() and VOICE.CanSpeak() then
 		VOICE.isTeam = false
--- a/gamemodes/terrortown/gamemode/client/cl_voice.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_voice.lua
@@ -55,7 +55,9 @@ end
 local function VoiceTeamTryEnable()
 	local client = LocalPlayer()
 
-	if hook.Run("TTT2CanUseVoiceChat", client, true) == false then return false end
+	if hook.Run("TTT2CanUseVoiceChat", client, true) == false then
+		return false
+	end
 
 	if not IsValid(client) then return false end
 
--- a/gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
@@ -247,12 +247,8 @@ function GM:PlayerCanSeePlayersChat(text, teamOnly, reader, sender)
 		return false
 	end
 
-	if not IsValid(sender) then
-		if IsEntity(sender) then
-			return true
-		else
-			return false
-		end
+	if not IsValid(sender) and IsEntity(sender) then
+		return true
 	end
 
 	local sTeam = sender:Team() == TEAM_SPEC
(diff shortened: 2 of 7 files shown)


gamemodes/terrortown/gamemode/server/sv_voice.lua
```diff
@@ -128,7 +143,7 @@ local function RoleGlobalVoice(ply, isGlobal)
 	local tm = ply:GetTeam()
 
 	ply[tm .. "_gvoice"] = isGlobal
-	ply.blockVoice = hook.Run("TTT2CanUseVoiceChat", ply, tm)
+	ply.blockVoice = hook.Run("TTT2CanUseVoiceChat", ply, not isGlobal) == false
```
Alf21: Where is this being used?

gamemodes/terrortown/gamemode/client/cl_voice.lua
```diff
@@ -53,7 +55,7 @@ end
 local function VoiceTeamTryEnable()
 	local client = LocalPlayer()
 
-	if not hook.Run("TTT2CanUseVoiceChat", client, true) then return false end
+	if hook.Run("TTT2CanUseVoiceChat", client, true) == false then return false end
```
Alf21: Return false in new line. I didn‘t mentioned it before because I thought you will look on your own into these files so that I don‘t need to spam ^^ nvm :p

gamemodes/terrortown/gamemode/server/sv_voice.lua
TimGoll: fabian added this hook a few days ago and added `tm` on server, while having a bool on clients. He said this was a bug

gamemodes/terrortown/gamemode/client/cl_voice.lua
TimGoll: This wasn't my code :P

gamemodes/terrortown/gamemode/server/sv_voice.lua
Alf21: So blockvoice is used, right? I‘m just wanna keep sure that there aren‘t any inconsistencies or unnecessary vars

TimGoll: I have no idea, but I think so. He said there was a double bug (missing `== false` and wrong variable) which lead to it still working

Alf21: @menof36go maybe this should clean up the situation :D

menof36go: blockVoice can be used to block a sender from using the voice chat. This could easily be used to replace the wrong hook the spy was using for example. blockVoice itself is used in PlayerCanHearPlayersVoice. It has to be a flag, otherwhise we would have to run the hook for every listener as well, which is quite costly

gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
```diff
@@ -233,48 +233,56 @@ end
...
 		return false
 	end
 
-	if not IsValid(speaker) then
-		if IsEntity(speaker) then
-			return true
-		else
-			return false
-		end
+	if not IsValid(sender) and IsEntity(sender) then
```
saibotk: What if the sender is not "valid" and also is not an entity? Then the `sender:Team()` call would error, wouldn't it? 
Also where is the `return false`? Or rather does this break anything?

gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
```diff
@@ -233,48 +233,56 @@ end
...
 		return true
 	end
 
 	return false
 end
 
+---
+-- Whether or not the @{Player} can receive the chat message.
+-- @param Player reader @{Player} who can receive chat
+-- @param Player isTeam @{Player} who speaks
```
saibotk: Should not be called `isTeam`, also wouldn't it rather be `... who chats` instead of `... who speaks`?

gamemodes/terrortown/gamemode/shared/sh_voice.lua
```diff
@@ -0,0 +1,14 @@
+---
+-- @section voice_manager
+
+---
+-- Whether or not the @{Player} use the voice chat. Hint: Has to be registered
```
saibotk: `Whether or not the @{Player} use the voice chat.` -> `Whether or not the @{Player} **can** use the voice chat.`

`Hint: Has to be registered` should be more specific, what do you mean? Or remove it if it is not that important.

gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
TimGoll: all hooks used `isTeam`, I just used the same wording

yes

gamemodes/terrortown/gamemode/shared/sh_voice.lua
TimGoll: The hook has to be registered on both client and server to make it work, see the next line

gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
saibotk: This is still wrong, as this is the player who speaks and not the bool, which is described in the line below, so the variable name is wrong ;)

gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
TimGoll: I have no idea, this code was in there previously, I just removed the part Alf wanted to have removed

gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
TimGoll: oh, I see now, this is a typo! thanks

gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
Alf21: But you removed the `IsEntity` part too. Maybe you should keep this as far as you don't know what it changes

gamemodes/terrortown/gamemode/shared/sh_voice.lua
Alf21: That isn't the code style or doc style we should go. Give a look into `@warning` or `@realm`

TimGoll: There is already `@realm shared` i just wanted to mention that the hook has to be registered on both client and server to function properly. I guess I could add a `@warning` though

TimGoll: There is already `@realm shared` i just wanted to mention that the hook has to be registered on both client and server to function properly. I guess I could add a `@warning` though

gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
TimGoll: Yes, I had a discussion with saibotk and he wanted to have it removed. But maybe I misunderstood him there. Like I said to him, I don't know about the purpose of this codeelement.

gamemodes/terrortown/gamemode/shared/sh_voice.lua
Alf21: That's a normal behaviour if using predicted hooks with different values on client and server.
Same with sprinting

PR #455, opened 2020-01-31
added preventSprint to doubletap feature

TheNickSkater:
I think preventSprint was forgotten here, added it back

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/sh_sprint.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_sprint.lua
@@ -73,7 +73,7 @@ else
 
 	function UpdateInputSprint(ply, key, pressed)
 		if pressed then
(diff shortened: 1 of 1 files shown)


Alf21:
@menof36go

menof36go:
I totally forgot about the flag. Luckily it seems to be unused as of now, so nobody should have experienced issues regarding the lack of the flag yet. I approve of this change

saibotk:
A general question: 
What does `preventSprint` do other than preventing the Sprint clientside? Or rather, what is the intended use case? @menof36go maybe you can give some insight. Id appreciate that

menof36go:
> 
> 
> A general question:
> What does `preventSprint` do other than preventing the Sprint clientside? Or rather, what is the intended use case? @menof36go maybe you can give some insight. Id appreciate that

Nothing, but it has always been there it seems. I'd just keep it for compatibility

PR #458, opened 2020-02-07
reworked targetID

TimGoll:
targetID works still the same, but ist now easier to use. The `TTTRenderEntityInfo` hook now receives a `TARGET_DATA` object that has many helper functions to make the addon devs life easier.

```lua
---
-- Add targetid info to a focused entity
-- @param @{TARGET_DATA} tdata The @{TARGET_DATA} data object to
-- which contains all information
-- @hook
-- @realm client
function GM:TTTRenderEntityInfo(tdata)

end
```
Additionally I added support for inline icons:
image


gamemodes/terrortown/gamemode/client/cl_target_data.lua
```diff
@@ -0,0 +1,164 @@
...
+-- Enables/Disables the outline around a focused entity
+-- @param [default=true] boolean enb_outline A boolean defining the outline state
+-- @realm client
+function TARGET_DATA:EnableOutline(enb_outline)
+	self.params.drawOutline = enb_outline or true
+end
+
+---
+-- Enables/Disables the targetID text and icons
+-- @param [default=true] boolean enb_text A boolean defining the text stat
```
Alf21: *`state`

Alf21: Could you pls explain why needing this function?

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -618,115 +678,129 @@ function HUDDrawTargetIDPlayers(data, params)
...
+			GetPT("corpse_hint", key_params)
+		)
 	elseif binoculars_useable then
-		params.displayInfo.subtitle.text = GetPT("corpse_binoculars", {key = Key("+attack", "ATTACK")})
+		tdata:SetSubtitle(
+			GetPT("corpse_binoculars", {key = Key("+attack", "ATTACK")})
+		)
 	else
-		params.displayInfo.subtitle.text = TryT("corpse_too_far_away")
+		tdata:SetSubtitle(
```
Alf21: one-line style as talked about :)

gamemodes/terrortown/gamemode/client/cl_target_data.lua
TimGoll: In case someone wants only an outline and no text

Alf21: and why not just leaving the text empty?

TimGoll: Because it is implemented this way. It was the same since targetID got implemented. I thought it would be the most straightforward way to enable/disable targetID

Especially if one addon adds the text, another addon can disable it without having to bother with the raw data.

Thinking about it, I should change it probably in a way that targetID isn't displayed once one addon sets it to false. What do you think about this?

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -274,34 +275,43 @@ function GM:HUDDrawTargetID()
 
 	-- preset a table of values that can be changes with a hook
```
Histalek: "changed"

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -402,150 +434,171 @@ function GM:HUDDrawTargetID()
 	draw.ShadowedLine(spacer_line_x, spacer_line_y, spacer_line_x, spacer_line_y + spacer_line_l, COLOR_WHITE)
 end
 
+---
+-- Add targetid info to a focused entity
+-- @param @{TARGET_DATA} tdata The @{TARGET_DATA} data object to which contains all information
```
Histalek: the "to" seems wrong here, but i'm not 100% sure

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -402,150 +434,171 @@ function GM:HUDDrawTargetID()
 	draw.ShadowedLine(spacer_line_x, spacer_line_y, spacer_line_x, spacer_line_y + spacer_line_l, COLOR_WHITE)
 end
 
+---
+-- Add targetid info to a focused entity
```
Histalek: "targetID" to be consistent with other comments

gamemodes/terrortown/gamemode/client/cl_target_data.lua
Alf21: If any add-on wanna prevent it (returning false), this prevention should count. imho.

TimGoll: yep, I've already changed it yesterday

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -618,115 +671,123 @@ function HUDDrawTargetIDPlayers(data, params)
...
+	-- enable targetID rendering
+	tData:EnableText()
+
+	-- add title and subtitle to the focused ent
+	local h_string, h_color = util.HealthToString(ent:Health(), ent:GetMaxHealth())
 
-	local h_string, h_color = util.HealthToString(data.ent:Health(), data.ent:GetMaxHealth())
+	tData:SetTitle(
+		ent:Nick() .. " " .. (disguised and string.upper(TryT("target_disg")) or ""),
+		disguised and COLOR_RED
```
saibotk: Here we might be missing a `or nil`

PR #461, opened 2020-02-09
Additional target id stuff

TimGoll:
Added targetID support for doors and expanded the entity metatable about door related stuff

image
Hint: In this image I accidentially interchanged the language strings.

This PR can only be merged after https://github.com/TTT-2/TTT2/pull/458 is merged. [Edit: It is ready to be merged]


gamemodes/terrortown/gamemode/shared/sh_entity_ext.lua
```diff
@@ -0,0 +1,184 @@
...
+-- @module Entity
+-- @ref https://wiki.facepunch.com/gmod/Entity
+-- @desc shared extensions to entity table
+
+local entmeta = FindMetaTable("Entity")
+if not entmeta then
+	assert(entmeta, "FAILED TO FIND ENTITY TABLE")
+
+	return
+end
```
menof36go:
```lua
local entmeta = assert(FindMetaTable("Entity"), "FAILED TO FIND ENTITY TABLE")
```

gamemodes/terrortown/gamemode/shared/sh_include.lua
```diff
@@ -122,7 +124,7 @@ function ttt_include(filename)
 	local fd = TTTFiles[filename]
 
 	if not fd then
-		error("[TTT2][ERROR] Tried to include a not existant file" .. filename)
+		error("[TTT2][ERROR] Tried to include a not existant file " .. filename)
```
menof36go: I don't think existant exists (no pun intended). non-existent or missing would be better.
Sth. like "[TTT2][ERROR] Tried to include a the non-existent file " .. filename or
"[TTT2][ERROR] Tried to include missing file " .. filename

menof36go:
The new target id system is an eye catcher, glad you keep improving it further. It seems subtle but it provides a very neat modern look

gamemodes/terrortown/gamemode/shared/sh_entity_ext.lua
TimGoll: Good to know, thanks! Why is it like this?

menof36go: If the first parameter is nil or false, it throws an error using the second parameter as message. If the first parameter is neither, it is returned instead

gamemodes/terrortown/gamemode/shared/sh_entity_ext.lua
```diff
@@ -0,0 +1,179 @@
...
+		if not ent:IsDoor() then continue end
+
+		door.__doors[#door.__doors + 1] = ent
+
+		-- set up synced states if on server
+		if CLIENT then continue end
+
+		ent:SetNWBool("ttt2_door_locked", ent:GetInternalVariable("m_bLocked") or false)
+		ent:SetNWBool("ttt2_door_forceclosed", ent:GetInternalVariable("forceclosed") or false)
+		ent:SetNWBool("ttt2_door_forceclosed", ent:GetInternalVariable("forceclosed") or false)
```
saibotk: duplicate line

gamemodes/terrortown/gamemode/shared/sh_entity_ext.lua
```diff
@@ -0,0 +1,179 @@
+--- DOORS MODULE STUFF ---
```
saibotk: Missing `@author`.
Should start with `---` followed by `-- This is a short description ..... `

gamemodes/terrortown/gamemode/shared/sh_door.lua
```diff
@@ -0,0 +1,184 @@
...
+---
+-- @module Entity
+-- @author Mineotopia
+-- @ref https://wiki.facepunch.com/gmod/Entity
+-- @desc shared extensions to entity table
+
+local entmeta = assert(FindMetaTable("Entity"), "FAILED TO FIND ENTITY TABLE")
+
+---
+-- Return wether this entiy is a door or not
```
Histalek: "Returns" "whether" "entity"

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -448,6 +449,33 @@ local key_params = {
 	walkkey = Key("+walk", "WALK")
 }
 
+-- handle looking at doors
+function HUDDrawTargetIDDoors(tData)
+	local client = LocalPlayer()
+	local ent = tData:GetEntity()
+
+	if not IsValid(client) or not client:IsTerror() or not client:Alive()
```
Alf21: If using a player metatable based function on an entity, you should check whether this ent is a player first.

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -448,6 +449,33 @@ local key_params = {
...
+	or not IsValid(ent) or not ent:IsDoor() or ent:IsDoorForceclosed() or tData:GetEntityDistance() > 100 then
```
Alf21: Two times `not IsValid`, this can be simplified with brackets

Alf21: Forceclosed or ForceClosed?

gamemodes/terrortown/gamemode/shared/sh_door.lua
```diff
@@ -0,0 +1,184 @@
+--- DOORS MODULE STUFF ---
+
+---
+-- @module door
+-- @author Mineotopia
+-- @desc A bunch of functions that handle all doors found on a map
+
+door = {}
+door.__doors = {}
```
Alf21: We never did anything related to this var naming, uff. Welcome to C

gamemodes/terrortown/gamemode/shared/sh_door.lua
```diff
@@ -0,0 +1,184 @@
...
+		timer.Create("ttt2_recheck_door_use_" .. ent:EntIndex(), 2.5, 1, function()
+			if not IsValid(ent) or ent:IsDoorLocked() then return end
+
+			ent:SetNWBool("ttt2_door_open", ent:InternalIsDoorOpen() or false)
+		end)
+	end
+end
+
+
+--- ENTITY EXTENSION STUFF ---
```
Alf21: Both in one file?

gamemodes/terrortown/gamemode/client/cl_targetid.lua
TimGoll: I don't thinkt it is better or more readable if combined in brackets as `not (IsValid(client) and IsValid(ent))`. I personally like the current approach better from a readability standpoint.

Forceclosed since the internal variable is named that way

gamemodes/terrortown/gamemode/shared/sh_door.lua
TimGoll: How do I name an internal car to make it clear that it should not be accessed directly. I guess I could make an internal local variable

gamemodes/terrortown/gamemode/shared/sh_door.lua
TimGoll: Yes since they both need access to the same local variables. I could make them public or define them twice. I don't like both approaches

gamemodes/terrortown/gamemode/client/cl_targetid.lua
saibotk: I do not understand what you mean, as the client (the entity we are talking about?) is retrieved using the `LocalPlayer()` function, thus i do not think an additional check for a player entity is needed. Or i might not understand your comment and you would need to further explain this.

gamemodes/terrortown/gamemode/shared/sh_door.lua
Alf21: That should be the way to prevent access

gamemodes/terrortown/gamemode/shared/sh_door.lua
Alf21: Ref. the previous comment, the current solution should be the best solution, you are right, but it‘s unconventional ^^

gamemodes/terrortown/gamemode/shared/sh_door.lua
```diff
@@ -113,7 +125,9 @@ local entmeta = assert(FindMetaTable("Entity"), "FAILED TO FIND ENTITY TABLE")
 -- @return boolean Returns true if it is a valid door
 -- @realm shared
 function entmeta:IsDoor()
-	if IsValid(self) and door.__valid_doors[self:GetClass()] then
+	local valid_door_cls = door.GetValid()
```
Alf21: why did you localized it?

TimGoll: to access it with `[self:GetClass()]`

Alf21: `door.GetValid()[self:GetClass()]` wouldn't work? ^^

TimGoll: I haven't tested it. I thought it wouldn't^^ I can test it

PR #462, opened 2020-02-10
Some more shop func rework

menof36go:
Removed another table.HasValue
Move affordability into EquipmentIsBuyable for consistency
Move CanOrder hook into EquipmentIsBuyable (opens it up to shared realm)
Add Hide hook, which can be used to hide any icon clientside. This turns out to be pretty useful in combination with CanOrder
Changing RandomShops to X now provides you with X random items + .NoRandom items, rather than X items total. This seems more intuitive. If I want an item to be non-random, I don't want less random items, I just want to make that one item always accessible. Currently if you have 5 .NoRandom items and RandomShops set to 4, you will have 0 random items in your shop, which is inconsistent and rather strange. Also, if you add an addon and that addon happens to contain equipment that is flagged .NoRandom, it shouldn't require you to readjust your shop settings
Add some default sorting to the shop

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_equip.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_equip.lua
@@ -313,7 +313,7 @@ local function CreateEquipmentList(t)
 	for k = 1, #itms do
 		local item = itms[k]
 
-		local buyable, icon, message = EquipmentIsBuyable(item, ply)
+		local buyable = EquipmentIsBuyable(item, ply)
 
 		local hide = hook.Run("TTT2HideEquipment", ply, item.ClassName, items.IsItem(item.ClassName), buyable)
 
--- a/gamemodes/terrortown/gamemode/client/cl_equip.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_equip.lua
@@ -273,13 +273,8 @@ local function CreateEquipmentList(t)
 		local wep = weps[i]
 
 		if wep.IsEquipment and wep:IsEquipment() then
-			owned_ids[#owned_ids + 1] = wep:GetClass()
+			owned_ids[wep:GetClass()] = true
 		end
-	end
-
-	-- Stick to one value for no equipment
-	if #owned_ids == 0 then
-		owned_ids = nil
 	end
 
 	local itms = {}
@@ -382,13 +377,8 @@ local function CreateEquipmentList(t)
 				-- If we cannot order this item, darken it
 				if not t.role and ((
 						-- already owned
-						table.HasValue(owned_ids, item.id)
-						or items.IsItem(item.id) and item.limited and ply:HasEquipmentItem(item.id)
-						-- already carrying a weapon for this slot
-						or ItemIsWeapon(item) and not CanCarryWeapon(item)
+						owned_ids[item.ClassName]
 						or not EquipmentIsBuyable(item, ply)
-						-- already bought the item before
-						or item.limited and ply:HasBought(item.id)
 					) or (item.credits or 1) > credits
 				) then
 					ic:SetIconColor(color_darkened)
--- a/gamemodes/terrortown/gamemode/client/cl_equip.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_equip.lua
@@ -39,7 +39,6 @@ local color_darkened = Color(255, 255, 255, 80)
 local eqframe = eqframe
 local dlist = dlist
 local curSearch = curSearch
-local SafeTranslate = LANG.TryTranslation
 
 --
 --     GENERAL HELPER FUNCTIONS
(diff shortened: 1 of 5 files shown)


menof36go:
I have tested it on my own, but I'd rather play a session or two with some friends first to see if they notice anything off

gamemodes/terrortown/gamemode/server/sv_shop.lua
```diff
@@ -97,18 +97,7 @@ local function OrderEquipment(ply, cls)
...
-	if message then
-		LANG.Msg(ply, message, nil, MSG_MSTACK_ROLE)
-	end
-
-	if allow == false then return end
-
-	if not ignoreCost then
-		ply:SubtractCredits(credits)
-	end
+	ply:SubtractCredits(credits)
```
saibotk: You missed the `ignoreCost` case, as the player is not charged any credits when ignoreCost is true

gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -273,26 +272,36 @@ local function CreateEquipmentList(t)
...
+			-- We need them anyways and this way we don't have to calculate them multiple times when sorting
+			tmp[i].cachedEquipmentTranslation = GetEquipmentTranslation(tmp[i].name, tmp[i].PrintName)
 			itms[#itms + 1] = tmp[i]
 		end
 	end
 
 	if #itms == 0 and not t.notalive then
 		ply:ChatPrint("[TTT2][SHOP] You need to run 'shopeditor' as admin in the developer console to create a shop for this role. Link it with another shop or click on the icons to add weapons and items to the shop.")
+	else
+		table.sort(itms, function(itmA, itmB)
```
saibotk: It might be useful to define this function somewhere, so it is easier to find / change and maybe even just use a reference, to be able to swap out sorting functions in the long run.

gamemodes/terrortown/gamemode/server/sv_shop.lua
menof36go: When I added that hook in #441 that was the intention, but it was problematic in a sense. You could let someone order something and then give it to them for free. You could however not let them order something they couldn't afford in the first place. Maybe EquipmentIsBuyable should just return an additional boolean ignoreCost, that we can use in the OrderEquipment function. That way we would keep the functionality and still open it up to shared realm. What do you think?

menof36go:
Before anyone mentions it: Yes, I did forget to remove the ignoreCost parameter from TTT2OrderedEquipment but that might be resolved soon when we have decided on what to with ignoreCost

gamemodes/terrortown/gamemode/server/sv_shop.lua
TimGoll: I'm not deep in this topic, but it sounds good to me

gamemodes/terrortown/gamemode/client/cl_equip.lua
menof36go: I just found a function that does what mine does in sh_init, but it was only ever used to sort the shopeditor. I dislike the way it sorts though, since it relies solely on the class name. I'll replace it with mine and test if it breaks anything
https://github.com/TTT-2/TTT2/blob/dc859bc6f4efb7a642a6443f337618fb71c3ef93/gamemodes/terrortown/gamemode/shared/sh_init.lua#L416-L438

gamemodes/terrortown/gamemode/server/sv_shop.lua
saibotk: I don't really like this approach, as the function is not meant to judge about whether or not the cost is to be ignored. The function is even called EquipmentIsBuyable. So we might have to think about another way to address this, but i get why you want to change it and that should definitely be addressed.

gamemodes/terrortown/gamemode/client/cl_equip.lua
saibotk: I wanted to emphasize on the possible variable solution to store a sorting function in, so it can be swapped out easily (even programmatcally). I did not want to say you should find an already defined function to replace yours with, just for optimization. You can still keep yours, thats fine, but just define it somewhere in the file instead and use a reference to it (saved on a variable)

gamemodes/terrortown/gamemode/server/sv_shop.lua
menof36go: All alternative I can think of lead to code redundancy

saibotk: Redundancy is not the ultimate evil, as sometimes redundancy is better than trying to "eliminate" it, by making the code less flexible or even much more unreadable etc... 
We should still think about another way or even rethink the whole function etc.

TimGoll:
Is there any news on this?

PR #464, opened 2020-02-11
Fix colleague color bug

menof36go:
Fixes the bug that your teammates will always show up as if they were talking in teamchat, even when talking in global chat

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_voice.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_voice.lua
@@ -120,11 +120,6 @@ function GM:TTT2CanHearVoiceChat(listener, speaker, isTeam)
 end
 
 local function SendRoleVoiceState(speaker)
-	local tm = speaker:GetTeam()
-	local isGlobal = speaker[tm .. "_gvoice"]
-
-	if isGlobal then return end
-
 	-- send umsg to living traitors that this is traitor-only talk
 	local rf = GetTeamMemberFilter(speaker, true)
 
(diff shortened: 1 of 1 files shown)


TimGoll:
Could you explain a bit about this bug?

menof36go:
> 
> 
> Could you explain a bit about this bug?

After reworking some of the voice chat functions, a player would always show up as speaking in team voice chat to their colleagues, even if they were actually just speaking in global chat. A traitor speaking in global voice would still show up red to his teammates, which would obviously confuse them.

TimGoll:
> > Could you explain a bit about this bug?
> 
> After reworking some of the voice chat functions, a player would always show up as speaking in team voice chat to their colleagues, even if they were actually just speaking in global chat. A traitor speaking in global voice would still show up red to his teammates, which would obviously confuse them.

Interesting. I've never noticed this during testing. Thanks for fixing!

saibotk:
I actually dont think that this is a bug to be honest

TimGoll:
> I actually dont think that this is a bug to be honest

Why should a traitor think his colleague speaks in the team chat if he is speaking in public voice?

saibotk:
I would need some more explanation on this, what was wrong with the colors? Is it different from the normal TTT now? Is this a regression?

TimGoll:
> I would need some more explanation on this, what was wrong with the colors? Is it different from the normal TTT now? Is this a regression?

@menof36go said

> After reworking some of the voice chat functions, a player would always show up as speaking in team voice chat to their colleagues, even if they were actually just speaking in global chat. A traitor speaking in global voice would still show up red to his teammates, which would obviously confuse them.

Based on this I would say in vanilla TTT it has always been like this and we just recently broke it. I've not used the voicechat much, but I'm quite sure that you could differentiate between team and global chat.

PR #466, opened 2020-02-12
added addon language file support

TimGoll:
In an attempt to clean up the chaotic language handling in TTT2 addons, I added language file support for addons. After this merge, addons can place their language files in `lua/lang/`.

Example for the jester:
image

```lua
L = LANG.GetLanguageTable("english")

L[JESTER.name] = "Jester"
L[JESTER.defaultTeam] = "TEAM Jesters"
-- ...
```

The TTT2 language files stay untouched and are loaded on file load. However addon language files are loaded in the `GM:Initialialize` hook to make sure all files are ready and available. The files are included prior to `LANG.Init()`.

I also fixed the docs for `function LANG.GetUnsafeNamed(name)`

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_lang.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_lang.lua
@@ -167,12 +167,29 @@ function LANG.GetUnsafeLanguageTable()
 end
 
 ---
--- Returns a translated @{string} text (if possible) directly from the language table
--- @param string name string key identifier for the translated @{string}
--- @return nil|string
+-- Returns the reference to a language table if it exists
+-- @param string name string key identifier for the language
+-- @return nil|table
 -- @realm client
-function LANG.GetUnsafeNamed(name)
-	return LANG.Strings[name]
+function LANG.GetUnsafeNamed(lang_name)
+	lang_name = lang_name and string.lower(lang_name)
+
+	return LANG.Strings[lang_name]
+end
+
+---
+-- Returns the reference to a language table if it exists, creates a new language if it did not exist
+-- @param string name string key identifier for the language
+-- @return nil|table
+-- @realm client
+function LANG.GetOrCreateLanguage(lang_name)
+	lang_name = lang_name and string.lower(lang_name)
+
+	if not LANG.Strings[lang_name] then
+		LANG.CreateLanguage(lang_name)
+	end
+
+	return LANG.Strings[lang_name]
 end
 
 ---
--- a/gamemodes/terrortown/gamemode/client/cl_lang.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_lang.lua
@@ -66,7 +66,7 @@ end
 -- @return string the inserted string_name parameter
 -- @realm client
 function LANG.AddToLanguage(lang_name, string_name, string_text)
-	lang_name = lang_name and string.lower(lang_name)
+	lang_name = lang_name and string.lower(lang_name) or nil
 
 	if not LANG.IsLanguage(lang_name) then
 		ErrorNoHalt(Format("Failed to add '%s' to language '%s': language does not exist.\n", tostring(string_name), tostring(lang_name)))
@@ -172,7 +172,7 @@ end
 -- @return nil|table
 -- @realm client
 function LANG.GetUnsafeNamed(lang_name)
-	lang_name = lang_name and string.lower(lang_name)
+	lang_name = lang_name and string.lower(lang_name) or nil
 
 	if not LANG.IsLanguage(lang_name) then
 		ErrorNoHalt(Format("Failed to get '%s': language does not exist.\n", tostring(lang_name)))
@@ -188,8 +188,8 @@ end
 -- @param string name string key identifier for the language
 -- @return nil|table
 -- @realm client
-function LANG.GetLanguageTable(lang_name)
-	lang_name = lang_name and string.lower(lang_name)
+function LANG.GetLanguageTableReference(lang_name)
+	lang_name = lang_name and string.lower(lang_name) or nil
 
 	if not LANG.IsLanguage(lang_name) then
 		LANG.CreateLanguage(lang_name)
@@ -239,7 +239,7 @@ end
 -- @param string lang_name the new language name
 -- @realm client
 function LANG.SetActiveLanguage(lang_name)
-	lang_name = lang_name and string.lower(lang_name)
+	lang_name = lang_name and string.lower(lang_name) or nil
 
 	if LANG.IsLanguage(lang_name) then
 		local old_name = LANG.ActiveLanguage
@@ -302,7 +302,7 @@ end
 function LANG.IsLanguage(lang_name)
 	if not lang_name then return end
 
-	return LANG.Strings[string.lower(lang_name)]
+	return LANG.Strings[string.lower(lang_name)] ~= nil
 end
 
 local function LanguageChanged(cv, old, new)
--- a/gamemodes/terrortown/gamemode/client/cl_lang.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_lang.lua
@@ -174,6 +174,12 @@ end
 function LANG.GetUnsafeNamed(lang_name)
 	lang_name = lang_name and string.lower(lang_name)
 
+	if not LANG.IsLanguage(lang_name) then
+		ErrorNoHalt(Format("Failed to get '%s': language does not exist.\n", tostring(lang_name)))
+
+		return
+	end
+
 	return LANG.Strings[lang_name]
 end
 
@@ -182,10 +188,10 @@ end
 -- @param string name string key identifier for the language
 -- @return nil|table
 -- @realm client
-function LANG.GetOrCreateLanguage(lang_name)
+function LANG.GetLanguageTable(lang_name)
 	lang_name = lang_name and string.lower(lang_name)
 
-	if not LANG.Strings[lang_name] then
+	if not LANG.IsLanguage(lang_name) then
 		LANG.CreateLanguage(lang_name)
 	end
 
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -192,6 +192,9 @@ function GM:Initialize()
 
 	hook.Run("TTT2FinishedLoading")
 
+	-- load addon language files
+	LANG.SetupFiles("lang/")
+
 	ShopEditor.SetupShopEditorCVars()
 	ShopEditor.CreateShopDBs()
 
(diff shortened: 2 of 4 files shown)


gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -167,12 +167,29 @@ function LANG.GetUnsafeLanguageTable()
...
+end
+
+---
+-- Returns the reference to a language table if it exists, creates a new language if it did not exist
+-- @param string name string key identifier for the language
+-- @return nil|table
 -- @realm client
-function LANG.GetUnsafeNamed(name)
-	return LANG.Strings[name]
+function LANG.GetOrCreateLanguage(lang_name)
```
Alf21: This should be named like `LANG.GetLanguage` or `LANG.GetLanguageTable`. It doesn‘t matter whether it internally creates and/or just returns the table.

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -167,12 +167,29 @@ function LANG.GetUnsafeLanguageTable()
...
+	lang_name = lang_name and string.lower(lang_name)
```
Alf21: or?

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -167,12 +167,29 @@ function LANG.GetUnsafeLanguageTable()
...
+function LANG.GetUnsafeNamed(lang_name)
+	lang_name = lang_name and string.lower(lang_name)
+
+	return LANG.Strings[lang_name]
+end
+
+---
+-- Returns the reference to a language table if it exists, creates a new language if it did not exist
+-- @param string name string key identifier for the language
+-- @return nil|table
```
Alf21: You should add whether it‘s the raw or a copied table

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -192,6 +192,9 @@ function GM:Initialize()
 
 	hook.Run("TTT2FinishedLoading")
 
+	-- load addon language files
```
Alf21: This comment ist wrong, if you wanna load lang files on server, you are doing something wrong.
You should just make it available for the clients

gamemodes/terrortown/gamemode/shared/sh_lang.lua
```diff
@@ -25,21 +25,25 @@ local net = net
...
+		if string.Right(fname, 3) == "lua" then
+			util.IncludeClientFile(path)
 
-		MsgN("Included TTT language file: " .. fname)
+			MsgN("Included TTT language file: " .. fname)
+		end
 	end
 end
 
+-- load default TTT2 language files
```
Alf21: You are including in sv_ and cl_main too, why here again?

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -167,12 +167,29 @@ function LANG.GetUnsafeLanguageTable()
...
 ---
--- Returns a translated @{string} text (if possible) directly from the language table
--- @param string name string key identifier for the translated @{string}
--- @return nil|string
+-- Returns the reference to a language table if it exists
+-- @param string name string key identifier for the language
+-- @return nil|table
+-- @realm client
+function LANG.GetUnsafeNamed(lang_name)
+	lang_name = lang_name and string.lower(lang_name)
```
Alf21: Or?

gamemodes/terrortown/gamemode/client/cl_lang.lua
TimGoll: I copied it from somewhere else in this file. I think it is to make sure the language stays nil if you forget to set `lang_name`

gamemodes/terrortown/gamemode/server/sv_main.lua
TimGoll: ? The language systems is run on client and server since the files have to be added with `AddCSLuaFile`.

gamemodes/terrortown/gamemode/shared/sh_lang.lua
TimGoll: I have only put this in a function. The TTT language system is designed this way. Not by me.

gamemodes/terrortown/gamemode/client/cl_lang.lua
TimGoll: same as above

gamemodes/terrortown/gamemode/client/cl_lang.lua
TimGoll: "Returns the reference to a language table"

gamemodes/terrortown/gamemode/client/cl_lang.lua
Alf21: Then you should add `or nil` as already talked about. But following the next try of indexing nil seems to be wrong imo

gamemodes/terrortown/gamemode/client/cl_lang.lua
Alf21: after the `@return` as the line i’ve created the comment

gamemodes/terrortown/gamemode/server/sv_main.lua
Alf21: Read the comment in detail. Making something available / downloadable isn‘t the same as loading a file

gamemodes/terrortown/gamemode/client/cl_lang.lua
Alf21: Same as above

gamemodes/terrortown/gamemode/shared/sh_lang.lua
Alf21: But you are doing this 3 times, one time shared, one time for client, one time for server. That‘s two times for esch realm

gamemodes/terrortown/gamemode/client/cl_lang.lua
TimGoll: I see your point. But it was like this forever. And I don't want to change it without knowing why it was done like this

gamemodes/terrortown/gamemode/shared/sh_lang.lua
TimGoll: no. Look at the paths. It's once done on client and server for the TTT2 language files and once on client an server for addon language files

gamemodes/terrortown/gamemode/client/cl_lang.lua
Alf21: It was done like this because the whole lang thing was a unstable construct you already mentioned on your own. As you did in the other functions, I suggest creating an own error to prevent some weird nil error. Just a suggestions as it improves the development and prevent issues. Additional checks here shouldn‘t lower the performance. If you don‘t think so, feel free to resolve this :)

gamemodes/terrortown/gamemode/client/cl_lang.lua
Alf21: The `or nil` we already talked about is missing

gamemodes/terrortown/gamemode/shared/sh_lang.lua
```diff
@@ -24,22 +24,53 @@ util.IncludeClientFile("terrortown/gamemode/client/cl_lang.lua")
...
+-- server and client!
+-- @param string lang_path The path to search in
+-- @param boolean deepsearch If true, subfolders are scanned
+-- @internal
+-- @realm shared
+function LANG.SetupFiles(lang_path, deepsearch)
+	local file_paths = {}
+
+	if deepsearch then
+		local sub_folders = ({file.Find(lang_path .. "*", "LUA")})[2] or {}
```
Alf21: haha smart one :D But here isn't a need for inline things, better use: `local _, sub_folders = file.Find(lang_path .. "*", "LUA") or (nil, {})`

But why should subfolders not be an empty table after calling this function?

gamemodes/terrortown/gamemode/shared/sh_lang.lua
```diff
@@ -24,22 +24,53 @@ util.IncludeClientFile("terrortown/gamemode/client/cl_lang.lua")
...
+
+		for k = 1, #sub_folders do
+			local subname = sub_folders[k]
+			local files = ({file.Find(lang_path .. subname .. "/*.lua", "LUA")})[1] or {}
```
Alf21: same as above: `local files = file.Find(lang_path .. subname .. "/*.lua", "LUA") or {}`

gamemodes/terrortown/gamemode/shared/sh_lang.lua
```diff
@@ -24,22 +24,53 @@ util.IncludeClientFile("terrortown/gamemode/client/cl_lang.lua")
...
+
+			for i = 1, #files do
+				file_paths[#file_paths + 1] = lang_path .. subname .. "/" .. files[i]
+			end
+		end
+	else
+		local files = ({file.Find(lang_path .. "*.lua", "LUA")})[1] or {}
```
Alf21: and here? :p

gamemodes/terrortown/gamemode/shared/sh_lang.lua
TimGoll: I saved a `if not sub_folders then return end` with this one

Alf21: yea and doing useless stuff :D

TimGoll: okay, I'll rewert it. I'll finish up the door PR first though

Alf21: ref.: https://wiki.facepunch.com/gmod/file.Find this seems to just return nil if the path is invalid, theoretically an empty table if there is no content (already). So I think this OR is useless, what do you think?

TimGoll: no I tested it. `#` fails if the table is nil.

Alf21: Maybe because https://wiki.facepunch.com/gmod/file.Exists would return false. YOu don't need to call this function additionally. Your solution should work, but doing useless stuff is a bad coding way, it's unreliable. If there is something that is invalid, it should be returned. But you are right too, if the dict is valid but empty, it would so the same. So I suggest using your solution but with the improved / simplified syntax :)

gamemodes/terrortown/gamemode/shared/sh_lang.lua
```diff
@@ -38,18 +38,24 @@ function LANG.SetupFiles(lang_path, deepsearch)
...
+		local _, sub_folders = file.Find(lang_path .. "*", "LUA")
+
+		if not sub_folders then return end
 
 		for k = 1, #sub_folders do
 			local subname = sub_folders[k]
-			local files = ({file.Find(lang_path .. subname .. "/*.lua", "LUA")})[1] or {}
+			local files = file.Find(lang_path .. subname .. "/*.lua", "LUA")
+
+			if not files then return end
```
Alf21: that would break the whole function. Maybe you missed `continue`?

TimGoll: Very good spot!

PR #478, opened 2020-02-22
Fallback (missing) icon for shopeditor/traitor shop

lebroomer:
* Replacing pink/black default buggy missing icon by a custom one

* Added caching of materials to improve performance instead of implicitely reinitializing a material through SetIcon


gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -979,6 +981,27 @@ function GM:OnContextMenuOpen()
 	end
 end
 
+-- Preload materials for the shop
+hook.Add("PostInitPostEntity", "TTTCacheEquipMaterials", function()
```
saibotk: I guess we should name new hook functions consistently: so i'd suggest `TTT2CacheEquipmentMaterials`

gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -979,6 +981,27 @@ function GM:OnContextMenuOpen()
...
+	local itms = ShopEditor.GetEquipmentForRoleAll()
```
saibotk: I think we do not need to get the items from the `ShopEditor`, i guess we could get them from the `items` module.

gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -309,7 +311,7 @@ local function CreateEquipmentList(t)
 			local ic = nil
 
 			-- Create icon panel
-			if item.material and item.material ~= "vgui/ttt/icon_id" then
+			if item.ttt2material then
```
saibotk: I'd suggest, to use a more fitting name, that expresses the function of the variable more accurately. It should show, that this is used for caching.

gamemodes/terrortown/gamemode/client/cl_equip.lua
Alf21: That's wrong, because Equipment isn't the same as an Item. An Equipment could be an Item AS WELL AS a weapon. Ref.: https://github.com/TTT-2/TTT2/blob/b7bfa4a03df2a2d000cf293a79587aafce7e3493/gamemodes/terrortown/gamemode/client/cl_shopeditor.lua#L72

gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -979,6 +981,27 @@ function GM:OnContextMenuOpen()
...
+			item.ttt2material = Material(item.material)
+			if item.ttt2material:IsError() then
+				-- Setting fallback material
+				item.ttt2material = fallback_mat
+			end
+		elseif item.model and item.model ~= "models/weapons/w_bugbait.mdl" then
+			--do not use fallback mat and use model instead
+			item.ttt2material = nil
+			item.ttt2model = model
+		else
```
TimGoll: Please take a look at the ci-output

```
gamemodes/terrortown/gamemode/client/cl_equip.lua: [Warning] line 999, column 3 - line 999, column 7: Empty else statement
```

gamemodes/terrortown/gamemode/client/cl_equip.lua
saibotk: This should still be moved out of the shopeditor , when we rework its code

PR #484, opened 2020-03-18
these are the tweaks i made

joeyjumper94:
it fixed the errors these files caused when they were autorun, it also simplified things too.


TimGoll:
Please take a look at the coding style. You can see the output here: https://ci.neoxult.de/job/TTT2/job/PR-484/1/

joeyjumper94:
that should be fixed now

TimGoll:
Yes, it would work. But we try to have a consitent style and decided on the lua style

lua/autorun/server/ttt2_force_download.lua
```diff
@@ -1,7 +1,7 @@
 -- This file forces clients to download the icons
 -- If you are distributing those files via FastDL, comment out the line below.
 
-if SERVER then
+if engine.ActiveGamemode() == "terrortown" then
```
TimGoll: it should still be server only

saibotk: That is ensured, as this file is in the /autorun/server folder

TimGoll: my bad

lua/autorun/ttt_bind_autorun.lua
```diff
@@ -1,6 +0,0 @@
--- Clientside bind lib
```
saibotk: Why did you remove this file?

lua/autorun/client/ttt_bind_autorun.lua
```diff
@@ -1,3 +1,4 @@
+if engine.ActiveGamemode() ~= "terrortown" then return end--block if TTT is not running
 ------------
```
saibotk: This file should not be moved and the `ttt_bind_autorun.lua` file should still be kept, as it will add the lua file to the files a client has to download.

Alf21: Why? I don‘t see any reason, autorun files are downloaded automatically afaik

saibotk: Yes, but we do not have "logic-heavy" files in the autorun folder as of now, and i think that is much cleaner as it is currently, with the bind code in its own folder.

saibotk:
This was done in 655f9e596d646f3e2941a85c60efc6e9ee3123da

PR #480, opened 2020-03-03
Vgui: Rework

TimGoll:
This PR completely overhauls the F1 menu while also providing a base to rework the other menues.

image

Things done in this PR:
- a new vgui design with theme support
- a base to enable easy vgui creation in the future
- restructured language files
- inverted some convars from disable to enable
- introduced a global scale factor
- introduced a resolution change handling, even when the gamemode is not loaded
- keeping compatibility to old TTT addons
- added hooks and an easy interface to populate (sub-)menues
- probably some more thing that I have forgotten

Of course there are many details that could be improved. However I decided to call this done and add some improvements in the future in smaller pull requests since this one already gotten out of hand.

Remark: Since all convars will be replaced soon, I did not bother to put them inside of tables etc.

PR #502, opened 2020-04-06
Added missing translations after Tim's cleanup

Wukerr:
I have manually corrected the Polish translation and it should now have everything that is needed, some of them like TargetID and VSin can't be translated.


TimGoll:
Nice, please take a look at the CI output

TimGoll:
> some of them like TargetID and VSin can't be translated

Can you explain what you mean with this?

Wukerr:
I mean "TargerID" and "VSkin" themself can not be translated and I don't understand CI output, I edited the file 2 times and I accidentally added space to it and it looks like an error for it. Don't know how to fix.

TimGoll:
Ah, I've kept those words in german as well because these are names. And in line `1208` a `"` is missing

Wukerr:
Everything should be fine right now

Histalek:
#480 needs to be merged before this

PR #508, opened 2020-04-15
Small ROLE improvements

Alf21:
Currently, there is no possibility to create an own abstract ROLE (because this still will be initialized like a normal ROLE)

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
@@ -79,32 +79,32 @@ end
 function plymeta:SetRole(subrole, team, forceHooks)
 	local oldRole = self:GetBaseRole()
 	local oldSubrole = self:GetSubRole()
-	local ord = self:GetSubRoleData()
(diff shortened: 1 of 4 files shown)


gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -79,20 +79,20 @@ end
 function plymeta:SetRole(subrole, team, forceHooks)
 	local oldRole = self:GetBaseRole()
 	local oldSubrole = self:GetSubRole()
+	local ord = self:GetSubRoleData()
```
saibotk: should be renamed to `oldRoleData`

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -79,20 +79,20 @@ end
...
 	local oldTeam = self:GetTeam()
+
 	local rd = roles.GetByIndex(subrole)
```
saibotk: Would be nice to also be renamed to `roleData`

PR #515, opened 2020-04-19
Added GAMEMODE.roundCount and cleaned GAMEMODE var usage

Alf21:
As MR to **discuss** var naming (`GAMEMODE.roundCount`) and whether it's required

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_main.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_main.lua
@@ -89,7 +89,8 @@ function GM:Initialize()
 
 	hook.Run("TTT2Initialize")
 
-	GAMEMODE.round_state = ROUND_WAIT
+	self.round_state = ROUND_WAIT
+	self.roundCount = 0
 
 	-- load addon language files
 	LANG.SetupFiles("lang/", true)
@@ -185,7 +186,7 @@ function GM:InitPostEntity()
 	-- make sure player class extensions are loaded up, and then do some
 	-- initialization on them
 	if IsValid(client) and client.GetTraitor then
-		GAMEMODE:ClearClientState()
+		self:ClearClientState()
 	end
 
 	-- cache players avatar
@@ -195,7 +196,7 @@ function GM:InitPostEntity()
 		draw.CacheAvatar(plys[i]:SteamID64(), "medium") -- caching
 	end
 
-	timer.Create("cache_ents", 1, 0, GAMEMODE.DoCacheEnts)
+	timer.Create("cache_ents", 1, 0, self.DoCacheEnts)
 
 	RunConsoleCommand("_ttt_request_serverlang")
 	RunConsoleCommand("_ttt_request_rolelist")
@@ -244,6 +245,8 @@ local function RoundStateChange(o, n)
 
 		-- reset cached server language in case it has changed
 		RunConsoleCommand("_ttt_request_serverlang")
+
+		GAMEMODE.roundCount = GAMEMODE.roundCount + 1
 
 		-- clear decals in cache from previous round
 		util.ClearDecals()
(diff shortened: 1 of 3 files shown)


Alf21:
This could be handled too with

```lua
local roundCount = GetConVar("ttt_round_limit"):GetInt() - GetGlobalInt("ttt_rounds_left")
```

Alf21:
```
[ERROR] addons/ttt2/gamemodes/terrortown/gamemode/client/cl_main.lua:396: attempt to index local 'self' (a number value)
  1. func - addons/ttt2/gamemodes/terrortown/gamemode/client/cl_main.lua:396
   2. unknown - lua/includes/extensions/net.lua:32
```

Throwing this error. Doesn't seem to make sense...

Alf21:
> 
> 
> Might be out of scope for this PR, but things like `round_state`, `FirstRound`, ... should be renamed to `roundState`, `firstRound`, ...

Yea thought the same, but this would break many add-ons and can't be done :(

gamemodes/terrortown/gamemode/client/cl_main.lua
```diff
@@ -195,7 +196,9 @@ function GM:InitPostEntity()
 		draw.CacheAvatar(plys[i]:SteamID64(), "medium") -- caching
 	end
 
-	timer.Create("cache_ents", 1, 0, GAMEMODE.DoCacheEnts)
+	timer.Create("cache_ents", 1, 0, function()
+		self:DoCacheEnts()
+	end)
```
Histalek, line 201: Is there a reason for defining this as an anonymous function?

gamemodes/terrortown/gamemode/client/cl_main.lua
```diff
@@ -89,7 +89,8 @@ function GM:Initialize()
 
 	hook.Run("TTT2Initialize")
 
-	GAMEMODE.round_state = ROUND_WAIT
+	self.round_state = ROUND_WAIT
+	self.roundCount = 0
```
Histalek, line 93: Isn't there already some kind of counter, which is used for the mapchange in X rounds? could that be used here instead? i don't know how that one is implemented though, but i'd rather not have two counters/vars for essentially the same thing if we can avoid it

gamemodes/terrortown/gamemode/client/cl_main.lua
```diff
@@ -426,11 +431,13 @@ function GM:ClearClientState()
...
 
-	if not GAMEMODE.ForcedMouse then return end
+	if not self.ForcedMouse then return end
 
 	gui.EnableScreenClicker(false)
 end
-net.Receive("TTT_ClearClientState", GM.ClearClientState)
+net.Receive("TTT_ClearClientState", function()
+	GAMEMODE:ClearClientState()
+end)
```
Histalek, line 440: Same as above

gamemodes/terrortown/gamemode/client/cl_main.lua
Alf21: Please give a look into my first comment here :)

gamemodes/terrortown/gamemode/client/cl_main.lua
Alf21: Because previously, it just called as `GM.DoCacheEnts`. This is a simple function reference WITHOUT referencing the GM / GAMEMODE. The problem is that the function in TTT requires GAMEMODE (it‘s written with `:`, not as a simple table function. Otherwise, self in this function wouldn‘t be GAMEMODE).

gamemodes/terrortown/gamemode/client/cl_main.lua
Alf21: - ' ' -

PR #523, opened 2020-04-24
improved epop

TimGoll:
improved the EPOP based on stuff learned while using it

- EPOPs are now always redered, even when dead since they are really good to inform players about upcoming respawns
- even if the system was originally designed to display always only one element, a queue was added since otherwise multiple addons trying to display text might clash
- messages now have a `blocking` flag, if set to false a new message can overwrite the message without waiting in a queue
- also I added a few convenience functions
- improved docs and coding style improvements

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
@@ -9,7 +9,7 @@ local defaultMessage = {
 	title = {
 		text = "A Test Popup, now with a multiline title, how NICE."
 	},
-	text = {
+	subtitle = {
 		text = "Well, hello there! This is a fancy popup with some special information. The text can be also multiline, how fancy! Ugh, I could add so much more text if I'd had any ideas..."
 	},
 	iconTable = {},
@@ -99,7 +99,7 @@ end
 -- @return table The message table
 -- @realm client
 function EPOP:GetMessage()
-	return self.messageQueue[1] or defaultMessage
+	return self.messageQueue[1] or table.Copy(defaultMessage)
 end
 
 function EPOP:RemoveMessageByIndex(index)
--- a/gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
@@ -72,7 +72,7 @@ function EPOP:ActivateMessage()
 
 	-- register a timer to remove the message
 	timer.Create(TIMER_IDENTIFIER, elem.displayTime, 1, function()
-		EPOP:RemoveMessageByIndex(1)
+		EPOP:RemoveMessage(elem.id)
 	end)
 
 	print("[TTT2] " .. elem.title.text .. " // " .. elem.subtitle.text or "")
--- a/gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
@@ -100,11 +100,18 @@ end
 
 ---
 -- Returns the neweset message in the queue that should be rendered right now.
--- If there is no message in the queue, a template message is returned.
 -- @return table The message table
 -- @realm client
 function EPOP:GetMessage()
-	return self.messageQueue[1] or table.Copy(defaultMessage)
+	return self.messageQueue[1]
+end
+
+---
+-- Returns the default message.
+-- @return table The message table
+-- @realm client
+function EPOP:GetDefaultMessage()
+	return table.Copy(defaultMessage)
 end
 
 ---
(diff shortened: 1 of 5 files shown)


gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
```diff
@@ -1,53 +1,150 @@
...
+	local elem = self.messageQueue[1]
+
+	elem.time = CurTime() + elem.displayTime
+
+	-- register a timer to remove the message
+	timer.Create(TIMER_IDENTIFIER, elem.displayTime, 1, function()
+		EPOP:RemoveMessageByIndex(1)
+	end)
+
+	print("[TTT2] " .. elem.title.text .. " // " .. elem.subtitle.text or "")
```
Alf21, line 78: Print?

TimGoll: it is the same in the mstack. therefore I added it so you can see it later in the log

gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
```diff
@@ -1,53 +1,150 @@
...
+function EPOP:ActivateMessage()
+	if #self.messageQueue == 0 then return end
+
+	local elem = self.messageQueue[1]
+
+	elem.time = CurTime() + elem.displayTime
+
+	-- register a timer to remove the message
+	timer.Create(TIMER_IDENTIFIER, elem.displayTime, 1, function()
+		EPOP:RemoveMessageByIndex(1)
```
saibotk, line 75: This should rather be RemoveMessage instead, as this timer should only be allowed to delete this exact element, to reduce sideeffects

gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
```diff
@@ -1,53 +1,150 @@
...
+-- @param [default=true]boolean blocking If this is false, this message gets instantly replaced if a new message is added
+-- @return string Returns a unique id generated for this message
 -- @realm client
-function EPOP:AddMessage(title, text, disp_time, icon_tbl)
-	self.msg = {
-		title = {},
-		text = {}
-	}
+function EPOP:AddMessage(title, subtitle, displayTime, iconTable, blocking)
+	local id = "epop_unique_id_" .. tostring(CurTime() * 10000)
```
lebroomer, line 33: I don't really like this way to create an unique identifier. Wouldn't it be better to just use a global variable, which is increased every time AddMessage is called?

TimGoll: and the benefit of this would be... ?

lebroomer: To be certain about unique identifiers.

TimGoll: I mean, this time is in 1/1000 seconds resolution

lebroomer: Sure, in practive, this won't probably ever result in a collision, but it is still bad style to assign an identifier in that way.

TimGoll: okay, I'll change it then

gamemodes/terrortown/gamemode/client/cl_main.lua
```diff
@@ -311,8 +311,29 @@ local function PlaySoundCue()
...
 end
 
-GM.TTTBeginRound = PlaySoundCue
-GM.TTTEndRound = PlaySoundCue
+---
+-- A hook that is called when the preparation phase starts
+-- @hook
+-- @realm client
+function GM:TTTPrepareRound()
+	EPOP:Clear()
```
saibotk, line 319: I think we should discuss, if this would not rather be placed in https://github.com/TTT-2/TTT2/blob/3008bec9301eff74bf07302736733d0cda34266d/gamemodes/terrortown/gamemode/client/cl_main.lua#L281

or if it even makes a difference, eg. is this due to being in a hook, also not deterministic and should this be the way to add code, when a hook usually is used to let external addons be called instead of ours? And why did TTT do it like that, with the `PlaySoundCue`. This seems inconsistent to me and I'd like to hear @Alf21's opinion in this.

TimGoll: > I think we should discuss, if this would not rather be placed in https://github.com/TTT-2/TTT2/blob/3008bec9301eff74bf07302736733d0cda34266d/gamemodes/terrortown/gamemode/client/cl_main.lua#L281
> 
> or if it even makes a difference, eg. is this due to being in a hook, also not deterministic and should this be the way to add code, when a hook usually is used to let external addons be called instead of ours? And why did TTT do it like that, with the `PlaySoundCue`. This seems inconsistent to me and I'd like to hear @Alf21's opinion in this.

Hm. I don't see why this isn't deterministic tbh. And there are many hooks used like this in TTT2.

PR #524, opened 2020-04-28
Printmessage overwrite and EPOP Proxy

TimGoll:
I added a server proxy for popups which features most features.

I added a printmessage overwrite, mainly to redirect `HUD_PRINTCENTER` to the popup system. While I were at it, I changed all types to the new systems because `PrintMessage` still used the old [umsg](https://wiki.facepunch.com/gmod/umsg) which is rather bad.

These functions can be used now without any problems, however I mainly only ported them for better support for old addons.

You can review it now, howewer it might be confusing before #523 is merged since it is based on this.


gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
```diff
@@ -0,0 +1,48 @@
...
+
+	if not subtitle or isstring(subtitle) then
+		net.WriteBool(false)
+		net.WriteString(subtitle or "")
+	elseif not IsValid(subtitle.color) then
+		net.WriteBool(false)
+		net.WriteString(subtitle.text or "")
+	else
+		net.WriteBool(true)
+		net.WriteString(titsubtitlele.text or "")
```
saibotk, line 36: `titsubtitlele` -> Yes, no, maybe ?

gamemodes/terrortown/gamemode/shared/sh_printmessage_override.lua
```diff
@@ -0,0 +1,63 @@
+---
+-- @desc Overwrites of the default PrintMessage function.
+-- This overwrite fixes some limitations with the default system (usage of
+-- archaic <a href="https://wiki.facepunch.com/gmod/umsg">umsg</a> system) while
+-- also moving `HUD_PRINTCENTER` to the EPOP system.
+-- @author Mineotopia
+
+if SERVER then
```
lebroomer, line 8: I don't really like the style. I think this should be either splitted up into two files, or combined better to reduce code/comment cloning. Especially the ply:PrintMessage function could be easily combined like this:

```PrintMessage(type, message, SERVER and self or nil)```, or just ```PrintMessage(type, message, self)```, since the client-side PrintMessage only supports 2 arguments.

File: gamemodes/terrortown/gamemode/client/cl_eventpopup.lua [hunk 3]
```diff
@@ -61,8 +62,8 @@ function EPOP:AddMessage(title, subtitle, displayTime, iconTable, blocking)
 	-- add the new message to the queue
 	self.messageQueue[queueSize + 1] = {
 		id = id,
-		title = isstring(title) and {text = title} or title,
-		subtitle = isstring(subtitle) and {text = subtitle} or subtitle,
+		title = (not title or isstring(title)) and {text = title or ""} or title,
```
Alf21, line 65: why not using `title = title or ""` in front to avoid the first check and the second `or` that does somehow the same?

gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
```diff
@@ -61,8 +62,8 @@ function EPOP:AddMessage(title, subtitle, displayTime, iconTable, blocking)
...
+		subtitle = (not subtitle or isstring(subtitle)) and {text = subtitle or ""} or subtitle,
```
Alf21, line 66: same here

gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
```diff
@@ -163,3 +164,23 @@ end
 function EPOP:Clear()
 	self.messageQueue = {}
 end
+
+net.Receive("ttt2_eventpopup", function()
+	local title, subtitle = {}, {}
+
+	if net.ReadBool() then
+		title.text = TryT(net.ReadString())
```
Alf21, line 172: put this out. It's the same as in the `else` section, so the else isn't needed

gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
```diff
@@ -163,3 +164,23 @@ end
...
+		title.color = net.ReadColor()
+	else
+		title.text = TryT(net.ReadString())
+	end
+
+	if net.ReadBool() then
+		subtitle.text = TryT(net.ReadString())
```
Alf21, line 179: same here

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -364,7 +364,9 @@ LANG.Styles = {
 		chat.AddText(warn_color, text)
 	end,
 
-	[MSG_CHAT_PLAIN] = chat.AddText
+	[MSG_CHAT_PLAIN] = chat.AddText,
+
```
Alf21, line 368: why an empty line? Previously, it made sense (because function declaration), but here it doesn't

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
```diff
@@ -0,0 +1,48 @@
...
+
+	if title or isstring(title) then
+		net.WriteBool(false)
+		net.WriteString(title or "")
+	elseif not IsValid(title.color) then
+		net.WriteBool(false)
+		net.WriteString(title.text or "")
+	else
+		net.WriteBool(true)
+		net.WriteString(title.text or "")
```
Alf21, line 24: couldn't this be simplified? In any case you write `title.text or ""`

Alf21: And is this `or ""` still needed?

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
```diff
@@ -0,0 +1,48 @@
...
+-- @param [default=4]number displayTime The render duration of the popup
+-- @param [default=true]boolean blocking If this is false, this message gets instantly replaced if a new message is added
+-- @realm server
+function EPOP:AddMessage(plys, title, subtitle, displayTime, blocking)
+	net.Start("ttt2_eventpopup")
+
+	if title or isstring(title) then
+		net.WriteBool(false)
+		net.WriteString(title or "")
+	elseif not IsValid(title.color) then
```
Alf21, line 19: Here, title could be invalid, This would lead to `nil` error

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
```diff
@@ -0,0 +1,48 @@
...
+	else
+		net.WriteBool(true)
+		net.WriteString(title.text or "")
+		net.WriteColor(title.color or COLOR_WHITE)
+	end
+
+	if not subtitle or isstring(subtitle) then
+		net.WriteBool(false)
+		net.WriteString(subtitle or "")
+	elseif not IsValid(subtitle.color) then
```
Alf21, line 31: same as above

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
```diff
@@ -0,0 +1,48 @@
...
+		net.WriteBool(false)
+		net.WriteString(subtitle.text or "")
+	else
+		net.WriteBool(true)
+		net.WriteString(subtitle.text or "")
+		net.WriteColor(subtitle.color or COLOR_WHITE)
+	end
+
+	net.WriteUInt(displayTime or 4, 16)
+	net.WriteBool(blocking == nil and false or blocking)
```
Alf21, line 41: `blocking == true` ?

gamemodes/terrortown/gamemode/shared/sh_lang.lua
```diff
@@ -17,6 +17,8 @@ MSG_CHAT_ROLE = 3
 MSG_CHAT_WARN = 4
 MSG_CHAT_PLAIN = 5
```
Alf21, line 19: Empty line needed?

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
```diff
@@ -0,0 +1,48 @@
+util.AddNetworkString("ttt2_eventpopup")
+
+EPOP = EPOP or {}
+
+---
+-- Adds a translated popup message to the @{EPOP}
+-- @param table|Player plys A table of player that should receive this popup, broadcasts it if nil
```
Alf21, line 7: Add `nil|` or `?` in front of `table|Player` if nil is possible

gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
TimGoll: because it has to be in a table if it is not already a table

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
Alf21: And can a color be valid? Why not using `iscolor` ?

gamemodes/terrortown/gamemode/shared/sh_lang.lua
TimGoll: yes since these are three groups

gamemodes/terrortown/gamemode/client/cl_eventpopup.lua
Alf21: That answer doesn't match my question

TimGoll: an aternative would be
```lua
title = title or ""
title = istable(title) and title or {text = title}
```
and I don't see any benefit here

Alf21:
```lua
title = title or ""
subtitle = subtitle or ""

-- add the new message to the queue
self.messageQueue[queueSize + 1] = {
	id = id,
	title = isstring(title) and {text = title} or title,
	subtitle = isstring(subtitle) and {text = subtitle} or subtitle,
	displayTime = displayTime or 4,
-- ...
```

TimGoll: this is exactly what I've just written in the previous comment and I don't see any benefit here except adding another line. Especially since title can be a table, the `string|table or ""` is rather confusing

Alf21: and `title = (not title or isstring(title)) and {text = title or ""} or title,` is better? :D

TimGoll: In my opinion, yes. It is a single line. And it says clearly that the title is converted to a table if is undefined or a string and kept if is a table

Alf21: Okay, it's a very small change, but here we go: Your solution needs more checks:
1. Is a title given? No.
2. Set text ? title. But title is nil. So set it "".

Mine:
1. Is a title given? No. Set it to "".

Additionally, while looking through the code, my solution makes clear it's a string if it's nil. And is showing that it's creating a table if it's a string.

It isn't every time the best solution to keep things compressed. In code, we can try to make it clear. Furthermore, the doc on top is telling the Dev, that a string or a table is useable, so that's an invalid arg imo.

TimGoll: > 1. Is a title given? No. Set it to "".

and 2. Is it a table? No. Set it to `{text = title}`

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
TimGoll: you're right here, it should have been `if not title or isstring(title)`

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
TimGoll: this is always valid. Or do you refer to the iscolor?

gamemodes/terrortown/gamemode/client/cl_lang.lua
TimGoll: the different print types (`MSTACK` and `CHAT`) always had an empty line between them

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
Alf21: Oh okay. A completely changed logic. Then you can encapsulate the `isstring` / istable check and do a simple if-else with title inside or smth

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
Alf21: Yep, same as above :D

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
Alf21: Above same as discussed before, you could use `title = title or ""` to reduce checks

TimGoll: Maybe I don't understand what you mean, but what's the benefit here? Only one of the two `title.text or ""` is executed since they are in different parts of the `if`.

gamemodes/terrortown/gamemode/shared/sh_printmessage_override.lua
Alf21: Even now it looks weird, calling on `CLIENT` a function of 2 params with 3

gamemodes/terrortown/gamemode/server/sv_eventpopup.lua
Alf21: You don't need to write `or ""` if the given input can't be nil. That what I'm asking. You do it in many parts, it looks like you don't know about the states of your own variables, the code looks inconsistent

TimGoll: No, the user can call the function however he likes. Therefore I don't know if this value is set or not

Alf21: Hmm you are right. But if no title and no subtitle is given, you even don't need networking anything... Additionally, the title check above (as already discussed before) can be improved anyways

gamemodes/terrortown/gamemode/shared/sh_printmessage_override.lua
TimGoll: So I should change it back to the first version?

Alf21: No but there should be a better way, this seems to be redundant :/

TimGoll: Then please suggest something because I have no better idea.

PR #525, opened 2020-04-28
introduced speed system

TimGoll:
This is a really simple PR that improves upon the newly included speedrun item. It now features a system similar to the armor system.

This makes it easier for roles etc to add a speed multiplier to players without having to sync all values. Additionally the icon now shows the overall speed (red for < 1, green > 1, no icon if equal to 1).

The function names are based on the armor function names.

image

image

EDIT: After the rework I removed the functions again and it now relys solely on the hook. However I moved the hook `TTTPlayerSpeedModifier` to be only serverside (with networking to the client), since this simplifies it for programmers while also "fixing" addons that set the factor only on the server. Additionally this ensures that I have the correct factor displayed in the HUD while also slightly improving the network performance.

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/sh_main.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_main.lua
@@ -64,9 +64,7 @@ end
 -- @hook
 -- @realm shared
 function GM:Move(ply, moveData)
-	if SERVER then
-		SPEED:HandleSpeedCalculation(ply, moveData)
-	end
+	SPEED:HandleSpeedCalculation(ply, moveData)
 
 	local mul = ply:GetSpeedMultiplier()
 
--- a/gamemodes/terrortown/gamemode/shared/sh_main.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_main.lua
@@ -85,6 +85,8 @@ function GM:Move(ply, mv)
 		mul = mul * ply.sprintMultiplier * sprintMultiplierModifier[1]
 	end
 
+	mul = mul * ply:GetSpeedMultiplier()
+
 	mv:SetMaxClientSpeed(mv:GetMaxClientSpeed() * mul)
 	mv:SetMaxSpeed(mv:GetMaxSpeed() * mul)
 end
--- a/gamemodes/terrortown/gamemode/shared/sh_main.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_main.lua
@@ -134,6 +134,10 @@ end
 -- @realm shared
 function GM:Think()
 	UpdateSprint()
+
+	if CLIENT then
+		EPOP:Think()
+	end
 end
 
 -- Drowning and such
--- a/gamemodes/terrortown/gamemode/shared/sh_main.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_main.lua
@@ -55,27 +55,20 @@ function GM:PlayerFootstep(ply, pos, foot, sound, volume, rf)
 end
 
 ---
--- Predicted move speed changes
+-- The Move hook is called for you to manipulate the player's MoveData.
+-- You shouldn't adjust the player's position in any way in the move hook. This is because due
+-- to prediction errors, the netcode might run the move hook multiple times as packets arrive late.
+-- Therefore you should only adjust the movedata construct in this hook.
+-- @param Player ply The player
+-- @param CMoveData moveData Movement information
 -- @hook
 -- @realm shared
-function GM:Move(ply, mv)
-	if not ply:IsTerror() then return end
-
-	local basemul = 1
-	local slowed = false
-
-	-- Slow down ironsighters
-	local wep = ply:GetActiveWeapon()
-
-	if IsValid(wep) and wep.GetIronsights and wep:GetIronsights() then
-		basemul = 120 / 220
-		slowed = true
+function GM:Move(ply, moveData)
+	if SERVER then
+		SPEED:HandleSpeedCalculation(ply, moveData)
 	end
 
-	local noLag = {1}
-
-	local mul = hook.Call("TTTPlayerSpeedModifier", GAMEMODE, ply, slowed, mv, noLag) or 1
-	mul = basemul * mul * noLag[1]
+	local mul = ply:GetSpeedMultiplier()
 
 	if ply.sprintMultiplier and (ply.sprintProgress or 0) > 0 then
 		local sprintMultiplierModifier = {1}
@@ -85,10 +78,8 @@ function GM:Move(ply, mv)
 		mul = mul * ply.sprintMultiplier * sprintMultiplierModifier[1]
 	end
 
-	mul = mul * ply:GetSpeedMultiplier()
-
-	mv:SetMaxClientSpeed(mv:GetMaxClientSpeed() * mul)
-	mv:SetMaxSpeed(mv:GetMaxSpeed() * mul)
+	moveData:SetMaxClientSpeed(moveData:GetMaxClientSpeed() * mul)
+	moveData:SetMaxSpeed(moveData:GetMaxSpeed() * mul)
 end
 
 local ttt_playercolors = {
--- a/gamemodes/terrortown/gamemode/shared/sh_main.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_main.lua
@@ -56,8 +56,8 @@ end
 
 ---
 -- The Move hook is called for you to manipulate the player's MoveData.
--- You shouldn't adjust the player's position in any way in the move hook. This is because due
--- to prediction errors, the netcode might run the move hook multiple times as packets arrive late.
+-- You shouldn't adjust the player's position in any way in the move hook. This is due to
+-- prediction errors, the netcode might run the move hook multiple times as packets arrive late.
 -- Therefore you should only adjust the movedata construct in this hook.
 -- @param Player ply The player
 -- @param CMoveData moveData Movement information
--- a/gamemodes/terrortown/gamemode/shared/sh_speed.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_speed.lua
@@ -54,7 +54,7 @@ if CLIENT then
 	-- It is called in @{GM:TTT2PlayerReady}.
 	-- @realm client
 	function SPEED:Initialize()
-		STATUS:RegisterStatus("ttt_walkspeed_status_good", {
+		STATUS:RegisterStatus("ttt_speed_status_good", {
 			hud = {
 				Material("vgui/ttt/perks/hud_speedrun.png")
 			},
@@ -64,7 +64,7 @@ if CLIENT then
 			end
 		})
 
-		STATUS:RegisterStatus("ttt_walkspeed_status_bad", {
+		STATUS:RegisterStatus("ttt_speed_status_bad", {
 			hud = {
 				Material("vgui/ttt/perks/hud_speedrun.png")
 			},
@@ -78,14 +78,14 @@ if CLIENT then
 			newval = math.Round(newval, 1)
 
 			if newval == 1.0 then
-				STATUS:RemoveStatus("ttt_walkspeed_status_good")
-				STATUS:RemoveStatus("ttt_walkspeed_status_bad")
+				STATUS:RemoveStatus("ttt_speed_status_good")
+				STATUS:RemoveStatus("ttt_speed_status_bad")
 			elseif newval > 1.0 and (not oldval or oldval <= 1.0) then
-				STATUS:RemoveStatus("ttt_walkspeed_status_bad")
-				STATUS:AddStatus("ttt_walkspeed_status_good")
+				STATUS:RemoveStatus("ttt_speed_status_bad")
+				STATUS:AddStatus("ttt_speed_status_good")
 			elseif newval < 1.0 and (not oldval or oldval >= 1.0) then
-				STATUS:RemoveStatus("ttt_walkspeed_status_good")
-				STATUS:AddStatus("ttt_walkspeed_status_bad")
+				STATUS:RemoveStatus("ttt_speed_status_good")
+				STATUS:AddStatus("ttt_speed_status_bad")
 			end
 		end)
 	end
--- a/gamemodes/terrortown/gamemode/shared/sh_speed.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_speed.lua
@@ -45,7 +45,8 @@ function SPEED:HandleSpeedCalculation(ply, moveData)
 end
 
 ---
--- A hook to modify the player speed, it is automatically networked.
+-- A hook to modify the player speed.
+-- @note This hook is predicted and should be therefore added on both server and client.
 -- @param Player ply The player whose speed should be modified
 -- @param boolean isSlowed Is true if the player uses iron sights
 -- @param CMoveData moveData The move data
--- a/gamemodes/terrortown/gamemode/shared/sh_speed.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_speed.lua
@@ -3,6 +3,8 @@ local plymeta = assert(FindMetaTable("Player"), "FAILED TO FIND ENTITY TABLE")
 SPEED = SPEED or {}
 
 if SERVER then
+	util.AddNetworkString("TTT2UpdateSpeedModifier")
+
 	function SPEED:HandleSpeedCalculation(ply, moveData)
 		if not ply:IsTerror() then return end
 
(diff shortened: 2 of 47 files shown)


gamemodes/terrortown/gamemode/shared/sh_walkspeed.lua
```diff
@@ -0,0 +1,94 @@
...
+	---
+	-- Increases the speed modifier by the given multiplier (the old value is multiplied with the new value)
+	-- @param number value The multiplier
+	-- @realm server
+	function plymeta:GiveSpeedMultiplier(value)
+		self:TTT2NETSetFloat("player_speed_multiplier", self:TTT2NETGetFloat("player_speed_multiplier", 1.0) * value)
+	end
+
+	---
+	-- Decreases the speed modifier by the given multiplier (the old value is devided by the new value)
```
Histalek, line 38: "Divides the (players) current speed modifier with the given value"

gamemodes/terrortown/gamemode/shared/sh_walkspeed.lua
```diff
@@ -0,0 +1,94 @@
+local plymeta = assert(FindMetaTable("Player"), "FAILED TO FIND ENTITY TABLE")
+
+SPEED = SPEED or {}
+
+if SERVER then
+	---
+	-- Sets the initial @{Player} speed multiplier, called in @{GM:TTTBeginRound} after (dumb) players
```
Histalek, line 7: These should be two sentences in my opinion. The first being what the function does and then the additional info on where, when and why it is used.

gamemodes/terrortown/gamemode/shared/sh_walkspeed.lua
```diff
@@ -0,0 +1,95 @@
...
+			hud = {
+				Material("vgui/ttt/perks/hud_speedrun.png")
+			},
+			type = "bad",
+			DrawInfo = function()
+				return math.Round(LocalPlayer():GetSpeedMultiplier(), 1)
+			end
+		})
+
+		TTT2NET:OnUpdate("players", function(oldval, newval, reversePath)
```
saibotk, line 72: Using only the "players" path, will lead to an update call for any player whose multiplier has changed and not only for your own player. You will need to append the Steamid64 to the path or use https://github.com/TTT-2/TTT2/blob/4947aa62ef89616b4e7350c195013e885ee8b200/gamemodes/terrortown/gamemode/client/cl_network_sync.lua#L370
which will take care of prefixing the correct path.

TimGoll: But then I need to register a function for every player which is rather problematic because I have to track latejoiners etc.

TimGoll:
> I noted some issues with the usage of the networking system, but other than that, i think we need to discuss the breaking change that is introduced with this system:
> You now are not able to track down what item causes which multiplier and if it has already been added or deleted etc. I think that might lead to some weird issues, and thus i think we need to talk about this.

It is the same as it is right now.

gamemodes/terrortown/gamemode/shared/sh_walkspeed.lua
saibotk: Please review the function of this code, which is only for the `LocalPlayer()` or am I wrong?

TimGoll: Yes, you're right. But It did not work for me with `TTT2NET:OnUpdateOnPlayer`, maybe I did something else wrong.

gamemodes/terrortown/gamemode/shared/sh_walkspeed.lua
```diff
@@ -0,0 +1,100 @@
...
+		ply:SetSpeedMultiplier(baseMultiplier * returnMultiplier * speedMultiplierModifier[1])
+	end
+
+	---
+	-- A hook to modify the player speed, it is automatically networked
+	-- @param Player ply The player whose speed should be modified
+	-- @param boolean isSlowed Is true if the player uses iron sights
+	-- @param CMoveData moveData The move data
+	-- @param table speedMultiplierModifier The speed modifier table. Modify the first table entry to change the player speed
+	-- @return [depricated]number The depricated way of changing the player speed
```
Histalek, line 32: deprecated

gamemodes/terrortown/gamemode/shared/sh_walkspeed.lua
```diff
@@ -0,0 +1,100 @@
...
+				STATUS:RemoveStatus("ttt_walkspeed_status_good")
+				STATUS:AddStatus("ttt_walkspeed_status_bad")
+			end
+		end)
+	end
+end
+
+---
+-- Returns the current player speed modifier
+-- @retrun number The speed modifier
```
Histalek, line 96: return

gamemodes/terrortown/gamemode/shared/sh_main.lua
```diff
@@ -55,27 +55,20 @@ function GM:PlayerFootstep(ply, pos, foot, sound, volume, rf)
 end
 
 ---
--- Predicted move speed changes
+-- The Move hook is called for you to manipulate the player's MoveData.
+-- You shouldn't adjust the player's position in any way in the move hook. This is because due
```
lebroomer, line 59:
```suggestion
-- You shouldn't adjust the players position in any way in the move hook. This is due
```

idiot's apostrophe :P

gamemodes/terrortown/gamemode/shared/sh_walkspeed.lua
```diff
@@ -0,0 +1,100 @@
...
+		end
+
+		local speedMultiplierModifier = {1}
+		local returnMultiplier = hook.Run("TTTPlayerSpeedModifier", ply, isSlowed, moveData, speedMultiplierModifier) or 1
+
+		ply:SetSpeedMultiplier(baseMultiplier * returnMultiplier * speedMultiplierModifier[1])
+	end
+
+	---
+	-- A hook to modify the player speed, it is automatically networked
```
lebroomer, line 27:
```suggestion
	-- A hook to modify the player speed, it is automatically networked.
```

gamemodes/terrortown/gamemode/shared/sh_main.lua
TimGoll: I have copied this exact text from the gmod wiki. However in english this is correct afaik.

lebroomer: Maybe I should return my English certificates...

saibotk:
Another thing that came to my mind while reviewing the other PRs:
Is it necessary or wanted to broadcast the speed value of each player to all other players? Does the prediction need this? Or is this usually only needed for the single client to know on how to interpolate between input and server result? If so, wouldn't it be better to not use TTT2NET for the speed synchronization, as this leads to the broadcast?

lua/terrortown/entities/items/item_ttt_speedrun.lua
```diff
@@ -9,11 +9,12 @@ ITEM.EquipMenuData = {
...
-ITEM.hud = Material("vgui/ttt/perks/hud_speedrun.png")
 ITEM.material = "vgui/ttt/icon_speedrun"
 
-hook.Add("TTTPlayerSpeedModifier", "SpeedrunModifySpeed", function(ply, _, _, refTbl)
-	if not IsValid(ply) or not ply:HasEquipmentItem("item_ttt_speedrun") then return end
+if SERVER then
+	hook.Add("TTTPlayerSpeedModifier", "TTT2SpeedRun", function(ply, _, _, speedMultiplierModifier)
+		if not IsValid(ply) or not ply:HasEquipmentItem("item_ttt_speedrun") then return end
 
-	refTbl[1] = refTbl[1] * 1.5 * (ply.speedrun_mul or 1)
```
Alf21, line 18: Why removed `speedrun_mul`?

gamemodes/terrortown/gamemode/shared/sh_speed.lua
```diff
@@ -0,0 +1,100 @@
...
+		ply:SetSpeedMultiplier(baseMultiplier * returnMultiplier * speedMultiplierModifier[1])
+	end
+
+	---
+	-- A hook to modify the player speed, it is automatically networked.
+	-- @param Player ply The player whose speed should be modified
+	-- @param boolean isSlowed Is true if the player uses iron sights
+	-- @param CMoveData moveData The move data
+	-- @param table speedMultiplierModifier The speed modifier table. Modify the first table entry to change the player speed
+	-- @return [deprecated]number The deprecated way of changing the player speed
```
Alf21, line 32: wrong documentation style

gamemodes/terrortown/gamemode/shared/sh_speed.lua
```diff
@@ -0,0 +1,100 @@
...
+	function GM:TTTPlayerSpeedModifier(ply, isSlowed, moveData, speedMultiplierModifier)
+
+	end
+
+	---
+	-- Sets the speed multiplier to an absolute value
+	-- @param number value The new value
+	-- @realm server
+	function plymeta:SetSpeedMultiplier(value)
+		if self.lastSpeedMultiplier and self.lastSpeedMultiplier == value then return end
```
Alf21, line 44: is `self.lastSpeedMultiplier` (~= nil) really needed?

lua/terrortown/entities/items/item_ttt_speedrun.lua
TimGoll: Because I don't see a reason to keep this. Why do we need this?

gamemodes/terrortown/gamemode/shared/sh_speed.lua
TimGoll: how do I mark a param as deprecated then?

gamemodes/terrortown/gamemode/shared/sh_speed.lua
TimGoll: probably not. However this will change because tkindanight noticed a bug in my implementation

Alf21:
And tbh, I'm definitely against automatical networking. It's the same as every time: Why trying to fix buggy add-ons with another add-on if that decreases the performance and increases the networking overflow... Things related to this should be handled by the server and by the client separately OR by gmod itself. Using a networking system that isn't made for things like this isn't recommended imho.

TimGoll:
> And tbh, I'm definitely against automatical networking. It's the same as every time: Why trying to fix buggy add-ons with another add-on if that decreases the performance and increases the networking overflow... Things related to this should be handled by the server and by the client separately OR by gmod itself. Using a networking system that isn't made for things like this isn't recommended imho.

I don't see how this decreases the performance. Actually I think it increases it. Prior to this change every addon had to do this manually, resulting in way more net messages than having one netmessage for all addons.

lua/terrortown/entities/items/item_ttt_speedrun.lua
Alf21: https://github.com/TTT-2/ttt2h-pack-default/blob/a110bc1a5fe614945bda6c38ca798bf3b5be1257/lua/classes/classes/class_dazzle.lua#L16

TimGoll: I know and I already updated the class locally since (at least in my opinion) it is cleaner to use the hook in the dazzle as well instead of a custom hacky variable.

gamemodes/terrortown/gamemode/shared/sh_speed.lua
Alf21: `-- @return[deprecated] number desc`, but returning isn't deprecated, it still does what it should do: Break the loop. Theoretically, the returned value should be taken. If there isn't a return value, the modified (dynamical) value should be used.

TimGoll: I mean, yes. The functionality is still there. But I wanted to mark that it shouldn't be used. But thanks, I'll change it.

Alf21:
> 
> 
> > And tbh, I'm definitely against automatical networking. It's the same as every time: Why trying to fix buggy add-ons with another add-on if that decreases the performance and increases the networking overflow... Things related to this should be handled by the server and by the client separately OR by gmod itself. Using a networking system that isn't made for things like this isn't recommended imho.
> 
> I don't see how this decreases the performance. Actually I think it increases it. Prior to this change every addon had to do this manually, resulting in way more net messages than having one netmessage for all addons.
> 
> Edit: Like discussed with tkindanight, I'll switch from his system to a normal netmessage.

okay... The calculations on the server are still the same. The calcs on the client could decrease because it is just networked, that's correct. BUT networking is something that shouldn't be done for useless stuff, it's all the time recommended keeping simple calculations done by the device itself than using another device to calculate it. It's just useful if the operation is complicated and the other device offers better performance.
This is NOT the case. Additionally, gmod automatically syncs the position of the clients by the server. That means that the "running" is done automatically by the server like every time. You are adding another networking to existing networking without any benefit except more networking. Calculations on clients should make it run smoothly too.

TimGoll:
> > > And tbh, I'm definitely against automatical networking. It's the same as every time: Why trying to fix buggy add-ons with another add-on if that decreases the performance and increases the networking overflow... Things related to this should be handled by the server and by the client separately OR by gmod itself. Using a networking system that isn't made for things like this isn't recommended imho.
> > 
> > 
> > I don't see how this decreases the performance. Actually I think it increases it. Prior to this change every addon had to do this manually, resulting in way more net messages than having one netmessage for all addons.
> > Edit: Like discussed with tkindanight, I'll switch from his system to a normal netmessage.
> 
> okay... The calculations on the server are still the same. The calcs on the client could decrease because it is just networked, that's correct. BUT networking is something that shouldn't be done for useless stuff, it's all the time recommended keeping simple calculations done by the device itself than using another device to calculate it. It's just useful if the operation is complicated and the other device offers better performance.
> This is NOT the case. Additionally, gmod automatically syncs the position of the clients by the server. That means that the "running" is done automatically by the server like every time. You are adding another networking to existing networking without any benefit except more networking. Calculations on clients should make it run smoothly too.

The calculation is done on the server in any case. If it is synced or not. So adding syncing here doesn't change much. Especially since it is only synced on change which doesn't happen that often.

Leaving it the old way makes it really hard for me to display any info at all in the HUD since the info on the client isn't necessarily the same as the info on the server.

Alf21:
> Also this fixes addons that have done it wrong and therefore removes the choppy walking which in itself is a huge plus for me.

You really should try to understand what I'm writing. Your last sentence is something we already cleared in many other discussions.

> Leaving it the old way makes it really hard for me to display any info at all in the HUD since the info on the client isn't necessarily the same as the info on the server.

And your second sentence relies on the same issue. If bugs exist, they should get fixed instead of trying to create a workaround.

> The calculation is done on the server in any case. If it is synced or not. So adding syncing here doesn't change much. Especially since it is only synced on change which doesn't happen that often.

And - as I already stated - it changes smth: Extra (unnecessary) networking.

> which doesn't happen that often

Any sprinting forcing an extra syncing - per player!

---

All in all, you are just adding a second layer over an already existing one. It's not necessary. Position checks are done by the server, and those `checks` (corrections) are done through a separate pipeline. That means that it can happen that this change could lead to a bigger lagging behavior than it had before. Doing it on the client is leading to a clean result. Things like visibility are things for the client. That's the reason why it should be handled by them.

TimGoll:
> Any sprinting forcing an extra syncing - per player!

No, sprinting is still handled shared like it has always been the case. Only the SpeedModifier hook is changed.

> And - as I already stated - it changes smth: Extra (unnecessary) networking.

Like I said. It removes x amounts of netmessages for adding one. Without this, many roles and addons sync it for themselves (at least the shinigami and supersoda). That's at least two netmessages removed for one added. So I don't see your point here.

Alf21:
> No, sprinting is still handled shared like it has always been the case. Only the SpeedModifier hook is changed.

So if I enable sprinting, all players will enable sprint?

> Like I said. It removes x amounts of netmessages for adding one. Without this, many roles and addons sync it for themselves (at least the shinigami and supersoda). That's at least two netmessages removed for one added. So I don't see your point here.

Maybe I don't get the point, but why should other add-ons sync it on their own. E.g. if I've an item and the `TTTPlayerSpeedModifier` hook is running, I don't need doing anything with syncing. I just need to change the value on the client in the same way as on the server.

TimGoll:
> So if I enable sprinting, all players will enable sprint?

No. The player pressing sprint will sprint, the other won't. Why would they all sprint if one player sprints?

> Maybe I don't get the point, but why should other add-ons sync it on their own

Because not everything is an item, sodas for example. They sync their state from server to client as well. Therefore needing syncing. This change removes that.

gamemodes/terrortown/gamemode/shared/sh_speed.lua
```diff
@@ -2,55 +2,59 @@ local plymeta = assert(FindMetaTable("Player"), "FAILED TO FIND ENTITY TABLE")
...
-	---
-	-- Sets the speed multiplier to an absolute value
-	-- @param number value The new value
-	-- @realm server
-	function plymeta:SetSpeedMultiplier(value)
-		if self:GetSpeedMultiplier() == value then return end
-
-		self.speedModifier = math.max(0, value)
+---
+-- A hook to modify the player speed, it is automatically networked.
```
Alf21, line 48: `automatically networked`?

PR #528, opened 2020-04-30
Revival Rework

TimGoll:
This started out as a small improvement over the default Revival function but grow bigger fast. Things I've done:

- improved the revival docs
- added a new parameter to the revival function, the position
- makes now sure the position is safe so that the player is not stuck in a wall
- added a few functions to get some informations about dead players (shared)
- moved all revival chat prints to the localized MSTACK
- added a HUD element that is shown while being revived, this greatly reduces the confusion for roles with a revival aspect
- added a function to add info to this HUD element, it is called on the server or client and supports strings with params inside
- some minor changes

image

image


gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -821,85 +826,192 @@ local function FindCorpsePosition(corpse)
...
+-- @param [default=false]boolean blockRound Stops the round from ending if this is set to true until the player is alive again
+-- @param [opt]function OnFail This @{function} is called if the revive fails
+-- @param [opt]Vector spawnPos The position where the player should be spawned, accounts for minor obstacles
+-- @realm server
+function plymeta:Revive(delay, OnRevive, DoCheck, needsCorpse, blockRound, OnFail, spawnPos)
 	local ply = self
 	local name = "TTT2RevivePlayer" .. ply:EntIndex()
 
-	if timer.Exists(name) or ply.reviving then return end
+	if timer.Exists(name) or ply:IsReviving() then return end
```
lebroomer, line 894: It shouldn't be possible that a timer ```TTT2RevivePlayer...``` exist and the player is not reviving. What is the reason for checking that timer explicitely?

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -821,85 +826,192 @@ local function FindCorpsePosition(corpse)
...
+		if not isfunction(DoCheck) or DoCheck(ply) then
 			local corpse = FindCorpse(ply)
 
-			if needcorpse and (not IsValid(corpse) or corpse:IsOnFire()) then
-				ply:ChatPrint("You have not been revived because your body no longer exists.")
+			if needsCorpse and (not IsValid(corpse) or corpse:IsOnFire()) then
+				OnReviveFailed(ply, OnFail, "message_revival_failed_missing_body")
 
-				timer.Remove(name) -- TODO needed?
+				timer.Remove(name)
```
lebroomer, line 920: The timer is automatically destroyed after the event is fired.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -821,85 +826,192 @@ local function FindCorpsePosition(corpse)
 	return false
 end
 
+local spawnPositions = {}
+for i = 0, 360, 22.5 do
+	spawnPositions[#spawnPositions + 1] = Vector(mathCos(i), mathSin(i), 0)
+end
+spawnPositions[#spawnPositions + 1] = Vector(0, 0, 1)
+
+local function MakeSpawnPositionSafe(unsafePos)
```
lebroomer, line 835: Add a small comment describing the function

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -821,85 +826,192 @@ local function FindCorpsePosition(corpse)
...
+
+		if not traceGround.HitWorld then continue end
+
+		return pos - Vector(0, 0, 0.5 * sizePlayer.z)
+	end
+
+	return unsafePos
+end
+
+local function OnReviveFailed(ply, failMessage)
```
lebroomer, line 867: Add a small comment describing the function

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -953,3 +953,83 @@ hook.Add("TTTEndRound", "TTTEndRound4TTT2TargetPlayer", function()
...
+	return self:TTT2NETGetFloat("player_revival_time", 1.0)
+end
+
+---
+-- Returns the death position of the player.
+-- @return [default=Vector(0.0, 0.0, 0.0)]Vector The death position
+-- @realm shared
+function plymeta:GetDeathPosition()
+	return Vector(
+		self:TTT2NETGetFloat("player_death_pos_x", 0.0),
```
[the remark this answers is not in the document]
TimGoll: Maybe. But this is totally out of scope of this PR

gamemodes/terrortown/gamemode/shared/hud_elements/tttrevivalinfo/pure_skin_revival.lua
```diff
@@ -0,0 +1,165 @@
...
+	function HUDELEMENT:GetDefaults()
+		const_defaults["basepos"] = {x = math.Round(ScrW() * 0.5 - self.size.w * 0.5), y = 0.5 * ScrH() + 100}
+
+		return const_defaults
+	end
+
+	function HUDELEMENT:ShouldDraw()
+		local client = LocalPlayer()
+
+		return HUDEditor.IsEditing or (not client:Alive() and client:IsReviving())
```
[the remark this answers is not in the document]
TimGoll: Should never happen. However I wanted to make sure that the element isn't displayed when not reviving.

gamemodes/terrortown/gamemode/shared/hud_elements/tttrevivalinfo/pure_skin_revival.lua
```diff
@@ -0,0 +1,165 @@
...
+				"PureSkinBar",
+				self.scale
+			)
+
+			revivalReasonLines = lines
+			lineHeight = textHeight / #revivalReasonLines
+
+			h = defaults.size.h + textHeight + self.pad
+		else
+			h = defaults.size.h
```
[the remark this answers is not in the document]
TimGoll: How do I change the size based on the contents without using defaults? Since it is not resizable in y direction, this should be fine.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: It has been this way. Maybe it has a reason.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: Last time I did this, Histaled meant they are unnecessary and I had to remove them

gamemodes/terrortown/gamemode/shared/hud_elements/tttrevivalinfo/pure_skin_revival.lua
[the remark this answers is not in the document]
lebroomer: But isn't that the task of the ```IsReviving()``` function?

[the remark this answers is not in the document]
TimGoll: Yes. But maybe someone abuses the `SetReviving()` function

gamemodes/terrortown/gamemode/shared/hud_elements/tttrevivalinfo/pure_skin_revival.lua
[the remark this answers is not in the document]
lebroomer: It is nevertheless bad style to use the defaults like this (bad for later maintenance and readibility). It would be better to just use a constant/member value to make it explicit.

gamemodes/terrortown/gamemode/shared/hud_elements/tttrevivalinfo/pure_skin_revival.lua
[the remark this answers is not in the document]
lebroomer: But this is not a problem, which should get tackled here in the HUDElement. Either move it to the ```IsReviving``` function or leave it as it. Code logic in a HUDelement it not the best style.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
lebroomer: Noone should create a timer named like that. If a timer with that name exists, it is completely fine to overwrite it with ```timer.Create()``` instead of canceling the revive.

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,290 @@
...
+	end
+
+	-- the table should be shuffled for each spawn point calculation to improve randomness
+	table.Shuffle(spawn.cachedSpawnEntities)
+
+	-- Optimistic attempt: assume there are sufficient spawns for all and one is free
+	for i = 1, #spawn.cachedSpawnEntities do
+		local spawnEntity = spawn.cachedSpawnEntities[i]
+
+		if not IsValid(ply) or not ply:IsTerror() then
```
saibotk, line 232: Why are these checks now needed?

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,290 @@
...
+		if riggedSpawnPoint then
+			local riggedSpawnEntity = ents.Create("info_player_terrorist")
+
+			riggedSpawnEntity:SetPos(riggedSpawnPoint)
+			riggedSpawnEntity:Spawn()
+
+			ErrorNoHalt("TTT2 WARNING: Map has too few spawn points, using a riggedSpawnPoints spawn for " .. tostring(ply) .. "\n")
+
+			-- this is an old TTT flag that I will keep for compatibilities sake
+			GAMEMODE.HaveriggedSpawnPointsSpawn = true
```
saibotk, line 258: Then please keep the original flag name `HaveRiggedSpawn`

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,290 @@
...
+spawn = spawn or {}
+
+spawn.cachedSpawnEntities = spawn.cachedSpawnEntities or {}
+
+---
+-- Checks if a given spawn point is safe. Safe means that a player can spawn without
+-- being stuck. Entities which are passable are ignored by the check.
+-- @param Vector pos The respawn position
+-- @param [default=false]boolean force Should the respawn be forced? This means killing other players that block this position
+-- @param [opt]table|Entity filder A table of entities or an entity that should be ignored by this check
```
saibotk, line 41: `filder` -> `filter`

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,290 @@
...
+-- @return bollean Teturns if the spawn point is safe
```
saibotk, line 42: `bollean` -> `boolean` 
`Treturns` -> `returns`

lua/ttt2/extensions/player.lua
```diff
@@ -0,0 +1,64 @@
+---
+-- player extension
```
saibotk, line 2: Actually, i dislike these functions, as they all do not really belong to the player library in my opinion, as they rather look like functions, with game logic, which is more suited in the appropriate code of the game state

lua/ttt2/extensions/player.lua
```diff
@@ -0,0 +1,64 @@
...
+-- @author Mineotopia
+
+---
+-- Resets the spawn position of all players.
+-- @realm server
+function player.ResetSpawnPosition()
+	local plys = player.GetAll()
+
+	for i = 1, #plys do
+		plys[i]:SetSpawnPosition(Vector(0, 0, 0))
```
saibotk, line 12: Setting the position to 0,0,0 should not be considered a "reset" rather use nil, to differentiate between unset / set

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -787,119 +789,191 @@ local function FindCorpse(ply)
...
+	self:TTT2NETSetFloat("player_death_pos_y", pos.y or 0.0)
+	self:TTT2NETSetFloat("player_death_pos_z", pos.z or 0.0)
+end
+
+---
+-- Sets the spawn position
+-- @param Vector pos The spawn position
+-- @realm server
+function plymeta:SetSpawnPosition(pos)
+	self:TTT2NETSetFloat("player_spawn_pos_x", pos.x or 0.0)
```
saibotk, line 954: You could put them in their own path, to group the values: `"player_spawn_pos_x"` -> `{"player_spawn_pos", "x"}`

saibotk: Also the system is capable of setting values to `nil`, which should be considered instead of 0.0 as a default value

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -787,119 +789,191 @@ local function FindCorpse(ply)
...
+	self:TTT2NETSetFloat("player_death_pos_x", pos.x or 0.0)
+	self:TTT2NETSetFloat("player_death_pos_y", pos.y or 0.0)
+	self:TTT2NETSetFloat("player_death_pos_z", pos.z or 0.0)
+end
+
+---
+-- Sets the spawn position
+-- @param Vector pos The spawn position
+-- @realm server
+function plymeta:SetSpawnPosition(pos)
```
saibotk, line 953: What is this needed for and why does the client need this information? Also what exactly does this do? The function name implies somehow, that this will actually affect a subsequent spawn to happen on the given position, which is not.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -787,119 +789,191 @@ local function FindCorpse(ply)
...
+	self:TTT2NETSetFloat("player_revival_time", time or 0.0)
+end
+
+---
+-- Sets the death position
+-- @param Vector pos The death position
 -- @realm server
-function plymeta:IsReviving()
-	return self.reviving or false
+function plymeta:SetDeathPosition(pos)
```
saibotk, line 943: Same questions as for the SpawnPosition

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
saibotk: Please also note, that this will result in all players knowing about the spawn position of each other player.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -787,119 +789,191 @@ local function FindCorpse(ply)
...
+end
+
+---
+-- Sets the if a player died in a round.
+-- @param boolean state The state
+-- @realm server
+function plymeta:SetDiedInRound(state)
+	if state and (GetRoundState() ~= ROUND_ACTIVE or not self:IsTerror()) then return end
+
+	self:TTT2NETSetBool("player_has_died_in_round", state or false)
```
saibotk, line 976: Please note, that this information is broadcasted to all players

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -787,119 +789,191 @@ local function FindCorpse(ply)
...
+end
+
+---
+-- Sets the if a player was active (TEAM_TERROR) in a round.
+-- @param boolean state The state
+-- @realm server
+function plymeta:SetActiveInRound(state)
+	if state and (GetRoundState() ~= ROUND_ACTIVE or not self:IsTerror()) then return end
+
+	self:TTT2NETSetBool("player_was_active_in_round", state or false)
```
saibotk, line 966: Please note, that this information is broadcasted to all players

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -787,119 +789,191 @@ local function FindCorpse(ply)
...
+
+	OnReviveFailed(self, failMessage or "message_revival_canceled")
+end
+
+---
+-- Sets the revival state.
+-- @param [default=false]boolean isReviving The reviving state
+-- @realm server
+function plymeta:SetReviving(isReviving)
+	self:TTT2NETSetBool("player_is_reviving", isReviving or false)
```
saibotk, line 912: Please note, that this information is broadcasted to all players

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -787,119 +789,191 @@ local function FindCorpse(ply)
...
+end
+
+---
+-- Sets the blocking revival state
+-- @param [default=false]boolean isBlockingRevival The blocking revival state
+-- @realm server
+function plymeta:SetBlockingRevival(isBlockingRevival)
+	self:TTT2NETSetBool("player_is_blocking_revival", isBlockingRevival or false)
```
saibotk, line 920: Please note, that this information is broadcasted to all players

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -787,119 +789,191 @@ local function FindCorpse(ply)
...
+end
+
+---
+-- Sets the revival start time
+-- @param [default=@{CurTime()}]number startTime The revival start time
+-- @realm server
+function plymeta:SetRevivalStartTime(startTime)
+	self:TTT2NETSetFloat("player_revival_start_time", startTime or CurTime())
```
saibotk, line 928: Please note, that this information is broadcasted to all players

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -787,119 +789,191 @@ local function FindCorpse(ply)
...
+end
+
+---
+-- Sets the revival time
+-- @param [default=0.0]number time The revival time
+-- @realm server
+function plymeta:SetRevivalTime(time)
+	self:TTT2NETSetFloat("player_revival_time", time or 0.0)
```
saibotk, line 936: Please note, that this information is broadcasted to all players

gamemodes/terrortown/gamemode/shared/sh_init.lua
```diff
@@ -525,6 +525,11 @@ include("ttt2/extensions/net.lua")
 include("ttt2/extensions/string.lua")
 include("ttt2/extensions/table.lua")
 
+if SERVER then
+	include("ttt2/extensions/player.lua")
```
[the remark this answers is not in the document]
TimGoll: I though all libraries/extensions are included in this file. It this is not the case, I can move the import.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -953,3 +953,120 @@ hook.Add("TTTEndRound", "TTTEndRound4TTT2TargetPlayer", function()
...
+end
+
+---
+-- Sets a revival reason that is displayed in the revival HUD element.
+-- It supports language identifier for translated strings.
+-- @param [default=nil]string name The text or the language identifer, nil to reset
+-- @param [opt]table params The params table used for @{LANG.GetParamTranslation}
+-- @realm shared
+function plymeta:SetRevivalReason(name, params)
+	if SERVER then
```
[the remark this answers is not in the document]
TimGoll: Since it has the same arguments on client and server, I think combining it is the better approach.

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -133,6 +132,9 @@ function GM:PlayerSpawn(ply)
 
 	SCORE:HandleSpawn(ply)
 
+	ply:SetSpawnPosition(ply:GetPos())
+	ply:SetDeathPosition(Vector(0, 0, 0))
```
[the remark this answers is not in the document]
TimGoll: I did not know your system supports nil

lua/ttt2/libraries/spawn.lua
TimGoll: These came from original TTT. They are needed because as a spectator, your spawnpoint should always be valid.

lua/ttt2/libraries/spawn.lua
TimGoll: Thanks. I don't know what happened here.

lua/ttt2/extensions/player.lua
TimGoll: I though `player` does stuff to all players. Like `player.GetAll()`

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: I though I needed this for the respawn caching, but didn't need it in the end. But I decided to keep it since maybe an addon wants to use this data.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: Death position is needed for respawn at death position.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: yes. This info is needed on all clients for the respawn indicator

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: yes, this info will be needed (in the future) for the respawn indicator

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: while it would be enough (for now) to sync this info to the one player, I see no problem with having it broadcasted

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: while it would be enough (for now) to sync this info to the one player, I see no problem with having it broadcasted

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: while it would be enough (for now) to sync this info to the one player, I see no problem with having it broadcasted

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -164,123 +166,31 @@ end
...
-
-		if force then
-			pl:Kill()
-		else
-			return false
-		end
-	end
-
-	return true
+	return spawn.IsSpawnPointSafe(spawnEntity:GetPos(), force)
```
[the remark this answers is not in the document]
TimGoll: You're right. This is an artifact of the old system

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -597,7 +602,7 @@ function plymeta:ResetViewRoll()
 end
 
 ---
--- Checks whether a @{Player} is able to spawn
+-- DoChecks whether a @{Player} is able to spawn
```
lebroomer: DoChecks again? ^^

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
lebroomer: I think they are neccessary, especially if the functions are non trivial

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1077,7 +1189,7 @@ function plymeta:Give(weaponClassName, bNoAmmo)
 end
 
 ---
--- Called to drop a weapon in a safe manner (e.g. preparing and space-check)
+-- Called to drop a weapon in a safe manner (e.g. preparing and space-DoCheck)
```
lebroomer: DoCheck

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1223,7 +1335,7 @@ function plymeta:PickupWeapon(wep, dropBlockingWeapon, shouldAutoSelect)
 	-- set in the runtime, we have to use the GM:PlayerCanPickupWeapon hook in a
 	-- slightly hacky way
 
-	-- first we have to check if the player can pick up the weapon at all by running the
+	-- first we have to DoCheck if the player can pick up the weapon at all by running the
```
lebroomer: DoCheck

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
lebroomer: I think they are neccessary, escpecially when a function is complicated.

TimGoll: I've already added it^^

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,290 @@
...
+		if not spawn.IsSpawnPointSafe(spawnPoint, false) then continue end
+
+		return spawnPoint
+	end
+end
+
+---
+-- Returns a list of all spawn entities
+-- @param boolean shouldShuffle whether the table should be shuffled
+-- @param boolean forceAll used unless absolutely necessary (includes info_player_start spawns)
```
Histalek, line 151: i can't comprehend what you mean by this

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,290 @@
...
+		if not spawn.IsSpawnPointSafe(spawnEntity:GetPos(), true) then continue end
+
+		return spawnEntity
+	end
+
+	return pickedSpawnEntity
+end
+
+---
+-- Removes all map spawn entities, shouldn't be used except you know what you're doing.
```
Histalek, line 277:
```suggestion
-- Removes all map spawn entities.
-- @warning Do not use this unless you know exactly what you are doing.
```

gamemodes/terrortown/gamemode/shared/hud_elements/tttrevivalinfo/pure_skin_revival.lua
```diff
@@ -0,0 +1,166 @@
...
+	local colorRevivingBar = Color(36, 154, 198)
+
+	local const_defaults = {
+		basepos = {x = 0, y = 0},
+		size = {w = 321, h = defaultHeight},
+		minsize = {w = 250, h = defaultHeight}
+	}
+
+	function HUDELEMENT:Initialize()
+		self.pad = pad
```
lebroomer, line 25: you should add self.defaultHeight here too, similar to self.pad

gamemodes/terrortown/gamemode/shared/sh_corpse.lua
```diff
@@ -72,3 +72,12 @@ function CORPSE.GetPlayer(rag)
 
 	return rag:GetDTEntity(dti.ENT_PLAYER)
 end
+
+---
+-- Checks if a tagdoll is a valid ragdoll because they could be a matress as well.
```
Histalek, line 77:
```suggestion
-- Checks if a ragdoll belongs to a player.
-- Because ragdolls can also be used for props (e.g. a mattress).
```

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,290 @@
...
+local mathSin = math.sin
+local mathCos = math.cos
+local entsFindByClass = ents.FindByClass
+
+local spawnPointVariations = {Vector(0, 0, 0)}
+for i = 0, 360, 22.5 do
+	spawnPointVariations[#spawnPointVariations + 1] = Vector(mathCos(i), mathSin(i), 0)
+end
+
+local sizePlayer = Vector(32, 32, 72)
```
lebroomer, line 16: I am not sure if you can just assume a player's size like that. Different playermodels (especially weird customized ones) can result in different player sizes.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -953,3 +953,120 @@ hook.Add("TTTEndRound", "TTTEndRound4TTT2TargetPlayer", function()
...
+---
+-- Returns the time when the ongoing revival started.
+-- @return [default=@{CurTime()}]number The time when the revival started in seconds
+-- @realm shared
+function plymeta:GetRevivalStartTime()
+	return self:TTT2NETGetFloat("player_revival_start_time", CurTime())
+end
+
+---
+-- Returns the time for the ongoing revival.
```
Histalek, line 982: do you mean the full duration of the revival process? the remaining duration until the revive happens? an absolute point in time?

lua/ttt2/libraries/spawn.lua
TimGoll: This is actually a great point, yes! It has been like this forever so I did not think about it at all. But yes, that's absolutely right!

Alf21:
Just an idea: how about adding a possibility to use a binded key to trigger an early respawn?

TimGoll:
> Just an idea: how about adding a possibility to use a binded key to trigger an early respawn?

Probably something we have to discuss. But since many addons will use this, I don't think it is something that should be enabled by default. See: defibrillator

lua/ttt2/extensions/table.lua
```diff
@@ -4,6 +4,9 @@
 
 AddCSLuaFile()
 
+local pairs = pairs
+local ipairs = ipairs
```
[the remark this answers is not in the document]
TimGoll: yes you miss something:

https://github.com/TTT-2/TTT2/blob/52a841e01313b2f036b1fd12e2cc01fdb5c02b8f/lua/ttt2/extensions/table.lua#L167

lua/ttt2/libraries/spawn.lua
TimGoll: This is copied from vanilla TTT, but you're right, I'll change it

lua/ttt2/extensions/player.lua
```diff
@@ -5,22 +5,22 @@
 ---
 -- Resets the spawn position of all players.
 -- @realm server
-function player.ResetSpawnPosition()
+function player.ReSetLastSpawnPosition()
```
lebroomer, line 8:
```suggestion
function player.ResetLastSpawnPosition()
```

lua/ttt2/extensions/player.lua
```diff
@@ -5,22 +5,22 @@
...
-		plys[i]:SetSpawnPosition(Vector(0, 0, 0))
+		plys[i]:SetLastSpawnPosition(nil)
 	end
 end
 
 ---
 -- Resets the death position of all players.
 -- @realm server
-function player.ResetDeathPosition()
+function player.ReSetLastDeathPosition()
```
lebroomer, line 19:
```suggestion
function player.ResetLastDeathPosition()
```

lua/ttt2/libraries/spawn.lua
```diff
@@ -27,21 +25,29 @@ local spawnTypes = {
 	"info_player_teamspawn"
 }
 
+-- returns the player size as a vector
+local function GetPlayerSize(ply)
+	local bottom, top = ply:GetHull()
 
+	return -1 * bottom + top
```
lebroomer, line 32: Does vector math support unary negation? If yes, ```return -bottom + top``` would fit better.

saibotk: or `return top - bottom`

TimGoll: I don't like lebroomers suggestion, therefore I used `-1  * x`, but saibotks solution is rather nice. Thanks. Too easy!

lua/ttt2/extensions/player.lua
TimGoll: the player library will be removed completely. I had talked about this with saibotk, so therefore this will no longer be important.

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -184,3 +184,75 @@ function GM:SetupMove(ply, mv, cmd)
...
+	LocalPlayer().revivalStartTime = net.ReadFloat()
+end)
+
+net.Receive("TTT2RevivalUpdate_RevivalDuration", function()
+	LocalPlayer().revivalDurarion = net.ReadFloat()
+end)
+
+---
+-- Returns if a player has a revival reason set.
+-- @return [default=false]boolean Returns if a player has a revival reason
```
Alf21, line 246: wrong doc style

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -184,3 +184,75 @@ function GM:SetupMove(ply, mv, cmd)
...
+-- @realm client
+function plymeta:HasRevivalReason()
+	return (self.revivalReason and self.revivalReason.name and self.revivalReason.name ~= "") or false
+end
+
+---
+-- Returns the current revival reason.
+-- @return [default={}]table The revival reason table
```
Alf21, line 254: wrong doc style

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -184,3 +184,75 @@ function GM:SetupMove(ply, mv, cmd)
 
 	hook.Run("TTT2PlayerReady", ply)
 end
+
+---
+-- Sets a revival reason that is displayed in the revival HUD element.
+-- It supports a language identifier for translated strings.
+-- @param [default=nil]string name The text or the language identifer, nil to reset
+-- @param [opt]table params The params table used for @{LANG.GetParamTranslation}
```
Alf21, line 192: wrong doc style

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -184,3 +184,75 @@ function GM:SetupMove(ply, mv, cmd)
 
 	hook.Run("TTT2PlayerReady", ply)
 end
+
+---
+-- Sets a revival reason that is displayed in the revival HUD element.
+-- It supports a language identifier for translated strings.
+-- @param [default=nil]string name The text or the language identifer, nil to reset
```
Alf21, line 191: wrong doc style

gamemodes/terrortown/gamemode/server/sv_entity.lua
```diff
@@ -2,6 +2,10 @@
 -- @module Entity
 -- @ref https://wiki.garrysmod.com/page/Category:Entity
 
+local safeCollisionGroups = {
+	[COLLISION_GROUP_WEAPON] = true
```
Alf21, line 6: if there is just one static entry, we don't need to use a table. Just return below `self:GetCollisionGroup() == COLLISION_GROUP_WEAPON`

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -1191,6 +1196,17 @@ function BeginRound()
...
 
+	local plys = player.GetAll()
+
+	for i = 1, #plys do
+		local ply = plys[i]
+
+		ply:SetDiedInRound(false)
+
+		-- a player should be considered "was active in round" if they received a role
+		ply:SetActiveInRound(ply:Alive() and ply:IsTerror())
```
Alf21, line 1207: Is it really needed? There are functions like `ply:GetForceSpec()` i think

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -929,9 +783,12 @@ end
 -- @param Player|Entity attacker @{Player} or @{Entity} that killed the victim
 -- @hook
 -- @realm server
--- @ref https://wiki.garrysmod.com/page/GM/PlayerDeath
+-- @ref https://wiki.facepunch.com/gmod/GM:PlayerDeath
 -- @local
 function GM:PlayerDeath(victim, infl, attacker)
+	victim:SetLastDeathPosition(victim:GetPos())
+	victim:SetDiedInRound(true)
```
Alf21, line 790: If we adding a storing variable here, why not using a counter instead? This could be useful in the future too.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
 -- Revives a @{Player}
--- @param number delay the delay of the revive
--- @param function fn the @{function} that should be run if the @{Player} revives
--- @param function check an additional checking @{function}
--- @param boolean needcorpse whether the dead @{Player} CORPSE is needed
--- @param boolean force if the @{Player} revive is already forced. Useful if you have multiple reviving equipments
--- @param function onFail this @{function} is called if the revive fails
--- @realm server
-function plymeta:Revive(delay, fn, check, needcorpse, force, onFail)
+-- @param [default=3]number delay The delay of the revive
```
Alf21, line 814: wrong doc style (all lines below too)

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
+-- @param [opt]function OnRevive The @{function} that should be run if the @{Player} revives
+-- @param [opt]function DoCheck An additional checking @{function}
+-- @param [default=false]boolean needsCorpse Whether the dead @{Player} @{CORPSE} is needed
+-- @param [default=false]boolean blockRound Stops the round from ending if this is set to true until the player is alive again
+-- @param [opt]function OnFail This @{function} is called if the revive fails
+-- @param [opt]Vector spawnPos The position where the player should be spawned, accounts for minor obstacles
+-- @param [opt]Angle spawnEyeAngle The eye angles of the revived players
+-- @realm server
+function plymeta:Revive(delay, OnRevive, DoCheck, needsCorpse, blockRound, OnFail, spawnPos, spawnEyeAngle)
 	local ply = self
```
Alf21, line 824: This isn't needed here

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
 	local name = "TTT2RevivePlayer" .. ply:EntIndex()
```
Alf21, line 825: this should be localized right after the if-check

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
-					ply:SetEyeAngles(Angle(0, corpse:GetAngles().y, 0))
-				end
+			if not spawnPos and IsValid(corpse) then
+				spawnPos = corpse:GetPos()
+				spawnEyeAngle = Angle(0, corpse:GetAngles().y, 0)
 			end
 
-			hook.Call("PlayerLoadout", GAMEMODE, ply)
+			spawnPos = spawnPos or ply:GetDeathPosition()
+			spawnPos = spawn.MakeSpawnPointSafe(self, spawnPos)
```
Alf21, line 864: this can be combined with the previous line

Alf21: You are possibly changing the `spawnPos`, but not the `spawnAngle` if given?

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
 
+				spawnPos = spawnEntity:GetPos()
+				spawnEyeAngle = spawnEntity:EyeAngles()
+			end
+
+			ply:SetPos(spawnPos)
+			ply:SetEyeAngles(spawnEyeAngle or Angle(0, 0, 0))
 			ply:SetMaxHealth(100)
 
+			hook.Run("PlayerLoadout", ply)
```
Alf21, line 877: Shouldn't this be handled automatically by `GM:PlayerSpawn`? Ref.: https://wiki.facepunch.com/gmod/GM:PlayerSpawn

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
+			OnReviveFailed(ply, "message_revival_failed")
 		end
 	end)
 end
 
 ---
--- Returns if a player is currently in a revival process started by @{Player:Revive}
--- @return boolean The revival status
+-- Cancel the ongoing revival process.
+-- @param [default="message_revival_canceled"]string failMessage The fail message that should be displayed for the client
```
Alf21, line 901: wrong doc style

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
+
+	net.Start("TTT2RevivalUpdate_RevivalDuration")
+	net.WriteFloat(self.revivalDurarion)
+	net.Send(self)
+end
+
+---
+-- Sends a revival reason that is displayed in the clients revival HUD element.
+-- It supports a language identifier for translated strings.
+-- @param [default=nil]string name The text or the language identifer, nil to reset
```
Alf21, line 971: wrong doc style

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
+	self.revivalStartTime = startTime or CurTime()
+
+	net.Start("TTT2RevivalUpdate_RevivalStartTime")
+	net.WriteFloat(self.revivalStartTime)
+	net.Send(self)
+end
+
+---
+-- Sets the revival duration.
+-- @param [default=0.0]number duration The revival time
```
Alf21, line 957: wrong doc style

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
+	self.isBlockingRevival = isBlockingRevival or false
+
+	net.Start("TTT2RevivalUpdate_IsBlockingRevival")
+	net.WriteBool(self.isBlockingRevival)
+	net.Send(self)
+end
+
+---
+-- Sets the revival start time.
+-- @param [default=@{CurTime()}]number startTime The revival start time
```
Alf21, line 944: wrong doc style

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
+	self.isReviving = isReviving or false
+
+	net.Start("TTT2RevivalUpdate_IsReviving")
+	net.WriteBool(self.isReviving)
+	net.Send(self)
+end
+
+---
+-- Sets the blocking revival state.
+-- @param [default=false]boolean isBlockingRevival The blocking revival state
```
Alf21, line 931: wrong doc style

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
+
+	timer.Remove("TTT2RevivePlayer" .. self:EntIndex())
+
+	OnReviveFailed(self, failMessage or "message_revival_canceled")
+end
+
+
+---
+-- Sets the revival state.
+-- @param [default=false]boolean isReviving The reviving state
```
Alf21, line 918: wrong doc style

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
+	return self.lastSpawnPosition
+end
+
+---
+-- Sets the if a player was active (TEAM_TERROR) in a round.
+-- @param boolean state The state
+-- @internal
+-- @realm server
+function plymeta:SetActiveInRound(state)
+	if state and (GetRoundState() ~= ROUND_ACTIVE or not self:IsTerror()) then return end
```
Alf21, line 1039: Is this check really needed? I mean if you wanna set it, you don't get feedback whether it was successfully.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,261 @@ local function FindCorpse(ply)
...
+	self:TTT2NETSetBool("player_was_active_in_round", state or false)
+end
+
+---
+-- Sets the if a player died in a round.
+-- @param boolean state The state
+-- @internal
+-- @realm server
+function plymeta:SetDiedInRound(state)
+	if state and (GetRoundState() ~= ROUND_ACTIVE or not self:IsTerror()) then return end
```
Alf21, line 1050: same as above

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -955,3 +955,60 @@ hook.Add("TTTEndRound", "TTTEndRound4TTT2TargetPlayer", function()
...
+function plymeta:HasDiedInRound()
+	return self:TTT2NETGetBool("player_has_died_in_round", false)
+end
+
+---
+-- Checks if a player was revived in the round.
+-- @return boolean Returns if the player was revived
+-- @realm shared
+function plymeta:WasRevivedInRound()
+	return self:HasDiedInRound() and self:IsTerror()
```
Alf21, line 1013: The second check isn't needed and would lead to the wrong result. A died terror is in spectator team.

lua/ttt2/extensions/table.lua
```diff
@@ -58,3 +61,116 @@ function table.SetWithPath(dataTable, path, value)
...
+	end
+
+	return out
+end
+
+---
+-- This @{function} adds missing values into a table
+-- @param table target
+-- @param table source
+-- @param boolean iterable
```
Alf21, line 162: add a `?` or `nil|` in front of `boolean`

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,303 @@
...
+-- @author Mineotopia
+
+local Vector = Vector
+local table = table
+local mathSin = math.sin
+local mathCos = math.cos
+local entsFindByClass = ents.FindByClass
+
+local spawnPointVariations = {Vector(0, 0, 0)}
+for i = 0, 360, 22.5 do
```
Alf21, line 12: empty line on front missing

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,303 @@
...
+	spawnPointVariations[#spawnPointVariations + 1] = Vector(mathCos(i), mathSin(i), 0)
+end
+
+local spawnTypes = {
```
Alf21, line 16: you should better use a key-true table instead of a list.

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,303 @@
...
+-- @realm server
+function spawn.GetSpawnPointsAroundSpawn(ply, pos)
+	local sizePlayer = GetPlayerSize(ply)
+
+	if not pos then return {} end
+
+	local positions = {}
+
+	for i = 1, #spawnPointVariations do
+		positions[#positions + 1] = pos + spawnPointVariations[i] * Vector(sizePlayer.x, sizePlayer.y, 0) * 1.5
```
Alf21, line 134: Better using `positions[i] = ...`

Alf21: The vector `Vector(sizePlayer.x, sizePlayer.y, 0)` should be stored outside this loop. You can multiply the 1.5 as well

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,303 @@
...
+-- @realm server
+function spawn.GetPlayerSpawnEntities(forceAll)
+	local tbl = {}
+
+	for i = 1, #spawnTypes do
+		local classname = spawnTypes[i]
+		local entsTbl = entsFindByClass(classname)
+
+		for k = 1, #entsTbl do
+			if entsTbl[k].markAsRemoved then continue end
```
Alf21, line 177: localize `entsTbl[k]`

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,303 @@
...
+	end
+
+	-- Don't use info_player_start unless absolutely necessary, because eg. TF2
+	-- uses it for observer starts that are in places where players cannot really
+	-- spawn well. At all.
+	if forceAll or #tbl == 0 then
+		local startTbl = entsFindByClass("info_player_start")
+
+		for k = 1, #startTbl do
+			if startTbl[k].markAsRemoved then continue end
```
Alf21, line 190: localize `startTbl[k]`

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,303 @@
...
+	-- cache. Otherwise, any riggedSpawnPoints spawn entities would not get reused, and
+	-- MORE new entities would be made instead. In reality, the map cleanup at
+	-- round start will remove our riggedSpawnPoints spawns, and we'll have to create new
+	-- ones anyway.
+	if #spawn.cachedSpawnEntities == 0 or not IsTableOfEntitiesValid(spawn.cachedSpawnEntities) then
+		spawn.cachedSpawnEntities = spawn.GetPlayerSpawnEntities(false)
+	end
+
+	if #spawn.cachedSpawnEntities == 0 then
+		Error("No spawn entity found!\n")
```
Alf21, line 232: Are you sure using `Error` instead of `ErrorNoHalt` ?

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,303 @@
...
+	end
+
+	-- the table should be shuffled for each spawn point calculation to improve randomness
+	table.Shuffle(spawn.cachedSpawnEntities)
+
+	-- Optimistic attempt: assume there are sufficient spawns for all and one is free
+	for i = 1, #spawn.cachedSpawnEntities do
+		local spawnEntity = spawn.cachedSpawnEntities[i]
+
+		if not IsValid(ply) or not ply:IsTerror() then
```
Alf21, line 244: shouldn't this be handled outside the for-loop?

lua/ttt2/libraries/spawn.lua
```diff
@@ -0,0 +1,303 @@
...
+
+	-- That did not work, so now look around spawns
+	local pickedSpawnEntity
+
+	for i = 1, #spawn.cachedSpawnEntities do
+		pickedSpawnEntity = spawn.cachedSpawnEntities[i]
+
+		local riggedSpawnPoint = spawn.MakeSpawnPointSafe(ply, pickedSpawnEntity:GetPos())
+
+		if riggedSpawnPoint then
```
Alf21, line 261: better using `if not riggedSpawnPoint then continue end`

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: could you explain what's wrong here?

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: same here, what's wrong with it?

gamemodes/terrortown/gamemode/server/sv_entity.lua
TimGoll: I know, but there are probably other safe groups that can be added at a later time

gamemodes/terrortown/gamemode/server/sv_main.lua
TimGoll: Maybe there is something that does the same already. I wasn't able to find anything and nobody in the dev group had an idea when I asked
Forced spec is something else though, latejoiners aren't forced spectators

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: nice idea, should this be reset on roundend? or on mapchange?

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: what's wrong with them?

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: This is from you^^

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: > this can be combined with the previous line

Hmm, I don't know how to combine it in a nice way. It would be
```lua
spawnPos = spawnPos or ply:GetDeathPosition() or spawn.MakeSpawnPointSafe(self, spawnPos or ply:GetDeathPosition())
```
wouldn't it?

TimGoll: > You are possibly changing the `spawnPos`, but not the `spawnAngle` if given?

And yes, the spawn angle doesn't rely matter. It's more or less just a visual thing, there are no unsafe spawn angles.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: I have no idea, this was in your revive function and I though you put your thought in it. I can test if it works without.

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
Alf21: Already told it so many times :D `@param[default=nil] string` not `@param [default=nil]string`

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: same here. I don't know what's wrong. This is the same style as everywhere

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: same here

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: same here

gamemodes/terrortown/gamemode/server/sv_main.lua
Alf21: LateJoiners should be forced specs as well, if not, it's a bug

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: I mean, not necessarily. But I though it would make the function safer.

gamemodes/terrortown/gamemode/server/sv_player.lua
Alf21: On roundend i think, same as you doing currently. On mapchange would be useless, therefor you can use the `ply:Death()` var (ref.: https://wiki.facepunch.com/gmod/Player:Deaths)

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
TimGoll: It should only return true if a player was revived in the round, hece this check. Otherwise it would still return true if the player is dead. Which is what `HasDiedInRound()` does.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
Alf21: Yea we came up localizing `self` before using it in timer, but this isn't needed outside the timer. But you are right

lua/ttt2/libraries/spawn.lua
TimGoll: I have no idea. It has been this way in the original TTT code. Probably an error is ok here if no player is able to spawn.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
Alf21: But if you change the position, wouldn't you have to change the angle as well? :o

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: I just checked the code and both styles are used 50:50. I guess one of them is wrong.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
Alf21: No, it would be `spawnPos = spawn.MakeSpawnPointSafe(self, spawnPos or ply:GetDeathPosition())`

gamemodes/terrortown/gamemode/server/sv_main.lua
TimGoll: aren't forced specs players that have set the "force spec" checkbox?

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
Alf21: They are used bc you submitted them i guess :')

gamemodes/terrortown/gamemode/server/sv_entity.lua
Alf21: Then we can change it later into a table.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: > No, it would be
> 
> ```lua
> spawnPos = spawn.MakeSpawnPointSafe(self, spawnPos or ply:GetDeathPosition())
> ```

Ah yes, I'm stupid!

> But if you change the position, wouldn't you have to change the angle as well? :o

I don't calculate the viewangle. In which way should I "correct" it?

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: Just checked it, most of them are by me, but not all of them. And you did not notice it while reviewing :P

gamemodes/terrortown/gamemode/server/sv_main.lua
Alf21: If so, there are bugs based by the idea that forced specs are players that were forced into the spec team (LateJoiners or CheckBox(by own or automatically if being idle))

TimGoll: Idle autosets the checkbox though, latejoining doesn't.
So there has to be a difference somewhere.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
Alf21: Yea I don't know, you are right, there is no data to calc the viewangle :D

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
Alf21: I can't see it in the old code, GitHub displays that you added it :o

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
Alf21: But based in the function name, A dead terror (spectator) could also have been revived in round...

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: yes. The angle is either given as parameter or based on the corpse position

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: Because I moved it to another file and put it in seperate functions. But the overall spawn logic isn't new, it's just a bit refactored

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
TimGoll: I'll have to check my intentions for this again and I'll reconsider the name and contents afterwars.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
saibotk: I think I'm with @Alf21, as this is unexpected behavior, and silently returning, makes debugging hard and logical mistakes even harder to find.

gamemodes/terrortown/gamemode/server/sv_entity.lua
saibotk: On this, I think leaving it as a table would keep it extensible and is no overhead that would have to be considered impactful, if it can even be noticed, which I suppose it won't.

gamemodes/terrortown/gamemode/server/sv_main.lua
Alf21: Then it's something we need to keep in mind here: https://github.com/TTT-2/TTT2/pull/517/files#diff-1512bbb3506493de161ace83dfe6b206R41

Alf21:
And pls give a look into these ideas too: https://github.com/TTT-2/TTT2/issues/475#issuecomment-589304855

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -184,3 +184,75 @@ function GM:SetupMove(ply, mv, cmd)
...
+	if not isReset then
+		name = net.ReadString()
+
+		local paramsAmount = net.ReadUInt(8)
+
+		if paramsAmount > 0 then
+			params = {}
+
+			for i = 1, paramsAmount do
+				client.revivalReason.params[net.ReadString()] = net.ReadString()
```
Alf21, line 215: `client.revivalReason.params` have to be `params`

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: See: <https://github.com/TTT-2/TTT2/pull/528/files#diff-bb7d466d24e307b3292a4722e8ade165L874>

gamemodes/terrortown/gamemode/server/sv_main.lua
TimGoll: Okay, I checked this code. Why should I change it to `ply:GetForceSpec()`? In my opinion my implementation is correct because a players has to be alive and not in the spectator team to be counted as "being active at round start". Forced spectator only sets the networked bool which might not cover all cases.

However `ply:SetActiveInRound(ply:IsTerror())` could be sufficient.

Alf21: Maybe you missed the entire discussion and the last meeting or the break was a bit too long 😅 You shouldn‘t replace it by GetForceSpec, but check whether code snippets that used GetForceSpec before need to change into your new system

TimGoll: But my system has nothing to do with spectators / forced specators

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,263 @@ local function FindCorpse(ply)
...
+	OnReviveFailed(self, failMessage or "message_revival_canceled")
+end
+
+
+---
+-- Sets the revival state.
+-- @param[default=false] boolean isReviving The reviving state
+-- @internal
+-- @realm server
+function plymeta:SetReviving(isReviving)
```
Alf21, line 920: return if the state is the same as before. Otherwise you could have senseless networking

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,263 @@ local function FindCorpse(ply)
...
+	net.Send(self)
+end
+
+---
+-- Sets the blocking revival state.
+-- @param[default=false] boolean isBlockingRevival The blocking revival state
+-- @internal
+-- @realm server
+function plymeta:SetBlockingRevival(isBlockingRevival)
+	self.isBlockingRevival = isBlockingRevival or false
```
Alf21, line 934: same as above

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,263 @@ local function FindCorpse(ply)
...
+	net.Send(self)
+end
+
+---
+-- Sets the revival start time.
+-- @param[default=@{CurTime()}] number startTime The revival start time
+-- @internal
+-- @realm server
+function plymeta:SetRevivalStartTime(startTime)
+	self.revivalStartTime = startTime or CurTime()
```
Alf21, line 947: same as above

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -789,119 +795,263 @@ local function FindCorpse(ply)
...
+	net.Send(self)
+end
+
+---
+-- Sets the revival duration.
+-- @param[default=0.0] number duration The revival time
+-- @internal
+-- @realm server
+function plymeta:SetRevivalDuration(duration)
+	self.revivalDurarion = duration or 0.0
```
Alf21, line 960: same as above

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
```diff
@@ -410,9 +410,11 @@ end
 -- @note Note that this is called both when a @{Player} spawns and when a round starts
 -- @hook
 -- @realm server
--- @ref https://wiki.garrysmod.com/page/GM/PlayerLoadout
+-- @ref https://wiki.facepunch.com/gmod/GM:PlayerLoadout
 -- @local
-function GM:PlayerLoadout(ply)
+function GM:PlayerLoadout(ply, isRespawn)
+	print("loadout: " .. tostring(isRespawn))
```
Alf21, line 416: This print doesn't say much for the person that receives it

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -852,8 +852,8 @@ end
 -- Returns whether a @{Player} was revived after beeing confirmed this round
```
Alf21, line 852: Bee's are cute, but better use `being`

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -852,8 +852,8 @@ end
...
 -- @return boolean
 -- @realm shared
-function plymeta:Revived()
-	return not self:TTT2NETGetBool("body_found", false) and self:OnceFound()
+function plymeta:WasRevivedAndConfirmed()
+	return self:WasRevivedInRound() and self:IsTerror() and self:OnceFound() and self:RoleKnown()
```
Alf21, line 856: Same as before, the `self:IsTerror` doesn't match the function name, even a current spectator could be revived and confirmed

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
TimGoll: this is a debug print by me, will be removed

PR #530, opened 2020-04-30
Radar Rework

TimGoll:
This is a farily small radar rework. These things were changed:

- removed old TTT code
- fixed a rare bug where the radar color had the wrong color
- moved the important radar stuff from client to server (like seriously, why was this on the client in the first place?!)
- added the possibility to give radar points a custom color
- added a convenience function for the creation of radar points: `RADAR.CreateNewTarget(ply, pos, ent, color)`
- added the possibility to change the radar time by either setting `ROLE.radarTime` or calling `RADAR.SetRadarTime(ply, time)`

### DIFF
--- a/lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
+++ b/lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
@@ -211,7 +211,7 @@ function RADAR:Draw(client)
 			if tgt.color then
 				surface.SetDrawColor(tgt.color.r, tgt.color.g, tgt.color.b, alpha)
 				surface.SetTextColor(tgt.color.r, tgt.color.g, tgt.color.b, alpha)
-			elseif c then
+			elseif tgt.hasSubrole and c then
 				surface.SetDrawColor(c.r, c.g, c.b, alpha)
 				surface.SetTextColor(c.r, c.g, c.b, alpha)
 			else
@@ -262,14 +262,19 @@ local function ReceiveRadarScan()
 			net.ReadInt(32)
 		)
 
-		local subrole = net.ReadInt(2 * ROLE_BITS)
+		local hasSubrole = net.ReadBool()
+		local subrole
+
+		if hasSubrole then
+			subrole = net.ReadUInt(ROLE_BITS)
+		end
 
 		local color
 		if net.ReadBool() then
 			color = net.ReadColor()
 		end
 
-		RADAR.targets[#RADAR.targets + 1] = {pos = pos, subrole = subrole, color = color}
+		RADAR.targets[#RADAR.targets + 1] = {pos = pos, subrole = subrole, hasSubrole = hasSubrole, color = color}
 	end
 
 	RADAR.enable = true
--- a/lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
+++ b/lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
@@ -9,7 +9,6 @@ local GetPTranslation
 local table = table
 local net = net
 local pairs = pairs
-local timer = timer
 local IsValid = IsValid
 
 local indicator = surface.GetTextureID("effects/select_ring")
@@ -18,11 +17,10 @@ local sample_scan = surface.GetTextureID("vgui/ttt/sample_scan")
 local det_beacon = surface.GetTextureID("vgui/ttt/det_beacon")
 local near_cursor_dist = 180
 
-RADAR = {}
+RADAR = RADAR or {}
 RADAR.targets = {}
 RADAR.enable = false
-RADAR.duration = 30
-RADAR.endtime = 0
+RADAR.startTime = 0
 RADAR.bombs = {}
 RADAR.bombs_count = 0
 RADAR.repeating = true
@@ -36,7 +34,7 @@ RADAR.called_corpses = {}
 -- @realm client
 function RADAR:EndScan()
 	self.enable = false
-	self.endtime = CurTime()
+	self.startTime = 0
 end
 
 ---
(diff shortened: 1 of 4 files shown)


lua/terrortown/entities/items/item_ttt_radar/init.lua
```diff
@@ -66,36 +46,139 @@ local function ttt_radar_scan(ply, cmd, args)
...
+end
+
+---
+-- Creates a new radar point
+-- @param Player ply The player that will see this radar point
+-- @param Vector pos The position of the radar point
+-- @param [opt]Entity ent The entity that is used for this radar point
+-- @param [opt]Color color A color for this radar point, this overwrites the normal color
+-- @realm server
+function RADAR.CreateNewTarget(ply, pos, ent, color)
```
lebroomer, line 123: This name is quite misunderstandable. I would expect a function, which is creating the target point on radar, not one which returns the processed radar data table

lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
```diff
@@ -271,32 +251,42 @@ end
...
 local function ReceiveRadarScan()
-	local num_targets = net.ReadUInt(8)
+	local num_targets = net.ReadUInt(16)
 
 	RADAR.targets = {}
 
 	for i = 1, num_targets do
-		local sr = net.ReadUInt(ROLE_BITS)
+		local pos = Vector(
+			net.ReadInt(32),
```
Alf21, line 260: Shouldn't `net.Write/ReadFloat` make more sense?

lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
```diff
@@ -271,32 +251,42 @@ end
...
+			net.ReadInt(32),
+			net.ReadInt(32)
+		)
 
-		local pos = Vector()
-		pos.x = net.ReadInt(32)
-		pos.y = net.ReadInt(32)
-		pos.z = net.ReadInt(32)
+		local hasSubrole = net.ReadBool()
```
Alf21, line 265: You could use a free role id too. Ref.: https://github.com/TTT-2/TTT2/blob/e131d95c08e7114e2d08515b2380999f0decc628/lua/includes/modules/roles.lua#L287

lua/terrortown/entities/items/item_ttt_radar/init.lua
```diff
@@ -66,36 +59,139 @@ local function ttt_radar_scan(ply, cmd, args)
...
+		subrole = subrole,
+		color = color
+	}
+end
+
+local function SetupRadarScan(ply)
+	timer.Create("radarTimeout_" .. ply:SteamID64(), ply.radarTime, 1, function()
+		if not IsValid(ply) or not ply:HasEquipmentItem("item_ttt_radar") then return end
+
+		if ply.radarDoesNotRepeat then return end
```
Alf21, line 150: This can be combined

lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
TimGoll: the positions are rounded. I don't know why, but it has been like this forever.

lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
TimGoll: I could do this. However for the readability sake I don't think it is nice to use a "random" number

Alf21: You are right, but regarding the upcoming change with `ROLE_NONE` or `ROLE_UNKNOWN` (...) this would help to remember :D

lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
Alf21: Yea I know that you didn't change it, but it doesn't make sense to me. And if they are rounded, a 32 bit int? o.O

lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
TimGoll: Yes, I added a boolean because someone said that the unknown role thing would be a secondary state, not a "fake role". I think it was saibotk. I'm not sure though.

lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
TimGoll: yes, they are indeed rounded. Isn't a float 32bit as well? Doesn't make sense to me.

Alf21: No I mean I can't imagine that a position of "2147483647" is available on a map :D That would break anything if it would be possible (because 2147483648 isn't far away). I like the solution with rounding somehow, but 32 bits yea...

lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
Alf21: Okay then feel free to resolve if it's clear :)

lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
TimGoll: image

source: https://www.reddit.com/r/gmod/comments/74eu8c/how_limited_is_the_map_size/

lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
TimGoll: I'll add a remark to this file in the trello ticket about the unknown role state.

PR #538, opened 2020-05-14
New strings and fixes

PinoMartirio:
Uploaded the old file without the VGUI rework and the new strings translated, with some fixes.


Histalek:
you need to remove the double dashes `--` from the lines you edited otherwise they will only be code comments and will not take effect :)

aside from that it looks good to me

PR #541, opened 2020-05-27
Adjust CLSCORE to fix clipping title

saibotk:
This implements the changes made by Kefta in https://github.com/Facepunch/garrysmod/pull/1677

The code is adjusted at some points to work with TTT2 and is not exactly the same.

Thank you @Kefta for letting us use this code and your contribution!

As a reference, this is the original commit message from Keftas commit:
- The CLSCORE panel now adjusts its width to the win text, fixing clipping with some languages (https://github.com/Facepunch/garrysmod/pull/1672). This required moving the win calculations to ``CLSCORE:ShowPanel`` from ``CLSCORE:BuildHilitePanel``
- ``CLSCORE:BuildHilitePanel(Panel base)`` -> ``CLSCORE:BuildHilitePanel(Panel base, WinTitle /*See definition at the bottom*/ title, Number starttime, Number endtime)``
- Changed Error calls to error as Error is no longer halting (https://github.com/Facepunch/garrysmod-issues/issues/2113)
- pairs loops of ``CLSCORE.Events`` are now numeric-for loops to make sure no stray string keys are considered
- Changed a usage of ``type`` to ``TypeID`` - using type is not safe for checking userdata types since the ``MetaName`` key of its metatable affects the output
- ``ValidAward`` now checks the types of its members instead of verifying they are simply non-nil
- ``CLSCORE:ShowPanel`` now deletes the last panel instead of creating a duplicate
- ``CLSCORE:ClearPanel`` now properly checks panel validity before trying to remove it
- Merged two loops over the event table in ``CLSCORE:Init``. Events with no traitor or detective tables no longer error from there
- Some hardcoded tables and colors are now members of CLSCORE so addons can modify them without overriding the whole system. These new members include:
-- Array<values = String> ``CLSCORE.ScorePanelNames``
-- Color ``CLSCORE.ScorePanelColor``
-- Table<keys = Number, values = WinTitle> ``CLSCORE.WinTypes``. This must have a string key ``Default`` pointing to a default WinTitle with all keys provided. WinTitle is defined as a Table<keys = String> {[String] ``Text`` /* Will be translated */, [Color] ``BoxColor``, [Color] ``TextColor``, [Color] ``BackgroundColor``}. All keys are optional and will fall back to the ``Default`` WinTitle table if not specified.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_scoring.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_scoring.lua
@@ -25,8 +25,25 @@ CLSCORE.Players = {}
 CLSCORE.EventDisplay = {}
 CLSCORE.StartTime = 0
 CLSCORE.Panel = nil
+CLSCORE.ScorePanelColor = Color(150, 50, 50)
+CLSCORE.ScorePanelNames = {
+	"",
+	"col_player",
+	"col_role",
+	"col_kills1",
+	"col_kills2",
+	"col_points",
+	"col_team",
+	"col_total"
+}
 
-surface.CreateFont("WinHuge", {font = "Trebuchet24", size = 72, weight = 1000, shadow = true})
+surface.CreateFont("WinHuge", {
+	font = "Trebuchet24",
+	size = 72,
+	weight = 1000,
+	shadow = true,
+	extended = true
+})
 
 -- so much text here I'm using shorter names than usual
 local T = LANG.GetTranslation
@@ -95,19 +112,19 @@ function CLSCORE.DeclareEventDisplay(event_id, event_fns)
 	-- basic input vetting, can't check returned value types because the
 	-- functions may be impure
 	if not tonumber(event_id) then
-		Error("Event ??? display: invalid event id\n")
+		error("Event ??? display: invalid event id")
 	end
 
 	if not event_fns or not istable(event_fns) then
-		Error(Format("Event %d display: no display functions found.\n", event_id))
+		error(string.format("Event %d display: no display functions found.", event_id))
 	end
 
 	if not event_fns.text then
-		Error(Format("Event %d display: no text display function found.\n", event_id))
+		error(string.format("Event %d display: no text display function found.", event_id))
 	end
 
 	if not event_fns.icon then
-		Error(Format("Event %d display: no icon and tooltip display function found.\n", event_id))
+		error(string.format("Event %d display: no icon and tooltip display function found.", event_id))
 	end
 
 	CLSCORE.EventDisplay[event_id] = event_fns
@@ -119,7 +136,10 @@ end
 -- @realm client
 -- @internal
 function CLSCORE:FillDList(dlst)
-	for _, e in pairs(self.Events) do
+	local events = self.Events
+
+	for i = 1, #events do
+		local e = events[i]
 		local etxt = self:TextForEvent(e)
 		local eicon, ttip = self:IconForEvent(e)
 		local etime = self:TimeForEvent(e)
@@ -187,9 +207,13 @@ function CLSCORE:BuildScorePanel(dpanel)
 	dlist:SetSortable(true)
 	dlist:SetMultiSelect(false)
 
-	local colnames = {"", "col_player", "col_roles", "col_teams", "col_kills1", "col_kills2", "col_points", "col_team", "col_total"}
+	local scorenames = self.ScorePanelNames
 
-	for _, name in pairs(colnames) do
+	for i = 1, #scorenames do
+		local name = scorenames[i]
+
+		if not isstring(name) then continue end
+
 		if name == "" then
 			-- skull icon column
 			local c = dlist:AddColumn("")
@@ -284,7 +308,7 @@ function CLSCORE:BuildScorePanel(dpanel)
 
 			if s.deaths > 0 then
 				surv = vgui.Create("ColoredBox", dlist)
-				surv:SetColor(Color(150, 50, 50))
+				surv:SetColor(self.ScorePanelColor)
 				surv:SetBorder(false)
 				surv:SetSize(18, 18)
 
@@ -310,7 +334,7 @@ function CLSCORE:BuildScorePanel(dpanel)
 			-- because images can't be sorted, so instead hack in a dummy value
 			local surv_col = l.Columns[1]
 			if surv_col then
-				surv_col.Value = type(surv_col.Value) == "Panel" and "1" or "0"
+				surv_col.Value = TypeID(surv_col.Value) == TYPE_PANEL and "1" or "0"
 			end
 		end
 	end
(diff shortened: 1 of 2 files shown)


gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -25,8 +25,25 @@ CLSCORE.Players = {}
 CLSCORE.EventDisplay = {}
 CLSCORE.StartTime = 0
 CLSCORE.Panel = nil
-
-surface.CreateFont("WinHuge", {font = "Trebuchet24", size = 72, weight = 1000, shadow = true})
+CLSCORE.ScorePanelColor = Color(150, 50, 50)
+CLSCORE.ScorePanelNames = {
```
TimGoll, line 29: I'm aware that all variables in this file seem to be names this way, but shouödn't they all start with a lowercase letter? Or do they have to stay this way for compatibility reasons?

saibotk: They are exposed, so people can change them. I just used the same names, as the PR does, for compatibility, but we could change them, if wanted, as we are already breaking many other things and this is not a big compatibility decrease.

TimGoll: I say keep it that way. I'll rework the panel at some point, so we don't have to do this twice

gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -95,19 +112,19 @@ function CLSCORE.DeclareEventDisplay(event_id, event_fns)
...
 
 	if not event_fns.text then
-		Error(Format("Event %d display: no text display function found.\n", event_id))
+		error(string.format("Event %d display: no text display function found.", event_id))
 	end
 
 	if not event_fns.icon then
-		Error(Format("Event %d display: no icon and tooltip display function found.\n", event_id))
+		error(string.format("Event %d display: no icon and tooltip display function found.", event_id))
 	end
```
Histalek, line 128: Couldn't all of these be asserts?

gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -500,15 +499,77 @@ function CLSCORE:BuildHilitePanel(dpanel)
...
+		end
+	end
+
+	local title = self:BuildTitle(wintype)
+
+	-- size the panel based on the win text w/ 88px horizontal padding and 44px veritcal padding
+	surface.SetFont("WinHuge")
+	local w, h = surface.GetTextSize(T(title.Text))
+
+	MsgN("Calculated width of text = " .. w)
```
Alf21, line 567: Debug isn't needed if this PR is ready

gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -500,15 +499,77 @@ function CLSCORE:BuildHilitePanel(dpanel)
...
+
+	-- w + padding (100) + DPropertySheet padding (8) + winlbl padding (30) + offset margin (margin * 2) + size margin (margin)
+	w, h = math.max(700, w + 138 + margin * 3), 500
```
Alf21, line 570: `700` is the max size. Just two ideas, we may discuss, even for the later rework: Shouldn't this be a dynamic value to support responsive design (`500` in height as well)? And shouldn't there be a minimum value as well?

gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -661,27 +725,34 @@ end
...
 	-- Get start time and traitors
-	local starttime
-	local tms
+	local starttime = 0
+	local tms = nil
+	local scores = {}
+	local nicks = {}
+
+	local game, selected, spawn = false, false, false
+	for i = 1, #events do
```
Alf21, line 734: Missing empty line in front

gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -500,15 +499,77 @@ function CLSCORE:BuildHilitePanel(dpanel)
...
+-- Converts the wintype into a title object to be used with @{CLSCORE:BuildHilitePanel}.
+--
+-- @internal
+-- @realm client
+-- @param string|number wintype
+-- @return table The title object table
+function CLSCORE:BuildTitle(wintype)
+	-- wintype can be both a WIN_ var or the team name directly, so we need to convert it to a teamname
+	local winnerTeam = TEAM_INNOCENT
+	if wintype == WIN_TIMELIMIT or wintype == WIN_INNOCENT then
```
Alf21, line 512: Missing empty line in front

gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -500,15 +499,77 @@ function CLSCORE:BuildHilitePanel(dpanel)
...
+	elseif wintype == WIN_TRAITOR then
+		winnerTeam = TEAM_TRAITOR
+	elseif wintype == WIN_NONE then
+		-- Special case, that should not be reachable, as WIN_NONE usually prevents a round from ending
+		-- If it should, we will show a tie.
+		return {
+			Text = "hilite_win_bees",
+			BoxColor = TEAMS[TEAM_INNOCENT].color,
+			TextColor = COLOR_WHITE,
+			BackgroundColor = Color(50, 50, 50, 255)
```
Alf21, line 523: Pls localize the Color

gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -500,15 +499,77 @@ function CLSCORE:BuildHilitePanel(dpanel)
...
+		}
+	elseif isstring(wintype) then
+		winnerTeam = wintype
+	end
+
+	return {
+		Text = "hilite_win_" .. winnerTeam,
+		BoxColor = TEAMS[winnerTeam].color,
+		TextColor = COLOR_WHITE,
+		BackgroundColor = Color(50, 50, 50, 255)
```
Alf21, line 533: Pls localize the Color

gamemodes/terrortown/gamemode/client/cl_scoring.lua
Kefta: I'm unfamiliar with TTT2's styling, but I avoided ``assert`` here so that string.format is only called in error cases. It saves a function call and string creation when there is no error case.

gamemodes/terrortown/gamemode/client/cl_scoring.lua
saibotk: math.max will just choose the highest number and does not limit the width to 700, instead it could rather be understood as a minimum.

Kefta:
I made one more change to the original PR: https://github.com/Facepunch/garrysmod/pull/1677/commits/4211b681ee0fd39f7253833767d0ac9cf45bc5ad

saibotk:
@Kefta Thanks for letting us know :+1:

Alf21:
@Kefta bc you have a lot of knowledge in this stuff, a question about micro-optimizations.
Is the first check here really needed? https://github.com/Facepunch/garrysmod/commit/4211b681ee0fd39f7253833767d0ac9cf45bc5ad#diff-6584fbd605c43889286045f89e710fc5R69
(`(not event_fns)`). 
Most of all it should be a valid table and if not, `istable(nil)` should return `false`. It's most of all a general question, bc maybe there are things I forgot.

Kefta:
Nah it's not necessary.

Issue #542, opened 2020-05-27
Double opening


dok441:
As you can see in the screenshot, if you open one door going down the stairs, then 2 doors open immediately. And so almost everywhere where there are doors that are located near each other open at the same time.
![Безымянный](https://user-images.githubusercontent.com/7480655/83014099-c6a31780-a026-11ea-9aff-471c3191ada4.png)

TimGoll:
There is a convar in ulx under the tab "TTT2 Doors" where you can disable that behaviour. Doors are grouped by proximity. However only doors with the same rotation should be grouped. Was one of the doors opened by default? Which version are you using?

TimGoll:
You could also give me the mapname so I can investigate this by myself. Maybe I can fix this.

dok441:
ttt_67thway_v7_j_l

TimGoll:
okay, IÄll take a look at it

ThePlatinumGhost:
I can confirm this bug happen also "ttt_rooftops_2016_v1"
https://steamcommunity.com/sharedfiles/filedetails/?id=534491717
Near the elevator

TimGoll:
I'm already at it and I think I have it. Give me half an hour.

TimGoll:
Fixed it, PR is ready to be reviewed

PR #543, opened 2020-05-27
fixed door grouping

TimGoll:
fixed double door grouping (#542). Previously I falsely simplified it to only group doors that don't have the same default rotation. Now only doors with inverted default rotation are grouped.

### DIFF
--- a/lua/ttt2/libraries/door.lua
+++ b/lua/ttt2/libraries/door.lua
@@ -54,7 +54,7 @@ local valid_doors = {
 local function GetClosedAngle(ent)
 	local data = ent:GetInternalVariable("m_angRotationClosed")
 
-	if not data then return end
+	if not data or #data < 3 then return end
 
 	return Angle(data[1], data[2], data[3])
 end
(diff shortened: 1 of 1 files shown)


lua/ttt2/libraries/door.lua
```diff
@@ -54,7 +54,7 @@ local valid_doors = {
 local function GetClosedAngle(ent)
 	local data = ent:GetInternalVariable("m_angRotationClosed")
 
-	if not data then return end
+	if not data or #data < 3 then return end
```
Alf21, line 57: To keep it save, pls add `istable()` around the first data check :)

TimGoll: I have to check if it returns true, because it isn't a table but userdata

Alf21: oh okay. Then it should be fine. Thought bc indexing e.g. a number would lead to an error

TimGoll: Tested it just now and no, userdata isn't counted as a table

PR #544, opened 2020-05-27
fixed radar on client

TimGoll:
fixed radar colors for undefined teams

### DIFF
--- a/lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
+++ b/lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
@@ -206,12 +206,12 @@ function RADAR:Draw(client)
 			subrole = tgt.subrole or ROLE_INNOCENT
 
 			local roleData = roles.GetByIndex(subrole)
-			local c = roleData.radarColor or TEAMS[tgt.team].color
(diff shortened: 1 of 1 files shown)


lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
```diff
@@ -206,12 +206,12 @@ function RADAR:Draw(client)
...
 
 			local roleData = roles.GetByIndex(subrole)
-			local c = roleData.radarColor or TEAMS[tgt.team].color
+			local c = roleData.radarColor or (TEAMS[tgt.team] and TEAMS[tgt.team].color or nil)
 
 			if tgt.color then
 				surface.SetDrawColor(tgt.color.r, tgt.color.g, tgt.color.b, alpha)
 				surface.SetTextColor(tgt.color.r, tgt.color.g, tgt.color.b, alpha)
-			elseif tgt.hasSubrole and c then
+			elseif IsColor(c) then
```
Alf21, line 214: This check isn't needed, you can combine it with the else below if you verify that c is a color everytime. That should be the case for teams color. If there is no team, then use a default color instead of this `or nil`.

TimGoll: This is exactly what I'm doing. If there is no color (because there is no team) the default color will be used

PR #549, opened 2020-05-30
Adjust CLSCORE to use the new stream net functions

saibotk:
This PR roughly adopts https://github.com/Facepunch/garrysmod/pull/1678
but uses the stream functions of TTT2 instead.

It moves the old custom implemented event stream to our new net.SendStream and net.ReceiveStream methods and removes now unused code / comments.
This also adjusts the events to store their exact date and other values, according to the PR linked above, as they can also now be sent without any problems (even though we are not relying on the Json function and are using pON instead).

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_scoring.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_scoring.lua
@@ -788,38 +788,15 @@ function CLSCORE:Toggle()
 	end
 end
 
-local buff = ""
-
-local function ReceiveReportStream(len)
-	local cont = net.ReadBit() == 1
-
-	buff = buff .. net.ReadString()
-
-	if cont then
-		return
+local function ReceiveReportStream(events)
+	if istable(events) then
+		table.SortByMember(events, "t", true)
+		CLSCORE:ReportEvents(events)
 	else
-		-- do stuff with buffer contents
-
-		local json_events = buff -- util.Decompress(buff)
-
-		if not json_events then
-			ErrorNoHalt("Round report decompression failed!\n")
-		else
-			-- convert the json string back to a table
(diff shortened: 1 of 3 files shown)


gamemodes/terrortown/gamemode/server/sv_scoring.lua
```diff
@@ -288,7 +276,9 @@ function SCORE:ApplyEventLogScores(wintype)
 	end
 
 	-- count deaths
-	for _, e in pairs(self.Events) do
+	local events = self.Events
+	for i = 1, #events do
```
Alf21, line 280: code style

gamemodes/terrortown/gamemode/server/sv_scoring.lua
```diff
@@ -328,64 +319,9 @@ function SCORE:Reset()
...
-		ErrorNoHalt("Round report event encoding failed!\n")
-
-		return false
-	else
-		return result
-	end
-end
-
 ---
 -- Steams the events list to all available @{Player}s
```
Histalek, line 323:
```suggestion
-- Streams the events list to all available @{Player}s
```

gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -788,38 +788,15 @@ function CLSCORE:Toggle()
...
-local function ReceiveReportStream(len)
-	local cont = net.ReadBit() == 1
-
-	buff = buff .. net.ReadString()
-
-	if cont then
-		return
+local function ReceiveReportStream(events)
+	if istable(events) then
+		table.SortByMember(events, "t", true)
```
TimGoll, line 793: shouldn't there be a new line?

PR #556, opened 2020-06-02
Fix max roles

saibotk:
The roleselection will now properly consider the max roles and max base roles convars.

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_roleselection.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_roleselection.lua
@@ -194,9 +194,6 @@ function roleselection.GetSelectableRolesList(maxPlys, rolesAmountList)
 		end
 	end
 
-	local curRoles = 2 -- amount of roles, start with 2 because INNOCENT and TRAITOR are all the time available
-	local curBaseroles = 2 -- amount of base roles, ...
-
 	-- we have to create a enumerable table to get random results easily
 	local availableBaseRolesTbl = {}
 	local availableSubRolesTbl = {}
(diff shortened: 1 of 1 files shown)


gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -194,16 +194,16 @@ function roleselection.GetSelectableRolesList(maxPlys, rolesAmountList)
...
-
 	-- we have to create a enumerable table to get random results easily
 	local availableBaseRolesTbl = {}
 	local availableSubRolesTbl = {}
 	local availableBaseRolesAmount = 0
 	local availableSubRolesAmount = 0
 
 	for roleData in pairs(rolesAmountList) do
+		-- exclude innocents and traitors, as they are already included, see below def. of selectableRoles.
+		if roleData == INNOCENT or roleData == TRAITOR then continue end
```
Alf21, line 205: With this change, `availableBaseRolesAmount`wouldn't increase. Same with `availableBaseRolesTbl`.

I understand that you adding them manually and that it should work in the same way, but in that way, it's hardcoded.

saibotk: This is needed, to avoid including them into the selectableRoles table. It doesn't matter if the availableXXXAmount is not increased, as they both are not used anywhere else and are just a local counter to insert the items into the respective table.
The roles have to be hardcoded, as they are also hardcoded in the selection code itself (eg. the part to select traitors and innocents first), otherwise, this code would break, as they (traitor and inno) would also have a chance to not be included in the selectableRoles table and therefore result in a nil comparison error. So i had to also ensure that both role's amount is always included.

PR #560, opened 2020-06-03
TargetID: Added hooks for modification

TimGoll:
Added two hooks to modify the targetID base data.

`GM:TTTModifyTargetedEntity(ent, distance)`: Modify the entity that is targeted. This is useful for addons like an "Identity Disguiser".

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_targetid.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_targetid.lua
@@ -243,14 +243,7 @@ function GM:HUDDrawTargetID()
 	endpos:Mul(MAX_TRACE_LENGTH)
 	endpos:Add(startpos)
 
-	local changedStartpos, changedEndpos = hook.Run("TTTModifyTargetTracedata", startpos, endpos)
-
-	if isvector(changedStartpos) and isvector(changedEndpos) then
-		startpos = changedStartpos
-		endpos = changedEndpos
-	end
-
-	local ent, distance
+	local ent, unchangedEnt, distance
 
 	-- if the user is looking at a traitor button, it should always be handled with priority
 	if TBHUD.focus_but and IsValid(TBHUD.focus_but.ent) and (TBHUD.focus_but.access or TBHUD.focus_but.admin) and TBHUD.focus_stick >= CurTime() then
@@ -279,7 +272,12 @@ function GM:HUDDrawTargetID()
 	-- only add onscreen infos when the entity isn't the local player
 	if ent == client then return end
 
-	ent = hook.Run("TTTModifyTargetedEntity", ent, distance) or ent
+	local changedEnt = hook.Run("TTTModifyTargetedEntity", ent, distance)
+
+	if IsValid(changedEnt) then
+		unchangedEnt = ent
+		ent = changedEnt
+	end
 
 	-- make sure it is a valid entity
 	if not IsValid(ent) or ent.NoTarget then return end
@@ -287,6 +285,7 @@ function GM:HUDDrawTargetID()
 	-- combine data into a table to read them inside a hook
 	local data = {
 		ent = ent,
+		unchangedEnt = unchangedEnt,
 		distance = distance
 	}
 
--- a/gamemodes/terrortown/gamemode/client/cl_targetid.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_targetid.lua
@@ -274,7 +274,7 @@ function GM:HUDDrawTargetID()
 
 	local changedEnt = hook.Run("TTTModifyTargetedEntity", ent, distance)
 
-	if IsValid(changedEnt) then
+	if changedEnt then
 		unchangedEnt = ent
 		ent = changedEnt
 	end
--- a/gamemodes/terrortown/gamemode/client/cl_targetid.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_targetid.lua
@@ -239,6 +239,12 @@ function GM:HUDDrawTargetID()
 
 	local startpos = client:EyePos()
 
+	local changedStartpos = hook.Run("TTTModifyTargetTracestart", startpos)
+
+	if isvector(changedStartpos) then
+		startpos = changedStartpos
+	end
+
 	local endpos = client:GetAimVector()
 	endpos:Mul(MAX_TRACE_LENGTH)
 	endpos:Add(startpos)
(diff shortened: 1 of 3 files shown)


gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -238,11 +238,18 @@ function GM:HUDDrawTargetID()
 	end
 
 	local startpos = client:EyePos()
-
 	local endpos = client:GetAimVector()
+
 	endpos:Mul(MAX_TRACE_LENGTH)
 	endpos:Add(startpos)
 
+	local changedStartpos, changedEndpos = hook.Run("TTTModifyTargetTracedata", startpos, endpos)
```
Alf21, line 246: `TTT2ModifyTargetTraceData`

For which scenario is this hook needed? Is it necessary to use startpos and endpos as arguments? Are they needed?

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -275,6 +282,12 @@ function GM:HUDDrawTargetID()
 	-- only add onscreen infos when the entity isn't the local player
 	if ent == client then return end
 
+	local changedEnt = hook.Run("TTTModifyTargetedEntity", ent, distance)
```
Alf21, line 285: Useful getting an entity here instead of the modified data? I know it's easier to implement, but just thinking about a way to modify other things later, that could lead to incompatibilities.

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -275,6 +282,12 @@ function GM:HUDDrawTargetID()
...
+
+	if IsValid(changedEnt) then
+		ent = changedEnt
```
Alf21, line 288: Better using
```
ent = changedEnt or ent

if not IsValid(ent) then return end
```

gamemodes/terrortown/gamemode/client/cl_targetid.lua
TimGoll: @lebroomer wanted a way to modify these for his supersheep

gamemodes/terrortown/gamemode/client/cl_targetid.lua
TimGoll: what do you mean by "modified data"?

Alf21: e.g. the text, etc. This current hook is just useful if you wanna switch the targeted player (which just makes sense for the disguiser, nothing else).

gamemodes/terrortown/gamemode/client/cl_targetid.lua
Alf21: Okay, and do he need these vars? It's totally okay if it's needed, then he doesn't need to recalculate. But if it's not needed, there should be as few arguments as possible

gamemodes/terrortown/gamemode/client/cl_targetid.lua
TimGoll: the ext etc has already its own hook, `TTTRenderEntityInfo`. Besides, the entity has to be changed prior to the creation of the text etc. At this point, only the entity is available.

gamemodes/terrortown/gamemode/client/cl_targetid.lua
TimGoll: I don't think he needs them. But why don't give the user the veraibles that are available? E.g. for a simple shift

Alf21: That argument won't count, imagine giving any data you have in TTTBeginRound hook etc :D

TimGoll: I thought these values are useful here. But I can remove them

gamemodes/terrortown/gamemode/client/cl_targetid.lua
Alf21: So is it really needed or couldn't this be solved with the already existing hooks?

TimGoll: The text can already be edited with existing hooks, the entity can't. Therefore a hook to edit the entity

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -271,19 +271,19 @@ function GM:HUDDrawTargetID()
...
-	if not IsValid(ent) or ent.NoTarget then return end
-
 	-- if a vehicle, we identify the driver instead
-	if IsValid(ent:GetNWEntity("ttt_driver", nil)) then
+	if IsValid(ent) and IsValid(ent:GetNWEntity("ttt_driver", nil)) then
 		ent = ent:GetNWEntity("ttt_driver", nil)
 	end
 
 	-- only add onscreen infos when the entity isn't the local player
 	if ent == client then return end
```
Alf21, line 280: How about checking if it isn't the spectated player as well?

TimGoll: The observer target is handles inside the drawcall itself:

https://github.com/TTT-2/TTT2/blob/f5539c32be05e16aa8249f3b0702373a357e4edd/gamemodes/terrortown/gamemode/client/cl_targetid.lua#L761

gamemodes/terrortown/gamemode/client/cl_targetid.lua
lebroomer: I don't think this hook is necessary. I really like the ```GM:TTTModifyTargetedEntity(ent, distance)``` hook. This can be used for addons like the supersheep too by simply calculating a traceline in the addon and inserting the information, which are calculated there.

TimGoll: OKay, then I'll remove this one completely since I don't need it as well :D

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -279,14 +272,20 @@ function GM:HUDDrawTargetID()
...
 
-	ent = hook.Run("TTTModifyTargetedEntity", ent, distance) or ent
+	local changedEnt = hook.Run("TTTModifyTargetedEntity", ent, distance)
+
+	if IsValid(changedEnt) then
```
Alf21, line 277: Just check `if changedEnt then`. It doesn't matter whether it's valid or not. If a script is going to change it, you shouldn't interrupt it.

PR #562, opened 2020-06-03
radar fix

TimGoll:
Simplified the role/team modification (used by spy for example). Additionally I added documentation.

### DIFF
--- a/lua/terrortown/entities/items/item_ttt_radar/init.lua
+++ b/lua/terrortown/entities/items/item_ttt_radar/init.lua
@@ -105,6 +105,17 @@ local function TriggerRadarScan(ply)
 end
 concommand.Add("ttt_radar_scan", TriggerRadarScan)
 
+---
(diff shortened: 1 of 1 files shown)


lua/terrortown/entities/items/item_ttt_radar/init.lua
```diff
@@ -105,6 +105,17 @@ local function TriggerRadarScan(ply)
 end
 concommand.Add("ttt_radar_scan", TriggerRadarScan)
 
+---
+-- This hook can be used to modify the radar dots of players.
+-- @param Player ply The player that receives the radar information
+-- @param Player target The Player whose info should be changed
+-- @return number, string The modified role, the modified team
```
Alf21, line 112: `number|string`

TimGoll: but it's not or, it's both

PR #564, opened 2020-06-04
Disguiser: Rework

TimGoll:
Cleaned up the disguiser code and moved the icon to the status system. This way it is only displayed while the player is disguised to remove the redudancy with the text displayed ob top of the HUD.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_keys.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_keys.lua
@@ -172,16 +172,3 @@ function GM:KeyRelease(ply, key)
 		UpdateInputSprint(ply, key, false)
 	end
 end
-
----
--- Called when a player releases a button.
--- @param Player ply @{Player} who released the button
--- @param number btn The button, see
--- <a href="https://wiki.garrysmod.com/page/Enums/BUTTON_CODE">BUTTON_CODE_Enums</a>
--- @hook
--- @realm client
--- @ref https://wiki.facepunch.com/gmod/GM:PlayerButtonUp
--- @local
-function GM:PlayerButtonUp(ply, btn)
-
-end
--- a/gamemodes/terrortown/gamemode/client/cl_keys.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_keys.lua
@@ -183,11 +183,5 @@ end
 -- @ref https://wiki.facepunch.com/gmod/GM:PlayerButtonUp
 -- @local
 function GM:PlayerButtonUp(ply, btn)
-	if not IsFirstTimePredicted() then return end
 
-	-- Would be nice to clean up this whole "all key handling in massive
-	-- functions" thing. oh well
-	if btn == KEY_PAD_ENTER then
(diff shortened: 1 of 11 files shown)


gamemodes/terrortown/gamemode/shared/lang/italian.lua
```diff
@@ -639,7 +639,7 @@ L.karma_min = "Irresponsabile"
 L.corpse = "Cadavere"
 L.corpse_hint = "Premi {usekey} per identificare. {walkkey} + {usekey} per identificare segretamente."
 
-L.target_disg = "(TRAVESTITO)"
+L.target_disg = "(travestito)"
```
Alf21, line 642: Those brackets shouldn't be a part of the translation

TimGoll: They were before and I though it was cheaper that adding two string concats each frame

Alf21: Hmm for sure, that's true, but anyways that's not a part of a translation by default...

TimGoll: I can add the brackets in the rendering hook if you like that more

gamemodes/terrortown/gamemode/client/cl_keys.lua
```diff
@@ -183,11 +183,5 @@ end
 -- @ref https://wiki.facepunch.com/gmod/GM:PlayerButtonUp
 -- @local
 function GM:PlayerButtonUp(ply, btn)
```
lebroomer, line 185: Why do we need the remaining Hook Stub here?

TimGoll: maybe for documentation? I dunno

lebroomer: But this is a default hook, we don't need it in our code - https://wiki.facepunch.com/gmod/GM:PlayerButtonUp

gamemodes/terrortown/gamemode/shared/lang/german.lua
```diff
@@ -1287,9 +1283,12 @@ L.hud_revival_time = "{time}s"
...
---2020-05-28
+-- 2020-05-28
 L.confirm_detective_only = "Nur Detektive können Leichen bestätigen"
 L.inspect_detective_only = "Nur Detektive können Leichen untersuchen"
 L.corpse_hint_no_inspect = "Nur ein Detektiv kann diesen Körper untersuchen."
 L.corpse_hint_inspect_only = "Drücke [{usekey}] zum Durchsuchen. Nur Detektive können diesen Körper bestätigen."
 L.corpse_hint_inspect_only_credits = "Drücke [{usekey}] zum Erhalten der Credits. Nur ein Detektiv kann diesen Körper untersuchen."
+
+-- 2020-06-04
+L.label_bind_disguiser = "Schalte Tarnung"
```
saibotk, line 1294: "Tarnung umschalten"

PR #565, opened 2020-06-04
Plymeta: Fixed revive fail function reference reset

TheNickSkater:


TheNickSkater:
> Please add an entry to the changelog :)

There you go

PR #568, opened 2020-06-05
AddonChecker: Reworked List and Cleanup

TimGoll:
I cleaned up the whole addon checker file. This cleanup can also be used in preparation for a future web based database. Additionally I added missing addons to this list. A new command `addonchecker` was also added to get the output at any time.

image


gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
```diff
@@ -1,167 +1,562 @@
...
-
-Addonchecker = {}
+-- @author Mineotopia
+
+ADDON_INCOMPATIBLE = 0
+ADDON_OUTDATED = 1
+
+addonChecker = addonChecker or {}
+
+addonChecker.curatedList = {
```
Alf21, line 11: Please make use of metatables

PR #571, opened 2020-06-07
Radar: added ForceScan

TimGoll:
Requested by @TheSuspect2306. This addds the possibility to force a radar rescan. This is especially useful if a player should have a custom radar time, because changing the radar time does not reset the current radar scan.

Example usage:
```lua
ply:GiveEquipmentItem("item_ttt_radar")
ply:SetRadarTime(5)
ply:ForceRadarScan()
```

While I was at it I also added a requested convar `ttt2_radar_charge_time`.


Alf21:
That looks really weird and should be moved to the player's metatable imo

TimGoll:
> That looks really weird and should be moved to the player's metatable imo

Why should it be moved to the player metatable and what's weird about it?

Alf21:
`RADAR.SetRadarTime(ply, 5)`. This function is focused on a Player object. So `ply:SetRadarTime(5)` or `ply:ForceRadarScan()` would make more sense.
Additionally, if RADAR isn't an instance, it should be names `radar`.

TimGoll:
The name radar was given by vanilla TTT and I think we should keep it for compatibility reasons.

And I don't know if I like moving it into the player metatable since this moves item code into the main TTT code.

Alf21:
<https://github.com/TTT-2/TTT2/blob/eda0b6688c10acfba373003dca0687734130d4bf/gamemodes/terrortown/gamemode/shared/sh_armor.lua#L14>

TimGoll:
Hm, you're right. The armor is baked in as well. If you say so, I can move it to the player table. Other opions on this? @saibotk @lebroomer

CHANGELOG.md
```diff
@@ -9,10 +9,13 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added Hooks to interact with door destruction
   - `GM:TTT2BlockDoorDestruction(doorEntity, activator)`: Hook to block the door destruction.
   - `GM:TTT2DoorDestroyed(doorPropEntity, activator)`: Hook that is called after the door is destroyed.
+- Added a new function to force a new radar scan: `ply:ForceRadarScan()`
+- Added a new convar to change the default radar time for players without custom radar times: `ttt2_radar_charge_time`
 
 ### Changed
 
 - Reworked the addonchecker and added a command to execute the checker at a later point
+- renamed `RADAR.SetRadarTime(ply, time)` to `ply:SetRadarTime(time)`
```
saibotk, line 18:
```suggestion
- Renamed `RADAR.SetRadarTime(ply, time)` to `ply:SetRadarTime(time)`
```

lua/terrortown/entities/items/item_ttt_radar/init.lua
```diff
@@ -212,3 +211,29 @@ net.Receive("TTT2RadarUpdateAutoScan", function(_, ply)
...
 	ply.radarDoesNotRepeat = not net.ReadBool()
 end)
+
+local plymeta = assert(FindMetaTable("Player"), "FAILED TO FIND PLAYER TABLE")
+
+---
+-- Sets the radar time interval, lets the current scan run out before it is changed.
+-- @param[default=ROLE.radarTime or 30] number time The radar time interval
+-- @realm server
+function plymeta:SetRadarTime(time)
```
saibotk, line 221: I think we should avoid default values on "Setter" functions, they are confusing and unnecessary.

TimGoll: but it would be way more complicated to handle the default value without this

saibotk: Where would that be then?

TimGoll: I would have to provide an additionaly function like `ResetRadarTime` (or something else) to make sure there is a way to reset the radar time. Would be ok too

lua/terrortown/entities/items/item_ttt_radar/init.lua
```diff
@@ -10,8 +10,6 @@ local hook = hook
 util.AddNetworkString("TTT2RadarUpdateAutoScan")
 util.AddNetworkString("TTT2RadarUpdateTime")
 
-local cv_radarTime = GetConVar("ttt2_radar_charge_time")
```
Alf21: I think we need an `ITEM:Initialize()` function to initialize a convar

PR #577, opened 2020-06-10
Revival: Added IsValid check to fix problem with reason reset on connect

TimGoll:


saibotk:
> Instead of making an `IsValid` check (client should be all the time valid if used correctly), removing this useless networking. Instead, setting the data for every client automatically (clientside). In that way, we removed a networking and a potential bug.

That is not easily possible and it can't just be removed, as it is a general round reset, which should happen from server side. Other than that, I do not think, the wiki pages you linked and their issues can be applied here, because if the player would not have been initialized (as stated in the LocalPlayer wiki), it wouldn't even return a player and return nil instead, so this can't be the reason as the error states, that it is in fact just missing the function itself on the object.

I don't really know why the error does even come up, as it doesn't seem plausible, that he can receive / execute code, when there is not yet all code loaded (eg. client receives net message before our code has been loaded with the new function on the plymeta table).

Alf21:
Hmm, then I gonna accept this hotfix. It's done this way on many code positions, even if I don't like working around an issue instead of getting the right reason.
If a client isn't initialized, that doesn't mean that he needs to be `nil`. Uninitialized player "objects" (...) could still have a variable, but the functions and vars don't have to be *initialized*.

Theoretically, an own IsInitialized check could help, but that wouldn't make sense bc an IsValid check is way faster and easier. All in all, maybe a gmod issue. I'm okay with the `IsValid` check.

TimGoll:
@Alf21 I was really busy the last few days, therefore I did not answer to anything in this topic. However I thought about resetting the data only on the client, since the BeginRound hook is already synced (and probably only called once the player is ready, I suppose).

Alf21:
> 
> 
> @Alf21 I was really busy the last few days, therefore I did not answer to anything in this topic. However I thought about resetting the data only on the client, since the BeginRound hook is already synced (and probably only called once the player is ready, I suppose).

That's another issue, a general one. Regarding every networking we are doing: The syncing of data, that even spectators should have, but didn't receive because of late joining. The solution should be the own network data system.

TimGoll:
> That's another issue, a general one. Regarding every networking we are doing: The syncing of data, that even spectators should have, but didn't receive because of late joining. The solution should be the own network data system.

Yes! This is something @saibotk started with his new networking system that either he or someone else should implement. This would solve many problems!

Issue #581, opened 2020-06-16
Remove plymeta:Is[Role] except inno, traitor, detective


TheNickSkater:
I don't think TTT2 needs to register the Is[Role] function for every role that is on the Server.. For compatibility IsTraitor IsDetective and IsInnocent should stay of course.

But, just as example.. If a Role has the name of some function, it gets overwritten.. This happened for me in the SpecDM, the Ghost role creates the "IsGhost" function, tough SpecDM creates this function itself and needs it.

Pustekuchen98:
I got a question about this. The Hitman is using the same function to exclude roles from being selected as a target. If this function would be removed wouldn't that break the hitman for example? And if I may ask, how would you go about fixing this issue then (hitman not able to exclude some roles after removing the is[Role] function)?

TimGoll:
The Hitman code could be changed to the `ply:GetSubRole() == ROLE_NAME`

Pustekuchen98:
ah okay ^^

TheNickSkater:
> The Hitman code could be changed to the `ply:GetSubRole() == ROLE_NAME`

I would actually say it SHOULD be done like that tbh.

TimGoll:
> > The Hitman code could be changed to the `ply:GetSubRole() == ROLE_NAME`
> 
> I would actually say it SHOULD be done like that tbh.

Yes, it is the style that we're using. I wasn't aware of the other functions at all. Probably something that was created in the early times of TTT2.

TheNickSkater:
> > > The Hitman code could be changed to the `ply:GetSubRole() == ROLE_NAME`
> > 
> > 
> > I would actually say it SHOULD be done like that tbh.
> 
> Yes, it is the style that we're using. I wasn't aware of the other functions at all. Probably something that was created in the early times of TTT2.

Probably.. I dont even find the code where this is created xD

TheNickSkater:
https://github.com/TTT-2/TTT2/blob/master/lua/includes/modules/roles.lua#L76 everything from this line just would need to be deleted.. The default functions are already hardcoded and will still work fine.. If everyone agrees at some point, I'd ofc make a PR.. or somebody else does it idk.

saibotk:
[...]

PR #582, opened 2020-06-17
Plymeta: Remove "Is{ROLENAME}" functions

TheNickSkater:
As discussed here https://github.com/TTT-2/TTT2/issues/581, this removes the creation of the IsROLE functions of all roles. Ofc IsInnocent/Detective/Traitor are still there.


TimGoll:
> Even if it seems to make the existing functions regarding default ttt functions inconsistent, you are right. It's anyways recommended to use the explicit methods `GetSubRole` and `GetBaseRole`. But to underline it here: This change could break some add-ons. This should be noted explicitly!

Yes. But I don't think there are (m)any addons that could break. Afaik only the role addon "ghost"

TheNickSkater:
> Even if it seems to make the existing functions regarding default ttt functions inconsistent, you are right. It's anyways recommended to use the explicit methods `GetSubRole` and `GetBaseRole`. But to underline it here: This change could break some add-ons. This should be noted explicitly!

Yes I did also think about that, and I do use many TTT2 addons and so far none of them broke for me. Sure, there might be 1 or 2, but I'm sure they will be able to change it then.

Issue #507, opened 2020-04-14
Incorrect countdown in miniscoreboard before mapchange


Reispfannenfresser:
## Your version of TTT2 (mandatory)
- Workshop version from Steam (v0.6.4b)

## Describe the bug (mandatory)
When the next map should be loaded (either when enough rounds were played or the timelimit is over) the miniscoreboard shows a countdown indicating when the next round will start.
There will be no next round though as the next map will be loaded.

## To reproduce
- Play the game until enough rounds were played and wait for the automatic map change.
- The miniscoreboard will show a countdown.

## Expected behaviour
It should count down to when the map will change. It should also show nothing when the TTT2LoadNextMap hook returned true.

PR #585, opened 2020-06-20
ttt_round_end now gets reset when the map changes

Reispfannenfresser:
Fix for [Issue 507](https://github.com/TTT-2/TTT2/issues/507)


TimGoll:
Out of interest: which time is now shown in the timer area? always 0? or the same as in your addon?

Reispfannenfresser:
> Out of interest: which time is now shown in the timer area? always 0? or the same as in your addon?

It's 0.
I've tried to make it changeable by mapvote addons, but it caused some problems:
1. The timers are not synced
2. It should be possible to create a mapvote addon without a timer
3. I would have to change the mapvote api a little (add a return value to the TTT2LoadNextMap hook) This could cause some incompatibilities.

Whenever mapvote addons want to change this timer they can still change the ttt_end_round variable on their own.

PR #591, opened 2020-06-21
Cleanup: Move round state hooks & remove hook add

saibotk:
This removes the random reseeding every round and also moves the buytable resets to their appropriate functions.

Also this moves the hooks with their documentation to the correct sh_main file, as they are shared.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_main.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_main.lua
@@ -342,29 +342,6 @@ local function ttt_print_playercount()
 end
 concommand.Add("ttt_print_playercount", ttt_print_playercount)
 
----
--- A hook that is called when the preparation phase starts.
--- @hook
--- @realm client
-function GM:TTTPrepareRound()
-
-end
-
----
--- A hook that is called when the round begins.
--- @hook
--- @realm client
-function GM:TTTBeginRound()
-
-end
-
--- A hook that is called when the round ends.
--- @hook
--- @realm client
-function GM:TTTEndRound()
-
-end
-
 -- usermessages
 
 local function ReceiveRole()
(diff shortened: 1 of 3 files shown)


Alf21:
@JustHev

JustHev:
In general the PRNG should only be seeded once and it must be seeded before random values are used. It is already seeded server-side in sv_main (which seems okay to me, but I'm not as familiar with the codebase as you guys) and on the client side in cl_scoring, although cl_scoring is only called after the first round.

A followup for clientside randomness is definitely needed, but not as high as a priority as serverside randomness. I'd recommend to not completely remove this call right now but move it to cl_main instead. This ensures that the client is seeded at init time, then once cl_scoring is reworked to not rely on randomseed for correctness (although other addons could rely on all client side PRNG's to have the same state, which would be bad) clientside randomness should be of an acceptable quality.

saibotk:
I explicitly removed it, as it does not make sense to seed it again after it was already seeded when the game mode was initialized.
This might also have caused weird issues with our randomness, as in the original TTT, there is also no reseeding besides the initial seed in the server side init file.

@JustHev About the client side, i think we would not need to move any of that seeding, as for now it really is not important and last but not least with this change it should be similar to the base TTT functionality (randomness-wise).

JustHev:
Fair point. Looking through the GMod lua I see that [they are already seeding the RNG in their lua includes](https://github.com/Facepunch/garrysmod/blob/cf262a1342ccb79831046b98452317ba1a1f9e3b/garrysmod/lua/includes/util.lua#L5) so I wonder if seeding it on the TTT2 side is needed at all.

saibotk:
> The changelos is missing though :P

I don't see a need for a changelog, do you?

TimGoll:
> I don't see a need for a changelog, do you?

Hmm, you're right. I think this fits the already added point in the changelog. However I wanted to be the one remembering you for once. :D

saibotk:
> > I don't see a need for a changelog, do you?
> 
> Hmm, you're right. I think this fits the already added point in the changelog. However I wanted to be the one remembering you for once. :D

Didn't you see i already added the `skip/changelog` label?

TimGoll:
> Didn't you see i already added the `skip/changelog` label?

Nope, I missed this. Thanks!

PR #600, opened 2020-06-26
Crosshair: Dot

### LINKED ISSUE #225: Crosshair dot (without side lines)
Would be nice if you could add a option to just enable a dot as a crosshair, without showing any side lines.

Alf21:
Added the client ConVar `ttt_crosshair_lines` to add the possibility to disable the crosshair lines (just showing the dot). The outlines need to be handled with the help of already existing crosshair ConVars.

Resolves #225


TimGoll:
LGTiM

PR #603, opened 2020-06-29
RoleSync: Changed when roles are getting synced to clients

TheNickSkater:
Roles will now only get synced to clients if they are actually known, not when only the body is confirmed.

This fixes the following issue: 
If a player (for example a Traitor) is confirmed but gets respawned, he will be shown as Innocent as soon as a SendFullStateUpdate is triggered (which happens quite often during a round).

I also changed the EndRound reset to a PrepareRound reset because bodys can still get confirmed after a round has ended.


TimGoll:
> this probably will fix this [TTT-2/ttt2-role_jes#47](https://github.com/TTT-2/ttt2-role_jes/issues/47) , right ?
> i hope so, anyway a huge thank you for you job @TheNickSkater

I don't think so as this only changes the miniscoreboard

PR #605, opened 2020-07-01
Sprint: airborne players can no longer replenish stamina

Reispfannenfresser:
The current sprint system allows players to sprint continuously by jumping to regain stamina.
This PR aims to fix this "issue"

Players have to be on ground to be able to sprint and players can't replenish their stamina when they're airborne.


gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -119,6 +119,8 @@ function UpdateSprint()
 
 	for i = 1, #plys do
 		local ply = plys[i]
+		if not ply:OnGround() then continue end
```
saibotk, line 122: Please leave an empty line in front

PR #607, opened 2020-07-04
Overhead Icons: Changed detective overhead icon to show to innocents

JustHev:
Innocents are not considered special, so no role icons were shown to
innocents, although Detective Icons should be shown to them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Changed
 - Roles are now only getting synced to clients if the role is known, not just the body being confirmed
 - Airborne players can no longer replenish stamina
+- Detective overhead icon is now shown to innocents
 
 ## [v0.7.2b](https://github.com/TTT-2/TTT2/tree/v0.7.2b) (2020-06-26)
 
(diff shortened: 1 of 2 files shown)


TimGoll:
Did you test this if this breaks the overhead icons of other roles?

JustHev:
Yes, the Cupid and Survivalist are not shown to innocents. I tested that by changing the role of some bots mid-round

Alf21:
Additionally to that commit, there should be the fix to display confirmed and revived inno overheads to other innos as well.

helaan:
> Additionally to that commit, there should be the fix to display confirmed and revived inno overheads to other innos as well.
> 
> (e.g. by adding here https://github.com/TTT-2/TTT2/pull/607/files#diff-e75c75314a90594eb6c031ab58bb239eR175 `or ply:RoleKnown()`).

I don't think that is related and should be fixed at this level, special innocents like the survivalists already show revived players role in overhead icons, the only change this PR makes is that normal innocents will also see any overhead icons.

JustHev:
I don't think that is related and should be fixed at this level, special innocents like the survivalists already show revived players role in overhead icons, the only change this PR makes is that normal innocents will also see any overhead icons.

TimGoll:
I was against this PR at first, but @JustHev is right. We should merge this and rework some parts of the role overhead icons in general. But this is out of scope for this community based PR

PR #612, opened 2020-07-08
Radar: Make bomb.team optional to fix compatibility

saibotk:
This will ensure that TTT items, which do not know about the new team property, will still be displayed.


lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
```diff
@@ -152,7 +152,7 @@ function RADAR:Draw(client)
 		surface.SetDrawColor(255, 255, 255, 200)
 
 		for _, bomb in pairs(self.bombs) do
-			if bomb.team == client:GetTeam() then
+			if bomb.team ~= nil and bomb.team == client:GetTeam() then
```
TimGoll, line 155: I think the `~= nil` can be removed

saibotk: Could be removed, but this is more explicit

PR #613, opened 2020-07-08
HUDManager: Fix saving forcedHUD and defaultHUD values to DB

### LINKED ISSUE #610: Server hud settings are not saving
Version: GitHub, 0.7.2b

Description:
Default/forced hud settings are not saving. Reset every map change.

1. Change default/forced hud. (as superadmin)
2. Wait for map change.
3. Check setting again and you can see that "pure_skin" is default again.

Expected behaviour:
Change settings, check it after map change and it should stay 

No errors at all. I can see only this: 
```
[TTT2][HUDManager] Storing data in database...
[TTT2][HUDManager] Storing data in database...
[TTT2][HUDManager] Storing data in database...
[TTT2][HUDManager] Storing data in database...

```

saibotk:
This fixes saving the values to the database, where the path was just wrong.

This fixes #610 and was tested by me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Roles are now only getting synced to clients if the role is known, not just the body being confirmed
 - Airborne players can no longer replenish stamina
 
+### Fixed
+- Fix HUDManager not saving forcedHUD and defaultHUD values
+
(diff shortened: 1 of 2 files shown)


plsbx:
What will happen if player with pure_skin selected join on my server with pure_skin disabled?

saibotk:
> What will happen if player with pure_skin selected join on my server with pure_skin disabled?

They will receive the forcedHUD if set, and otherwise they are given the defaultHUD, as it always was, this just fixes the saving issues.

PR #614, opened 2020-07-08
Corpses: Fix multiple corpses

### LINKED ISSUE #611: Calling kill on a dead player causes another corpse to spawn

- Workshop version from Steam (v0.7.2b)

Calling [kill](https://wiki.facepunch.com/gmod/Player:Kill) on a dead player causes another corpse to spawn.

## To reproduce

1. Kill a player with an explosion for example
2. Right after it, kill the player with the [kill](https://wiki.facepunch.com/gmod/Player:Kill) function

## Expected behaviour

Only one corpse spawns with explosion as the kill reason.

## Additional context

It was working correctly in TTT. It may be caused by a different implementation of [GM:PlayerDeath](https://wiki.facepunch.com/gmod/GM:PlayerDeath).

saibotk:
This removes the timer in sv_player, that would delay handling the death to the next frame, as this leads to multiple corpses etc, when a player dies from damage and a `ply:Kill` command in the same frame.

Don't know why this timer even was introduced, maybe @Alf21 knows something about this?

Fixes #611

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Roles are now only getting synced to clients if the role is known, not just the body being confirmed
 - Airborne players can no longer replenish stamina
 
+### Fixed
+- Fixed death handling spawning multiple corpses when killed multiple times in the same frame
+
 ## [v0.7.2b](https://github.com/TTT-2/TTT2/tree/v0.7.2b) (2020-06-26)
 
 ### Added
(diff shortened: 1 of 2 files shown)


Alf21:
The timer was introduced because `victim:Extinguish()` was called too early, so that a player that burned continued being on fire even if switched into the spec team. Don't know if it's still a thing or got fixed by gmod.

saibotk:
> The timer was introduced because `victim:Extinguish()` was called too early, so that a player that burned continued being on fire even if switched into the spec team. Don't know if it's still a thing or got fixed by gmod.

I checked the TTT source code and they still do it like that, seems to be not that big of a problem and is unavoidable to remove the timer at least for all the other functionality, that is used to handle the death

TimGoll:
@Alf21 why not move `victim:Extinguish()` into the next frame? Wouldn't this solve both problems?

saibotk:
@TimGoll i think we should just keep it as it is done on TTT for now, if the issue comes up again, we can try to fix this with the ugly timer workaround

PR #616, opened 2020-07-12
Update sv_addonchecker.lua: Removed updated and now compatible Addons

mexikoedi:
The Golden Deagle V1, V2 and the Juggernaut Suit were updated by Zaratusa. They are now compatible with TTT2 and can be removed from the Addonchecker.

PR #617, opened 2020-07-12
Update CHANGELOG.md: Added new entry for #616

mexikoedi:
Added new entry for the changes at the addonchecker. Look PR #616


saibotk:
This should be done in the PR, where it belongs, not in another PR. Also i dont think it is needed to mention this change in the changelog. We could think about writing in the changelog, that we updated the Addonchecker, to include future changes.

PR #624, opened 2020-07-31
Fileloader: Introduced a new file loader to TTT2

TimGoll:
This new file loader is heavily based on the old language file loader but is now more widely used.

CHANGED:
Language files are moved from `lua/lang/` to `lua/terrortown/lang/`. However the old path is loaded as well for now with a deprecation warning.

ADDED:
Added a new fileloader for addon files that were previously placed inside of `lua/autorun/` and were loaded in all gamemodes. This caused problems for players that switched between gamemodes. Additionally files inside those new folders are loaded directly after all TTT2 gamemode files are loaded. This has the added benefit that right from the time the files are loaded, all gamemode functions and library extensions are available.
The new path for files is `lua/terrortown/autorun` with three seperate subfolders for `client`, `server` and `shared`.

With this change, all TTT2 specifix files are now located in `lua/terrortown` (`/lang`, `/autorun`, `/entities`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Added a new custom file loader that loads lua files from `lua/terrortown/autorun/`
   - it basically works the same as the native file loader
-  - there are three subfolder: `client`, `server` and `shared`
+  - there are three subfolders: `client`, `server` and `shared`
   - the files inside this folder are loaded after all TTT2 gamemode files and library extensions are loaded
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,22 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Added
+
+- Added a new custom file loader that loads lua files from `lua/terrortown/autorun/`
+  - it basically works the same as the native file loader
+  - there are three subfolder: `client`, `server` and `shared`
+  - the files inside this folder are loaded after all TTT2 gamemode files and library extensions are loaded
+
 ### Changed
+
 - Roles are now only getting synced to clients if the role is known, not just the body being confirmed
 - Airborne players can no longer replenish stamina
 - Detective overhead icon is now shown to innocents
+- moved language files from `lua/lang/` to `lua/terrortown/lang`
 
 ### Fixed
+
 - Fixed death handling spawning multiple corpses when killed multiple times in the same frame
 - Radar now shows bombs again, that do not have the team property set
 - Fix HUDManager not saving forcedHUD and defaultHUD values
--- a/gamemodes/terrortown/gamemode/client/cl_main.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_main.lua
@@ -122,7 +122,7 @@ function GM:Initialize()
 
 	-- load addon language files
 	fileloader.LoadFolder("lang/", true, CLIENT, function(path)
-		ErrorNoHalt("[DEPRECATION WARNING]: Loaded language file from 'lang/', this folder is deprecated. Please switch to `terrortown/lang/`\n")
+		MsgN("[DEPRECATION WARNING]: Loaded language file from 'lang/', this folder is deprecated. Please switch to 'terrortown/lang/'")
 		MsgN("Added TTT2 language file: ", path)
 	end)
 
--- a/gamemodes/terrortown/gamemode/client/cl_main.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_main.lua
@@ -81,11 +81,11 @@ ttt_include("cl_damage_indicator")
 ttt_include("sh_armor")
 ttt_include("cl_weapon_pickup")
 
-fileloader.LoadFolder("terrortown/autorun/client/", false, CLIENT, function(path)
+fileloader.LoadFolder("terrortown/autorun/client/", false, CLIENT_FILE, function(path)
 	MsgN("Added TTT2 client autorun file: ", path)
 end)
 
-fileloader.LoadFolder("terrortown/autorun/shared/", false, SHARED, function(path)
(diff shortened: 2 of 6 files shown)


TimGoll:
I converted my three roles to use this system and I tested it. Works without any problems for me, even hot reloading is no problem.

lua/ttt2/libraries/fileloader.lua
```diff
@@ -0,0 +1,76 @@
...
+
+fileloader = fileloader or {}
+
+---
+-- Sets up files by scanning through directories and including them into the runtime.
+-- @note Has to be run on both server and client for client and shared files.
+-- @param string path The path to search in
+-- @param[default=false] boolean deepsearch If true, subfolders are scanned
+-- @param[default=SHARED] number realm The realm where the file should be included
+-- @param[opt] funcion callback A function that is called after the file is included
```
Histalek, line 20:
```suggestion
-- @param[opt] function callback A function that is called after the file is included
```

lua/ttt2/libraries/fileloader.lua
```diff
@@ -0,0 +1,76 @@
...
+local fileFind = file.Find
+local stringRight = string.Right
+
+fileloader = fileloader or {}
+
+---
+-- Sets up files by scanning through directories and including them into the runtime.
+-- @note Has to be run on both server and client for client and shared files.
+-- @param string path The path to search in
+-- @param[default=false] boolean deepsearch If true, subfolders are scanned
```
Histalek, line 18: This means it is not fully recursive but only one layer deep right? we should state that explicitly imo

lua/ttt2/libraries/fileloader.lua
```diff
@@ -0,0 +1,76 @@
...
+
+local fileFind = file.Find
+local stringRight = string.Right
+
+fileloader = fileloader or {}
+
+---
+-- Sets up files by scanning through directories and including them into the runtime.
+-- @note Has to be run on both server and client for client and shared files.
+-- @param string path The path to search in
```
Histalek, line 17: I'd personally like to be a bit more explicit here about the form of the path e.g.:
```suggestion
-- @param string path The absolute path to search in, has to end with `/`
```

CHANGELOG.md
```diff
@@ -4,12 +4,22 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Added
+
+- Added a new custom file loader that loads lua files from `lua/terrortown/autorun/`
+  - it basically works the same as the native file loader
+  - there are three subfolder: `client`, `server` and `shared`
```
```suggestion
  - there are three subfolders: `client`, `server` and `shared`
```

lua/ttt2/libraries/fileloader.lua
TimGoll: yes, it is. I'll update the description

gamemodes/terrortown/gamemode/client/cl_main.lua
```diff
@@ -113,7 +121,14 @@ function GM:Initialize()
 	self.roundCount = 0
 
 	-- load addon language files
-	LANG.SetupFiles("lang/", true)
+	fileloader.LoadFolder("lang/", true, CLIENT, function(path)
+		ErrorNoHalt("[DEPRECATION WARNING]: Loaded language file from 'lang/', this folder is deprecated. Please switch to `terrortown/lang/`\n")
```
Alf21, line 125: I wouldn't make this an Error. This will popup for every user using an add-on that handles language files in the old way. Better making it a `MsgN` as well.

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -192,7 +204,14 @@ function GM:Initialize()
 	hook.Run("TTT2FinishedLoading")
 
 	-- check for language files to mark them as downloadable for clients
-	LANG.SetupFiles("lang/", true)
+	fileloader.LoadFolder("lang/", true, CLIENT, function(path)
+		ErrorNoHalt("[DEPRECATION WARNING]: Loaded language file from 'lang/', this folder is deprecated. Please switch to `terrortown/lang/`\n")
```
Alf21, line 208: Same as above

gamemodes/terrortown/gamemode/shared/sh_init.lua
```diff
@@ -10,6 +10,9 @@ GM.Customized = true
 
 TTT2 = true -- identifier for TTT2. Just use "if TTT2 then ... end"
 
+-- extend CLIENT and SERVER flags
+SHARED = 2
```
Alf21, line 14: Is this really needed? You could simply use no argument or `CLIENT or SERVER`. An existing convar like this (handled in another way than CLIENT and SERVER are done) could result in big confusion

TimGoll: I thought it was nicer in the code to be honest. But no, it is not really needed.

Alf21: Yea but doing `if SHARED then` will return true EVERYTIME, even if just done on the SERVER or CLIENT.

TimGoll: yes, but it wasn't intended for that usage. See my linked example. I somehow have to pass to the function that it is shared. This seemed to be the best solution to me.

Alf21: Then make it a boolean and just use hardcoded `true` or `CLIENT or SERVER` (true)

TimGoll: But I need to differentiate between three states, not just two:
- include on the server
- include on the server and distribute to the client
- only distribute to the client

Alf21: local var or no var (nil == shared)

TimGoll: To be honest, this is pretty janky.

PR #627, opened 2020-08-07
Spectators: Stop teleporting players to players they're not spectating

Alf21:
If you crouch to stop spectating a player and want to slowly move while free-roaming and hold duck, you are teleported back to the person you were previously spectating. There is no need for this since if you want to return to the person you were spectating, you can just use your reload bind.

Change regarding https://github.com/Facepunch/garrysmod/pull/1696

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_player.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player.lua
@@ -379,7 +379,7 @@ function GM:KeyPress(ply, key)
 
 		local target = ply:GetObserverTarget()
 
-		if IsValid(target) and target:IsPlayer() then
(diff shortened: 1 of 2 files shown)


gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -379,7 +379,7 @@ function GM:KeyPress(ply, key)
 
 		local target = ply:GetObserverTarget()
 
-		if IsValid(target) and target:IsPlayer() then
+		if IsValid(target) and target:IsPlayer() and ply:GetObserverMode() ~= OBS_MODE_ROAMING then -- Only set the spectator's position to the player they are spectating if they are in chase or eye mode. They can use the reload key if they want to return to the person they're spectating
```
TimGoll, line 382: To be honest, I'd move the comment to a different position and break it into a few lines

PR #628, opened 2020-08-08
Shop: Doubleclick buying, Coroutine loading and small fixes

Alf21:
- Added double-click buying
- Moved equipment list generation into a coroutine
- Fixed small bugs


TimGoll:
Took a look at your new changes. Good catch finding those issues!

PR #629, opened 2020-08-08
Fetch: improvements in avatar fetching / some internal structure changes

Reispfannenfresser:
- The regex is fixed to now work with the new steam avatar host and maybe even future steam avatar hosts.

Because the steamID64 function returns nil for bots on clients i changed the variables from some netmessages to set the id to "" when sending them to the client. The client then sets them to nil.
- When searching a dead bot's corpse
- In the TTT2PlayerAuthedShared netmessage

Changes inside the library
- FetchAvatarAsset now returns the material
- FetchAvatarAsset returns nil when the material wasn't cached before
- renamed fetch_asset to FetchAsset
- FetchAsset now returns nil when the material wasn't cached before
- the default avatar and the bot avatar are now vtf files instead of being fetched them from weird sites
- Stuff shouldn't fail that easily. (There were weird edge cases that spaghettised themselves to work somehow)
- It will not cache the default avatar for players where the fetching failed. (It will try to fetch again on another call)

(There will probably be some new icons coming along with this PR, but these are not ready yet)


lua/autorun/client/b-draw_lib.lua
```diff
@@ -55,48 +57,42 @@ local function fetch_asset(url, fallback)
 	return mats[url]
 end
 
-local function fetchAvatarAsset(id64, size, onFetched)
-	id64 = id64 or "BOT"
-	size = size == "medium" and "medium" or size == "small" and "" or size == "large" and "full" or ""
+local function fetch_avatar_asset(id64, size, fallback)
```
TimGoll, line 60: rename to `FetchAvatarAsset`

lua/autorun/client/b-draw_lib.lua
```diff
@@ -55,48 +57,42 @@ local function fetch_asset(url, fallback)
...
+	if id64 then print(id64) else print("nil") end
```
TimGoll, line 61: don't make this one line

lua/autorun/client/b-draw_lib.lua
```diff
@@ -55,48 +57,42 @@ local function fetch_asset(url, fallback)
...
+	if not id64 then return _bot_avatar end
```
TimGoll, line 62: don't make this in a single line

gamemodes/terrortown/gamemode/client/cl_main.lua
```diff
@@ -729,16 +731,15 @@ net.Receive("TTT2PlayerAuthedShared", function(len)
...
 
+	-- checking for bots
+	if steamid64 == "" then
+		steamid64 = nil
+	end
+
 	hook.Run("TTT2PlayerAuthed", steamid64, name)
 end)
 
 hook.Add("TTT2PlayerAuthed", "TTT2CacheAvatar", function(steamid64, name)
```
saibotk, line 742: It would be cool, if you could also make this a named function and call it inside the function above instead of adding it to the hook :) (even if that does not have to be done in this PR)

gamemodes/terrortown/gamemode/client/cl_main.lua
```diff
@@ -729,16 +731,15 @@ net.Receive("TTT2PlayerAuthedShared", function(len)
...
-	local ply = player.GetBySteamID64(steamid64)
-
-	if not IsValid(ply) or ply:IsBot() then
-		steamid64 = nil
-	end
-
 	draw.CacheAvatar(steamid64, "medium") -- caching
```
saibotk, line 743: Also, is this still needed? Or rather is this whole function even needed?

gamemodes/terrortown/gamemode/client/cl_search.lua
```diff
@@ -673,12 +673,14 @@ local function TTT2ConfirmMsg()
...
-
-	if sid64 ~= "" then
-		img = draw.GetAvatarMaterial(sid64, "medium", img)
+	-- checking for bots
+	if sid64 == "" then
+		sid64 = nil
 	end
 
+	local img = Material("vgui/ttt/icon_corpse")
+	img = draw.GetAvatarMaterial(sid64, "medium", img)
```
saibotk, line 682: I think fallbacks should be handled in place rather than invisible inside the function, we might want to think about changing that.

lua/autorun/client/b-draw_lib.lua
```diff
@@ -55,48 +57,43 @@ local function fetch_asset(url, fallback)
 	return mats[url]
 end
 
-local function fetchAvatarAsset(id64, size, onFetched)
-	id64 = id64 or "BOT"
-	size = size == "medium" and "medium" or size == "small" and "" or size == "large" and "full" or ""
-
-	if fetchedavatars[id64 .. " " .. size] then
-		return fetchedavatars[id64 .. " " .. size]
+local function FetchAvatarAsset(id64, size, fallback)
```
saibotk, line 60: I dislike that this accepts a fallback as a parameter instead of returning nil.

lua/autorun/client/b-draw_lib.lua
```diff
@@ -55,48 +57,43 @@ local function fetch_asset(url, fallback)
...
+		local link = body:match("<avatarIcon><%!%[CDATA%[(.*)%]%]><%/avatarIcon>")
 
-		fetchedavatars[id64 .. " " .. size] = link:Replace(".jpg", (size ~= "" and "_" .. size or "") .. ".jpg")
+		if not link then return end
 
-		if isfunction(onFetched) then
-			onFetched(fetchedavatars[id64 .. " " .. size])
-		end
+		fetched_avatar_urls[key] = link:Replace(".jpg", size .. ".jpg")
+		FetchAsset(fetched_avatar_urls[key])
```
saibotk, line 80: Why is this fetched here? (if legitimate, add a comment explaining it)

lua/autorun/client/b-draw_lib.lua
```diff
@@ -145,24 +142,38 @@ end
...
 -- @param number width
 -- @param number height
 -- @param Color color
--- @param Angle ang
--- @param boolean corner
-function draw.SteamAvatar(id64, size, x, y, width, height, color, ang, corner)
-	draw.WebImage(fetchAvatarAsset(id64, size), x, y, width, height, color, ang, corner)
+-- @param Angle angle
+-- @param boolean cornerorigin if it is set to <code>true</code>, the WebImage will be centered based on the x- and y-coordinate
+function draw.SteamAvatar(id64, size, x, y, width, height, color, angle, cornerorigin)
```
saibotk, line 154: This is close to the code of the `draw.WebImage` function and only differs in their material, maybe create a shared draw function and pass the material to be used via a parameter?

lua/autorun/client/b-draw_lib.lua
```diff
@@ -55,48 +57,43 @@ local function fetch_asset(url, fallback)
...
+		if not link then return end
 
-		if isfunction(onFetched) then
-			onFetched(fetchedavatars[id64 .. " " .. size])
-		end
+		fetched_avatar_urls[key] = link:Replace(".jpg", size .. ".jpg")
+		FetchAsset(fetched_avatar_urls[key])
 	end)
+
+	return FetchAsset(fetched_avatar_urls[key], fallback)
```
saibotk, line 83: If there is a fetch in progress, this will usually always return the fallback as the code above is asynchronous.

gamemodes/terrortown/gamemode/client/cl_main.lua
Reispfannenfresser: All avatars should be fetched before they are first used.

When a client joins it caches the avatars of all existing players.
To this client any new joining players would be avatarless though and that's why these are cached.

I should probably also add 
```
draw.CacheAvatar(steamid64, "small")
draw.CacheAvatar(steamid64, "large")
```

gamemodes/terrortown/gamemode/client/cl_main.lua
Reispfannenfresser: It can be useful globally, but where should i put it?

TimGoll: I'm not that deep in this problem, but if it is something that can be put inside of `xxx.yyy`, then put it in `lua/ttt2/libraries`

lua/autorun/client/b-draw_lib.lua
Reispfannenfresser: I'm fetching them here because they wouldn't be cached elsewise.
Since fetching is asynchronous i want to have them fetched before they are first used.

(I copied this behaviour from the old function)

lua/autorun/client/b-draw_lib.lua
Reispfannenfresser: I'll change it to return nil then.

gamemodes/terrortown/gamemode/client/cl_main.lua
Reispfannenfresser: To summarize the problem:
https://wiki.facepunch.com/gmod/Player:SteamID64
→ SteamID64 returns different values on client and server for bots.
→ When the SteamID is sent via a netmessage the client gets confused easily.

It would either need 2 functions:
- A function for the client to set the received steamid to nil, when it's X.
- A function for the server to set a steamid to X when it's a bots before sending.

or one function:
- A function for the client which detects if it's a bots id and sets it to nil

TimGoll:
Please add the icons to the forcedownload file as well

PR #635, opened 2020-08-10
Language: Add a Japanese translation.

westooooo:
I made a Japanese translation. Sorry, I'm late...


saibotk:
No problem :D 

And a huge thank you for your work from me and the whole team :heart: 

PS: Could you add an entry to the CHANGELOG.md or should we do that for you?

gamemodes/terrortown/gamemode/shared/lang/japanese.lua
```diff
@@ -0,0 +1,1274 @@
+---- Japanese language strings
+
+local L = LANG.CreateLanguage("japanese")
```
TimGoll, line 3: all the other languages defined their language with their name. So for example "deutsch" for "german". Maybe you want to replace this as well?

PR #637, opened 2020-08-11
CI: Move glualint workflow to github action

saibotk:
This adds a CI workflow using Github Actions for the glualint tool.
The Jenkins build will then no longer be needed and all checks / output is available right here on Github, while also reducing the needed maintenance.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Download & extract glualint
       run: |
         wget -c https://github.com/FPtje/GLuaFixer/releases/download/${GLUALINT_VERSION}/glualint-${GLUALINT_VERSION}-linux.zip -O glualint.zip
-        unzip -u glaulint.zip
+        unzip -u glualint.zip
 
     - name: Check code with glualint
       run: ./glualint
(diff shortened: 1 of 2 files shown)


gamemodes/terrortown/gamemode/client/vgui/cl_sb_row.lua
```diff
@@ -20,7 +20,6 @@ local ipairs = ipairs
 local IsValid = IsValid
 local surface = surface
 local vgui = vgui
-
```
Histalek, line 23: i don't think this is an intentional change, is it?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
```
Histalek, line 17: is there any reason to not call this job `lint` or something similar to better reflect what the job does?

saibotk:
> I guess it is alright? :D
> 
> (Nice work as always with your infrastructure projects! Will this work for all TTT2 repos?)

It can be applied, if we want to. But this specific action is currently only active on this repo, as it is defined here. 
You can always just copy the workflow file to add it to another repository.

PR #641, opened 2020-08-21
Language: Spanish translation update

DennisWolfgang:
-Added a translation for the 2020-08-07 and 2020-08-11 text lines.

-Fixed a word missplacement in the death confirmation notification, showing the word "Era" twice (in english being like "They were were a Traitor!").

-Changed a duplicated translation for the detectives only inspect and confirm corpse game mechanic, being almost the same, so now you can tell the difference which one is enabled.

-Improved typing and grammar where needed for "better a sounding" translation.


saibotk:
Thank you very much :rocket:

PR #647, opened 2020-09-10
Scoreboard: Team indicator

Alf21:
grafik

Added a team indicator in front of every name in the scoreboard. I thought about a role indicator (instead of / as well), but came to this solution (a discussion is always welcome).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Added ConVar to toggle double-click buying
 - Added Japanese translation (by @Westoon)
+- Added a team indicator in front of every name in the scoreboard (just known teams will be displayed)
 
 ### Changed
 
--- a/gamemodes/terrortown/gamemode/client/vgui/cl_sb_row.lua
+++ b/gamemodes/terrortown/gamemode/client/vgui/cl_sb_row.lua
@@ -196,6 +196,11 @@ function PANEL:Init()
 	self.nick = vgui.Create("DLabel", self)
 	self.nick:SetMouseInputEnabled(false)
 
+	self.team2 = vgui.Create("DImage", self)
+	self.team2:SetSize(iconSizes, iconSizes)
+	self.team2:SetMouseInputEnabled(false)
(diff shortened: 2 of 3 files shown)


TimGoll (approved):
Looks good to me! However you might want to add a tooltip with "Team Traitor" etc to make it clear it isn't the role

Alf21:
> 
> 
> Looks good to me! However you might want to add a tooltip with "Team Traitor" etc to make it clear it isn't the role

https://github.com/TTT-2/TTT2/pull/647/files#diff-9da617f14da6d325fe596fb9f097a7beR399 :D

TimGoll:
Oh. I somehow missed this. But it doesn't show that it is a team icon. It just shows "traitor"
Might be confusing if a hitman is confirmed

TimGoll:
closed? o.O

Alf21:
> 
> 
> Oh. I somehow missed this. But it doesn't show that it is a team icon. It just shows "traitor"
> Might be confusing if a hitman is confirmed

Idk what you wanna say ^^. There are no special team icons. And it says for traitors "TEAM Traitors" as defined here: https://github.com/TTT-2/TTT2/blob/37541be47ac40c8828b78be95906f35d140b364a/gamemodes/terrortown/gamemode/shared/lang/english.lua#L16

TimGoll:
Ah, you use the team lang string, of course. Completely forgot this is a thing. Sorry! 
And there are special team icons. Each team has their own icon and you even use it. It is always the main role icon, yes. But it's still a team icon.

But that doesn't matter since I was stupid and you did the right thing

Alf21:
Probably we should add a shadow to match the appearance of the nickname as well as the default icons?

TimGoll:
Btw, a changelog is missing

TimGoll:
> Probably we should add a shadow to match the appearance of the nickname as well as the default icons?

I thought about this as well. Sounds like a good idea. I did not add this as a request since the scoreboard will be reworked at some point anyway.

Alf21:
Added
grafik

Histalek (approved):
LGTM

saibotk (approved):

PR #648, opened 2020-09-13
Util functions: Added table.ExtractRandomEntry(tbl, filterFn)

Alf21:
Added a useful utility function and made some improvements to existing util-functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Added ConVar to toggle double-click buying
 - Added Japanese translation (by @Westoon)
+- Added `table.GetRandomEntry(tbl, filterFn)` function
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Added ConVar to toggle double-click buying
 - Added Japanese translation (by @Westoon)
-- Added `table.GetRandomEntry(tbl, filterFn)` function
+- Added `table.ExtractRandomEntry(tbl, filterFn)` function
 
 ### Changed
 
--- a/gamemodes/terrortown/gamemode/shared/sh_util.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_util.lua
@@ -145,18 +145,38 @@ end
 -- @note The returned entry will get removed from the given @{table}. If you wanna keep the original table untouched, create a copy for this function.
 -- @realm shared
 function table.GetRandomEntry(tbl, filterFn)
-	local entry = nil
-	local isInvalidFilterFn = not isfunction(filterFn)
+	local cTbl = #tbl
 
-	repeat
-		if #tbl <= 0 then return end
+	-- if no filterFn is defined, get a any random entry of the given @{table}
+	if not isfunction(filterFn) then
+		local index = rand(cTbl)
+		local entry = tbl[index]
 
-		local rnd = math.random(#tbl)
-		entry = tbl[rnd]
-		table.remove(tbl, rnd)
-	until isInvalidFilterFn or filterFn(entry)
+		table.remove(tbl, index)
 
-	return entry
+		return entry
+	end
+
+	local tmpTbl = {}
+
+	-- create a temporary table used for easy and fast access after shuffling
+	for i = 1, cTbl do
+		tmpTbl[i] = i
+	end
+
+	table.Shuffle(tmpTbl)
+
+	for i = 1, cTbl do
+		local index = tmpTbl[i]
+
+		if filterFn(tbl[index]) then
+			local entry = tbl[index]
+
+			table.remove(tbl, index)
+
+			return entry
+		end
+	end
 end
 
 ---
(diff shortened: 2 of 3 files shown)


gamemodes/terrortown/gamemode/shared/sh_util.lua
```diff
@@ -107,27 +119,44 @@ end
...
+	local entry = nil
+	local isInvalidFilterFn = not isfunction(filterFn)
+
+	repeat
+		if #tbl <= 0 then return end
+
+		local rnd = math.random(#tbl)
+		entry = tbl[rnd]
+		table.remove(tbl, rnd)
+	until isInvalidFilterFn or filterFn(entry)
```
Alf21, line 157:
```suggestion
	local finalTbl = tbl
	
	if isfunction(filterFn) then
		local tmp = {}

		for i = 1, #tbl do
			if filterFn(tbl[i]) then
				tmp[#tmp + 1] = tbl[i]
			end
		end
		
		finalTbl = tmp
	end

	return #finalTbl > 0 and finalTbl[math.random(#finalTbl)] or nil
```

Would be the alternative.

AaronMcKenney: `return #finalTbl > 0 and finalTbl[math.random(#finalTbl)] or nil`

I'm still new to Lua, but how would this line handle the case where finalTbl has only one entry, which is zero or false (i.e. {0} or {false})? Wouldn't the condition return nil in this case (as opposed to 0 or false)?

I think something like this would be more readable in any case:
```
if #finalTbl > 0 then
    return finalTbl[math.random(#finalTbl)]
else
    return nil
end
```

Alf21: `finalTbl[math.random(#finalTbl)]` is a shortened code for `finalTbl[math.random(1, #finalTbl)]`. So if `#finalTbl` is `1`, then `math.random(1, 1)` is `1` (in 100%) too. So you are right so far.
Let's continue: if `finalTbl[1]` is `0` or `false`, then the function will return `0` or `false` (as intended). The function has no error-state or false-return in order to detect any issue. It's a simple getter method. So if you insert a list of numbers or booleans, it's intended that `0` or `false` will possibly return.

Just if the function returns `nil`, nothing was found.

---

Regarding your last code suggestion: Simple things should be kept simple. The `or nil` just is an addition to make it more readable. In your suggested code, you can cut the whole `else`-clause, without worrying about any changed mechanic. Running `return nil` or `return` at the end of a function isn't necessary.

so

```lua
if #finalTbl > 0 then
    return finalTbl[math.random(#finalTbl)]
end
```

could be an alternative. If writing lua-code on github, keep in mind to enable code highlighting (` ```lua [code] ``` `)

gamemodes/terrortown/gamemode/shared/sh_util.lua
```diff
@@ -107,27 +119,44 @@ end
...
+	return alive[1]
+end
+
+---
+-- Returns a random entry of the given @{table}
+-- @param table tbl the @{table} that contains the data
+-- @param nil|function filterFn the @{function} that has to return true on the given entry
+-- @return any the entry that returned true on the given @{function}
+-- @note The given @{table} has to be iterable.
+-- @note The returned entry will get removed from the given @{table}. If you wanna keep the original table untouched, create a copy for this function.
```
Histalek, line 145: I would really like this to be a warning or even rename the function because a getter function indicates value retrieval (undistructive) and not value extraction (destructive) imo

Histalek (changes requested):
Changes to existing functions LGTM
The new function needs to be discussed i feel

gamemodes/terrortown/gamemode/shared/sh_util.lua
saibotk: @Alf21 I like the alternative more, but then I think we should make this a utility function for purely applying a filter on a table instead. So it can be reused in a more general use case and would separate the concerns better. (Your shortcut, on checking if a value is actually in a filter might also be realized, if you would want to keep this as a whole including the part of retrieving an item, by just checking the filter function result and early returning on true)

Histalek (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Added ConVar to toggle double-click buying
 - Added Japanese translation (by @Westoon)
+- Added `table.GetRandomEntry(tbl, filterFn)` function
```
saibotk, line 11:
```suggestion
- Added `table.ExtractRandomEntry(tbl, filterFn)` function
```

gamemodes/terrortown/gamemode/shared/sh_util.lua
```diff
@@ -107,27 +120,64 @@ end
...
+
+---
+-- Returns a random entry of the given @{table}
+-- @param table tbl the @{table} that contains the data
+-- @param nil|function filterFn the @{function} that has to return true on the given entry
+-- @return any the entry that returned true on the given @{function}
+-- @note The given @{table} has to be iterable.
+-- @warning The returned entry will get removed from the given @{table}. If you wanna keep the original table untouched, create a copy for this function.
+-- @realm shared
+function table.ExtractRandomEntry(tbl, filterFn)
```
saibotk, line 148: I think this function should be moved to `/lua/ttt2/extensions/table.lua`

saibotk (changes requested):

saibotk (approved):

PR #650, opened 2020-09-18
[WIP] ShopEditor: Changed how .minPlayers considers active players

TheNickSkater:
.minPlayers will now only consider players that were active in a round. For this I added the global int "ttt2_active_players". I did not use a reset for this, as it is set on every round begin, and I dont think you need to access a "0" in prepare/post.

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -1215,14 +1215,22 @@ function BeginRound()
 
 	local plys = player.GetAll()
 
+	local activeCount = 0
+
 	for i = 1, #plys do
 		local ply = plys[i]
 
 		ply:ResetRoundDeathCounter()
 
+		local isActive = ply:Alive() and ply:IsTerror()
+
 		-- a player should be considered "was active in round" if they received a role
-		ply:SetActiveInRound(ply:Alive() and ply:IsTerror())
+		ply:SetActiveInRound(isActive)
+
+		activeCount = isActive and activeCount+1 or activeCount
 	end
+
+	SetGlobalInt("ttt2_active_players", activeCount)
 
 	hook.Call("TTTBeginRound", GAMEMODE)
 
--- a/gamemodes/terrortown/gamemode/shared/sh_init.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_init.lua
@@ -663,20 +663,9 @@ function EquipmentIsBuyable(tbl, ply)
 	end
 
 	if tbl.minPlayers and tbl.minPlayers > 1 then
-		local choices = {}
-		local plys = player.GetAll()
-
-		for i = 1, #plys do
-			local v = plys[i]
-
-			-- everyone on the forcespec team is in specmode
-			if not IsValid(v) or v:GetForceSpec() then continue end
-
-			choices[#choices + 1] = v
-		end
-
-		if #choices < tbl.minPlayers then
-			return false, " " .. #choices .. " / " .. tbl.minPlayers, "Minimum amount of active players needed."
+		local activePlayers = GetGlobalInt("ttt2_active_players", 0)
+		if activePlayers < tbl.minPlayers then
+			return false, " " .. activePlayers .. " / " .. tbl.minPlayers, "Minimum amount of active players needed."
 		end
 	end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,13 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Added ConVar to toggle double-click buying
 - Added Japanese translation (by @Westoon)
+- Added new Global Integer "ttt2_active_players"
 
 ### Changed
 
 - The weapon pickup system has been improved to increase stability and remove edge cases in temporary weapon teleportation
 - Updated Spanish translation (by @DennisWolfgang)
+- Changed so .minPlayers now only considers the active count of players a round had at start
 
 ### Fixed
 


gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -1215,15 +1215,23 @@ function BeginRound()
...
 
 		ply:ResetRoundDeathCounter()
 
+		local isActive = ply:Alive() and ply:IsTerror()
+
 		-- a player should be considered "was active in round" if they received a role
-		ply:SetActiveInRound(ply:Alive() and ply:IsTerror())
+		ply:SetActiveInRound(isActive)
+
+		activeCount = isActive and activeCount+1 or activeCount
```
TimGoll, line 1230: Did you test this? I'm pretty sure lua fails with the `+1` and the `or` here, you should add brackets.

On a different note, please always add `spaces` around the `+` sign.

gamemodes/terrortown/gamemode/shared/sh_init.lua
```diff
@@ -663,20 +663,9 @@ function EquipmentIsBuyable(tbl, ply)
...
-
-			-- everyone on the forcespec team is in specmode
-			if not IsValid(v) or v:GetForceSpec() then continue end
-
-			choices[#choices + 1] = v
-		end
-
-		if #choices < tbl.minPlayers then
-			return false, " " .. #choices .. " / " .. tbl.minPlayers, "Minimum amount of active players needed."
+		local activePlayers = GetGlobalInt("ttt2_active_players", 0)
```
TimGoll, line 666: I'd much prefer having a shared getter method for this `GlobalInt`. This greatly improves the documentation

TimGoll (changes requested):

gamemodes/terrortown/gamemode/server/sv_main.lua
TheNickSkater: I did test it and it worked like intendet. I will add spaces.

gamemodes/terrortown/gamemode/shared/sh_init.lua
TheNickSkater: > I'd much prefer having a shared getter method for this `GlobalInt`. This greatly improves the documentation

Sure, where should I put them? @Alf21, could you tell me (us) where I should best put the getter and setter function?

Alf21: This won't work at all. The client should never know the exact amount of alive players. So a shared global doesn't make sense. Instead, just the available equipment should be sent by the server. As you see, the problem is a way more complex.

Alf21: Additionally, I don't see any issue here. Players that aren't alive but forced spectators were alive at some time in the round. As a result, the equip should be available, because the requirements have been filled.

Alf21 (changes requested):

gamemodes/terrortown/gamemode/shared/sh_init.lua
TheNickSkater: The problem without any changes is, that even players that join during a around count as players. I though about using WasActiveInRound, but this will lead to the item not being buyable if somebody leaves and the minPlayers amount is not fullfiled anymore. Also, I mean a player does already "know" how many players are alive? Pretty much because he knows the alive state.

Alf21: Yea you are right, sad that `Alive` is a shared function for a multiplayer game... Anyways, I meant the combination with `IsTerror`. Wouldn't a clientside combination of `Alive() and GetForceSpec()` doing the trick? If not, then it's an issue with Alive() which probably should get patched

TheNickSkater: A combination would still not work as it should, as it would still not count players that disconnect from the Server, which would be pretty unfair for Traitors suddenly not being able to buy an item anymore, just because somebody left.

Alf21: Caching the exact amount of active players at the beginning of the round should do the trick.

TheNickSkater: Well, I do this in this PR xD?

Alf21: Yea, but then there still isn't any need for a shared global var (the issue we are discussing about here).

TheNickSkater: But EquipmentIsBuyable is called on the client aswell, so it would not grey out the Item, but just give out an error when trying to buy it.

Alf21: Then we have to modify it. There is no need of adding an additional networked var if the client is already aware of the attributes needed for further "calculations".

TheNickSkater: Okay well.. This is getting a bit over my head rn. I do understand, but I do not understand how exactly you'd like to have it.
Sure, not GlobalInt, alright. You said caching, but this also needs to be done in client and Server. Soo? Should it be a global variabel which is shared?

Alf21: Yea, a shared var that isn't synced (because we don't need a sync if the client and the server are already aware of the attributes needed to get to the same state).

TheNickSkater: Okay well.. I could now go over "but what if a client joins after round begun, and gets respawned? He wont know the start player count" thingy.. But if thats okay, sure.. I just need to know, where I should initialize the variable, and how to call it.

TheNickSkater:
I will close this for now as I will probably redo it anyway sometimes.

PR #652, opened 2020-09-23
Call Detective Hook: added serverside hook on "call detective"

TimGoll:
This is a really small change that adds a serverside hook that is called once a player pressed the "call detective" button. This was requested by @blackmagicfine on discord.

Additionally I changed `LANG.Msg` to `LANG.MsgAll`. The previous usage of the function already resulted in a message to everybody, however it isn't obvious to someone new and therefore bad coding style.

Please don't merge this until it is tested. While it doesn't throw errors and looks good to me, the functionality has to be confirmed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added ConVar to toggle double-click buying
 - Added Japanese translation (by @Westoon)
 - Added a team indicator in front of every name in the scoreboard (just known teams will be displayed)
+- Added a hook (`TTT2ModifyCorpseCallRadarReceipients`) that is called once "call detective" is pressed
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added ConVar to toggle double-click buying
 - Added Japanese translation (by @Westoon)
 - Added a team indicator in front of every name in the scoreboard (just known teams will be displayed)
-- Added a hook (`TTT2ModifyCorpseCallRadarReceipients`) that is called once "call detective" is pressed
+- Added a hook `TTT2ModifyCorpseCallRadarRecipients` that is called once "call detective" is pressed
 
 ### Changed
 
--- a/gamemodes/terrortown/gamemode/server/sv_corpse.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_corpse.lua
@@ -237,7 +237,7 @@ local function ttt_call_detective(ply, cmd, args)
 		if CORPSE.GetFound(rag, false) then
 			local plyTable = GetRoleChatFilter(ROLE_DETECTIVE, true)
 
-			hook.Run("TTT2ModifyCorpseCallRadarReceipients", plyTable, rag, ply)
+			hook.Run("TTT2ModifyCorpseCallRadarRecipients", plyTable, rag, ply)
 
 			-- show indicator in radar to detectives
 			net.Start("TTT_CorpseCall")
@@ -261,7 +261,7 @@ concommand.Add("ttt_call_detective", ttt_call_detective)
 -- @param Player ply The player that pressed the "call detective" button
 -- @hook
 -- @realm server
-function GM:TTT2ModifyCorpseCallRadarReceipients(notifiedPlayers, ragdoll, ply)
+function GM:TTT2ModifyCorpseCallRadarRecipients(notifiedPlayers, ragdoll, ply)
 
 end
 
--- a/gamemodes/terrortown/gamemode/server/sv_corpse.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_corpse.lua
@@ -235,18 +235,35 @@ local function ttt_call_detective(ply, cmd, args)
 
 	if IsValid(rag) and rag:GetPos():Distance(ply:GetPos()) < 128 then
 		if CORPSE.GetFound(rag, false) then
+			local plyTable = GetRoleChatFilter(ROLE_DETECTIVE, true)
+
+			hook.Run("TTT2ModifyCorpseCallRadarReceipients", plyTable, rag, ply)
+
 			-- show indicator in radar to detectives
 			net.Start("TTT_CorpseCall")
 			net.WriteVector(rag:GetPos())
-			net.Send(GetRoleChatFilter(ROLE_DETECTIVE, true))
+			net.Send(plyTable)
 
-			LANG.Msg("body_call", {player = ply:Nick(), victim = CORPSE.GetPlayerNick(rag, "someone")}, MSG_CHAT_PLAIN)
+			LANG.MsgAll("body_call", {player = ply:Nick(), victim = CORPSE.GetPlayerNick(rag, "someone")}, MSG_CHAT_PLAIN)
 		else
 			LANG.Msg(ply, "body_call_error")
 		end
 	end
 end
 concommand.Add("ttt_call_detective", ttt_call_detective)
+
+---
+-- This hook is called after a players pressed the "call detective" button and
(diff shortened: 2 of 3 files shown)


Alf21 (approved):

gamemodes/terrortown/gamemode/server/sv_corpse.lua
```diff
@@ -235,19 +235,36 @@ local function ttt_call_detective(ply, cmd, args)
...
+			hook.Run("TTT2ModifyCorpseCallRadarReceipients", plyTable, rag, ply)
+
 			-- show indicator in radar to detectives
 			net.Start("TTT_CorpseCall")
 			net.WriteVector(rag:GetPos())
-			net.Send(GetRoleChatFilter(ROLE_DETECTIVE, true))
+			net.Send(plyTable)
 
-			LANG.Msg("body_call", {player = ply:Nick(), victim = CORPSE.GetPlayerNick(rag, "someone")}, MSG_CHAT_PLAIN)
+			LANG.MsgAll("body_call", {player = ply:Nick(), victim = CORPSE.GetPlayerNick(rag, "someone")}, MSG_CHAT_PLAIN)
```
Alf21, line 247: But I don't like this change or the function at all, it's a useless function redirect, because https://github.com/TTT-2/TTT2/blob/01b27b75ef8b9eec8eeabdd5523ef23a0726b93f/gamemodes/terrortown/gamemode/shared/sh_lang.lua#L50 already does the same and looks a bit inconsistent.

TimGoll: I mean, I can leave it. However do you see on first glance that `LANG.Msg` with three arguments and a stringa as the first argument means that it is broadcasted? Because I don't. And `LANG.MsgAll` clearly shows this.

Alf21: Then why not change `LANG.MsgAll` directly instead of broadcasting it with a string check inside the function / redirecting it? That's what I meant with 'inconsistency'

TimGoll: I mean, I don't write that `LANG.Msg` function. It's ancient and is probably used by many addons. `Lang.MsgAll` calls `LANG.Msg` with `nil` as its first argument btw.

Alf21: Yep, that's a redirect because this function (A) is called, and calls another function (B) that doing a thing it would do even without calling function A.

Changing `MsgAll` into calling `LANG.ProcessMsg(nil, arg1, arg2, arg3)` directly would make more sense imo and should help to clear unnecessary nesting. But that's for another PR.

gamemodes/terrortown/gamemode/server/sv_corpse.lua
```diff
@@ -235,19 +235,36 @@ local function ttt_call_detective(ply, cmd, args)
...
+---
+-- This hook is called after a players pressed the "call detective" button and
+-- all requirements were met. It can be used to modify the table of players that
+-- should receive the corpse radar ping.
+-- @param Player notifiedPlayers The table of players that will be notified
+-- @param Entity ragdoll The ragdoll whose coordinates are about to be sent
+-- @param Player ply The player that pressed the "call detective" button
+-- @hook
+-- @realm server
+function GM:TTT2ModifyCorpseCallRadarReceipients(nofifiedPlayers, ragdoll, ply)
```
AaronMcKenney, line 264: `nofifiedPlayers` should be `notifiedPlayers`

AaronMcKenney (changes requested):
Ok, so this technically works, but there is one typo (nofifiedPlayers) and one UI thing missing from it:

In cl_init.lua there is this if statement:
```lua
if client:IsActiveRole(ROLE_DETECTIVE) and not table.IsEmpty(self.called_corpses) then
```
Which determines who can see the call notification that hovers over the corpse. However, since we are now allowing non-detectives to use the call button, it should be something like:
```lua
if not table.IsEmpty(self.called_corpses) then
```
So that non-detectives of interest can also see the called corpses.

gamemodes/terrortown/gamemode/server/sv_corpse.lua
TimGoll: Oh great, thanks. I'll rework it in a minute and let you know. Are you blackmagicfire on discord?

AaronMcKenney: Yup, I am blackmagicfine.

TimGoll: So, I thought about this for a while. At first I thought your suggestion might be a problem because it can be weird with role changes.
However I came to the conclusion that it doesn't matter. In the worst case, you see a corpse position for up to 45 seconds after your role change which is no problem in my opinion. Thanks for your input!

Alf21: This anyways isn't an item issue. It depends on the role, if the radar was bought and the role changed, the radar should be kept so it can still be used. If the radar is included in a role and that role got changed, then the old role has to handle this on it's own (removing the item in the right hook).

TimGoll: I'm not that deep into the code, but is it really based on the radar? To me it looks like the code happens to be in the radar file, but has nothing to do with the actual radar. This fact is supported by the situation where a detective has a corpse point in world without having a radar.

gamemodes/terrortown/gamemode/server/sv_corpse.lua
```diff
@@ -235,19 +235,36 @@ local function ttt_call_detective(ply, cmd, args)
 
 	if IsValid(rag) and rag:GetPos():Distance(ply:GetPos()) < 128 then
 		if CORPSE.GetFound(rag, false) then
+			local plyTable = GetRoleChatFilter(ROLE_DETECTIVE, true)
+
+			hook.Run("TTT2ModifyCorpseCallRadarReceipients", plyTable, rag, ply)
```
Histalek, line 240:
```suggestion
			hook.Run("TTT2ModifyCorpseCallRadarRecipients", plyTable, rag, ply)
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added ConVar to toggle double-click buying
 - Added Japanese translation (by @Westoon)
 - Added a team indicator in front of every name in the scoreboard (just known teams will be displayed)
+- Added a hook (`TTT2ModifyCorpseCallRadarReceipients`) that is called once "call detective" is pressed
```
Histalek, line 12: typo + parenthesis are not needed imo
```suggestion
- Added a hook `TTT2ModifyCorpseCallRadarRecipients` that is called once "call detective" is pressed
```

gamemodes/terrortown/gamemode/server/sv_corpse.lua
```diff
@@ -235,19 +235,36 @@ local function ttt_call_detective(ply, cmd, args)
...
+---
+-- This hook is called after a players pressed the "call detective" button and
+-- all requirements were met. It can be used to modify the table of players that
+-- should receive the corpse radar ping.
+-- @param Player notifiedPlayers The table of players that will be notified
+-- @param Entity ragdoll The ragdoll whose coordinates are about to be sent
+-- @param Player ply The player that pressed the "call detective" button
+-- @hook
+-- @realm server
+function GM:TTT2ModifyCorpseCallRadarReceipients(notifiedPlayers, ragdoll, ply)
```
Histalek, line 264:
```suggestion
function GM:TTT2ModifyCorpseCallRadarRecipients(notifiedPlayers, ragdoll, ply)
```

Histalek (changes requested):
small naming problem :)

also i didn't test it (yet)

AaronMcKenney (approved):
I have tested the latest changes and they work!

Histalek (approved):

PR #654, opened 2020-09-26
DNA Scanner Crash: Added a check to prevent crash if using an old weapon base

Alf21:
This should fix the crash while providing an error that should help to fix further connected problems.

### DIFF
--- a/gamemodes/terrortown/entities/weapons/weapon_ttt_wtester.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_ttt_wtester.lua
@@ -116,7 +116,7 @@ function SWEP:Initialize()
 			self:AddTTT2HUDHelp("dna_help_primary", "dna_help_secondary")
 			self:AddHUDHelpLine("dna_help_reload", Key("+reload", "R"))
 		else
-			ErrorNoHalt("[TTT2][ERROR] You are using an add-on that overwrites the 'weapon_tttbase.lua' file while not provide essential functions. That will lead to several incompatibilites.")
+			ErrorNoHalt("[TTT2][ERROR] You are using an add-on that overwrites the 'weapon_tttbase.lua' file while not providing essential functions. That will lead to several incompatibilites.")
 		end
 	end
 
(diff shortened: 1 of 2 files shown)


TimGoll (approved):
I approve this, but I'm not 100% happy with this check. But I guess it is the only way to make sure the game stops crashing.
But maybe we should keep it crashing? I think an error like this is probably ignored by most users.

gamemodes/terrortown/entities/weapons/weapon_ttt_wtester.lua
```diff
@@ -99,20 +99,25 @@ local dna_screen_circle = Material("models/ttt2_dna_scanner/screen/circle")
...
 
 		self:SetSubMaterial(0, "!scanner_screen_mat")
 
 		surface.CreateAdvancedFont("DNAScannerDistanceFont", {font = "Trebuchet24", size = 32, weight = 1200})
+
+		if isfunction(self.AddTTT2HUDHelp) then
+			self:AddTTT2HUDHelp("dna_help_primary", "dna_help_secondary")
+			self:AddHUDHelpLine("dna_help_reload", Key("+reload", "R"))
+		else
+			ErrorNoHalt("[TTT2][ERROR] You are using an add-on that overwrites the 'weapon_tttbase.lua' file while not provide essential functions. That will lead to several incompatibilites.")
```
TimGoll, line 119: "while not provid**ing**"

Alf21: Could you please make a suggestion 👍 :D

Alf21:
> 
> 
> I approve this, but I'm not 100% happy with this check. But I guess it is the only way to make sure the game stops crashing.
> But maybe we should keep it crashing? I think an error like this is probably ignored by most users.

Thought about it as well, but an error should be visible in the top left corner as well as an error in the console. Further, everything else should work if the default TTT weapon base functions are provided. If not, it wouldn't work at all. So imo we don't need to keep crashing, an error should work so far.

gamemodes/terrortown/entities/weapons/weapon_ttt_wtester.lua
TimGoll:
```suggestion
			ErrorNoHalt("[TTT2][ERROR] You are using an add-on that overwrites the 'weapon_tttbase.lua' file while not providing essential functions. That will lead to several incompatibilites.")
```

saibotk (approved):

PR #655, opened 2020-09-28
Speed: Fix initialization of oldval

saibotk:
This patch avoids the invalid nil state of the oldval variable in case the users speedMultiplier is not 1 at the initialization.
This is now fixed, by using the GetSpeedMultiplier method, which returns either the speedModifier or 1.0


TimGoll (approved):
A simple but good change, I like it!

Histalek (approved):
LGTM

Alf21 (approved):

Issue #657, opened 2020-09-28
[Feature Request] Additional Hook called after PlayerTakeDamage should output real damage


ZenBre4ker:
**Is your feature request related to a problem? Please describe.**
The hitdamage numbers addon only shows the raw damage before karma and other effects get considered, as it only hooks to the GM:EntityTakeDamage. 
Even trying to mod it in by using the net.receive("ttt2_damage_received")-function, wouldn't work as the GM:PlayerTakeDamage only sends the "amount" instead of dmginfo:GetDamage() to the cl_damage_indicator.

**Describe the solution you'd like**
My solution would be to create a new function like GM:EntityTookScaledDamage to hook to, which just contains all the processed information after GM:EntityTakeDamage and GM:PlayerTakeDamage got called.

**Describe alternatives you've considered**
Alternatively you could send the information like this:
net.Start("ttt2_damage_received")
net.WriteFloat(dmginfo:GetDamage())
net.Send(ent)
but I'm unsure if it would be possible to have more than one receive function. Normally a packet can be only sent once and would need to be resend with the same information, which sound unnecessary.

**Important Files**
TTT2\gamemodes\terrortown\gamemode
..\server\sv_player.lua
..\client\cl_damage_indicator

TimGoll:
I actually like your proposal of an additional hook as I've seen the same problem as well. This hook could be used in our addons as well.

ZenBre4ker:
Feature bereits in Garrys Mod:
https://wiki.facepunch.com/gmod/GM:PostEntityTakeDamage

PR #670, opened 2020-10-08
Playerdamage: Adding Post EntityTakeDamage Hook

ZenBre4ker:
#657 
To be able to read out or manipulate the damage after TTT2 processed it, this simple hook is called after calculating the reduced damage to players due to karma.


ZenBre4ker:
Can you use return hook.run() like this, so that hooks to "TTT2PostEntityTakeDamage" could also return a value to block the whole damage event? 
Or does this always lead to a block?

TimGoll:
> Can you use return hook.run() like this, so that hooks to "TTT2PostEntityTakeDamage" could also return a value to block the whole damage event?
> Or does this always lead to a block?

This should work because it will pass `nil` as well. However You should test your PR before we review it.

ZenBre4ker:
Finally back and the new Hook now is tested and allows to show damage decrease due to Karma. Returning this hook allows to cancel the damage event by returning a value.

Here's an example performed with a M98B from the M9k weapons, which deals 200 on the head and with 0 karma should deal 90% less, so only 20. See the effect ingame and the output in the server console, which states the real damage.
With old Hook:
Headshot with 0 Karma
grafik

With New Hook:
Headshot with 0 Karma
grafik
Headshot with 1000 Karma
grafik

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1477,3 +1481,15 @@ end
...
+-- Called by @{GM:EntityTakeDamage} after damage output was changed due to internal events. You can modify all parts of the damage info in this hook again.
+-- @param Entity ent The @{Entity} taking damage
+-- @param CTakeDamageInfo dmginfo Damage info
+-- @return boolean Return true to completely block the damage event
+-- @note e.g. no damage if a value after Karma calculation is too low, etc
+-- @hook
+-- @realm server
+-- @local
+function GM:TTT2PostEntityTakeDamage(ent, dmginfo)
+end
```
TimGoll, line 1495:
```suggestion
-- Called by @{GM:EntityTakeDamage} after damage output was changed due to internal events. You can modify all parts of the damage info in this hook again.
-- @param Entity ent The @{Entity} taking damage
-- @param DamageInfo dmgInfo The damage info
-- @return boolean Return true to completely block the damage event
-- @note There is no damage value if it is too low after the karma calculation.
-- @hook
-- @realm server
-- @local
function GM:TTT2PostEntityTakeDamage(ent, dmgInfo)

end
```

ZenBre4ker: The note was meant as an example what you could do with the post damage hook. But I removed that note instead.

TimGoll:
Okay. It looks good to me. One thing though:
While it is ok that you can modify the damage in this hook, I don't think it is a good idea to mention or event emphasis on it. Doesn't modifying/canceling the damage in this hook defete the whole purpose why you added it?

ZenBre4ker:
On one hand I also wasn't sure about that, because it would be nice if all changes would be made before TTT2 applies its changes and therefore only a copy of the data should be sent.
On the other hand an addon may want to change damage information after Karma has been calculated.

So I'm unsure what's better. Just not mentioning, that you can change the damage afterwards or just not passing the original reference and therefore making it read only.
But as I can't imagine any purpose of changing it afterwards instead of before addons, I should probably make that hook read-only.

What do you think?

TimGoll:
In my opinion a read only hook is the way to go. Is there a scenario where you want to modify the damage after the karma? At least for my addons it doesn't matter. At least in my opinion armor and karma should be factored in AFTER all damage multipliers are calculated.
But that's just my opninion. What's your opinion @Alf21 @saibotk @Histalek ?

ZenBre4ker:
Changed it, because I agree with you. Can't think of a good use, therefore making it safe to hook to this and read out the real damage.

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -1210,7 +1210,8 @@ function GM:EntityTakeDamage(ent, dmginfo)
 	end
 
 	--Call a second Hook after Damage got processed to output the real Damage that has been dealt
-	return hook.Run("TTT2PostEntityTakeDamage", ent, dmginfo)
+	local tmpdmginfo =  table.Copy(dmginfo)
```
ZenBre4ker, line 1213: This is not possible, noticed it too late....

ZenBre4ker (commented):
How do I copy dmgInfo though?

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: yes, it is probably userinfo, isn't it?

ZenBre4ker:
Feature ist bereits in Garrys Mod enthalten:
https://wiki.facepunch.com/gmod/GM:PostEntityTakeDamage

PR #672, opened 2020-10-14
Weapon Pickup: Fixed pickup bug

TheNickSkater:
Fixed bug where some weapons would stay in inventory even though they should get dropped, leading to a player being able to pickup pretty much infinite weapons. (Yes I know its just a timer)

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player_ext.lua
@@ -1369,6 +1369,8 @@ function plymeta:SafePickupWeapon(wep, ammoOnly, forcePickup, dropBlockingWeapon
 		if dropWeapon == wep then return end
 
 		timer.Simple(0, function()
+			if not IsValid(self) or not IsValid(dropWeapon) then return end
+
 			self:SafeDropWeapon(dropWeapon, true)
 		end)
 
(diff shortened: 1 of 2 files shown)


TimGoll:
seems like a reasonable change. Did you verify that this fixes the problem

TheNickSkater:
> seems like a reasonable change. Did you verify that this fixes the problem?

I did, I let [SirZento] Nico test it and I tested it myself as it did also affect my weapons.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1368,7 +1368,9 @@ function plymeta:SafePickupWeapon(wep, ammoOnly, forcePickup, dropBlockingWeapon
 		-- Very very rarely happens but definitely breaks the weapon and should be avoided at all costs
 		if dropWeapon == wep then return end
 
-		self:SafeDropWeapon(dropWeapon, true)
+		timer.Simple(0, function()
```
TimGoll, line 1371: please add a `if not IsValid(self) or not IsValid(dropWeapon) then return end` check here to prevent further errors. Otherwise it looks good to me

TimGoll (changes requested):

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TheNickSkater: Added

TimGoll (approved):

Alf21 (approved):
Anyways, there's still a bug/inconsistency that it's possible picking up infinite weapons. There definitely need to be another check preventing a pickup if the slot has reached it's limit.

TimGoll:
> Anyways, there's still a bug/inconsistency that it's possible picking up infinite weapons. There definitely need to be another check preventing a pickup if the slot has reached it's limit.

I'm not entirely sure what you're refering to as this is already a thing:

https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/server/sv_player_ext.lua#L1408-L1410

PR #673, opened 2020-10-15
[WIP] Migrations: Added migration library to enable schema changes

Histalek:
This library should allow for changes of the database schema in a reversible and controlled manner.

Right now only the addition of migrations is implemented but the other functions are already documented, so let me know if you miss any functionality.

Depending on which other included files will use the migration library we may need to include it sooner (in `gamemodes/terrortown/gamemode/shared/sh_init.lua`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,15 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Added
+
+- Added `migrations` library to enable easy schema changes of database tables
+
 ### Changed
 
 - Added global alias for IsOffScreen function to util.IsOffScreen
 - Updated Japanese localization (by @Westoon)
 - Moved rendering modules to libraries
-
 
 ## [v0.8.1b](https://github.com/TTT-2/TTT2/tree/v0.8.1b) (2021-02-19)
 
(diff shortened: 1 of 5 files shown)


Histalek:
Aside from the error-handling in `migrations.Add()` and this not being tested yet - both of which i will most likely tackle tomorrow - this should be ready.

I'll leave it as a draft for now. Questions & reviews are welcome :)

saibotk:
Nice to see progress!

Some feedback though:
What is the point of storing the queries in the database? I would suggest handling the loaded migration files as the single source of truth, as this could also cause issues, with different queries in the database than in the file themselves. (note: Changing the code semantically inside a migration file after it has been in production should be considered as disallowed)

Also, I would instead suggest using a table based approach, like we do with the items, roles etc. but without the inheritance. 
Using something like a `MIGRATION` table that we expect, would allow defining functions like `MIGRATION:up()` and `MIGRATION:down()`, that define the migration cleanly in one place in its file and allows for a cleaner auto-load of migrations.

This proposal is really inspired by Laravels approach, and I recommend checking out their repository too ( https://github.com/laravel/laravel/tree/8.x/database/migrations ).

Histalek:
Storing the queries in the database was my take on discouraging to change migrations once in production. I didn't think about possible discrepancies between the definitions in lua and the definitions in the database, having a single source of truth is definitely not a bad idea and i will think about the table based approach or possible alternatives.

Regarding the 'one file per migration' i'm not really sure if the file-limit should be a concern for us (frankly i don't know how high/low it exactly is and how close we are to reaching it) but i wanted to keep it simply one file per addon/table/whatever you decide to throw into a namespace.

Histalek:
Not even 4 years old ... R.I.P. :'(

Issue #676, opened 2020-10-16
[Feature Request] Use a hook to determine who can and can't transfer credits.


AaronMcKenney:
**Is your feature request related to a problem? Please describe.**
This request is in relation to The Defective role that I have released and The Banker role which I have coded up.

In the case of The Defective, the problem is that currently a player can only transfer credits to other players who they believe are on their same team. This leads to the odd situation in which The Detective can transfer credits to The Defective (as The Detective sees The Defective as a Detective), but the reverse is not true (as The Defective sees that they themselves are not a Detective).

The Banker is a new detective role which obtains all credits that all non-bankers spend. This leads to the interesting situation in which The Banker can transfer credits to The Detective, who pays The Banker for their items, who then transfers the credits back to The Detective. In otherwords, The Banker can work with an aligned shopping role to generate infinite credits. I think that it is important to add a ConVar to disable these types of transactions. I believe that requires a server hook however.

**Describe the solution you'd like**
Generally speaking, a solution to both problems involves the client being more lax in suggesting which players can receive credits and giving the server more flexibility in how it handles these transactions. 

Explicitly, the UI in cl_transfer.lua can be modified to allow for credit transferring across roles (as long as the client believes that their recipient is a shopper), and a hook can be added in TransferCredits to determine if the transfer should be rejected or not.

One thing to note about this solution is that it would change how existing roles can transfer credits: It would allow for Survivalists, Traitors, and Jackals alike to give credits to The Detective (it would not allow for the reverse to occur, since The Detective does not know what these other players' roles are). I think that this would allow for more interesting interactions to occur (ex. shopping roles bribing The Detective).

PR #677, opened 2020-10-16
Small rework of how credits are transfered (now includes a hook).

AaronMcKenney:
#676 This pull request implements the solution expressed in the credit transfer feature request. It adds a new hook that can prevent  the transferring of credits and allows for shopping roles to potentially transfer credits across teams.


gamemodes/terrortown/gamemode/server/sv_shop.lua
```diff
@@ -213,6 +213,10 @@ local function TransferCredits(ply, cmd, args)
 
 	if credits == 0 then return end
 
+	local allow = hook.Run("TTT2CanTransferCredits", ply, target, credits)
+
+	if allow == false then return end
```
Alf21, line 218:
```suggestion
	if hook.Run("TTT2CanTransferCredits", ply, target, credits) == false then return end
```

gamemodes/terrortown/gamemode/client/cl_transfer.lua
```diff
@@ -50,7 +50,7 @@ function CreateTransferMenu(parent)
 	for i = 1, #plys do
 		local ply = plys[i]
 
-		if ply ~= client and ply:IsActive() and (not ply:GetSubRoleData().unknownTeam or ply:IsRole(ROLE_DETECTIVE) and client:IsRole(ROLE_DETECTIVE)) and ply:IsInTeam(client) then
+		if ply ~= client and ply:IsActiveShopper() and (not ply:GetSubRoleData().unknownTeam or ply:GetBaseRole() == ROLE_DETECTIVE) then
```
Alf21, line 53: A Detective should be a shopper in every way, so the new check is useless. The goal of the old check have been to make sure just Detectives can send credits to each other AND all in all just teammates.

Additionally, preventing something on the server but not on the client is bad practice, because in that way the situation could happen that credits won't decrease / transfer even if it's a valid option for the client. So add the hook here as well or make it possible to transfer credits to everyone OR to transfer credits just to teammates. I don't think it's wrong to be able transferring credits to the Defective if thinking it's a Detective. Otherwise you could prove him.

Alf21 (changes requested):

gamemodes/terrortown/gamemode/client/cl_transfer.lua
AaronMcKenney: > A Detective should be a shopper in every way, so the new check is useless

So then should the check be just "ply ~= client and ply:IsActiveShopper()" ?

> So add the hook here as well.

Perhaps then the check would be something like: 
```lua
if ply ~= client and ply:IsActiveShopper() and hook.Run("TTT2CanTransferCredits", client, ply, 1) ~= false then
```
I can make that change and re-test.

> or make it possible to transfer credits to everyone

That's more or less what this code change means to do. The only catches are: the sender must have some idea as to what role the recipient has and that role must be a shopping role.

> I don't think it's wrong to be able transferring credits to the Defective if thinking it's a Detective. Otherwise you could prove him.

Agreed. The change here is a fix for the reverse: currently a Defective cannot transfer credits to a Detective because the Defective sees that they are on different teams. This change allows for cross-team credit transferring.

TimGoll: I totally agree with the shared thingy.

But why is the detective always a shopper? Isn't it possible to remove the detective shop with the shopeditor?

Alf21: Hmmm... I thought about the topic again and regarding that amount of different roles, shouldn't it rather be possible transferring credits to EVERYONE? For a detective, it's still hard to transfer credits to a survivalist even if that role is proven (e.g. through gamestyle oder priest), because the only way to detect that role as proven is by killing the player and confirming (credits transfer at this point doesn't make sense anymore lmao)

TimGoll: yeah, transfering it to everyone seems to be the easiest and cleanest solution. I like it

AaronMcKenney: By everyone does this include non-shopping roles? Can non-shoppers even transfer credits currently? I'm all for it, but I'm curious about what kind of changes this would require.

TimGoll: Yes, I think this is the best solution. Otherwise you might be able to detect the innocent player. If you so desire to spend your credits someone that can not use them, it's your own fault.

TimGoll:
Is this PR still being worked on?

AaronMcKenney:
Hey Tim. I had to take a break from this one for a couple of unrelated reasons, but I do have a solution of sorts that I have not fully tested (partially due to needing other people to test it with). I'm planning to show it off before that stale-bot closes the issue next week. I may as well run down what I've done so far (I may have gone overboard):
1. Allow players to transfer credits to players of different roles/teams, as previously discussed.
2. Allow non-shoppers to receive credits from transfers, as previously discussed.
3. Allow corpses to receive credits. This had to be done in case a detective wanted to transfer a credit to a dead player who they think is alive. The credit will appear on their dead body if possible and can be picked up by another player who interacts with their body (this is keep those credits in circulation. Unspent credits lead to uninteresting games).
4. If a dead non-shopper revives, they receive the credits that were on their corpse, so that those credits remain in circulation.
5. Allow non-shoppers to pick up credits from corpses.
6. Allow non-shoppers to transfer credits in the same manner as shoppers. This involves giving non-shoppers access to the context menu as well as the transfer credit tab (but not the shop tab), and adding a GUI element that shows how many credits they currently have (similar to the shop tab). The implementation of this on my end has the added benefit of giving non-shoppers access to other components of this GUI, such as the radar tab if they have a radar equipped.

(5) and (6) technically a separate discussion. For (6) I felt that if it is allowed for non-shoppers to receive credits, then it should only be fair to allow them to transfer credits to players who can actually use them. Also if they can't transfer their credits, then those credits will remain unspent for as long as they're alive, which would make for a less interesting game. For (5) I think this one might be controversial, as it will most likely hamper the detective's ability to receive credits, but I added it mostly to make the rules of credits apply in the same manner for everyone. If you want me to take out (5) I can do so.

saibotk:
Hey @AaronMcKenney,
Thank you for your status update, 
I agree with 1-4 but 5 and 6 should be discussed elsewhere beforehand, as they would be a bigger gameplay change.

AaronMcKenney:
Hey @saibotk that's fair. I'll amend what I have now to just be 1-4 for now.

AaronMcKenney:
After some testing, I have updated this fork, which has the 1-4 changes (5+6 were taken out).
Something that I forgot to mention were the changes I made to the help message in the credit transfer tab. The hook has two return values now: "allow" and "msg", the latter of which can be used to inform the user as to why their transaction may have failed (through the dhelp panel). The existing messaging logic in this tab had to be refactored to make this work.

You can see the hook in action here:
https://github.com/AaronMcKenney/ttt2-role_bank/blob/master/lua/terrortown/entities/roles/banker/shared.lua

saibotk (changes requested):
Could you also add the parameters to the hook documentation? That would be nice, otherwise this looks good! Thank you!

AaronMcKenney:
@saibotk 

Should I also add return info? For example, something like this?
```lua
---
-- Called to check if a transaction between two players is allowed.
-- @param Player that wants to send credits
-- @param Player that would receive the credits
-- @param Number of credits that would be transferred
-- @return[default=nil] Boolean which disallows a transaction when false
-- @return[default=nil] String for the client which offers info related to the transaction 
-- @realm client
local allow, msg = hook.Run("TTT2CanTransferCredits", client, ply, CREDITS_PER_XFER)
```

Also, will I need to add the hook documentation to https://api-docs.ttt2.neoxult.de/ and if so, how?

saibotk:
Exactly. 
Don't worry about the api-docs ;) once the PR is merged the site is auto deployed and generated from the annotations :+1:

saibotk (changes requested):

saibotk (approved):

gamemodes/terrortown/gamemode/client/cl_transfer.lua
```diff
@@ -2,8 +2,48 @@
...
+local dsubmit
+local dhelp
+local dform
+local selected_sid
+
+local function UpdateTransferSubmitButton()
+	local client = LocalPlayer()
+	if client:GetCredits() <= 0 then
+		dhelp:SetText(GetTranslation("xfer_no_credits"))
+		dsubmit:SetDisabled(true)
```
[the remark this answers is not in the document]
Alf21: below the same

Alf21 (changes requested):

gamemodes/terrortown/gamemode/server/sv_shop.lua
```diff
@@ -221,9 +223,31 @@ local function TransferCredits(ply, cmd, args)
...
+	-- @param Number of credits that would be transferred
+	-- @return[default=nil] Boolean which disallows a transaction when false
+	-- @return[default=nil] String for the client which offers info related to the transaction
+	-- @realm server
+	local allow, _ = hook.Run("TTT2CanTransferCredits", ply, target, credits)
+	if allow == false then return end
+
 	ply:SubtractCredits(credits)
-	target:AddCredits(credits)
+	if target:Alive() then
```
Alf21, line 238: Forced spectators are alive as well. `IsTerror` check is missing

gamemodes/terrortown/gamemode/server/sv_shop.lua
```diff
@@ -221,9 +223,31 @@ local function TransferCredits(ply, cmd, args)
...
+		target:AddCredits(credits)
+	else
+		-- The would be recipient is dead, which the sender may not know.
+		-- Instead attempt to send the credits to the target's corpse, where they can be picked up.
+		local rag = target:FindCorpse()
+		if IsValid(rag) then
+			CORPSE.SetCredits(rag, CORPSE.GetCredits(rag, 0) + credits)
+		end
```
Alf21, line 246: Maybe we should discuss about what happens if the corpse was burned as well or got destroyed somehow else. Should we display a message as well about the lost credits? That would e.g. add a new mechanic to check for alive players with spending credits. But yea, that could break many things. So are we sure we just wanna let those credits get lost?

saibotk: I'd say to just let them be lost. So we do not introduce a new mechanic (which could also still be done by an add-on)

AaronMcKenney:
Hey this is a bit odd. I realized that I ought to re-test after running into a syntax bug during lint processing (whoops). However, upon doing so (with up-to-date code, as opposed to code from 2 months ago), the game now breaks after searching corpses?

This is the error that's thrown in the console:
```
[ttt2] addons/ttt2/lua/ttt2/extensions/util.lua:74: attempt to call upvalue 'sentsGetStored' (a nil value)
  1. WeaponForClass - addons/ttt2/lua/ttt2/extensions/util.lua:74
   2. unknown - addons/ttt2/gamemodes/terrortown/gamemode/client/cl_search.lua:158
    3. PreprocSearch - addons/ttt2/gamemodes/terrortown/gamemode/client/cl_search.lua:248
     4. ShowSearchScreen - addons/ttt2/gamemodes/terrortown/gamemode/client/cl_search.lua:486
      5. func - addons/ttt2/gamemodes/terrortown/gamemode/client/cl_search.lua:655
       6. unknown - lua/includes/extensions/net.lua:32
```

I don't think this is related to the changes brought on by code review though. Could it be related to the ClassBuilder commit from 11 days ago or so?

ZenBre4ker:
This is an error pointed out by zac on discord.

"Also I think its because of this line. Is this meant to be?
local sentsGetStored = scripted_ents.GetStored

instead of 
local sentsGetStored = sentsGetStored

https://github.com/TTT-2/TTT2/blob/ffcb50b3772ffab101d4d4fc65842ef4ac4e2517/lua/ttt2/extensions/util.lua#L23
"

TimGoll:
I will fix this one today after breakfast

saibotk:
@AaronMcKenney This is now fixed, but you will have to update the PR with the new language changes on the master. Sorry :/

Alf21 (approved):
Thanks for the improvements!

AaronMcKenney:
Ok, I have merged latest and re-tested with a friend. It is all good to go.

PR #681, opened 2020-10-17
C4: Removed defuse restriction for teammates

### LINKED ISSUE #679: Innocents can't defuse other innocents C4

Leave one of the following:
- Latest GitHub version

## Video
https://youtu.be/fCwWP37c8OI

When an innocent places a c4 other innocents can't defuse them
and the message that appears is wrong

## Expected behaviour

This should be intended only for Team Traitor(?)

Histalek:
This PR removes the restriction on teammates not being able to defuse a c4 if the owner is still alive.

The original Intention of this was most likely to combat trolling of teammates. But as I can think of legit reasons why you want to defuse a teammates c4 (e.g. save yourself from dying, save the c4 because no opponent would be killed by it) i think this is unjustified and should be removed.
This would also fix #679 and the possibility to confirm another players team.

(i'm not sure what happened with the japanese lang file, probably a different linebreak character?)


saibotk (approved):

Alf21 (approved):
Yea regarding japanese lang file, that happens every time, caused by different LF (EOL) i guess

Issue #653, opened 2020-09-26
Role-selection bug: Forced roles are not deducted from the available roles amount list


Alf21:
e.g. here https://github.com/TTT-2/TTT2/blob/cee9c123983d7ba9e666f1d666ce5fd987663a12/gamemodes/terrortown/gamemode/server/sv_roleselection.lua#L496, a decrease of the roles' amount is missing, as result, the function is not working properly (the roles' amount limit is exceeded)

ZenBre4ker:
To solve this issue you only need to decrease the roleCount of selectableRoles in the function SelectForcedRoles().

One idea would be to do it like this:
```lua
local function SelectForcedRoles(plys, selectableRoles)
	--[...]
	for subrole, forcedPlys in pairs(transformed) do
		--[...]
		for i = 1, amount do
                        --[...]
			if roleCount <= curCount then break end -- if the limit is reached, stop selection for this role
                        --[...]
		end
		
		--NEW CODE
		if roleCount <=curCount then
		        table.remove(selectableRoles, rd)
		else
			roleCount = roleCount-curCount
		        selectableRoles[rd]=roleCount
	        end
		--END
	end

	--NEW CODE
	roleSelection.selectableRoles = selectableRoles --Also update selectableRoles
        --END

	roleselection.forcedRoles = {}

	return selectedPlys
end 
```

This is not perfect as you could instead just decrement roleCount on the go instead of incrementing curCount and update it in the end, but it should be the solution to the problem and work fine with the rest of the code.

Don't know yet how Github works otherwise I'd try to commit it and make a pullrequest.

Alf21:
The problem is that modifying the original table could lead to some unexpected behavior (if you wanna use this list for later selections / in-round selections as well). That's why I thought about decreasing it "manually" (if calling a function).

Anyways, thanks for your input, always appreciated.
Yea feel free to try new things here, probably you/we can learn some new things. 👍

ZenBre4ker:
Okay, you are right and I just had to go further down the code to find a problem which could give a nil value.....

Well, how about copying selectableRoles and changing that to selectedForcedRoles, which counts the already assigned roles, so that it is only used later for the calculation of amount in SelectBaseRolePlayers()
My idea would look like this:
SelectForcedRoles only needs a third parameter and assigns at the end the curCount to the new parameter
```lua
local function SelectForcedRoles(plys, selectableRoles,selectedForcedRoles)
        --[...]
[...]

PR #685, opened 2020-10-20
Roleselection: Fixed forced roles not being subtracted from the roles amount

ZenBre4ker:
#653 By adding another table that keeps track of all the forced roles, it is possible to subtract them from the available roles amount and therefore keep a limit on the roles.

I also tested it with hardforcenr, which always keeps the limit until you hardforce more roles than available. It is also prior to the normal forcenr.

And ignore all the commits before 0ff6061, don't know how to delete history, because those were just me deleting stuff and updating to the master branch regularly.....


TimGoll:
probably something that @Alf21 should review.
However, just out of curiosity, this would allow a traitor pass addon, wouldn't it?

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Fixed
 
 - Fixed weapon pickup bug, where weapons would not get dropped but stayed in inventory
+- Fixed #653:Role-selection bug, where Forced roles are not deducted from the available roles amount list
```
Histalek, line 31:
```suggestion
- Fixed a roleselection bug, where forced roles would not be deducted from the available roles
```

Alf21 (changes requested):

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -672,7 +676,8 @@ function roleselection.SelectRoles(plys, maxPlys)
 	local selectableRoles = roleselection.GetSelectableRolesList(maxPlys, allAvailableRoles) -- update roleselection.selectableRoles table
 
 	-- Select forced roles at first
-	local selectedForcedPlys = SelectForcedRoles(plys, selectableRoles) -- this updates roleselection.finalRoles table and returns a key based list of selected players
+	local selectedForcedRoles = {}
+	local selectedForcedPlys = SelectForcedRoles(plys, selectableRoles, selectedForcedRoles) -- this updates roleselection.finalRoles table and returns a key based list of selected players
```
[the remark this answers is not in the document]
ZenBre4ker: This was a relict of the old idea. Changed it. This makes more sense.

Alf21:
Okay so the current state is working imo / I'd approve, but just for the base roles. Subroles still aren't working afaik / regarding your changes

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -592,19 +607,25 @@ end
...
+		if selectedForcedRoles[sub] then
+			roleAmount = roleAmount - selectedForcedRoles[sub]
+		end
+		if roles.GetByIndex(sub).baserole == subrole and roleAmount > 0 then
 			availableRoles[#availableRoles + 1] = sub
 		end
 	end
 
-	SetSubRoles(plys, availableRoles, selectableRoles)
+	SetSubRoles(plys, availableRoles, selectableRoles, selectedForcedRoles)
```
ZenBre4ker, line 628: Then I guess you like this more general approach better. Subroles are now also being deducted by forced Roles.
We trade efficiency vs. reusability. The selectedForcedRoles can be an empty list and the selectableRoles are no longer being affected by calculations inside these functions.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -578,11 +584,20 @@ local function SelectForcedRoles(plys, selectableRoles)
...
 			selectedPlys[ply] = true
 		end
+
+		-- assigning the amount of forced players per role
+		selectedForcedRoles[subrole] = curCount
+
+		-- now assign amount of forced players per baserole if this is only a subrole
+		local baserole = roles.GetByIndex(subrole).baserole
+		if baserole and baserole ~= subrole then
+			selectedForcedRoles[baserole] = curCount
```
Alf21, line 594: Here you overwriting the amount of baseroles by the amount of subroles

Alf21 (changes requested):

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: Oops, wonder why I didn't notice this while testing. Theoretically I should have gotten less baseroles, when forcing subroles.....

saibotk (approved):
Looks fine to me, thank you for taking your time and diving this deep into the code! This is much appreciated :heart:

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -498,6 +499,9 @@ local function SetSubRoles(plys, availableRoles, selectableRoles)
 		local subrole = availableRoles[rolePick]
 		local roleData = roles.GetByIndex(subrole)
 		local roleCount = tmpSelectableRoles[subrole]
+		if selectedForcedRoles[subrole] then
```
Alf21, line 502:
```suggestion

		if selectedForcedRoles[subrole] then
```

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -578,11 +584,20 @@ local function SelectForcedRoles(plys, selectableRoles)
 
 			selectedPlys[ply] = true
 		end
+
+		-- assigning the amount of forced players per role
+		selectedForcedRoles[subrole] = curCount
+
+		-- now assign amount of forced players per baserole if this is only a subrole
+		local baserole = roles.GetByIndex(subrole).baserole
+		if baserole and baserole ~= subrole then
```
Alf21, line 593:
```suggestion
		if not roles.GetByIndex(subrole):IsBaseRole() then
```
ref. https://github.com/TTT-2/TTT2/blob/953f060584d0ad6975b3a2bce4b85ae5ad894e11/lua/terrortown/entities/roles/ttt_role_base/shared.lua#L89

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -592,19 +607,25 @@ end
...
+	local roleAmount = 0
 
 	-- now upgrade this role if there are other subroles
 	for sub in pairs(selectableRoles) do
-		if roles.GetByIndex(sub).baserole == subrole then
+		roleAmount = selectableRoles[sub]
+		if selectedForcedRoles[sub] then
+			roleAmount = roleAmount - selectedForcedRoles[sub]
+		end
+		if roles.GetByIndex(sub).baserole == subrole and roleAmount > 0 then
```
Alf21, line 623:
```suggestion
		if not roles.GetByIndex(sub).baserole == subrole then continue end

		roleAmount = selectableRoles[sub]

		if selectedForcedRoles[sub] then
			roleAmount = roleAmount - selectedForcedRoles[sub]
		end

		if roleAmount > 0 then
```

Alf21 (changes requested):
Just some performance and code-style change requests

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: The function is nice, but is it faster? Because I still need the baserole index at line 594, though.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: agreed

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -556,7 +563,7 @@ local function SelectForcedRoles(plys, selectableRoles)
 		-- if it's not a selectable role, continue
 		if not roleCount then continue end
 
-		local curCount = 0
+		local curCount = (selectedForcedRoles[subrole] or 0)
```
ZenBre4ker, line 566: Didn't notice this before, but this makes the order of forcing roles indepedent of each other.
I seem to miss so much on this topic. There are all these small little depndecies, which are easy to miss.

Hopefully that's the last little bug.

ZenBre4ker: And it's not. Just found another issue, when forcing baseroles first to the maximum and then force subroles, they come additionally, even if their baserole is maxed out...... oof

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -556,7 +563,25 @@ local function SelectForcedRoles(plys, selectableRoles)
...
+			-- take the minimum of baseroles or subroles, if baseroles are available, else continue
+			if baseroleCount > 0 then
+				roleCount = math.min(baseroleCount, roleCount)
+			else
+				continue
+			end
+		end
+
+		-- Consider number of roles, that were forced by other subroles
+		local curCount = (selectedForcedRoles[subrole] or 0)
```
ZenBre4ker, line 584: Had to change this section here. Otherwise it was still possible to force more baseroles than available, depending on force order, when more subroles got chosen, than available baseroles.

This now always takes the minimum of baseroleCount and the subroleCount. As a maximum of available roles.
The currentCount just gets increased according to already forced subroles.

Seems to work, if not I probably have to come up with an automated testing, because the testing here is what takes most of the time. There are so many possibilities one doesn't know have an actually different behavior.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
Alf21: Not directly faster, but even not a big difference while it's better to read. I suggests storing `roles.GetByIndex(subrole)` in another var so you don't need to loop through every role again. And later, you can use `roleData:GetBaseRole()` to get the connected baserole while not accessing .baserole directly but using the recommended functions.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -563,6 +563,24 @@ local function SelectForcedRoles(plys, selectableRoles)
 		-- if it's not a selectable role, continue
 		if not roleCount then continue end
 
+		local isBaseRole = roles.GetByIndex(subrole):IsBaseRole()
```
Alf21, line 566: `roles.GetByIndex(subrole)` should be stored

Alf21 (changes requested):

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: Generally in the complete roleselection or only in this function, because its only used twice in SelectForcedRoles?

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -526,17 +531,40 @@ local function SetSubRoles(plys, availableRoles, selectableRoles)
...
+
+			-- assign amount of forced players per baserole if this is only a subrole
+			if not isBaseRole then
+				local baserole = curRole.baserole
+				selectedForcedRoles[baserole] = (selectedForcedRoles[baserole] or 0) + 1
+			end
+		end
+	end
+	return selectedForcedRoles
+end
```
ZenBre4ker, line 556: This now puts baseroles of hardforced subroles into the selectedForcedRoles-List.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -672,15 +735,16 @@ function roleselection.SelectRoles(plys, maxPlys)
...
+	SelectForcedRoles(plys, selectableRoles, selectedForcedRoles) -- this updates roleselection.finalRoles table and the list of forced Roles
 
 	-- We need to remove already selected players
 	local plysFirstPass, plysSecondPass = {}, {} -- modified player table
 
 	for i = 1, #plys do
 		local ply = plys[i]
 
-		if selectedForcedPlys[ply] then continue end
+		if roleselection.finalRoles[ply] then continue end
```
ZenBre4ker, line 747: This seemed kinda strange to remove already selected Players, but only check for selectedForcedPlys, therefore I changed that to the finalRoles List, which should contain both.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -575,14 +622,20 @@ local function SelectForcedRoles(plys, selectableRoles)
...
+		-- now assign amount of forced players per baserole if this is only a subrole
+		if not isBaseRole then
+			selectedForcedRoles[baserole] = (selectedForcedRoles[baserole] or 0) + curCount
 		end
 	end
 
 	roleselection.forcedRoles = {}
 
-	return selectedPlys
+	return selectedForcedRoles
```
ZenBre4ker, line 638: Need to delete that return statement as it's unused.

Histalek (approved):
LGTM

Thank you for the contribution!

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -526,17 +531,40 @@ local function SetSubRoles(plys, availableRoles, selectableRoles)
...
+			local isBaseRole = curRole:IsBaseRole()
+
+			-- assign amount of forced players per baserole if this is only a subrole
+			if not isBaseRole then
+				local baserole = curRole.baserole
+				selectedForcedRoles[baserole] = (selectedForcedRoles[baserole] or 0) + 1
+			end
+		end
+	end
+	return selectedForcedRoles
```
Alf21, line 555:
```suggestion

	return selectedForcedRoles
```

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -526,17 +531,40 @@ local function SetSubRoles(plys, availableRoles, selectableRoles)
...
+-- Ensure, that baseroles of hardforced subroles are accounted for (eg. received by ULX etc).
+--
+-- @return table List with a count of forced @{ROLE}s
+-- @realm server
+-- @internal
+local function GetHardForcedBaseRoles()
+	local selectedForcedRoles = {}
+
+	for ply, subrole in pairs(roleselection.finalRoles) do
+		if IsValid(ply) then
```
Alf21, line 544: That's not necessary

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -556,7 +584,26 @@ local function SelectForcedRoles(plys, selectableRoles)
 		-- if it's not a selectable role, continue
 		if not roleCount then continue end
 
-		local curCount = 0
+		local curRole = roles.GetByIndex(subrole)
+		local isBaseRole = curRole:IsBaseRole()
+		local baserole = nil
+
+		--Consider maximum number of roles, that are available to the corresponding baserole
```
Alf21, line 591:
```suggestion
		-- Consider maximum number of roles, that are available to the corresponding baserole
```

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -556,7 +584,26 @@ local function SelectForcedRoles(plys, selectableRoles)
...
+			baserole = curRole.baserole
+
+			local baseroleCount = selectableRoles[baserole] - (selectedForcedRoles[baserole] or 0)
+
+			-- take the minimum of baseroles or subroles, if baseroles are available, else continue
+			if baseroleCount > 0 then
+				roleCount = math.min(baseroleCount, roleCount)
+			else
+				continue
+			end
```
Alf21, line 602:
```suggestion
			if baseroleCount < 1 then continue end

			roleCount = math.min(baseroleCount, roleCount)
```

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -556,7 +584,26 @@ local function SelectForcedRoles(plys, selectableRoles)
...
+		end
+
+		-- Consider number of roles, that were forced by other subroles
+		local curCount = (selectedForcedRoles[subrole] or 0)
```
Alf21, line 606:
```suggestion
		local curCount = selectedForcedRoles[subrole] or 0
```

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
Alf21: So you don't need this var localized here anymore? https://github.com/TTT-2/TTT2/pull/685/files#diff-d2698bbe621f7660fa9806073e9a4fe3e10977cb1c48a345babc478d6f3e44a9R736

Alf21 (changes requested):

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: Well I not for the here marked part, but for this input https://github.com/TTT-2/TTT2/pull/685/files#diff-d2698bbe621f7660fa9806073e9a4fe3e10977cb1c48a345babc478d6f3e44a9R737

Or should it be put into the function itself or instead of a parameter use the function.
Depends on the style you like more. I'd prefer it localized there to make it clear what its used for.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: What happens if you hardforce a player and then he leaves the game? Thought this would be safer or is it handled somewhere else?

Alf21: Yea that's something I forget, the list have to get cleaned up before the roleselection begins :)

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
Alf21: I think it would match the best directly inside the SelectForcedRoles fn

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: Okay, added a cleanup before forcing players.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -712,11 +788,17 @@ function roleselection.SelectRoles(plys, maxPlys)
 
 			if not roleData:IsBaseRole() or not selectableRoles[subrole] then continue end
 
-			local baseRolePlys = SelectBaseRolePlayers(plysFirstPass, subrole, selectableRoles[subrole])
+			local amount = selectableRoles[subrole]
+
+			if selectedForcedRoles[subrole] then
+				amount = amount - selectedForcedRoles[subrole]
+			end
```
ZenBre4ker, line 795: Here I actually need the localized variable.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -671,16 +745,18 @@ function roleselection.SelectRoles(plys, maxPlys)
 	local allAvailableRoles = roleselection.GetAllSelectableRolesList(maxPlys)
 	local selectableRoles = roleselection.GetSelectableRolesList(maxPlys, allAvailableRoles) -- update roleselection.selectableRoles table
 
+	updateFinalRoles() -- Update the roleselection.finalRoles table by removing all invalid players
```
Alf21, line 748:
```suggestion
	UpdateFinalRoles() -- Update the roleselection.finalRoles table by removing all invalid players
```

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -592,19 +643,42 @@ end
...
+-- Update the roleselection.finalRoles table by removing all invalid players
+--
+-- @realm server
+-- @internal
+local function updateFinalRoles()
+	for ply, subrole in pairs(roleselection.finalRoles) do
+		if not IsValid(ply) then
+			table.remove(roleselection.finalRoles, ply)
+		end
+	end
```
Alf21, line 681:
```suggestion
local function UpdateFinalRoles()
	local tbl = {}

	for ply, subrole in pairs(roleselection.finalRoles) do
		if IsValid(ply) then
			tbl[ply] = subrole
		end
	end

	roleselection.finalRoles = tbl
```

Alf21: Iterating over the whole list and resizing it every time wouldn't just take a lot of time but could break the pairs iteration too. Creating a valid list instead and then reassigning it should be the way to go.

Alf21 (changes requested):

saibotk (approved):

TimGoll:
you should update your branch.

@Alf21 it seems like he has done everything you requested

ZenBre4ker:
Yeah he was fine with the changes. Just need to update and properly test it again.
Didn't find time yet to test all forces again.
@TimGoll

ZenBre4ker:
Okay, finally got around to test it again. The list really helped quickly iterating over many possible scenarios. So everything I could think of, woks now.

Alf21 (approved):

Alf21:
Thanks for your hard work and the in-depth investigations! ❤️

PR #688, opened 2020-10-21
Roleselection: Alternative Solution: Fixed forced roles not being subtracted from the roles amount

ZenBre4ker:
#653 Fixed the roles not being deducted by subtracting forced roles directly from the selectableRoles table, which is now only a copy of the global one, so that it can still be used later.
Also if a subrole is forced the baserole gets deducted as well, otherwise it was possible to have more traitors, by forcing a subrole.

In general I tested this with various use cases and right now it works with the tested hitman and detective.
This is a new way compared to #685


gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -473,8 +473,9 @@ function roleselection.GetSelectableRolesList(maxPlys, rolesAmountList)
 	hook.Run("TTT2ModifySelectableRoles", selectableRoles)
 
 	roleselection.selectableRoles = selectableRoles
+	local tmpSelectableRoles = table.Copy(selectableRoles)
 
-	return selectableRoles
+	return tmpSelectableRoles
```
Alf21, line 478: Can be combined

Alf21: Then you've to edit the first lines of this func as well i guess

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -578,6 +579,23 @@ local function SelectForcedRoles(plys, selectableRoles)
...
+
+		-- now decrease amount of the baserole if this is only a subrole
+		local baserole = roles.GetByIndex(subrole).baserole
+		if baserole and baserole ~= subrole then
+			roleAmount = selectableRoles[baserole] - curCount
+			selectableRoles[baserole] = roleAmount
+			if roleAmount < 1 then
+				selectableRoles[baserole] = nil
+			end
+		end
```
ZenBre4ker, line 598: Roles get removed from the list now, if they reach zero.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -497,7 +497,7 @@ local function SetSubRoles(plys, availableRoles, selectableRoles)
 		local rolePick = math.random(availableRolesAmount)
 		local subrole = availableRoles[rolePick]
 		local roleData = roles.GetByIndex(subrole)
-		local roleCount = tmpSelectableRoles[subrole]
+		local roleCount = selectableRoles[subrole]
```
Alf21, line 500: so you editing the roles list directly? That's the opposite we talked about. ^^

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -599,7 +609,7 @@ local function UpgradeRoles(plys, subrole, selectableRoles)
 
 	-- now upgrade this role if there are other subroles
 	for sub in pairs(selectableRoles) do
-		if roles.GetByIndex(sub).baserole == subrole then
+		if selectableRoles[sub] >= 1 and roles.GetByIndex(sub).baserole == subrole then
```
Alf21: Empty roles shoudln't be in here

Alf21 (changes requested):

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: This is outdated. Already changed that with the latest update. It's back to 
if roles.GetByIndex(sub).baserole == subrole then
as roles now get removed if they are empty

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: Yeah I know, but it works so smoothly with the local list..... It's direct and can be easily edited.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: Oh yeah, didnt see that, changed it now.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: And what your marked there is just a performance increase. If needed this could stay as a copied table. Only the force needs to change it, the rest works lile this too.

ZenBre4ker:
Then I will close this here.

PR #689, opened 2020-10-25
Update japanese.lua

westooooo:
updated Japanese translation.


TimGoll (approved):

Histalek (approved):
Looks good to me. I'd say we don't need a changelog for this, do we?

PR #692, opened 2020-11-02
Classbuilder: Added a new classbuilder module

TimGoll:
This pullrequest introduces a module: a class builder. This classbuilder aims to remove the duplicated code in this repository by unifying it to a simle builder. I also migrated the newly introduced event system to this module.

The reason why I added this: While creating the roundendscreen it occured to me to use this class system to create menues instead of the existing hooks. To prevent yet another code duplication, I created this new module.


lua/ttt2/extensions/table.lua
```diff
@@ -286,3 +286,28 @@ function table.ExtractRandomEntry(tbl, filterFn)
...
+function table.DeepInherit(t, base)
+	if not base then
+		return t
+	end
+
+	for k, v in pairs(base) do
+		if t[k] == nil then
+			t[k] = v
+		elseif k ~= "BaseClass" and istable(t[k]) then
+			table.Inherit(t[k], v, SpecialCheck)
```
Alf21, line 306: `SpecialCheck`?

lua/ttt2/libraries/classbuilder.lua
```diff
@@ -0,0 +1,82 @@
...
+	-- Now a global scope has to be created.
+	_G[scope] = {}
+
+	-- This variable will be used to store the loaded classes from the folder.
+	local classTable = {}
+
+	fileloader.LoadFolder(path, false, realm, function(filePath)
+		-- get the filename from the file path
+		local pathArray = stringSplit(filePath, "/")
+		local name = stringLower(stringSub(pathArray[#pathArray], 0, -5))
```
Alf21, line 53: Wouldn't run a `.lua` check and `.` split better?

lua/ttt2/libraries/classbuilder.lua
```diff
@@ -0,0 +1,82 @@
...
+
+		-- copy the table from the loaded class file
+		classTable[name] = tableCopy(_G[scope])
```
Alf21, line 56: I think it would be better not copying the scope regarding checks if using classTable later. Instead assigning the real table should work

lua/ttt2/libraries/classbuilder.lua
```diff
@@ -0,0 +1,82 @@
...
+	end)
+
+	-- In the last step, the cached scope should be reset to the global scope.
+	_G[scope] = cachedScope
+
+	-- Now that all classes are loaded, they should inherit from their base
+	-- class if enabled.
+	if shouldInherit then
+		for name, class in pairs(classTable) do
+			local base = classTable[class.base]
```
Alf21, line 73: Coudldn't it be possible that `.base` is `nil`?

lua/ttt2/libraries/classbuilder.lua
```diff
@@ -0,0 +1,82 @@
...
+
+			if isfunction(SpecialCheck) and not SpecialCheck(class, base) then continue end
+
+			classTable[name] = tableDeepInherit(class, base)
```
Alf21, line 77: Does a table change ofa a table in a loop of the same table work (I don't know how pairs works internally regarding this)? Tested?

lua/ttt2/libraries/events.lua
```diff
@@ -358,32 +279,38 @@ else --CLIENT
...
 
 		MsgN("Added TTT2 event file: ", path)
 	end,
-	nil,
-	function() -- post folder loaded
-		events.DeinitVariables()
+	true, -- should inherit
+	function(t, base) -- special inheritance check
+		return t.base ~= t.type
 	end
```
Alf21, line 315: Those both functions can be localized regarding duplication above

Alf21 (changes requested):

lua/ttt2/extensions/table.lua
TimGoll: artifact, removed

lua/ttt2/libraries/classbuilder.lua
TimGoll: I thought about this as well. But you/ttt used this method in the previuos functions. using a dot to split might cause problems with `file.bla.lua`

Alf21: Agreed, BUT anyways, that's the file's name in that case: file.bla. So you just have to merge every split except the last one.

lua/ttt2/libraries/classbuilder.lua
TimGoll: I wasn't sure either. However I used my code to test and it worked. So I guess it works, yes

lua/ttt2/libraries/classbuilder.lua
TimGoll: Are you sure? Because we copied the table everywhere in the code

lua/ttt2/libraries/classbuilder.lua
TimGoll: Yes. But it would be worse performancewise

lua/ttt2/libraries/classbuilder.lua
Alf21: Yea. The reference is deleted later, so that should work - theoretically. For sure, we have to test, but using a table comparation inside the EVENT with a loop through other events would fail - because a copied table has another memory address than the real table (as you are aware for sure)

lua/ttt2/libraries/classbuilder.lua
Alf21: Those files are loaded on startup, so that doesn't really matter. Additionally, files with an ending <> 3 chars would break this while the "engine" tries to load these files. That's just a stability thing tbh.

lua/ttt2/libraries/classbuilder.lua
Alf21: Now `base` could be `nil` if providing a wrong `.base` ^^

lua/ttt2/libraries/events.lua
```diff
@@ -358,32 +279,38 @@ else --CLIENT
...
-		events.Initialize(path)
+	"EVENT", -- class scope
+	function(class, path, name) -- on class loaded
+		class.type = name
+		class.base = class.base or "base_event"
+
+		_G["EVENT_" .. string.upper(name)] = name
 
 		MsgN("Added TTT2 event file: ", path)
 	end,
```
Alf21, line 297: That function is repeating as well, maybe add an option to accept a table of multiple paths into `classbuilder.BuildFromFolder`?

Alf21 (changes requested):

lua/ttt2/libraries/events.lua
TimGoll: I thought about removing (in a future PR) all duplicated paths. Right now we have always a `shared/xyz` and a `lua/terrortown/xyz` folder. I'd prefer to only keep the latter.

lua/ttt2/libraries/classbuilder.lua
TimGoll: `table.DeepInherit` returns the table if the base is nil

lua/ttt2/libraries/events.lua
Alf21: Then probably localizing one of these two exact function calls while accepting the path as an arg and calling this function manually two times with the two paths

Histalek (approved):
LGTM

Alf21 (approved):

lua/ttt2/libraries/events.lua
```diff
@@ -358,32 +279,38 @@ else --CLIENT
...
 
-fileloader.LoadFolder(
+table.Merge(eventTypes, classbuilder.BuildFromFolder(
 	"terrortown/events/",
-	false,
 	SHARED_FILE,
-	function(path)
-		events.Initialize(path)
+	"EVENT", -- class scope
+	function(class, path, name) -- on class loaded
```
saibotk, line 306: Can this function be localized and then referenced twice? Instead of two functions with the same content.

saibotk (changes requested):

lua/ttt2/libraries/events.lua
Alf21: https://github.com/TTT-2/TTT2/pull/692#discussion_r520749256

Histalek: small reminder to check out the linked comment and possibly update your review :)

saibotk: I did see that, but my comment still holds.

TimGoll:
As discussed with @saibotk yesterday, I'll merge the PR after localizing the function.

saibotk (approved):

PR #694, opened 2020-11-03
Lang: Cleanup and Menu Additions

TimGoll:
I cleaned up the language file further and removed all duplicates that were found by henk on discord.

Additionally I renamed the language indentifiers from their original localized version to an english identifier. This fixes the linter problem seen with languages like russian in classes and minigames. I added a new translated language identifier to the language table such that the language identifier in the F1 menu is still translated.

Additionally I added a small info about the language coverage to motivate users to contribute to the translations:
image


TimGoll:
@helaan You talked about changing from "german" to the default abrevations like "de_de". This PR could be the perfect situation to change it. I'd appreciate your input on this.

saibotk:
> @helaan You talked about changing from "german" to the default abrevations like "de_de". This PR could be the perfect situation to change it. I'd appreciate your input on this.

I think this would be fine and a needed change.
Is there anything, that might break? Or rather, what would have to be changed exactly?

TimGoll:
> > @helaan You talked about changing from "german" to the default abrevations like "de_de". This PR could be the perfect situation to change it. I'd appreciate your input on this.
> 
> I think this would be fine and a needed change.
> Is there anything, that might break? Or rather, what would have to be changed exactly?

Everything that would break by this change would breake by the changes in this PR anyway. We have to update the language identifiers in our addons. But I think that is ok.
I can revert this change if you prefer. However I came to dislike the fact that we used language identifiers that did not only use `aA-zZ`.

saibotk:
Yes, i really dislike this too, and welcome the change. But couldn't we add an alias (also defined in the language file), to help users migrate to the new language system? Something like L._alias = "german" and then the language table will also contain the same table reference with the old key to provide backwards compatibility. (Maybe even throw out a log message about deprecated language identifiers)

TimGoll:
Albeit not nice, this would work, yes. Give me a minute.

TimGoll:
done

helaan:
What I'm running into with Weblate is that Weblate expects an [ISO 639-1 language identifier](https://docs.weblate.org/en/latest/admin/languages.html#language-code) for its input files: you can make it work without it and map them yourself, but I'd rather not.

I'd suggest the following on top of the current changes: rename the files to its ISO 639-1 code (en, de, these codes are pretty simple to find, with the exception of Simplified Chinese, which is zh_Hans) which is a mostly* neutral identifier, which would make it significantly easier for me to map between Weblate files and Lua files. [I've already created a mapping internally though](https://github.com/helaan/tttranslate/commit/04f9c4bd6650115923ae099b0e62c6a6af6bee73), so although this change would help me kill some code and save me from writing a few lines in integration scripts, it is not strictly necessary. It could be useful for nesting translations though, for example if we ever want to support both American and British English, but we need more code on the TTT2 side for that.

I'd also prefer that we change the internal identifier to the ISO 639-1 identifier as well, but that would break compatibility with TTT. This means that this PR would break compatibility with TTT, forcing addons to either translate for TTT or for TTT2, which I think is undesireable. **Therefore, please revert the identifier change**

Have you asked the translators about which of the duplicate strings fit best?

I like the entry where translators can specify how they want their language to be identified in the menu, this makes the identifier legacy data that doesn't need to be translated and making it easier to change this name in future if needed.

* https://en.wikipedia.org/wiki/ISO_639-3#Criticism although note that we are not using ISO 639-3, just ISO 639-1

TimGoll:
@helaan Thanks to @saibotk s suggestion, the new system is compatible with the old system while also having new internal identifiers.

In all but one cases the duplication came by a deprecation. Therefore it was easy to see which one was the correct one.

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -65,7 +65,7 @@ end
 -- @return string the inserted string_name parameter
 -- @realm client
 function LANG.AddToLanguage(lang_name, string_name, string_text)
-	lang_name = lang_name and string.lower(lang_name) or nil
+	lang_name = lang_name and LANG.GetNameFromAlias(string.lower(lang_name)) or nil
```
Histalek: would it make sense to move the `string.lower` into the function itself?

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -177,12 +177,33 @@ function LANG.GetUnsafeLanguageTable()
 end
 
 ---
--- Returns the reference to a language table if it exists
--- @param string name string key identifier for the language
+-- Caused by a recent rename of the language names, old addons loos compatibility.
+-- To prevent annoyance, a compatibiliy mode is introduced.
+-- @param string lang_name string key identifier for the language
```
Histalek: please capitalize the first word of the description

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -177,12 +177,33 @@ function LANG.GetUnsafeLanguageTable()
...
+			MsgN("[DEPRECATION WARNING]: Language name identifier deprecated, please switch from '" .. lang_name .. "' to '" .. name .. "'.")
+
+			return name
+		end
+	end
+end
+
+---
+-- Returns the reference to a language table if it exists.
+-- @param string lang_name string key identifier for the language
```
Histalek: capitalize as well please

helaan:
It is compatible, but it still calls the old method "deprecated", which I don't think is fair towards addon devs who try to support both.

If you do want to change the identifiers though, I suggest you adopt ISO 639-1 codes as identifiers, which are more neutral than just their english name and also allows stacking languages in future as described in the previous comment.

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -194,12 +215,12 @@ function LANG.GetUnsafeNamed(lang_name)
 end
 
 ---
--- Returns the reference to a language table if it exists, creates a new language if it did not exist
--- @param string name string key identifier for the language
+-- Returns the reference to a language table if it exists, creates a new language if it did not exist.
+-- @param string lang_name string key identifier for the language
```
Histalek: capitalize

Histalek (changes requested):
small doc nitpick and one question/suggestion

TimGoll:
Hmm. Fair point. However it only prints a message in the chat once, it isn't declared as an error. And most addons that use localization are created for TTT2 anyway. I haven't seen many addons in the workshop that aren't hardcoded.

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -162,59 +173,84 @@ end
...
 function LANG.GetUnsafeLanguageTable()
-	return cached_active
+	return cachedActive
+end
+
+---
+-- Caused by a recent rename of the language names, old addons loos compatibility.
+-- To prevent annoyance, a compatibiliy mode is introduced.
+-- @param string langName string key The identifier for the language
+-- @return string The lang name in the new format
```
Histalek:
```suggestion
---
-- Caused by a recent rename of the language names, old addons lost compatibility.
-- To prevent annoyance, a compatibility mode is introduced.
-- @param string langName The identifier for the language
-- @return string The language name in the new format
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -162,59 +173,84 @@ end
...
+	end
+
+	return langName
 end
 
 ---
--- Returns the reference to a language table if it exists
--- @param string name string key identifier for the language
+-- Returns the reference to a language table if it exists.
+-- @param string langName string key The identifier for the language
```
Histalek:
```suggestion
-- @param string langName The identifier for the language
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -162,59 +173,84 @@ end
...
 
-	return LANG.Strings[lang_name]
+	return LANG.Strings[langName]
 end
 
 ---
--- Returns the reference to a language table if it exists, creates a new language if it did not exist
--- @param string name string key identifier for the language
+-- Returns the reference to a language table if it exists, creates a new language if it did not exist.
+-- @param string langName string key The identifier for the language
```
Histalek:
```suggestion
-- @param string langName The identifier for the language
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -162,59 +173,84 @@ end
...
-	return LANG.Strings[lang_name]
+	return LANG.Strings[langName]
 end
 
 ---
--- Returns a copy of a selected language table
+-- Returns a copy of a selected language table.
 -- @note Safe and slow access, not sure if it's ever useful.
--- @param string lang_name the language name
+-- @param string langName the language name
```
Histalek, line 239:
```suggestion
-- @param string langName The language name
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -229,37 +265,37 @@ function SetFallback(tbl)
...
 	setmetatable(tbl, {
-		__index = cached_default
+		__index = cachedDefault
 	})
 end
 
 ---
 -- Switches the active language
--- @param string lang_name the new language name
+-- @param string langName the new language name
```
Histalek, line 274:
```suggestion
-- @param string langName The new language name
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -270,38 +306,38 @@ end
...
 	end
 
-	LANG.SetActiveLanguage(lang_name)
+	LANG.SetActiveLanguage(langName)
 end
 
 ---
 -- Returns whether the given language name is the default server language
--- @param string lang_name the language name
+-- @param string langName the language name
```
Histalek, line 322:
```suggestion
-- @param string langName The language name
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -270,38 +306,38 @@ end
...
+	langName = string.lower(langName)
 
-	return lang_name == "server default" or lang_name == "auto"
+	return langName == "server default" or langName == "auto"
 end
 
 ---
 -- Returns whether the given language is a valid language (already exists)
--- @param string lang_name the language name
+-- @param string langName the language name
```
Histalek, line 333:
```suggestion
-- @param string langName The language name
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -270,38 +306,38 @@ end
...
 -- @return table the language table, same as @{LANG.GetUnsafeNamed}, but without
```
Histalek, line 334:
```suggestion
-- @return table The language table, same as @{LANG.GetUnsafeNamed}, but without
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -454,6 +490,31 @@ function LANG.ProcessMsg(name, params, mode)
 	LANG.ShowStyledMsg(text, LANG.GetStyle(name, mode))
 end
 
+---
+-- Returns the percentage of the coverage of the given language compared to the
+-- default language (english).
+-- @param string langName The name of the language is question
+-- @return number The percentage of the coverage
+-- @realm client
```
Histalek, line 498:
```suggestion
-- Returns the coverage in percent of the given language compared to the
-- default language (english).
-- @param string langName The name of the language in question
-- @return number The coverage in percent
-- @realm client
```

Histalek (changes requested):
please ignore the those 3 suggestions you just fixed, github wont show them to me anymore .-.

Histalek (approved):

saibotk (approved):

helaan:
Would it also be possible to rename the localization files from `english.lua` to `en.lua`? This completely removes internal usage of the old names

TimGoll:
> Would it also be possible to rename the localization files from `english.lua` to `en.lua`? This completely removes internal usage of the old names

The names do not matter for our prupose. So yes, this is no problem!

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -148,11 +148,22 @@ LANG.GetPTranslation = LANG.GetParamTranslation
...
--- @param string lang_name the language name
+-- @param string langName The language name
 -- @return nil|string
 -- @realm client
-function LANG.GetTranslationFromLanguage(name, lang_name)
-	return rawget(LANG.Strings[lang_name], name)
+function LANG.GetTranslationFromLanguage(name, langName)
+	langName = langName and string.lower(langName) or nil
+
+	return rawget(LANG.Strings[langName], name)
```
Alf21, line 157:
```suggestion
	if langName == nil then return end

	return rawget(LANG.Strings[string.lower(langName)], name)
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -164,59 +175,84 @@ end
...
+
+---
+-- Caused by a recent rename of the language names, old addons lost compatibility.
+-- To prevent annoyance, a compatibility mode is introduced.
+-- @param string langName The string key identifier of the language
+-- @return string The language name in the new format
+-- @internal
+-- @realm client
+function LANG.GetNameFromAlias(langName)
+	langName = langName and string.lower(langName) or nil
```
Alf21, line 189:
```suggestion
	if langName == nil then return end

	langName = string.lower(langName)
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -164,59 +175,84 @@ end
...
 function LANG.GetUnsafeLanguageTable()
-	return cached_active
+	return cachedActive
+end
+
+---
+-- Caused by a recent rename of the language names, old addons lost compatibility.
+-- To prevent annoyance, a compatibility mode is introduced.
+-- @param string langName The string key identifier of the language
+-- @return string The language name in the new format
```
Alf21, line 185:
```suggestion
-- @return string|nil The language name in the new format
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -164,59 +175,84 @@ end
...
--- Returns the reference to a language table if it exists
--- @param string name string key identifier for the language
+-- Returns the reference to a language table if it exists.
+-- @param string langName The string key identifier of the language
 -- @return nil|table
 -- @realm client
-function LANG.GetUnsafeNamed(lang_name)
-	lang_name = lang_name and string.lower(lang_name) or nil
+function LANG.GetUnsafeNamed(langName)
+	langName = LANG.GetNameFromAlias(langName)
```
Alf21, line 212: What if `langName` is `nil`? Same below

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -231,39 +267,39 @@ function LANG.SetFallback(tbl)
...
 
 ---
 -- Switches the active language
--- @param string lang_name the new language name
+-- @param string langName The new language name
 -- @realm client
-function LANG.SetActiveLanguage(lang_name)
-	lang_name = lang_name and string.lower(lang_name) or nil
+function LANG.SetActiveLanguage(langName)
+	langName = langName and string.lower(langName) or nil
```
Alf21, line 279:
```suggestion
	if langName == nil then return end

	langName = string.lower(langName)
```

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -460,6 +496,31 @@ function LANG.ProcessMsg(name, params, mode)
...
+-- @realm client
+function LANG.GetDefaultCoverage(langName)
+	-- if language is set to auto, get server default
+	if langName == "auto" then
+		langName = LANG.ServerLanguage
+	end
+
+	local englishTbl = LANG.Strings[LANG.DefaultLanguage]
+
+	return table.GetEqualEntriesAmount(LANG.Strings[langName], englishTbl) / table.Count(englishTbl)
```
Alf21, line 513: No rounding regarding floats / decimals?

gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
```diff
@@ -664,14 +665,14 @@ function SKIN:PaintHelpLabelTTT2(panel, w, h)
...
-	local textTranslated = TryT(panel:GetText())
+	local textTranslated = ParT(panel:GetText(), panel:GetParams())
 	local textWrapped = drawGetWrappedText(
 		textTranslated,
 		w - 2 * panel.paddingX,
 		panel:GetFont()
 	)
 
-	local _, heightText = drawGetTextSize(textTranslated, panel:GetFont())
+	local _, heightText = drawGetTextSize("", panel:GetFont())
```
Alf21, line 675: Explanation? Probably as a in-code comment

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
```diff
@@ -268,9 +268,13 @@ function PANEL:MakeComboBox(data)
 			RunConsoleCommand(slf.m_strConVar, tostring(rawdata or value))
 		end
 
-		if isfunction(data.OnChange) then
-			data.OnChange(slf, index, value, rawdata)
-		end
+		-- run the callback function in the next frame since it takes
+		-- one frame to update the convar if one is set.
```
Alf21, line 272: oof that doesn't seem to be really reliable...

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
```diff
@@ -385,14 +390,14 @@ function PANEL:MakeHelp(data)
...
-		local textTranslated = LANG.TryTranslation(slf:GetText())
+		local textTranslated = LANG.GetParamTranslation(slf:GetText(), slf:GetParams())
 
 		local textWrapped = draw.GetWrappedText(
 			textTranslated,
 			w - 2 * slf.paddingX,
 			slf:GetFont()
 		)
-		local _, heightText = draw.GetTextSize(textTranslated, slf:GetFont())
+		local _, heightText = draw.GetTextSize("", slf:GetFont())
```
Alf21, line 400: Same as some comments above

Alf21 (changes requested):
Just some `nil` inconsistencies. Additionally, I've to ask why we aren't using `de_de`, but `de` without regional support?

gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
TimGoll: whyt should be explained here?

Alf21: Why using an empty string?

TimGoll: Because it doesn't change anything (only the height is relevant)
But in another part of the code I had to use an empty string because and the time of the calculation, the text is not yet known. I therefore changed it here as well to keep it consistent.

Alf21: Theoretically yes, but the width differs as well regarding e.g. `w` and `i`. So if you are sure, I gonna close this.

TimGoll: As I'm only using the height, this does not matter. I veriefied it for the texts used in the vgui rework and they all looked good to me.

saibotk (approved):
Can be merged, as soon as @Alf21's comments are resolved

gamemodes/terrortown/gamemode/client/cl_lang.lua
TimGoll: `LANG.GetNameFromAlias(langName)` returns the provided lang name if no alias was found. Therefore returns `nil` if `nil` was provided

gamemodes/terrortown/gamemode/client/cl_lang.lua
TimGoll: it is rounded in the UI. I thought it is ok if this function privides maximal accuracy and the user can decide if and how he wants to round it.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
TimGoll: Worked for me. Convars are updated in the next frame as it seems. There are many things that are handled like this in Gmod. So I don't see a problem with this.

Alf21 (approved):

PR #701, opened 2020-12-19
TargetID-Fix: Initializing convars later

ZenBre4ker:
Now doesnt depend on earlier convar creation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
   - added a hook `TTT2AddedEvent` that is called after an event was added
 - Added `orm` library to simplify database access
 - Added French translation (by @MisterClems)
+- Added a `targetid` library, that can be used to draw TargetIDs for entities
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


TimGoll:
To be honest, I prefer the solution with the convar loader being moved infront of the library loader. This would solve this issue as well. DIdnÄt be discuss this yesterday and decided it was the best solution?

ZenBre4ker:
I actually thought we discussed, that we shouldn't Initialize that at all, because moving libraries shouldn't be a solution. Or in other words, libraries shouldn't depend on each other.
So this is a hotfix until I find more time to properly change the library.

TimGoll:
Yes we discussed to either move the convars to the start or make this library completely independant of thos things.

But I see, this should only be a hotfix and will be later improved?

ZenBre4ker:
Yeah kind of. I didn't really like the option to just move libraries. And so this is still a bad approach. Only problem I see here, is that alf doesn't like TODOs, but this is one.....

saibotk (approved):

Histalek (approved):
Please fix the Changelog file, it seems to be

Alf21 (approved):

PR #704, opened 2021-01-01
Lang: Fix variable name in SetActiveLanguage method

saibotk:
This patch fixes the variable name cached_active to the now renamed cachedActive.


TimGoll (approved):
Thanks! Seems like a merging error that slipped me.

Histalek (approved):

PR #710, opened 2021-01-08
Language: English localization changes and improvements

Satton2:


gamemodes/terrortown/gamemode/shared/lang/en.lua
```diff
@@ -1119,19 +1119,19 @@ L.submenu_appearance_interface_title = "Interface"
...
 L.help_hud_special_settings = "These are specific settings of this HUD."
-L.help_vskin_info = "VSkin (vgui skin) is the skin applied to all menu elements like the current one. Skins can be easily created with a simple lua script and can change colors and some size parameters."
-L.help_targetid_info = "TargetID is the information rendered when focusing an entity. A fixed color can be set in the general settings panel."
-L.help_hud_default_desc = "Set the default HUD for all players. Players that have not yet selected a HUD will receive this HUD as their default. This won't change the HUD of players that have already selected their HUD."
-L.help_hud_forced_desc = "Force a HUD for all players. This disables the HUD selection feature for everyone."
+L.help_vskin_info = "VSkin (VGUI skin) is the skin applied to all menu elements like the current one. They can be easily created with a simple Lua script and can change colors and some size parameters."
+L.help_targetid_info = "TargetID is the information rendered when pointing your crosshair on an entity. Its color can be configured in the 'General' tab."
```
Histalek, line 1132:
```suggestion
L.help_targetid_info = "TargetID is the information rendered when pointing your crosshair at an entity. Its color can be configured in the 'General' tab."
```

gamemodes/terrortown/gamemode/shared/lang/en.lua
```diff
@@ -1237,17 +1237,17 @@ L.item_no_fall_damage_desc = [[Makes you immune to fall damage.]]
 L.item_no_fire_damage = "No Fire Damage"
 L.item_no_fire_damage_desc = [[Makes you immune to fire damage.]]
 L.item_no_hazard_damage = "No Hazard Damage"
-L.item_no_hazard_damage_desc = [[Makes you immune to hazard damage such as posion, radiation and acid.]]
+L.item_no_hazard_damage_desc = [[Makes you immune to hazard damage such as poison, radiation, and acid.]]
 L.item_no_energy_damage = "No Energy Damage"
-L.item_no_energy_damage_desc = [[Makes you immune to energy damage such as lasers, plasma and lightning.]]
+L.item_no_energy_damage_desc = [[Makes you immune to energy damage such as lasers, plasma, and lightning.]]
```
Histalek, line 1242: While I do like the oxford comma i don't think we use it anywhere else. We should decide if we want to use it or not and use it consistently (or don't use it respectively)

gamemodes/terrortown/gamemode/shared/lang/en.lua
```diff
@@ -1276,7 +1276,7 @@ L.dna_help_primary = "Collect a DNA sample"
 L.dna_help_secondary = "Switch the DNA slot"
 L.dna_help_reload = "Delete a sample"
 
-L.binoc_help_pri = "Identify a body."
+L.binoc_help_pri = "Search a body."
```
Histalek, line 1279:
```suggestion
L.binoc_help_pri = "Inspect a body."
```
Regarding line 1266 this seems like the most consistent wording

gamemodes/terrortown/gamemode/shared/lang/en.lua
```diff
@@ -1087,17 +1087,17 @@ L.menu_administration_title = "Administration"
...
+L.menu_changelog_description = "A list of changes and fixes in recent versions."
+L.menu_guide_description = "Helps you to get started with TTT2 and explains some things about gameplay, roles and other stuff."
+L.menu_bindings_description = "Bind specific features of TTT2 and its addons to your own liking."
+L.menu_language_description = "Select the language of the gamemode."
+L.menu_appearance_description = "Tweak the appearance and performance of the UI."
+L.menu_gameplay_description = "Avoid roles and tweak some features."
+L.menu_addons_description = "Configure local addons to your liking."
+L.menu_legacy_description = "A panel with converted tabs from the original TTT that should be ported over to the new system."
+L.menu_administration_description = "General settings for HUDs, shops, etc."
+L.menu_equipment_description = "Set credits, limitations, availability, and other stuff."
```
Histalek, line 1099: same as the other comment regarding the oxford comma

Histalek (changes requested):
Thank you for your work! I really like the changes :)

just some small nitpicks and one team decision i guess =D

gamemodes/terrortown/gamemode/shared/lang/en.lua
Satton2: Regarding string 1040 (and many others), there should be the word «search». While the word «inspect» is used only in one string (1266 and currently 1279).

Histalek: you are right, in that case line 1266 should be `search` i guess. unless i'm missing something regarding the meaning behind these two

gamemodes/terrortown/gamemode/shared/lang/en.lua
Satton2: I think it should be used since it was used almost in every project where I was translating something. I have also seen that many study books and mass media use it, but not every, of course.

gamemodes/terrortown/gamemode/shared/lang/en.lua
Satton2: I think it should be used since it was used almost in every project where I was translating something. I have also seen that many study books and mass media use it, but not every, of course.

gamemodes/terrortown/gamemode/shared/lang/en.lua
Satton2: Is it the same action as in line 1267, right? If yes, then I will change it to "search".

gamemodes/terrortown/gamemode/shared/lang/en.lua
```diff
@@ -1119,19 +1119,19 @@ L.submenu_appearance_interface_title = "Interface"
 L.submenu_appearance_miscellaneous_title = "Misellaneous"
 
 L.submenu_gameplay_general_title = "General"
-L.submenu_gameplay_avoidroles_title = "Avoid Role Selection"
+L.submenu_gameplay_roles_title = "Avoid role selection"
```
TimGoll, line 1122: Headings are always with capital letters for every word. Same holds true for all of your changes.

gamemodes/terrortown/gamemode/shared/lang/en.lua
TimGoll: Afaik search and inspect are used interchangeably. Just make sure search/inspect are not mixed with confirm.

gamemodes/terrortown/gamemode/shared/lang/en.lua
TimGoll: Additionally you can't just change the string identifier without changing it in the code as well.

Satton2: I forgot about this rule, yeah. For string 1218 it doesn't really matter because of this:
image


> Additionally you can't just change the string identifier without changing it in the code as well.

Wait, how did I... I haven't change it...

gamemodes/terrortown/gamemode/shared/lang/en.lua
Alf21: Tbh, most of all I'm against the oxford style. It feels unnutural, most of all caused by the short grammatical constructs, which could easily be avoided and so would go hand in hand with many of other languages. Additionally, oxford style theoretically is just needed if the context is unclear. But "such as" already implies a list or single item, but not a deeper desccription with another list of an example.

Satton2: I don't split writing styles on Oxford and non-Oxford. In the Russian language, we have consistent and clear grammar rules, while in the English language, some things are unclear. I was taught to use this comma in such places, I see that comma many times, and that's why I want to put it there. If the TTT2 team doesn't want a comma in such places, I will not understand that decision, but I will accept it.

saibotk: I don't really favor any of them. I'm fine with both.

saibotk (approved):
Thank you very much for taking the time to go through all the strings :rocket:

gamemodes/terrortown/gamemode/shared/lang/en.lua
Histalek: after thinking about it again, i'm inclined to agree with alf here. in this case the oxford comma doesn't serve a purpose.
furthermore i'm afraid that it might confuse future translators who aren't familiar with it. The English language file is our 'reference' language afterall and therefore should be as simple as possible and only as complex as needed

Satton2: > furthermore i'm afraid that it might confuse future translators who aren't familiar with it. The English language file <…> should be as simple as possible and only as complex as needed

For me, it sounds like someone just said that you can translate without enough knowledge... heh, translation work isn't easy. Also, there can be more confusing things than just a little comma. (especially in the Russian language ;) )

I have to say that this comma doesn't really matter for my language since we have different grammar rules, so we don't put it there (_I bet that for other languages it doesn't matter too._). It only matters for players who use English. This comma is usually used in American English. For example, Steam and Valve games use it, while Garry's Mod... well, there are not enough in-game strings to understand if it uses it or not, but GMod's blogs use it too. So, I don't see what can cause confusion here for translators, except for the lack of knowledge.

Alf21: @TimGoll @lebroomer Any additional personal oppinion?

saibotk: We decided to not use the oxford comma in the translations.

gamemodes/terrortown/gamemode/shared/lang/en.lua
Satton2: The comma is now removed.

gamemodes/terrortown/gamemode/shared/lang/en.lua
Satton2: The comma is now removed.

Histalek (approved):
Thank you very much :)

Alf21 (approved):
❤️

PR #711, opened 2021-01-11
Roleselection: Fix Layering

ZenBre4ker:
Before this fix, there was a for-loop that should have been repeated with the same index, if certain conditions were met. As lua creates a fixed set of indizes, which it internally iterates through changes to the index i in this case had no effect. Therefore if an empty layer was found, the next layer got also skipped as a consequence.

The Fix therefore now safely iterates through the layers until a valid layer is found and chooses a role out of it.
In case it still has baseroles left, but doesnt find a valid layer, the rest of available base roles are chosen after that. That's why I switched from if-else to two if-statements.

Still missing is a fix to the subroles, which are a bit more tricky. But this is my first draft for discussion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Fixed weapon pickup bug, where weapons would not get dropped but stayed in inventory
 - Fixed a roleselection bug, where forced roles would not be deducted from the available roles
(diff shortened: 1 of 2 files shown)


ZenBre4ker:
Tested now base and subrole layers with some roles. It seems to be working now.

saibotk (approved):
Looks good to me!

Thank you very very much :rocket: awesome, that you found the issue :heart:

Alf21 (approved):

Issue #686, opened 2020-10-20
Deactivating Karma doesn't reset Damage reduction


ZenBre4ker:
## Your version of TTT2 (mandatory)
- Latest GitHub version of 18.10.2020

## Describe the bug (mandatory)
If your Karma is below 1000, where your damage gets reduced, you then disable Karma per Convar/Ulx and restart the round, the damage reduction stays active. This only gets refreshed by a mapchange.

## To reproduce

Steps to reproduce the behaviour:

1. Change your Karma to 0
2. Start or Restart a round
3. Deactivate Karma per Convar/Ulx
4. Restart the round
5. Take a sniper rifle for example, which deals 200 damage headshot. Console says 20 now.

## Expected behaviour
Normally the damage reduction shouldnt be there, when Karma is deactivated.

ZenBre4ker:
Is this fixed now?

saibotk:
Yes it should, if not, let me know and i will reopen the issue :) 

(Thanks didn't see this was not auto closed)

PR #713, opened 2021-01-13
Karma: Fixed karma damage bug

TheNickSkater:
Fixed so karma will no longer be accounted for in the damage if the system was disabled.
Related to https://github.com/TTT-2/TTT2/issues/686


TimGoll (approved):
Great, this is a nice and simple fix!

saibotk (approved):
Thank you very much :rocket:

PR #721, opened 2021-02-07
Update CHANGELOG.md

westooooo:
I fixed japanese translation. but I'll update it soon again.


saibotk:
@westooooo Hint: You need to change this file on the same branch as your other PR is, by default Github will always create a new branch if you do not select anything :)

PR #722, opened 2021-02-08
Classbuilder: Fixed inheriting from same base

TimGoll:
Right now inheriting on the same base from different calls does not work because the base is restricted to the one call for one specific folder. I change it in such a way that a table can be passed through.

Additionally I improved the event loading print.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Fixed
 
-- Inheriting from the same base in different folders did not work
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Fixed
+
+- Inheriting from the same base in different folders did not work
```
saibotk, line 9:
```suggestion
- Inheriting from the same base using the classbuilder in different folders did not work
```

saibotk (approved):
:rocket:

Alf21 (approved):

PR #725, opened 2021-02-19
Credits: Made TTT2CanTransferCredits a part of GM

AaronMcKenney:
The TTT2CanTransferCredits hook from my previous pull request seems to have messed with the API Documentation somewhat:
Credit_Hook_Whoops

Based on the layout, it seems like it should be a part of the GM section, so I went ahead and made that change. Not sure why the right panel has "()" and nothing else.


TimGoll (approved):
Nice catch, thanks!

TimGoll:
Another thing: `number` has to be lowercase in the params list

saibotk (approved):

PR #728, opened 2021-02-19
Language: Updated Japanese localization

TimGoll:
Thanks to @westooooo and @miniduikboot


TimGoll:
I already verified that it has the same lines as the english translation

Histalek (approved):
Thanks!

saibotk (approved):

PR #737, opened 2021-03-08
Fixed outline library

### LINKED ISSUE #736: Outline Bug/Error

Leave one of the following:

- Latest GitHub version

## Video
https://youtu.be/KNdcDz8HdAg

When a player become traitor and there is another traitor alive, the screen turn black e appars an error in console 
Is for sure correlated with traitor outline or generally speaking with outline system
This issue is really serious because it does not allow you to play traitors
This error do not appear if is not installed "Teammates glowing addon"

## To reproduce

1. Force Traitor or Become traitor
2. Force another Player/bot traitor

## Expected behaviour
No error and the screen should be not black screen

- Collection link of your Server: https://steamcommunity.com/sharedfiles/filedetails/?id=1806066645
- **All** Server errors in the server console
- 
```
[ttt2-master] addons/ttt2-master/lua/ttt2/libraries/outline.lua:116: bad argument #2 to 'IsLineOfSightClear' (Vector expected, got no value)
  1. IsLineOfSightClear - [C]:-1
   2. Render - addons/ttt2-master/lua/ttt2/libraries/outline.lua:116
    3. fn - addons/ttt2-master/lua/ttt2/libraries/outline.lua:192
     4. unknown - addons/ulib-v2_63/lua/ulib/shared/hook.lua:109 (x12750) 
```

## Additional context

Also @ZenBre4ker can confirm this issue I practically discovered it with him

ZenBre4ker:
This will fix #736 and therefore makes the outline library finally work as intended.
By inputting the client as self into the function call, we now really get if the line of sight is clear.
Before that, the call broke as it didnt get enough arguments.

Also explaining why it didn't occur before with addons like glowing teammates and probably the glowing jester:
Global variables of modules created with module("name",package.seeall) need to be accessed via their modulename:
So outline.OUTLINE_MODE_NOTVISIBLE instead of just the variablename.

This might lead to addons not working as intended anymore. They should evaluate if they might need the OUTLINE_MODE_NOTVISIBLE mode instead now, so that teammates are visible only through walls or the other way round with OUTLINE_MODE_VISIBLE only if visible/inlineofsight or even always with OUTLINE_MODE_BOTH.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Inheriting from the same base using the classbuilder in different folders did not work
 - Fixed a docstring in the vskin module
 - Made TTT2CanTransferCredits hook a part of GM for API Documentation sanity
+- Fixed outline library not working
 
 ## [v0.8.0b](https://github.com/TTT-2/TTT2/tree/v0.8.0b) (2021-02-06)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - TTT: fix instant reload of dropped weapon (by @svdm)
 - TTT: fix ragdoll pinning HUD for innocents (by @Flapchik)
+- Fixed outline library not working
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


Histalek (approved):
LGTM :rocket:

Alf21 (approved):
👍

saibotk (approved):

PR #740, opened 2021-03-24
Gamemode: Added category from the upcoming update

Satton2:
https://wiki.facepunch.com/gmod/Gamemode_Creation#category


Histalek (approved):
Thanks for taking care of this!

Histalek:
@Alf21 we might want to cut a release because of this, before the next gmod update. So that we don't lose this serverbrowser functionality. WDYT?

saibotk (approved):

Alf21 (approved):
Thanks for the contribution!
Yea, we should release this update.

PR #795, opened 2021-04-09
Localization: Mini update for localization

TheNickSkater:
Changed this as its not "Settings" but "Gameplay" in F1


saibotk (approved):
Good catch, Thanks!

Histalek (approved):
:+1:

PR #799, opened 2021-04-24
WIN_NONE: Fixed draw

Alf21:
This fixes the bug of an endless round (until time is up) even if no one is alive anymore (draw)


TimGoll (approved):
Looks good to me, quick and simple fix

saibotk (approved):

PR #802, opened 2021-05-02
Reworked networking

Alf21:
Removed function:
- SendRoleListMessage
- SendSubRoleList
- SendRoleList
- SendTeamList
- SendConfirmedTeam
- SendPlayerToEveryone
- SendRoleReset

Removed role's params:
- .disableSync
- .visibleForTeam
- .networkRoles

Removed network messages:
- TTT_Role
- TTT_RoleList

Removed global vars:
- TTT2NETTABLE

Removed global hooks:
- TTT2OverrideDisabledSync

Removed cmd
- _ttt_request_rolelist

TODO:
- Test
- Implement syncing queue (list support and sending "team" and "subrole" in the same message)
- Confirmation system (syncing)
- VoiceChat syncing


TimGoll:
Okay. I looked through your code and I *really* like what you did. Awesome stuff!

There is one thing that I'm not so sure about: You define a special role syncing hook for the team mates of the traitor. In my opinion a role flag to show a role to their team mates should be included because there are so many roles that are able to see their team mates.

Alf21:
> Similar question here: Would a Vampire appear as a regular Traitor to a fellow Traitor here?

As a normal T currently, to avoid issues with the spy etc.

Alf21:
> But does this mean that the dete isn't synced to innos and Ts @Alf21 ?

https://github.com/TTT-2/TTT2/pull/802/files#diff-392431d2a41d3c3722c73117fa8c3f6e1893e4e16097e9af1f555a4916e07cebR36-R41

If the target is a D, it is set as a D

TimGoll:
> > But does this mean that the dete isn't synced to innos and Ts @Alf21 ?
> 
> https://github.com/TTT-2/TTT2/pull/802/files#diff-392431d2a41d3c3722c73117fa8c3f6e1893e4e16097e9af1f555a4916e07cebR36-R41
> 
> If the target is a D, it is set as a D

So the dete is now a hidden role to innos and Ts?

Alf21:
> > > But does this mean that the dete isn't synced to innos and Ts @Alf21 ?
> > 
> > 
> > https://github.com/TTT-2/TTT2/pull/802/files#diff-392431d2a41d3c3722c73117fa8c3f6e1893e4e16097e9af1f555a4916e07cebR36-R41
> > If the target is a D, it is set as a D
> 
> So the dete is now a hidden role to innos and Ts?

No. There is no check regarding the role of the current player. As a result, this hook (runned for every player) just checks for the role of the target.

PR #806, opened 2021-05-14
Fixed c4 plant event references for GetText()

ZacharyHinds:
Removed .owner as the Event doesn't have an owner sub table. Alternatively, could change the self:Add() to add the parameters in an owner table.


TimGoll (approved):
Thanks for this quick fix, much appreciated!

saibotk (approved):

PR #809, opened 2021-05-17
Map Integration: Map role check

TimGoll:
Modified the `ttt_logic_role` and `ttt_traitor_check` entity to call a hook before the check takes place. This can be used to hide/modify the real player role.

Additionally from now on all evil roles are counted as a traitor for these role checks. There is a convar to disable this. This might cause problems, so this has to be tested.

Based on work done by @ZenBre4ker


gamemodes/terrortown/entities/entities/ttt_logic_role.lua
```diff
@@ -35,24 +37,36 @@ function ENT:KeyValue(key, value)
...
+
 	if name == "TestActivator" then
-		if IsValid(activator) and activator:IsPlayer() then
-			local activator_role = (GetRoundState() == ROUND_PREP) and ROLE_INNOCENT or activator:GetBaseRole()
+		if not IsValid(activator) or not activator:IsPlayer() then return end
+
+		local activator_role = (GetRoundState() == ROUND_PREP) and ROLE_INNOCENT
+			---
+			-- @realm server
+			or roles.GetByIndex(hook.Run("TTT2ModifyLogicCheckRole", ply, self, activator, caller, data) or ply:GetSubRole()):GetBaseRole()
```
ZenBre4ker, line 56: The end-bracket is in the wrong place.
Also why not move the ply:GetSubRole():GetBaseRole() into the hook. Thats the default case and therefore could be embedded directly inside or not?

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
```diff
@@ -35,24 +37,36 @@ function ENT:KeyValue(key, value)
...
 
-			if self.Role == ROLE_NONE or self.Role == activator_role then
-				Dev(2, activator, "passed logic_role test of", self:GetName())
+		if self.Role == ROLE_TRAITOR
+			and (cv_evil_roles:GetBool() and (activator_role ~= ROLE_INNOCENT or activator_role ~= ROLE_DETECTIVE)
+				or not cv_evil_roles:GetBool() and (activator_role == ROLE_TRAITOR))
+			or self.Role == ROLE_NONE or self.Role == activator_role
```
ZenBre4ker, line 61: Maybe change the order. As it should check the Role None and activator role first, if they are  correct it works directly.
The cv_evil_roles is doubled. if the first is false, the second statement automatically is true, therefore it only needs no be activator_role == ROLE_TRAITOR.

Rest looks good.

ZenBre4ker: My fault, that is necessary in case the others are false, but evil_role isnt.....

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
TimGoll: no, the hook can return any role that isn't a baserole. So we have to make sure we get the baserole here. And what do you mean with "The end-bracket is in the wrong place"?

gamemodes/terrortown/entities/entities/ttt_traitor_check.lua
```diff
@@ -0,0 +1,97 @@
...
+-- @param Player ply The player whose role is checked
+-- @param Entity ent The entity that is used (either ttt_logic_role` or `ttt_traitor_check`)
+-- @param Entity|Player activator The initial cause for the input getting triggered (e.g. the player who pushed a button)
+-- @param Entity caller The entity that directly triggered the input (e.g. the button that was pushed)
+-- @param string data The data passed
+-- @return[default=nil] Return the role of the player that should be used for this check
+-- @hook
+-- @realm server
+function GAMEMODE:TTT2ModifyLogicCheckRole(ply, ent, activator, caller, data)
+
```
ZenBre4ker, line 96: I mean we could just
return ply:GetSubRole()):GetBaseRole()
here. Thats the default case, when nobody changes the hook or not?

TimGoll: That's right, yes. I don't like putting code in there, but there is nothing that speaks against it

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
ZenBre4ker: Its ply:GetSubRole() ):G.... not ply:GetSubRole():G.... )

TimGoll: Yes, this is intended. What speaks against it?

ZenBre4ker: oh nvm..... I am confused by numbers and roles......

ZenBre4ker: Ply is not available here, I guess this needs to be activator there too.

TimGoll: > Ply is not available here, I guess this needs to be activator there too.

yes, copy paste error

TimGoll:
From my point of view this PR is done and tested. If @ZenBre4ker agrees, it is ready to be reviewed

ZenBre4ker:
Looks good, the additional check seems necessary. 👍

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
```diff
@@ -24,35 +26,46 @@ function ENT:KeyValue(key, value)
...
+-- @realm server
+function ENT:AcceptInput(name, activator, caller, data)
 	if name == "TestActivator" then
-		if IsValid(activator) and activator:IsPlayer() then
-			local activator_role = (GetRoundState() == ROUND_PREP) and ROLE_INNOCENT or activator:GetBaseRole()
+		if not IsValid(activator) or not activator:IsPlayer() then return end
+
+		---
+		-- @realm server
+		local role, team = hook.Run("TTT2ModifyLogicRoleCheck", activator, self, activator, caller, data)
```
saibotk, line 53: Why is activator used twice?

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
```diff
@@ -24,35 +26,46 @@ function ENT:KeyValue(key, value)
...
+		local activatorRole = roles.GetByIndex(role):GetBaseRole()
```
saibotk, line 54: A check / assertion that this role index is actually available would be useful for other developers when debugging

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
```diff
@@ -24,35 +26,46 @@ function ENT:KeyValue(key, value)
...
+		local activatorTeam = (GetRoundState() == ROUND_PREP) and TEAM_INNOCENT or team
 
-			if self.Role == ROLE_NONE or self.Role == activator_role then
-				Dev(2, activator, "passed logic_role test of", self:GetName())
+		if self.checkingRole == ROLE_TRAITOR and util.IsEvilTeam(activatorTeam)
+			or self.checkingRole == ROLE_INNOCENT and not util.IsEvilTeam(activatorTeam)
```
saibotk, line 58: This should explicitly check for team innocent imo. As the convar for counting evil teams would also affect this. Or remove this special case completely

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
```diff
@@ -24,35 +26,46 @@ function ENT:KeyValue(key, value)
...
 	if name == "TestActivator" then
-		if IsValid(activator) and activator:IsPlayer() then
-			local activator_role = (GetRoundState() == ROUND_PREP) and ROLE_INNOCENT or activator:GetBaseRole()
+		if not IsValid(activator) or not activator:IsPlayer() then return end
+
+		---
+		-- @realm server
+		local role, team = hook.Run("TTT2ModifyLogicRoleCheck", activator, self, activator, caller, data)
+		local activatorRole = roles.GetByIndex(role):GetBaseRole()
+		local activatorTeam = (GetRoundState() == ROUND_PREP) and TEAM_INNOCENT or team
```
saibotk, line 55: Mhh why would we want to pin this to innocent? Shouldnt this be handled through the role assignment at the beginning / reset to role_none?

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
```diff
@@ -24,35 +26,46 @@ function ENT:KeyValue(key, value)
...
 
-			if self.Role == ROLE_NONE or self.Role == activator_role then
-				Dev(2, activator, "passed logic_role test of", self:GetName())
+		if self.checkingRole == ROLE_TRAITOR and util.IsEvilTeam(activatorTeam)
+			or self.checkingRole == ROLE_INNOCENT and not util.IsEvilTeam(activatorTeam)
+			or self.checkingRole == activatorRole and not (self.checkingRole == ROLE_TRAITOR or self.checkingRole == ROLE_INNOCENT)
```
saibotk, line 59: Why do you also check traitor and innocent again? Wouldnt those only introduce edge cases and arent they wrong anyways?

gamemodes/terrortown/entities/entities/ttt_traitor_check.lua
```diff
@@ -0,0 +1,94 @@
...
+-- @param string value The new value
+-- @realm server
+function ENT:KeyValue(key, value)
+	if key == "TraitorsFound" then
+		-- this is our output, so handle it as such
+		self:StoreOutput(key, value)
+	end
+end
+
+local function VectorInside(vec, mins, maxs)
```
saibotk, line 22: Move to util?

lua/terrortown/entities/roles/ttt_role_base/shared.lua
```diff
@@ -99,6 +100,15 @@ function ROLE:IsBaseRole()
...
 end
 
+---
+-- Checks whether it is a special @{ROLE}.
+-- @note This just returns <code>false</code> if the role is Innocent!
+-- @return boolean Returns true if the player has a special role
+-- @realm shared
+function ROLE:IsSpecialRole()
+	return self.subrole ~= ROLE_INNOCENT
+end
```
saibotk, line 110: Somehow i thin this function needs to be justified, why do we need it at all? What is a special role and why would a role overwrite this anyways? I dont really see the usecase here

saibotk (changes requested):

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
TimGoll: Because for this entity there only is one entity. The activator and the player that triggered the check are always the same. For the other entity they can be different. Therefore I doubled it here.

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
TimGoll: Didn't we decide that we want "hard fails"?

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
TimGoll: yes, you're absolutely right on this one!

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
TimGoll: *shrug*
it was done like this in TTT and I did not want to break anything

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
TimGoll: Because otherwise this does fun things if you return `ROLE_INNOCENT, TEAM_TRAITOR`. While this is a stupid combination, it should be understandable.

gamemodes/terrortown/entities/entities/ttt_traitor_check.lua
TimGoll: ok

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
ZenBre4ker: REmoving is no option here, as we check for innocent role what detectives definitely arent, but should be included.
This check makes sure, that in the classic way without the convar everything is inno, thats not a traitor and in the modern way makes sure only the innocent team is allowed.

So I'd say it should stay like this.

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
saibotk: No rather fail early. I just think an assertion in a place that responds to others developing with such hooks makes more sense

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
saibotk: Since this is a role check entity i dont really think innocent checks should be special. Would you mind explaining the usecase behind this team stuff and special handling in detail?

lua/terrortown/entities/roles/ttt_role_base/shared.lua
saibotk: I don't think that a role is to decide whether it is special or not that really depends on the viewer

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
saibotk: Actually i dont think this is wrong since it is only a role check and if i have the innocent role i should be included even if you return a stupid team

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
ZenBre4ker: The maps use different approaches to check for traitor. Here are some examples in the "classic" way:
1) Waterworld: Checks for Innocent, so every other role including the detective is invalid/a "traitor"
2) Minecraft: Counts all Traitors, so only traitors are included as valid
3) Minecraft: Checks for Traitor, so everyone else beside traitors are invalid
4) Innocentmotel: Checks for Detective, only he is valid

So our idea was to get those cases, that are used on maps valid in two ways.
The Classic case means, that really only traitors are counted as traitor and not inno.
The New Case with the Convar activated means, that only inno teams are counted as innocent and not traitor

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
ZenBre4ker: One case thats a bit off, but is future-proof, would be a traitor, thats in another team, like the pirate.
He should be working for the right team and therefore count as them in a test. But because of his role he still would have access to the other stuff or count as traitor/valid even if hes in the right team.

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
TimGoll: I think I see the problem @saibotk is describing here. Think about the following scenario: The convar is set to `false` and the player is `TEAM_NECROMANCER`.

https://github.com/TTT-2/TTT2/pull/809/files#diff-96dc1dc56ce17268ec8dd91f20b2ccfcf7747bd07ce4f11c40e01e32fc089c3bR457-R477

This function then returns `false`, which means this function passes a necromancer as an innocent. Is this the behavior we want in this case? Honestly I'm not 100% sure, it might be ok.

ZenBre4ker: I know, but thats why we implemented this convar in the first place. The old system didnt know about other roles, but they handled everything via traitor, inno and dete. Therefore we decided that a check for inno, which I've only seen in one place is actually a check for not traitor. So that old cases are still valid with the convar set to false. In my opinion I would just get rid of the convar and make it modern all the way, but this would be my idea of implementing it. If we decide for innocents we only use the modern way kinda seems wrong.

TimGoll: I don't think the existence of the convar changes anything really. The way it was handled previously was that the check just failed, now it doesn't. Tbh both are wrong, because the maps aren't designed for aything but T, dete and inno. Therefore I'd say our implementation here is alright, because it isn't better or worse for this specific edge case, just different.

ZenBre4ker: I just think that we should deal with this one way.
Convar false: only traitor is "traitor" and everything else is "innocent"
Convar true: Only Team Inno is "innocent" and everything else is traitor

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
TimGoll: Hm, `roles.GetByIndex` never fails because it returns `ROLE_NONE` as a fallback. So I don't think an assert is needed here

lua/terrortown/entities/roles/ttt_role_base/shared.lua
TimGoll: okay, I move it back to the player then

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
```diff
@@ -24,35 +26,46 @@ function ENT:KeyValue(key, value)
...
 
-		self.Role = tonumber(value)
+		self.checkingRole = tonumber(value)
 
-		if not self.Role then
+		if not self.checkingRole then
 			ErrorNoHalt("ttt_logic_role: bad value for Role key, not a number\n")
 
-			self.Role = ROLE_NONE
+			self.checkingRole = ROLE_NONE
```
Alf21, line 34: Wouldn't this break compatibility?

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
Alf21: Shouldn't it be `ROLE_INNOCENT` by default regarding your following comment?

lua/ttt2/extensions/util.lua
```diff
@@ -454,6 +454,40 @@ end
...
+-- @param string team The team that should be tested
+-- @return boolean Returns if a team is evil
+-- @realm shared
+function util.IsEvilTeam(team)
+	-- players without a team are counted as neutral
+	if not team or team == TEAM_NONE or TEAMS[team].alone then
+		return false
+	end
+
+	if GetConVar("ttt2_rolecheck_all_evil_roles"):GetBool() then
```
Alf21, line 469: That's something I definitely don't like. It differs the gameplay into two completely different ways. As well, it's a hidden option. Core functions and mechanics shouldn't be adjustable as players does not expect such changes. Additionally, I don't like the Evil approach, Imo, every role or better said, every team should decide on it's own whether it's "evil" -> And even that can't lead to a better handling, because every role interaction can be unique. That's why I said I'd like to modify things via hooks instead of role attributes.

TL;DR:
All roles except innocent are "evil" || "evil" == `IsSpecial()`

Alf21 (changes requested):

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
TimGoll: not sure which addon would modify this entity

lua/ttt2/extensions/util.lua
TimGoll: We can leave the convar out. I just thought about the moaning players that want the default case back.

And we thought about defining an isEvil team flag, but we decided against it. We don't want to create a comiplicated system. A player is evil, if he isn't innocent. E.g. in a traitor tester not only trators should trigger it, but also necros, jackals etc. This was our intention.

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
TimGoll: I don't get what you mean here, sorry. `roles.GetByIndex` returns `ROLE_NONE`, not `ROLE_INNOCENT`

Alf21: The default TTT and default maps are using `ROLE_INNOCENT` as well. If the reason to keep `TEAM_INNOCENT` is to have backward compatibility, then it should be `ROLE_INNOCENT` here as well

TimGoll: So you mean I should change the fallback `ROLE_NONE` here to `ROLE_INNOCENT`?

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
Alf21: Would be cool if @TheNickSkater could test this with DamageLogs as that add-on does somethings with role data :D

lua/ttt2/extensions/util.lua
Alf21: Yea, so we don't need an extra function but a simple inline check `Player:GetTeam() ~= TEAM_INNOCENT` :D That should be the most common case. Currently, I can't imagine a different behavior by different maps and if so, a convar wouldn't be the way to go, then it would need to be a map-based config

TimGoll: If we don't have a convar, we don't need this function, yes. However we should also consider `TEAM_NONE` like we did. It is easier to read if both checks are inside a function

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
Alf21: not exactly. The fallback of the function still should be ROLE_NONE as most of TTT role add-ons are done by our community and so are up-to-date. BUT the fallback / default value of the return here should be `ROLE_INNOCENT`, if I did not got things wrong.

lua/ttt2/extensions/util.lua
Alf21: Hmmm... That's a good point tho. Anyways, then please keep in mind that a team with `.alone` is still evil. Regarding:

```lua
if not team or team == TEAM_NONE or TEAMS[team].alone then
	return false
end
```

e.g. the SERIALKILLER has TEAM.alone

TimGoll: ah-oh. Good to know. I didn't know what's the reason of `.alone`

Alf21: TL;DR: Teams with `.alone` should just be able to win with one (and only one) alive team player. As a result, if there is more than one Serialkiller, a WIN should just trigger if one of them died.

ZenBre4ker: I'd support leavin the convar out. I like doing it the modern way ^^

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
TimGoll: This should have been dealt with by now because I removed the convar.

TimGoll:
I think all your points should be cleared now @Alf21 and @saibotk

Alf21 (approved):

CHANGELOG.md
```diff
@@ -40,6 +42,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Code cleanup and removed silly negations
 - Extended some ttt2net functions
 - Changed `bees` win to `nones` win
+- By default all evil roles are now counted as traitor roles for map related checks (with convar to disable)
```
saibotk, line 45: There is no convar anymore.

saibotk (changes requested):

saibotk:
@ZenBre4ker are you fine with the latest changes?

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
   - Enables the possibility to give Innocents access to a custom shop (`shopeditor`)
 - Karma now stores changes
   - Is shown in roundend menu
+- Count all evil roles as traitor roles for map related entity checks if enabled
```
ZenBre4ker, line 19: "if enabled" also nothing there anymore

ZenBre4ker (commented):
LGTM :)

saibotk (approved):

Issue #810, opened 2021-05-19
Nerf not making sound when crouch bhoping 


broiiler:
title is pretty self explanatory make it so when you crouch bhop on the ground it makes sound for accelerating at certaint amount of speed

broiiler:
https://github.com/Facepunch/garrysmod/blob/be251723824643351cb88a969818425d1ddf42b3/garrysmod/gamemodes/terrortown/gamemode/shared.lua#L158

ZenBre4ker:
As crouching is limited to something under 150 speed. I will test if it actually works or it had a reason why they put the crouch statement in there. Usually it should be fine if we just check for the speed.

ZenBre4ker:
Can be closed I guess.

TimGoll:
thanks, yes

PR #811, opened 2021-05-19
Footsteps: Fixed silenced footsteps when moving fast while crouching

ZenBre4ker:
#810
Instead of limiting to the players max speed, which doesnt change while crouching, we now check for the real speed, which is 60 for crouching, 100 for slowwalking and 220 for normal walking. Everything below 150 is now muted.
If someone can bhop and is faster than 150 you can hear him. Therefore cant be abused anymore.

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/sh_main.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_main.lua
@@ -8,7 +8,7 @@ local UpdateSprint = UpdateSprint
 
 local MAX_DROWN_TIME = 8
 
-local sneakSpeedSquared = math.pow(150,2)
+local sneakSpeedSquared = math.pow(150, 2)
 
 ---
 -- Initializes TTT2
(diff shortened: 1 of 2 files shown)


TimGoll (approved):
interesting, I think this is a good thing ~~, but you should add a changelog entry~~

saibotk (approved):
Thank you as always Sven. 

It is a pleasure working with you :rocket:

gamemodes/terrortown/gamemode/shared/sh_main.lua
```diff
@@ -8,6 +8,8 @@ local UpdateSprint = UpdateSprint
 
 local MAX_DROWN_TIME = 8
 
+local sneakSpeedSquared = math.pow(150,2)
```
Alf21, line 11:
```suggestion
local sneakSpeedSquared = math.pow(150, 2)
```

Alf21 (approved):

ZenBre4ker:
done ^^

PR #812, opened 2021-05-22
TeamChat: Fixed issue where base innocents could bypass chat hooks

AaronMcKenney:
This pull request fixes two issues:
- Fixed issue where base innocents could bypass the TTT2AvoidGeneralChat and TTT2AvoidTeamChat hooks with the team chat key.
- Fixed issue where roles with unknownTeam could see messages sent with the team chat key

Explicitly, checks for determining whether a player can use/read team chat messages involved the "IsSpecial" function, which only checks if the player is a non-innocent (and further lead to strange behavior with the team chat key). I altered the functionality to be more like voice chat: unknownTeam is used instead now. If the player is on an unknownTeam (Innocents, Detectives, Amnesiacs, etc.), they can technically open up the text box to type messages with the team chat key, but no message will be displayed or sent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed a German translation string (by @FaRLeZz)
 - Fixed a Polish translation by adding new lines (by @Wuker)
 - Fixed a data initialization bug that appeared on the first (initial) spawn
+- Fixed issue where base innocents could bypass the TTT2AvoidGeneralChat and TTT2AvoidTeamChat hooks with the team chat key
+- Fixed issue where roles with unknownTeam could see messages sent with the team chat key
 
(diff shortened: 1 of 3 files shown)


TimGoll (approved):
really impressive that you were able to go through these aweful nested conditions without stroking out. Great work, thanks!

saibotk (approved):

AaronMcKenney:
> really impressive that you were able to go through these aweful nested conditions without stroking out. Great work, thanks!

Oh I totally did though. I only became confident that it worked after a testing session with a friend.

Alf21 (approved):
As a "hotfix", I'm more than happy approving this PR. But the whole voice chat definitely needs a better concept. I don't like the '.unknownTeam' and similar role's attributes as they aren't precise enough, I'd prefer using hooks or more specified attributes to get a more reliable and adjustable voice chat system.

PR #816, opened 2021-06-04
Shopeditor: Move edit equipments into F1 Menu

ZenBre4ker:
Still need to fix setting the variables correctly.

For some reason using the ShopEditor.GetEquipmentForRoleAll() function too early leads to the equipment not being fully loaded. Probably the Equipment isnt registered at that point of time. It also destroys the normal shopeditor in its functions....
~~Also settings the variables with the net-message "TTT2SESaveItem" seems to be incorrect. Gotta look further into that.~~

Heres the first design at least:
grafik


saibotk (changes requested):
I noted some style stuff. But otherwise im super happy that you took care about that and that it works so well with our new vgui system.

My only technical feedback is about the virtual submenus: I would try to replace the item (the data) that is displayed and react to the change instead of creating an array of all views (the base menu items) with the base item table being copied in for each one. But i dont know if it is easy to implement in our current system. Just wanted to leave this as an idea / comment.

gamemodes/terrortown/gamemode/shared/sh_shopeditor.lua
```diff
@@ -4,13 +4,34 @@
...
+		order = 5, typ = "bool", default = 0,
+		name = "player_limited", b_desc = false
+	},
+	minPlayers = {
+		order = 6, typ = "number", bits = 6, default = 0, min = 0, max = 63,
+		name = "min_players", b_desc = false
+	},
+	credits = {
+		order = 7, typ = "number", bits = 5, default = 1, min = 1, max = 20,
+		name = "credits", b_desc = false
```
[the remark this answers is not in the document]
ZenBre4ker: What do you mean with that?
In the last PR I did the same style from line 37-62.

gamemodes/terrortown/gamemode/shared/sh_shopeditor.lua
```diff
@@ -116,7 +137,8 @@ end
 -- @return string name of the equipment
 -- @return ITEM|Weapon equipment table
 -- @realm shared
-function ShopEditor.ReadItemData()
+-- @deprecated
+function ShopEditor.ReadItemDataOld()
```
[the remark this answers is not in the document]
ZenBre4ker: My idea was to change the old shopeditor as less as possible and take as much out of it as possible.
As it gets deleted after the full rework anyway I thought just marking it as deprecated and changing the function would suffice. Because yes this is a breaking change but I should have changed everything that depended on it.

lua/terrortown/menus/gamemode/equipment.lua
```diff
@@ -1,12 +1,61 @@
...
+
+-- overwrite the normal submenu function to return our custom virtual submenus
+function CLGAMEMODEMENU:GetSubmenus()
+	lang = GetConVar("ttt_language"):GetString()
+	if not self.isInitialized or lang ~= self.lang then
+		self.isInitialized = true
+		self.lang = lang
+		self:InitializeData()
+	end
+	return virtualSubmenus
```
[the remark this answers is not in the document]
ZenBre4ker: Yeah this is still a mess. Waiting for Tim to tell me how to fix the translation problem and for a general opinion about setting up the virtual menu only if you are trying to access it. Normally it's done in initialization.

ZenBre4ker:
> 
> My only technical feedback is about the virtual submenus: I would try to replace the item (the data) that is displayed and react to the change instead of creating an array of all views (the base menu items) with the base item table being copied in for each one.

I dont fully get it, besides that you want to reduce the stored data?
Generally I only need some stuff, like name and id. But as I copied mainly the code out of the old shopeditor, it would need a deeper rework, as I am writing some stuff directly into the items.

gamemodes/terrortown/gamemode/shared/lang/de.lua
```diff
@@ -1359,8 +1359,22 @@ L.shopeditor_desc_random_shops = [[Zufalls-Shops geben jedem Spieler eine zufäl
...
 L.shopeditor_name_random_shop_reroll_per_buy = "Automatisches Auswürfeln nach jedem Kauf"
+
+-- 2021-06-04
+L.header_equipment_setup = "Equipment Einstellungen"
+L.header_equipment_value_setup = "Balance Einstellungen"
+
+L.itemeditor_name_not_buyable = "Nicht kaufbar"
+L.itemeditor_desc_not_buyable = [[Nicht zufälliges Equipment wird im Zufalls-Shop immer enthalten sein.
+Equipments können auf einen Kauf limitiert werden. Entweder Global, im Team oder nur für einen Spieler.]]
+L.itemeditor_name_not_random = "Nicht zufällig"
```
TimGoll, line 1375: I think "nicht zufällig" isn't really descriptive. Equipment marked with `.noRandom` is always included in the shop, even if a random shop is used. Maybe you should make this more clear?

gamemodes/terrortown/gamemode/shared/sh_shopeditor.lua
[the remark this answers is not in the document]
TimGoll: either everything should be in a single line or every entry should have its own line

gamemodes/terrortown/gamemode/shared/sh_shopeditor.lua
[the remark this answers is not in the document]
TimGoll: I tend to agree with @saibotk. However @ZenBre4ker is right as well. As soon as the shop editor is moved as well, this will be removed anyways. So it isn't that important.

lua/terrortown/menus/gamemode/equipment.lua
```diff
@@ -1,12 +1,61 @@
...
 
 CLGAMEMODEMENU.icon = Material("vgui/ttt/vskin/helpscreen/equipment")
 CLGAMEMODEMENU.title = "menu_equipment_title"
 CLGAMEMODEMENU.description = "menu_equipment_description"
 CLGAMEMODEMENU.priority = 49
 
+CLGAMEMODEMENU.isInitialized = false
+CLGAMEMODEMENU.lang = "en"
+
+function CLGAMEMODEMENU:Initialize()
```
TimGoll, line 17: you don't need the initialize function if you don't use it.

lua/terrortown/menus/gamemode/equipment.lua
```diff
@@ -1,12 +1,61 @@
...
+
 CLGAMEMODEMENU.base = "base_gamemodemenu"
 
 CLGAMEMODEMENU.icon = Material("vgui/ttt/vskin/helpscreen/equipment")
 CLGAMEMODEMENU.title = "menu_equipment_title"
 CLGAMEMODEMENU.description = "menu_equipment_description"
 CLGAMEMODEMENU.priority = 49
 
+CLGAMEMODEMENU.isInitialized = false
+CLGAMEMODEMENU.lang = "en"
```
TimGoll, line 15: where do you use this?

TimGoll (changes requested):

gamemodes/terrortown/gamemode/shared/lang/de.lua
TimGoll: Thinking about it a bit more, keep it like you did. Once you finished your todos here, I'll go over the menu and add a few explanatory things. It's easier to do it by myself

lua/terrortown/menus/gamemode/equipment.lua
ZenBre4ker: In GetSubmenus to make sure the virtual menus are only created at first call or when the language did change.
Thinking about making that even more effective and would start with lang = nil.

gamemodes/terrortown/gamemode/shared/lang/de.lua
ZenBre4ker: "Always available"/"Immer verfügbar" would be better. I guess because it had to fit without description and in the old shopeditor .norandom was born.
We could change that now to a better name, but it would destroy every config if we change the NoRandom-name directly.
Instead of course we could just change the translation.

In general I dislike the negation in those variables. Didnt we decide that we wanna do all variables in a non-negated way?

TimGoll: yes. It should be done for buyable, I think it is okay for the other variables though. Teamlimited is not really negated for example.

But you can keep it that way and I'll do it quick

gamemodes/terrortown/gamemode/shared/sh_shopeditor.lua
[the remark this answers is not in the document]
ZenBre4ker: Ihr habt recht, war auch nicht schwer... Reine Faulheit es drin zu lassen.
Ist zwar jetzt für die alte Version sehr unnötig, aber dafür clean.

TimGoll:
I added the ingame documenation. If you fix the problem with the reset button, I'm fine with it

TimGoll (approved):
Tested and I'm fine with it, thanks as always. However before this is merged, you have to fix the default behaviour (or disable the reset button)

gamemodes/terrortown/gamemode/shared/sh_shopeditor.lua
```diff
@@ -75,39 +152,40 @@ function ShopEditor.WriteItemData(messageName, name, item, plys)
 
 	if not name or not item then return end
 
-	net.Start(messageName)
-	net.WriteString(name)
-
 	for key, data in pairs(ShopEditor.savingKeys) do
+		net.Start(messageName)
+		net.WriteString(name)
+		net.WriteString(key)
```
Alf21, line 158: That's a ton of nw-messages, what are the reasons changing this?

ZenBre4ker: Mainly because of the single variable change in the new UI compared to the bunch of Saves of the old one. I rewrote the send and save method so it would fit.

ZenBre4ker: But it was difficult to keep the overview and contain the old structure, so maybe this approach isnt ideal. Actually instead of the old hardcoded number of saves, i could make an individual message size. That would work too.

TimGoll: okay, I have to agree with Alf. Why do you send a new message for every key and not one message for every item?

ZenBre4ker: Because of ReadItemData. Its used for several use-case and therefore is unified into one function. Its optimized for the new UI, but as I said, I came up with the idea to make that message size variable. You just write and additional parameter that contains the number of keys to save. Then you can have the old system still with all keys and the new system with only one key per save.

In general it's just a bit annoying to keep an old system intact, so I just changed it to thos inefficient way. But maybe I'll need the upper function anyway and then an optimization should be one message per item instead.
Gotta look more closely at the code to be sure :)

TimGoll: Ah I understand. So in the new system you mostly send only one key so it is already the most performant way. Only in the old UI it has an overhead. Did I get this right?

ZenBre4ker: Yes, in the old UI it's definitely bad and it is optimized for the new UI one way (Client->Server), but what I forgot was, that you send data back to the clients and maybe there I didnt change the system to one key per message. So need to look at that anyway. :)

So definite answer is, I need to check this part too, as it might be still needed for the new UI and therefore produces too many

TimGoll: It is ok for me then since the focus should be on the new UI anyway

saibotk (approved):

gamemodes/terrortown/gamemode/shared/sh_shopeditor.lua
Alf21: https://github.com/TTT-2/TTT2/pull/816/files#diff-eb89179dfc4da23810ef7f998dd5d73c8c3b380007153cbcf4bca2301f451692R155

That means you are looping through every key in ShopEditor every time `ShopEditor.WriteItemData` is called:
https://github.com/TTT-2/TTT2/pull/816/files#diff-eb89179dfc4da23810ef7f998dd5d73c8c3b380007153cbcf4bca2301f451692R6-R67

If that's right and intended, then feel free merging it :)

ZenBre4ker:
Okay correct default values are now implemented. Works for those I tested.

Also fixed the vast amount of messages by adding an UInt for the message size. Now the shopeditor can just save one key and everything else can synchronize as many keys as they want per message.

Only thing that could be improved is the sql saving and only synchronizing changed data. Right now, in every sync, all savingKeys get sent over again, even if only one might be changed. As that is some deeply integrated stuff, I didnt further investigate. (Too lazy and it's not real-time performance heavy)

TimGoll:
Took a look at your changes and I still approve them. Maybe @Alf21 wants to voice his opinion so we can merge this.

gamemodes/terrortown/gamemode/shared/sh_shopeditor.lua
```diff
@@ -66,6 +66,19 @@ ShopEditor.savingKeys = {
...
 
+local function countSavingKeys()
+	local count = 0
+
+	for key, data in pairs(ShopEditor.savingKeys) do
+		count = count + 1
+	end
+
+	return count
+end
```
Alf21, line 77: https://wiki.facepunch.com/gmod/table.Count

Alf21 (changes requested):

gamemodes/terrortown/gamemode/shared/sh_shopeditor.lua
ZenBre4ker: Done 😆

Alf21 (approved):

PR #820, opened 2021-06-16
F1 Menu: Legacy Addons: Fix Null Panel

ZenBre4ker:
As the panel always gets deleted, when it's parent gets deleted, we have to recreate it everytime we fetch a submenu.

This is a dirty workaround to get the job done. Better would be to free the child of its parent before clearing it. This also calls the hook  `TTTSettingsTabs` again everytime we open another legacy addon submenu.


TimGoll (approved):
Really nice! @TheNickSkater should confirm that it works

saibotk (approved):
Thank you for all of your hard work recently 🤩

Alf21 (approved):

PR #821, opened 2021-06-18
Cleanup library structure

saibotk:
This cleans up the autorun folder from the BEM database functions and the bind library.
It also moves the includes/modules/* files to the libraries folder for consistency.


TimGoll (approved):
Really nice!

A general question though: Should we move those files from `lua/ttt2/libraries` to `lua/terrortown/libraries`? This seems more consistent to me.

Alf21 (approved):

PR #822, opened 2021-06-18
Credit transfer: Sort teammates first and show indicator

### LINKED ISSUE #734: ON or OFF button/CVAR for the 'send a credit to anyone, even those players who aren't not on your team' feature
**Is your feature request related to a problem? Please describe.**
Although for the many extra roles it's great letting Detectives & Traitors and other roles give creds to about any player, for those servers who don't use those roles and want to only be able to give points to their teammates in a quick set of clicks, having the entire player list screws up their speed and maybe even the round if they are too hasty and forget how it's no longer only their teammates in the **transfer credit list**.
On some of the session of my server we had some situations where one T would give a credit to the detective by mistake then got called KOS and their entire round was ruined.

**Describe the solution you'd like**
Adding a simple button or a console command(cvar) to turn ON or OFF the ability to send credits to everyone or only to your team would suffice.

**Describe alternatives you've considered**
Letting people know where to change it/remove it on the code/config files would be the other way of letting them opt out of this new feature, if they would want to change it according to their needs.

**Additional context**
ttt_credit_transfer

saibotk:
This patch will list teammates first in the credit transfer menu and prepend a "(Team)" string behind their names to make them easily distinguishable for sending out credits to people that are important.

This fixes #734

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Shopeditor equipment is now available in F1 menu
 - Moved the role layering menu to the F1 menu (administration submenu)
   - removed the command `ttt2_edit_rolelayering`
+- Sort teammates first in credit transfer selection and add an indicator to them
 
 ### Removed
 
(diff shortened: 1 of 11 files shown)


TimGoll (approved):
kooks good to me

Alf21 (approved):

PR #823, opened 2021-06-19
Cleanup: Folder structure

TimGoll:
Inspired by #821 I moved `lang/`, `vskin/` and `events/` from `gamemodes/terrortown/shared/` to `lua/terrortown/`. Addons always used the latter part, but for some reason we used a different path for internal files, resulting in different structures for the basegame and addons. Therefore I cleaned this up.

In my opinion `huds/` and `hud_elements/` should be moved as well, but I've never really worked with those systems. So maybe @saibotk could move them. Moving them might break addons though. Therefore I'm not 100% sure.


Alf21 (approved):

TimGoll:
image

welp, don't merge it yet :D

TimGoll:
Everything fixed and ready

saibotk (approved):

PR #835, opened 2021-06-25
F1 Menu: Added general Searchbar and Fullsize icons for submenus

ZenBre4ker:
Equipment-editor has now full sized icons.
Default Searchfunction searches titles:
grafik

Filters out all submenus in real-time while typing:
grafik

Searchfunction can be extended to submenu-specific items, here shown for text and date:
grafik

(Ignore branch name, was intended for a cleanup, but got stuck with feature creep :3)


gamemodes/terrortown/gamemode/client/cl_help.lua
```diff
@@ -339,71 +336,27 @@ function HELPSCRN:ShowSubmenu(menuClass)
 
 	-- BUILD GENERAL BOX STRUCTURE
 	local navArea = vgui.Create("DNavPanelTTT2", frame)
-	navArea:SetSize(widthNav, heightNav - vskin.GetHeaderHeight() - vskin.GetBorderSize())
+	navArea:SetSize(widthNav)
```
TimGoll, line 339:
```suggestion
	navArea:SetWide(widthNav)
```

So it is a bit more readable

ZenBre4ker:
Interesting feature could be to mark the found strings. The polish-marking was made by myself and is not automatically done.

lua/terrortown/menus/gamemode/addons.lua
```diff
@@ -6,3 +6,7 @@ CLGAMEMODEMENU.icon = Material("vgui/ttt/vskin/helpscreen/addons")
 CLGAMEMODEMENU.title = "menu_addons_title"
 CLGAMEMODEMENU.description = "menu_addons_description"
 CLGAMEMODEMENU.priority = 94
+
+function CLGAMEMODEMENU:Initialize()
+	self:EnableSearchbar(true)
+end
```
TimGoll, line 12: EOF

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -1,11 +1,16 @@
 ---
 -- @class CLGAMEMODEMENU
 
+local TryT = LANG.TryTranslation
+local lowerString = string.lower
+local findString = string.find
```
TimGoll, line 6: you don't have to change this, but I'd recommend renaming those to `stringLower` and `stringFind` respectively

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -102,6 +107,56 @@ function CLGAMEMODEMENU:AddSubmenu(submenu)
...
+	self.searchbar = active
+end
+
+---
+-- Checks if the menu has a searchbar enabled.
+-- @return boolean Returns true if searchbar is set
+-- @realm client
+function CLGAMEMODEMENU:HasSearchbar()
+	return self.searchbar
+end
```
TimGoll, line 124: In my opinion I'd rewrite this. All other settings are based on baseclass overwrites, not on variables. So just a `HasSearchbar()` that can be overwritten should be used imho

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -102,6 +107,56 @@ function CLGAMEMODEMENU:AddSubmenu(submenu)
...
+---
+-- Sets the used searchfunction.
+-- Parameters for that function are submenuClasses and the searchText
+-- @param function searchFunction
+-- @realm client
+function CLGAMEMODEMENU:SetSearchFunction(searchFunction)
+	if not isfunction(searchFunction) then return end
+
+	self.searchFunction = searchFunction
+end
```
TimGoll, line 135: I'm not sure I like this setter here, why not define the function directly on the class? Is there a reason I'n not seeing right now?

lua/terrortown/menus/gamemode/changelog.lua
```diff
@@ -1,6 +1,9 @@
 --- @ignore
 
 local tableCopy = table.Copy
+local lowerString = string.lower
+local findString = string.find
```
TimGoll, line 5: same here with the names

TimGoll (changes requested):
`EnableSearchbar()` and `SetSearchFunction()` should really be chanegd imo. This is a completely different style then mine.
Besides this, this is of course awesome work as always.

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
ZenBre4ker: I agree, only later saw the style you use for this. Just leaving the dot out.

lua/terrortown/menus/gamemode/equipment.lua
```diff
@@ -16,6 +16,10 @@ CLGAMEMODEMENU.isInitialized = false
 CLGAMEMODEMENU.langConVar = nil
 CLGAMEMODEMENU.lang = nil
 
+function CLGAMEMODEMENU:Initialize()
+	self:EnableSearchbar(true)
```
ZenBre4ker, line 20: Well the main idea was to let the developer have the flexibility to just use a default searchbar like here by just enabling it. But if he wants to change it, he easily can set the searchfunction manually.
Im absolutely no fan of those class overwrites you did for example in the submenu with shouldShow. Generally I like it more if you can just set a simple variable with a setter like self:shouldShow(enabled) and the backend handles it for you. Otherwise you have to write all logic in that overwrite instead of just calling a setter somewhere in your code where you need it.

But I agree, that there should be a consistent style, so I can change it into a class overwrite for 
-hasSearchBar
-getSearchBar without setter but an overwrite

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -102,6 +106,31 @@ function CLGAMEMODEMENU:AddSubmenu(submenu)
...
+function CLGAMEMODEMENU:GetSearchFunction()
+	return function(submenuClass, searchText)
+			local txt = stringLower(searchText)
+			local title = stringLower(TryT(submenuClass.title))
+			local start = stringFind(title, txt)
+
+			return start or false and true
+		end
+end
+
```
ZenBre4ker, line 133: So like this, I guess :) Unsure about the function indentation though.

lua/terrortown/menus/gamemode/equipment.lua
TimGoll: I see your point. Maybe someone else can voice his opinion here as well. I first encoutered this style in Minecraft Modding and I really liked it. This is the reason I still stick to it.

What's the difference between GetSearchBar and HasSearchBar?

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
TimGoll: I still don't see the point here, sorry. Why not just:

```lua
function CLGAMEMODEMENU:GetSearchFunction(submenuClass, searchText)
	local txt = stringLower(searchText)
	local title = stringLower(TryT(submenuClass.title))
	local start = stringFind(title, txt)

	return start or false and true
end
```

Also your indentation is one too deep for our style

lua/terrortown/menus/gamemode/equipment.lua
ZenBre4ker: getSearchFunction i meant. See in the comment below.

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
ZenBre4ker: Ahh, thats what you mean. It was my idea of localization. So you can just store that searchfunction. But I guess, the way you did it, I can still store it with submenuClass.SearchMenu ^^

gamemodes/terrortown/gamemode/client/cl_help.lua
```diff
@@ -339,71 +336,27 @@ function HELPSCRN:ShowSubmenu(menuClass)
...
-			settingsButton.DoClick = function(slf)
-				HELPSCRN:SetupContentArea(contentArea, subMenuClasses[i])
-				HELPSCRN:BuildContentArea()
+	-- MAKE SEPARATE SUBMENULIST ON THE NAVAREA WITH A CONTENT AREA
+	local submenuList = vgui.Create("DSubmenuListTTT2", navArea)
+	submenuList:Dock(FILL)
+	submenuList:SetPadding(self.padding)
+	submenuList:SetSubmenuClasses(menuClass:GetVisibleSubmenus(), contentArea)
+	submenuList:EnableSearchBar(menuClass:HasSearchbar())
+	submenuList:SetSearchFunction(menuClass:GetSearchFunction())
```
ZenBre4ker, line 356: Here I just want to store it. But yeah, as I said, I can just write menuClass.SearchFunction instead

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
TimGoll: Obviously you spent much more time with the problem on hand. So if you think that your solution is the better one, I won't force you to change it. However with my insight based on this PR it seems to me as if your solution is unnecessary complicated

ZenBre4ker: No, I see your point. Im still stuck in a different mindset ^^
It's weird to be able to address function in a table directly, so I always feel like I want to give an object which is a function instead.

gamemodes/terrortown/gamemode/client/cl_help.lua
TimGoll: I see what you mean. Should the search function be really defined like this? I mean the search is always dependant on the menu you are in. So why do you need to link this function anyways?

ZenBre4ker: Mainly because I access it in a for-loop, so it's called very often. Localizing it should make it more efficient as far as I understood your performance improvements on older code.

TimGoll: Okay, hear me out. Take this part of the code:

https://github.com/TTT-2/TTT2/pull/835/files#diff-b5666cfb68877262c5ef28ff03124e5a00e32b08ced4be0a298615727238c6b2R86-R94

And move it to the menu search code. Then you only have to make one function call. Would this work?

TimGoll: Something on the lines of:

```lua
function CLGAMEMODEMENU:GetMatchingSubmenus(searchString)
	-- iterate over all submenus and return a list of matching menus
end
```

This would also fit the style of the submenu lists and virtual menus

TimGoll: I see your reasoning that this would create code duplication. You could make the search string matching a secondary function on `CLGAMEMODEMENU`. In normal cases only the matching function gets overwritten. But if one wants to, they can also overwrite the whole search system.

ZenBre4ker: I thought about this, but it gives to much freedom to the user. And everytime I want to make a custom searchfunction, I have to copy a lot of code and look out for edge-cases. I commited the better version, because you are right with the complicated style I used.

ZenBre4ker: Yeah that might allow to rewrite it, but I thought about making the search way more effective with caching for long lists and searches. I rather want to keep the outer logic to TTT2 and only allow the user to define when a search results in true and when not ^^

TimGoll: Is too much freedom to the developer a bad thing? And if you have a `CLGAMEMODEMENU:GetMatchingSubmenus` and a `CLGAMEMODEMENU:MatchSearchString` function where you overwrite only the latter, would this really be more copy and paste then your initial one? Because the matching function basically is exactly your initial function, you just moved the search iteration code from the menu element to the menu class. Which is not a bad thing imho, because I think the menu logic should be contained in the menu class.

TimGoll: > Yeah that might allow to rewrite it, but I thought about making the search way more effective with caching for long lists and searches. I rather want to keep the outer logic to TTT2 and only allow the user to define when a search results in true and when not ^^

That's okay with me. But I don't see what you can cache in the current version, that you can't cache in the menu-class-version

TimGoll: Looking at your most recent commit I see that you have the matching function like I suggested, only the search function is not in the class. I'm fine either way and I think I voiced my opinion on this enough. Let's see what the rest of the team has to say about this.

lua/terrortown/menus/gamemode/changelog.lua
```diff
@@ -30,3 +33,23 @@ end
...
+
+-- overwrite and do a custom search inside title, date and content
+function CLGAMEMODEMENU:SearchFunction(submenuClass, searchText)
+	local txt = stringLower(searchText)
+	local change = submenuClass.change
+
+	if stringFind(stringLower(TryT(submenuClass.title)), txt) then return true end
+
+	if change.date > 0
+	and stringFind(stringLower(os.date("%Y/%m/%d", change.date)), txt) then return true end
```
TimGoll, line 50:
```suggestion
	if change.date > 0
		and stringFind(stringLower(os.date("%Y/%m/%d", change.date)), txt)
	then
		return true
	end
```

If you want to lines, then do it like this. Also returns with a value should never be in line.

gamemodes/terrortown/gamemode/client/cl_help.lua
ZenBre4ker: Your last point is also valid. I see how I could easily make the same performance-optimization like this. Then you can only change what you want. I still have my simple matchsearchstring function, where I don't have to rewrite the outer loop and anybody could change it anyways :D

TimGoll: I know the feeling of someone critisizing the code that one thinks is so brilliant. Maybe give this problem a day to think about and then decide if you like your solution or my suggestion more.
I often noticed that I hated such suggestions in the beginning but came around liking them in the end after thinking about it for a while

ZenBre4ker: Nah, thats not my problem ^^
I like your suggestions so far, you convinced me quite fast xD

It's just really difficult to see clear after a few days of deep coding. I already refactored a lot and am always surprised of what I didnt think of. The code on the base_gamemodemenu was the first I did and then didnt touch anymore, because it just worked :stuck_out_tongue_closed_eyes:

ZenBre4ker: I really appreciate it, that you care about the code and look out for the style, many other people would just ask:
"Does it work?"
"Yes"
"Okay, merge it!" 😄

TimGoll: I learned from the best @Alf21 and @saibotk who always take my PRs apart, lol :D

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -102,6 +106,60 @@ function CLGAMEMODEMENU:AddSubmenu(submenu)
...
+-- @return bool Returns 
+-- @realm client
+function CLGAMEMODEMENU:MatchesSearchString(submenuClass, searchText)
+	local txt = stringLower(searchText)
+	local title = stringLower(TryT(submenuClass.title))
+	local start = stringFind(title, txt)
+
+	return start or false and true
+end
+
```
ZenBre4ker, line 162: Refactored that, I liked your idea and already found a major performance boost, if you dont search, you dont need to go through the complete for loop to get the same list as before 😄

TimGoll: Personally I'd add a note to `CLGAMEMODEMENU:GetMatchingSubmenus` to make it clear to devs that they probably do not want to overwrite that function. Otherwise I really like it, great work!

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenulist_ttt2.lua
```diff
@@ -0,0 +1,217 @@
...
+end
+
+---
+-- This function enables or disables the searchBar.
+-- @param bool active
+-- @realm client
+function PANEL:EnableSearchBar(active)
+	if not active then
+		if self.searchBar then self.searchBar:Clear() end
+		return
```
TimGoll, line 72: add an ampty line in front of this line

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenulist_ttt2.lua
```diff
@@ -0,0 +1,217 @@
...
+-- @realm client
+function PANEL:AddSubmenuButton(submenuClass)
+	local settingsButton = self.navAreaScrollGrid:Add("DSubmenuButtonTTT2")
+
+	settingsButton:SetTitle(submenuClass.title or submenuClass.type)
+	settingsButton:SetIcon(submenuClass.icon, submenuClass.iconFullSize)
+
+	settingsButton.PerformLayout = function(panel)
+		local width = panel:GetParent():GetSize()
+		panel:SetSize(width, heightNavButton)
```
TimGoll, line 109:
```suggestion
		panel:SetSize(panel:GetParent():GetWide(), heightNavButton)
```

I *think* this should work as well

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenulist_ttt2.lua
```diff
@@ -0,0 +1,217 @@
...
+	end
+
+	settingsButton.DoClick = function(slf)
+		HELPSCRN:SetupContentArea(self.contentArea, submenuClass)
+		HELPSCRN:BuildContentArea()
+
+		-- handle the set/unset of active buttons for the draw process
+		if self.lastActive and self.lastActive.SetActive then self.lastActive:SetActive(false) end
```
TimGoll, line 117: Conditions with content should not be in a single line, also the following line should have an empty line in front

lua/terrortown/menus/gamemode/changelog.lua
```diff
@@ -30,3 +33,30 @@ end
...
+function CLGAMEMODEMENU:MatchesSearchString(submenuClass, searchText)
+	local txt = stringLower(searchText)
+	local change = submenuClass.change
+
+	if stringFind(stringLower(TryT(submenuClass.title)), txt) then
+		return true
+	end
+
+	if change.date > 0
+			and stringFind(stringLower(os.date("%Y/%m/%d", change.date)), txt)
```
TimGoll, line 52: one tab too much here

TimGoll (approved):

gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
```diff
@@ -1352,5 +1356,46 @@ function SKIN:PaintRoleLayeringReceiverTTT2(panel, w, h)
...
+	local heightPad = topPad + bottomPad
+
+	if not panel:IsEnabled() then
+		colorBox = ColorAlpha(colorBox, alphaDisabled)
+		colorBar = ColorAlpha(colorBar, alphaDisabled)
+		colorText = ColorAlpha(colorText, alphaDisabled)
+	end
+
+	-- Draw custom box background for the searchBar
+	drawBox(leftPad, h * (1 - heightMult) / 2 + topPad, w - widthPad, h * heightMult - heightPad, colorBox)
```
Alf21, line 1381:
```suggestion
	drawBox(leftPad, h * (1 - heightMult) * 0.5 + topPad, w - widthPad, h * heightMult - heightPad, colorBox)
```

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsearchbar_ttt2.lua
```diff
@@ -0,0 +1,166 @@
...
+
+	self:PerformLayout()
+end
+
+---
+-- @param string newFont
+-- @realm client
+function PANEL:SetFont(newFont)
+	self.textEntry:SetFont(newFont)
+	font = newFont
```
Alf21, line 85:
```suggestion

	font = newFont
```

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsearchbar_ttt2.lua
```diff
@@ -0,0 +1,166 @@
...
+end
+
+---
+-- @ignore
+function PANEL:PerformLayout()
+	local width, height = self:GetSize()
+	local heightMult = self:GetHeightMult()
+
+	self.textEntry:SetSize(width, height * heightMult)
+	self.textEntry:SetPos(0, height * (1 - heightMult) / 2)
```
Alf21, line 151:
```suggestion
	self.textEntry:SetPos(0, height * (1 - heightMult) * 0.5)
```

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenubutton_ttt2.lua
```diff
@@ -69,9 +70,11 @@ end
 
 ---
 -- @param Material iconMat
+-- @param boolean iconFullSize
 -- @realm client
-function PANEL:SetIcon(iconMat)
+function PANEL:SetIcon(iconMat, iconFullSize)
 	self.contents.icon = iconMat
+	self.contents.iconFullSize = iconFullSize and true or false
```
Alf21, line 77:
```suggestion
	self.contents.iconFullSize = iconFullSize
```
😆 

If you wanna make sure it's never `nil`, you could cut out the `and true`.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenulist_ttt2.lua
```diff
@@ -0,0 +1,219 @@
...
+	self.searchBar:SetSize(widthBar, heightBar)
+end
+
+---
+-- This function enables or disables the searchBar.
+-- @param bool active
+-- @realm client
+function PANEL:EnableSearchBar(active)
+	if not active then
+		if self.searchBar then self.searchBar:Clear() end
```
Alf21, line 71:
```suggestion
		if self.searchBar then
			self.searchBar:Clear()
		end
```

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenulist_ttt2.lua
```diff
@@ -0,0 +1,219 @@
...
+		local widthContent = self.contentArea:GetSize()
+
+		labelNoContent:SetText("label_menu_not_populated")
+		labelNoContent:SetSize(widthContent - 40, 50)
+		labelNoContent:SetFont("DermaTTT2Title")
+		labelNoContent:SetPos(20, 0)
+	else
+		for i = 1, #submenuClasses do
+			local submenuClass = submenuClasses[i]
+
```
Alf21, line 147:
```suggestion
```

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -102,6 +106,61 @@ function CLGAMEMODEMENU:AddSubmenu(submenu)
...
+	local submenuClasses = self:GetVisibleSubmenus()
+
+	if searchText == "" then
+		return submenuClasses
+	end
+
+	local filteredSubmenuClasses = {}
+
+	local counter = 0
+	for i = 1, #submenuClasses do
```
Alf21, line 135:
```suggestion

	for i = 1, #submenuClasses do
```

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -102,6 +106,61 @@ function CLGAMEMODEMENU:AddSubmenu(submenu)
...
+-- @param menuClass submenuClass
+-- @param string searchText
+-- @return bool Returns 
+-- @realm client
+function CLGAMEMODEMENU:MatchesSearchString(submenuClass, searchText)
+	local txt = stringLower(searchText)
+	local title = stringLower(TryT(submenuClass.title))
+	local start = stringFind(title, txt)
+
+	return start or false and true
```
Alf21, line 161: Regarding <https://wiki.facepunch.com/gmod/string.find>, a number is returned if found.
Regarding <https://wiki.facepunch.com/gmod/Global.tobool>, tobool() should work, so `tobool(start)` or `start ~= nil`.

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -102,6 +106,61 @@ function CLGAMEMODEMENU:AddSubmenu(submenu)
...
+end
+
+---
+-- Determines the used searchfunction, when a HasSearchbar returns true.
+-- Parameters for that function are submenuClasses and the searchText
+-- Per default only titles are searched and compared to the searchtext in lowercase letters.
+-- @note This function can be overwritten to use a custom searchfunction.
+-- @param menuClass submenuClass
+-- @param string searchText
+-- @return bool Returns
```
Alf21, line 154: That's not very descriptive :D

Alf21 (changes requested):
Great work, really clean and logical code. Just some small requests.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenubutton_ttt2.lua
ZenBre4ker: nah, nil is not the problem, I wanted to make sure its always a bool, for some reason, that someone inputs a material for example. So I guess tobool() would suffice. But yeah as I only check it and dont use it, just leave it away :stuck_out_tongue_closed_eyes:

ZenBre4ker:
Done and tested :)

saibotk (approved):
Wow thank you very very much for your work :)

TimGoll:
@Alf21 you have to reapprove the changes

Alf21 (approved):
Thanks :)

PR #836, opened 2021-06-28
pure_skin: consistent text positioning in pure_skin bar

LukasMandok:
I noticed, that the text positioning in a pure_skin bar is not really consistent at different screen resolutions.
This has caused problems especially with a new HUD element, but it is also noticeable in the player info panel. 
I have tested this at various screen resolutions between 720p and 3000x2000, and it seems to be positioned consistently with the proposed change. 
It also looks good with the Vampire and Bodyguard role. But I have not tested more than that.

comparison - text in pure_skin bar
This is at 1080p and 2560x1600.

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/hud_elements/base_elements/pure_skin_element.lua
+++ b/gamemodes/terrortown/gamemode/shared/hud_elements/base_elements/pure_skin_element.lua
@@ -62,7 +62,7 @@ if CLIENT then
 
 		-- draw text
 		if t then
-			draw.AdvancedText(t, "PureSkinBar", x + 14, y + 1, util.GetDefaultColor(c), TEXT_ALIGN_LEFT, TEXT_ALIGN_LEFT, true, s)
+			draw.AdvancedText(t, "PureSkinBar", x + 14, y + h/2 - 1, util.GetDefaultColor(c), TEXT_ALIGN_LEFT, TEXT_ALIGN_CENTER, true, s)
 		end
 	end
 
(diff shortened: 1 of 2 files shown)


gamemodes/terrortown/gamemode/shared/hud_elements/base_elements/pure_skin_element.lua
```diff
@@ -62,7 +62,7 @@ if CLIENT then
 
 		-- draw text
 		if t then
-			draw.AdvancedText(t, "PureSkinBar", x + 14, y + 1, util.GetDefaultColor(c), TEXT_ALIGN_LEFT, TEXT_ALIGN_LEFT, true, s)
+			draw.AdvancedText(t, "PureSkinBar", x + 14, y + h/2 - 1, util.GetDefaultColor(c), TEXT_ALIGN_LEFT, TEXT_ALIGN_CENTER, true, s)
```
TimGoll, line 65: please change `h / 2` to `0.5 * h` because that is faster in lua

TimGoll (changes requested):

gamemodes/terrortown/gamemode/shared/hud_elements/base_elements/pure_skin_element.lua
LukasMandok: Good to know. I changed it. 
(I probably need to go through my other lua code as well.)

TimGoll (approved):

TimGoll:
You should add an entry to the changelog

saibotk (approved):
Thank you very much for your contribution 🤩

PR #845, opened 2021-07-10
Revive: Prevent spectators about to revive from gathering player info

AaronMcKenney:
This code change was made for reviving roles such as Pharaoh, Occultist, and Cursed. Each of these roles are likely to revive after death, but not immediately. During that 10 or so seconds they can browse the roles of other players, which they shouldn't know since they're about to re-enter play.

So this change prevents that. While the revive timer is going, the spectator can move around, but can no longer enter first-person view of an alive player to see their role, or look at unidentified corpses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Changed
 - Revise and additions simplified Chinese (by @TEGTianFan)
+- Prevent spectators from gathering info on players if they're about to revive (by @AaronMcKenney)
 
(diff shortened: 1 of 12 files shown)


Histalek (approved):
Very nice addition!

TimGoll (approved):
While I approve this, I want to start a short discussion. In my opinion a MSTACK message should be shown to explain why the player is unable to enter spectator mode

AaronMcKenney:
Warning message has been added.

TimGoll:
if you add a new line in front of the two `return` statements I'm happy!

AaronMcKenney:
> if you add a new line in front of the two `return` statements I'm happy!

Done.

PR #850, opened 2021-08-03
Hotreload: Weapon hotreloading fix and persistent data across reloads

ZenBre4ker:
This fixes the hotreloading problems with ttt2 missing certain key-values like .id or ttt2_cached_material.
It also cleans up the old code and reduces number of loops for weapons.GetList().

As a bonus I added the possibility to add HotReloadableKeys as a list of strings to make data saved via weapons.GetStored() inside the original SWEP-files to be persistent across hotreloads, if you need some variables to keep on counting.

As this was a deep going fix, I'd appreciate some tests and am open for critics as it might be that I dont see clear on this matter anymore 😄 To see how the weapons get registered and the hook "PreRegisterSWEP" gets ran look into garrysmod files directly under \garrysmod\lua\includes\modules\weapons.lua

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_equip.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_equip.lua
@@ -1104,23 +1104,27 @@ function GM:OnContextMenuOpen()
 	end
 end
 
+function TTT2CacheEquipMaterials(item)
+	--if there is no material or model, the item should probably not be available in the shop
+	if item.material and item.material ~= "vgui/ttt/icon_id" then
+		item.ttt2_cached_material = Material(item.material)
+		if item.ttt2_cached_material:IsError() then
+			-- Setting fallback material
+			item.ttt2_cached_material = fallback_mat
+		end
+	elseif item.model and item.model ~= "models/weapons/w_bugbait.mdl" then
+		--do not use fallback mat and use model instead
+		item.ttt2_cached_material = nil
+		item.ttt2_cached_model = model
+	end
+end
+
 -- Preload materials for the shop
 hook.Add("PostInitPostEntity", "TTT2CacheEquipMaterials", function()
 	local itms = ShopEditor.GetEquipmentForRoleAll()
 
 	for _, item in pairs(itms) do
-		--if there is no material or model, the item should probably not be available in the shop
-		if item.material and item.material ~= "vgui/ttt/icon_id" then
-			item.ttt2_cached_material = Material(item.material)
-			if item.ttt2_cached_material:IsError() then
-				-- Setting fallback material
-				item.ttt2_cached_material = fallback_mat
-			end
-		elseif item.model and item.model ~= "models/weapons/w_bugbait.mdl" then
-			--do not use fallback mat and use model instead
-			item.ttt2_cached_material = nil
-			item.ttt2_cached_model = model
-		end
+		TTT2CacheEquipMaterials(item)
 	end
 end)
 
--- a/gamemodes/terrortown/gamemode/client/cl_equip.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_equip.lua
@@ -1104,6 +1104,10 @@ function GM:OnContextMenuOpen()
 	end
 end
 
+---
+-- Caches the material or model
+-- @param table item
+-- @realm client
 function TTT2CacheEquipMaterials(item)
 	--if there is no material or model, the item should probably not be available in the shop
 	if item.material and item.material ~= "vgui/ttt/icon_id" then
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -505,7 +505,7 @@ function GM:InitPostEntity()
 	LoadShopsEquipment()
 
 	MsgN("[TTT2][INFO] Shops initialized...")
-	isShopFallbackInitialized = true
+	TTT2ShopFallbackInitialized = true
 
 	WEPS.ForcePrecache()
 
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -270,7 +270,7 @@ fileloader.LoadFolder("terrortown/autorun/server/", false, SERVER_FILE, function
 	MsgN("Added TTT2 server autorun file: ", path)
 end)
 
-CHANGED_EQUIPMENT = {}
+CHANGED_EQUIPMENT = CHANGED_EQUIPMENT or {}
 
 ---
 -- Called after the gamemode loads and starts.
@@ -425,18 +425,17 @@ function GM:InitPostEntity()
 	-- load all HUD elements
 	hudelements.OnLoaded()
 
-	InitDefaultEquipment()
-
 	local itms = items.GetList()
-	local sweps = weapons.GetList()
 
-	-- load and initialize all SWEPs and all ITEMs from database
-	if sql.CreateSqlTable("ttt2_items", ShopEditor.savingKeys) then
-		for i = 1, #itms do
-			local eq = itms[i]
+	local isSqlTableCreated = sql.CreateSqlTable("ttt2_items", ShopEditor.savingKeys)
 
-			ShopEditor.InitDefaultData(eq)
+	for i = 1, #itms do
+		local eq = itms[i]
 
+		InitDefaultEquipment(eq)
+		ShopEditor.InitDefaultData(eq)
+
+		if isSqlTableCreated then
 			local name = GetEquipmentFileName(WEPS.GetClass(eq))
 			local loaded, changed = sql.Load("ttt2_items", name, eq, ShopEditor.savingKeys)
 
@@ -447,38 +446,37 @@ function GM:InitPostEntity()
 			end
 		end
 
-		for i = 1, #sweps do
-			local wep = sweps[i]
+		CreateEquipment(eq) -- init items
 
-			ShopEditor.InitDefaultData(wep)
+		eq.CanBuy = {} -- reset normal items equipment
 
-			local name = GetEquipmentFileName(WEPS.GetClass(wep))
-			local loaded, changed = sql.Load("ttt2_items", name, wep, ShopEditor.savingKeys)
-
-			if not loaded then
-				sql.Init("ttt2_items", name, wep, ShopEditor.savingKeys)
-			elseif changed then
-				CHANGED_EQUIPMENT[#CHANGED_EQUIPMENT + 1] = {name, wep}
-			end
-		end
+		eq:Initialize()
 	end
 
-	for i = 1, #itms do
-		local itm = itms[i]
-
-		CreateEquipment(itm) -- init items
-
-		itm.CanBuy = {} -- reset normal items equipment
-
-		itm:Initialize()
-	end
+	local sweps = weapons.GetList()
 
 	for i = 1, #sweps do
-		local wep = sweps[i]
+		local eq = sweps[i]
 
-		CreateEquipment(wep) -- init weapons
+		-- Insert data into role fallback tables
+		InitDefaultEquipment(eq)
 
-		wep.CanBuy = {} -- reset normal weapons equipment
+		/*ShopEditor.InitDefaultData(eq)
+
+		if isSqlTableCreated then
+			local name = GetEquipmentFileName(WEPS.GetClass(eq))
+			local loaded, changed = sql.Load("ttt2_items", name, eq, ShopEditor.savingKeys)
+
+			if not loaded then
+				sql.Init("ttt2_items", name, eq, ShopEditor.savingKeys)
+			elseif changed then
+				CHANGED_EQUIPMENT[#CHANGED_EQUIPMENT + 1] = {name, eq}
+			end
+		end
+
+		CreateEquipment(eq) -- init weapons*/
+
+		eq.CanBuy = {} -- reset normal weapons equipment
 	end
 
 	-- init hudelements fns
@@ -522,6 +520,7 @@ function GM:InitPostEntity()
 	LoadShopsEquipment()
 
 	MsgN("[TTT2][INFO] Shops initialized...")
+	isShopFallbackInitialized = true
 
 	WEPS.ForcePrecache()
 
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -461,21 +461,6 @@ function GM:InitPostEntity()
 		-- Insert data into role fallback tables
 		InitDefaultEquipment(eq)
 
-		/*ShopEditor.InitDefaultData(eq)
-
-		if isSqlTableCreated then
-			local name = GetEquipmentFileName(WEPS.GetClass(eq))
-			local loaded, changed = sql.Load("ttt2_items", name, eq, ShopEditor.savingKeys)
-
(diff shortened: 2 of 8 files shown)


gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -1104,23 +1104,27 @@ function GM:OnContextMenuOpen()
 	end
 end
 
+function TTT2CacheEquipMaterials(item)
+	--if there is no material or model, the item should probably not be available in the shop
+	if item.material and item.material ~= "vgui/ttt/icon_id" then
+		item.ttt2_cached_material = Material(item.material)
```
TimGoll, line 1110: while you're at it, you should probably change the name in the TTT2 project. The material is names `.iconMaterial` for roles and I think it should be the same here. Also the model should have a smiliar name,

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -811,47 +816,24 @@ function AddToShopFallback(fallback, subrole, eq)
...
+	tbl[#tbl + 1] = base
 
-	-- mark custom items
-	for _, i in pairs(tbl) do
-		if i and i.id then
-			i.custom = not table.HasValue(DefaultEquipment[roleData.index], i.id) -- TODO
-		end
+	-- mark custom equipment
+	if base and base.id then
+		base.custom = not table.HasValue(DefaultEquipment[roleData.index], base.id) -- TODO
```
TimGoll, line 836: has anyone here an idea what this todo means? It should be removed and put into trello

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -881,38 +863,89 @@ end
 -- table.HasValue is still supported, but please just check if the key is not nil instead.
 -- @internal
 -- @realm shared
-local function CleanUpDefaultCanBuyIndices()
-	local itms = items.GetList()
+local function CleanUpDefaultCanBuyIndices(eq)
```
TimGoll, line 866: update doc for this function

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -881,38 +863,89 @@ end
...
 
-		wep.CanBuy = wep.CanBuy or {}
+	if addEquipment then
+		local tableHasValue = false
 
-		ValueToKey(wep.CanBuy)
+		for i = 1, tblSize do
+			if tbl[i].id == base.id then
+				tableHasValue = true
+				break
```
TimGoll, line 904: add empty line in front

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -881,38 +863,89 @@ end
...
+		if not tableHasValue then
+			tbl[tblSize + 1] = base
+		end
+	else
+		local counter = 0
+
+		for i = 1, tblSize do
+			if tbl[i].id == base.id then
+				continue
+			end
```
TimGoll, line 917: can be single line

gamemodes/terrortown/gamemode/shared/sh_main.lua
```diff
@@ -10,6 +10,137 @@ local MAX_DROWN_TIME = 8
...
+				if not istable(keys) then continue end
+
+				local continueOuterLoop = false
+				local counter = 0
+				local saveKey = keys[1]
+
+				for _, key in pairs(keys) do
+					if not isstring(key) or not oldKeyField then
+						continueOuterLoop = true
+						break
```
TimGoll, line 54: add empty line in front

gamemodes/terrortown/gamemode/shared/sh_main.lua
```diff
@@ -10,6 +10,137 @@ local MAX_DROWN_TIME = 8
...
+					counter = counter + 1
+					eqKeyField[key] = eqKeyField[key] or {}
+					oldKeyField = oldKeyField[key]
+
+					-- To keep eqKeyField as reference check for tables or create one
+					if counter < #keys and not istable(eqKeyField[key]) then
+						eqKeyField[key] = {}
+					elseif counter == #keys then
+						saveKey = key
+						break
```
TimGoll, line 68: add empty line in front

gamemodes/terrortown/gamemode/shared/sh_main.lua
```diff
@@ -10,6 +10,137 @@ local MAX_DROWN_TIME = 8
...
+
+				eqKeyField[saveKey] = oldKeyField
+			end
+		end
+
+		ResetDefaultEquipment(equipment)
+	end
+
+	if SERVER and sql.CreateSqlTable("ttt2_items", ShopEditor.savingKeys) then
+		if name == "weapon_ttt_demonicsheep" then print("3") end
```
TimGoll, line 86: do not put into single line, also add an empty line after the end, the same holds true for the following lines

gamemodes/terrortown/gamemode/shared/sh_main.lua
```diff
@@ -10,6 +10,137 @@ local MAX_DROWN_TIME = 8
...
+		elseif changed then
+			if name == "weapon_ttt_demonicsheep" then print("5") end
+			local counter = #CHANGED_EQUIPMENT + 1
+
+			if isShopFallbackInitialized then
+				if name == "weapon_ttt_demonicsheep" then print("6") end
+				for i = 1, #CHANGED_EQUIPMENT do
+					if CHANGED_EQUIPMENT[i][1] == name then
+						counter = i
+						break
```
TimGoll, line 101: add empty line in front

gamemodes/terrortown/gamemode/shared/sh_main.lua
```diff
@@ -10,6 +10,137 @@ local MAX_DROWN_TIME = 8
...
+					end
+				end
+			end
+			CHANGED_EQUIPMENT[counter] = {name, equipment}
```
TimGoll, line 105: leave empty line after `end`

TimGoll (changes requested):
Really nice, thanks for again for your awesome work. Besides the linter warnings, here are a few things that I've spotted

gamemodes/terrortown/gamemode/shared/sh_main.lua
ZenBre4ker: oops, thats a leftover debug statement ^^

TimGoll: now that I look at it again, it seems rather obvious that it is. :D

gamemodes/terrortown/gamemode/client/cl_equip.lua
ZenBre4ker: grafik
It would be material. But I could change that.

ZenBre4ker: Your ttt2 Items do use material though

TimGoll: > Your ttt2 Items do use material though

What do you mean?

ZenBre4ker: grafik

TimGoll: ah, but this is a path, not a real material. Bad name

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -56,20 +56,13 @@ EquipmentItems = EquipmentItems or setmetatable(
...
 -- @internal
 -- @realm shared
-function GetEquipmentBase(data, eq)
-	if not eq or eq.inited then
-		return eq
-	end
-
-	local name = WEPS.GetClass(eq)
-	if not name then return end
+function AddStandardKeyValues(eq, name)
```
saibotk, line 64: This method name should contain something equipment related (otherwise this sounds too generic but is actually specific to be used with equipment only)

(ideally those functions would be transformed into a module or something like that, but that might be out of the scope of this PR)

ZenBre4ker: I agree and changed the name, in general your trello mentioned, that sh_equip_items needs a refactor. With that, I atleast made some functions more clear in purpose. But the overloading of functions doing mutliple stuff at once is a bit weird. There is a base getter, that assigns the keyvalues, which is really weird. Didnt want to break anything, so I left it there. In general the getter shouldnt be abused to assign values to uninitialized equipment......

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
ZenBre4ker: https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/shared/sh_init.lua#L126-L127

Seems to come from here, there are several such todos, that apparently check for default roles.
grafik

gamemodes/terrortown/gamemode/client/cl_equip.lua
ZenBre4ker: So, I dont really know what this would do, if I change it. TTT2 only uses it in some places, so maybe do it in another refactor, so that this change doesnt introduce new bugs. ^^

TimGoll: Don't change it for now, this might break compatibility to other addons

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
TimGoll: ahh, hmm @Alf21 is this something that has to be done? Or can it be removed?

Alf21: custom marks the equipment as a custom created one (star or something similar). That doesn't make sense, as nearly every item is a custom one. The user has no advantage from it. That's why I added a TODO to think about it and use this already integrated feature in another way

TimGoll: Ah, I see
this is something that will be changed with the planned shop UI rework. Thanks! I'll write something in Trello so you @ZenBre4ker can remove the todo here

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -522,6 +505,7 @@ function GM:InitPostEntity()
 	LoadShopsEquipment()
 
 	MsgN("[TTT2][INFO] Shops initialized...")
+	isShopFallbackInitialized = true
```
saibotk, line 508: Can we rename those to `TTT2ShopFallbackInitialized` since they are global?

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -876,43 +870,99 @@ local function ValueToKey(tbl)
...
+		if diff > 0 then
+			for i = 1, diff do
+				table.remove(tbl)
+			end
+		end
+	end
+
+	-- mark custom equipment
+	if base and base.id then
+		base.custom = not table.HasValue(DefaultEquipment[roleData.index], base.id) -- TODO
```
saibotk, line 950: This todo can also be removed now

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -876,43 +870,99 @@ local function ValueToKey(tbl)
...
+
+			counter = counter + 1
+			tbl[counter] = tbl[i]
+		end
+
+		local diff = tblSize - counter
+
+		if diff > 0 then
+			for i = 1, diff do
+				table.remove(tbl)
```
saibotk, line 943: This seems to be missing parameters

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -876,43 +870,99 @@ local function ValueToKey(tbl)
...
+			end
+		end
```
saibotk, line 945: Why are you doing that / what should this do?
Should this remove the equipment from the table if it exists? 
If so then this should not be based on counters / the amount and instead remove the item once the match is found. (Also this seems to be wrong atm)

saibotk (changes requested):

gamemodes/terrortown/gamemode/server/sv_main.lua
ZenBre4ker: Good idea.

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
ZenBre4ker: Actually it just removes the last entry as far as I understood the documentation and it worked as I tested it. But I can check again ^^

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
ZenBre4ker: At the moment I wrote it, I wasnt aure how the table exactly was structured and therefore went the safe way and iterated over all items to remove them. But I guess this is unnecessary if you can rely on the fact that an item is at max one time inside the table. This generally cleans the ShopFallback up if after a hotreload the default loadout changes. Maybe you could do it more efficient by comparing shop-default changes before hotreload.
But this is generally more complicated than it needs to be, as I wasnt sure first why hotreloads dont work. Probably will rework this. Otherwise it only happens if you change a weapon live, so speed is not of the highest importance. This is safe but not efficient.

saibotk: I think thats fine with iterating over the table, but i do not get why you are sure that all instances of the equipment are at the end of the table and that you can just work with a counter. To me this can just be removed while looping through all items instead of continuing.

ZenBre4ker: Ahh okay, yeah I am still unsure about how tables are working inside lua and as it is an indexed table I was afraid, that when I remove an item, that there are gonna be missing indexes. So I wrote my own remover, that just skips all duplicated items and because the counter only increases when it is not continued, those places get filled with non-skipped items.
tbl[counter] = tbl[i]
so I only iterate once over it (like one iteration of bubble sort) and at the end, I know, that the rest is free for clearing. Removing at the end can be done in O(1) instead of O(n). And removing while iterating would lead to changing table size, which would also need to be handled with something else thats not a for-loop.

In general you could just remove those you find, but then the gmod table.remove function would do the same I did, but for the whole table for every item you find.

I thought this would the most efficient way, if you would need to remove more than one index.

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -876,43 +870,105 @@ local function ValueToKey(tbl)
...
 
-		ValueToKey(wep.CanBuy)
+		local diff = tblSize - counter
+
+		-- Remove last table entries
+		if diff > 0 then
+			for i = 0, diff - 1 do
+				table.remove(tbl, tblSize - i)
+			end
+		end
```
ZenBre4ker, line 951: Okay, added some documentation, hope this helps to understand. And added a second argument to table.remove to make it clear, that the last item is removed.

saibotk (approved):

TimGoll (approved):
Looks really nice. Good work as always!

PR #852, opened 2021-08-11
Convars: Change server convars clientside as admin

ZenBre4ker:
To be able to further improve our F1 menu, I extended the cvars library such that we can easily change serverside convars directly from the client.
Also implemented the first test for the slider and tested it for the randomshop, which before this change used "TTT2UpdateCVar", which only could change hardcoded registered convars for the shopeditor.

As I need some input on safety and general implementation, I stop it at this point and wait for your comments.
Also should we discuss if we change some areas of the F1 menu to superadmin only. Currently an admin can already change them, but in ulx serverside convars can only be changed by a superadmin. So maybe we should do that too.

This time I tried to make my commits stepwise and contain complete changes. Might help reviewing it.

Looking forward to what you say, may it be as harsh as can be 😜

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
@@ -195,28 +195,30 @@ end
 -- @param string cvar
 -- @realm client
 function PANEL:SetConVar(cvar)
-	if not cvar or cvar == "" then return end
+	if not cvar or cvar == "" or istable(cvar) then return end
 
-	if isstring(cvar) then
-		self.conVar = GetConVar(cvar)
+	self.conVar = GetConVar(cvar)
 
-		self:SetValue(self.conVar:GetFloat())
-		return
-	end
+	self:SetValue(self.conVar:GetFloat())
+end
 
-	if not istable(cvar) then return end
+---
+-- @param string cvar
+-- @realm client
+function PANEL:SetServerConVar(cvar)
+	if not cvar or not istable(cvar) or not isstring(cvar.name) then return end
 
 	self.serverConVar = cvar
 
-	cvars.RegisterServerConVar(cvar.name, cvar.type or "float", cvar.bitCount)
+	cvars.RegisterServerConVar(cvar.name, cvar.type or "string", cvar.bitCount)
 
 	local function OnReceiveFunc(wasSuccess, value)
 		if wasSuccess and value then
-			self:SetValue(value, true)
+			self:SetValue(tonumber(value), true)
 		end
 	end
 
-	cvars.ServerConVarGetValue(cvar.name, OnReceiveFunc, cvar.type or "float", cvar.bitCount)
+	cvars.ServerConVarGetValue(cvar.name, OnReceiveFunc, cvar.type or "string", cvar.bitCount)
 end
 
 ---
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
@@ -132,7 +132,7 @@ end
 ---
 -- @param any val
 -- @realm client
-function PANEL:SetValue(val)
+function PANEL:SetValue(val, ignoreConVar)
 	if not val then return end
 
 	val = math.Clamp(tonumber(val) or 0, self:GetMin(), self:GetMax())
@@ -143,6 +143,8 @@ function PANEL:SetValue(val)
 	self.m_fValue = val
 
 	self:ValueChanged(self.m_fValue)
+
+	if ignoreConVar then return end
 
 	if self.conVar then
 		self.conVar:SetFloat(self.m_fValue)
@@ -210,7 +212,7 @@ function PANEL:SetConVar(cvar)
 
 	local function OnReceiveFunc(wasSuccess, value)
 		if wasSuccess and value then
-			self:SetValue(value)
+			self:SetValue(value, true)
 		end
 	end
 
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
@@ -222,6 +222,7 @@ function PANEL:SetServerConVar(cvar)
 	local function OnServerConVarChangeCallback(conVarName, oldValue, newValue)
 		if not IsValid(self) then
 			cvars.RemoveChangeCallback(conVarName, "TTT2F1MenuServerConVarChangeCallback")
+
 			return
 		end
 
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
@@ -131,6 +131,7 @@ end
 
 ---
 -- @param any val
+-- @param bool ignoreConVar To avoid endless loops, separated setting of convars and UI values
 -- @realm client
 function PANEL:SetValue(val, ignoreConVar)
 	if not val then return end
@@ -151,7 +152,7 @@ function PANEL:SetValue(val, ignoreConVar)
 	end
 
 	if self.serverConVar then
-		cvars.ChangeServerConVar(self.serverConVar.name, val)
+		cvars.ChangeServerConVar(self.serverConVar, tostring(val))
 	end
 end
 
@@ -197,7 +198,7 @@ end
 -- @param string cvar
 -- @realm client
 function PANEL:SetConVar(cvar)
-	if not cvar or cvar == "" or istable(cvar) then return end
+	if not cvar or cvar == "" then return end
 
 	self.conVar = GetConVar(cvar)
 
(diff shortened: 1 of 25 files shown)


TimGoll:
I look into it later, but to answer your questions:
- in my opinion convars should be changeable for admins, but maybe someone like @TheNickSkater can chime in on this
- for security it is enough to check if the player is valid and admin
- we should probably add more new submenus to the admin section. I have some time today to discuss the structure later today

TheNickSkater:
> I look into it later, but to answer your questions:
> 
> * in my opinion convars should be changeable for admins, but maybe someone like @TheNickSkater can chime in on this
> * for security it is enough to check if the player is valid and admin
> * we should probably add more new submenus to the admin section. I have some time today to discuss the structure later today

I would say the administration section should be available for superadmins only tbh. Just because imo those settings belong to the actual highest administrator of the server.
Though maybe we could add a ULX Permission check for this?

TimGoll:
There will only be one rank. Either all settings are for admins, or all are for superadmins. And it will be completely independant from ulx. So no ulx permission check

TheNickSkater:
> There will only be one rank. Either all settings are for admins, or all are for superadmins. And it will be completely independant from ulx. So no ulx permission check

Hm okay, I would have found it a bit better to let the user decide. But in this case I would say superadmin should have it, not admin.

ZenBre4ker:
Okay, changed rights needed to superadmin. Also Checkboxes and Sliders are now compatible with serverConVars. Didnt do comboboxes yet, as I have nothing to test it with, yet.

Ready for review, I guess ^^

lua/ttt2/extensions/cvars.lua
```diff
@@ -0,0 +1,174 @@
...
+local cvars = cvars
+
+local messageIdentifier = -1
+local identityBitCount = 8
+local maxUInt = 2 ^ identityBitCount
+
+local serverConVars = {}
+local functionCache = {}
+
+local function IsAdmin (ply)
```
saibotk, line 19: I don't see the necessity for this function

lua/ttt2/extensions/cvars.lua
```diff
@@ -0,0 +1,174 @@
...
+end
+
+local cvars = cvars
+
+local messageIdentifier = -1
+local identityBitCount = 8
+local maxUInt = 2 ^ identityBitCount
+
+local serverConVars = {}
+local functionCache = {}
```
saibotk, line 17:
```suggestion
local functionCache = functionCache or {}
```

I guess this might otherwise be cleared on hot reload, but I'm not 100% sure.

saibotk (changes requested):
Thank you for your hard work!
I love the quality of your code, but i have some papercuts that we need to discuss.

A major thing this system misses are ConVar changes that are made manually or by scripts or other code since we don't update clients with the new server value as we do not hook into the changes of server cvars. This might lead to state issues (but is not overly dramatic because it is only in this edge case and does not really affect 99% use-cases).

Regarding my ttt2net suggestion I'm always available for a chat about its usage etc, msg me for details.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
```diff
@@ -144,9 +145,15 @@ function PANEL:SetValue(val)
...
 	self:ValueChanged(self.m_fValue)
 
+	if ignoreConVar then return end
+
 	if self.conVar then
 		self.conVar:SetFloat(self.m_fValue)
 	end
+
+	if self.serverConVar then
+		cvars.ChangeServerConVar(self.serverConVar, tostring(val))
```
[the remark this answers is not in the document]
ZenBre4ker: Generally to make the structure clear to other programmers, that copy the code. Because I didnt know, that convars were saved as strings and such conversions might help. The cvars-library makes the same conversion, so of course this is unnecessary. In case of the checkbox a boolean need to be converted to 1 or 0.
So depends what you think makes sense. I'd leave it inside to have an overall similar structure, but it's redundant.

lua/ttt2/extensions/cvars.lua
ZenBre4ker: yeah true. Same with serverConVars

lua/ttt2/extensions/cvars.lua
ZenBre4ker: It is not necessary. But we didnt know which rights we wanted to give, so I let that be easily changeable.
As we kinda discussed it should be superadmin, I can delete it.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
```diff
@@ -195,7 +202,33 @@ function PANEL:SetConVar(cvar)
...
+
+---
+-- @param string cvar
+-- @realm client
+function PANEL:SetServerConVar(cvar)
+	if not cvar or cvar == "" then return end
+
+	self.serverConVar = cvar
+
+	cvars.ServerConVarGetValue(cvar, function (wasSuccess, value)
```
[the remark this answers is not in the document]
ZenBre4ker: Yeah thats a point. I didnt want to hardcode the necessary convars and instead make it on demand only.
I will look into ttt2net, but generally I thought of something like a buffer, that gets checked every frame and and the end send all requests at once.
But it would only do this once per F1 menu, as they get cached. So gotta see, whats the easiest approach 😄

TimGoll:
> A major thing this system misses are ConVar changes that are made manually or by scripts or other code since we don't update clients with the new server value as we do not hook into the changes of server cvars. This might lead to state issues (but is not overly dramatic because it is only in this edge case and does not really affect 99% use-cases).

Interesting that you mention it, because it was my first criticism as well. However I said to Sven that I see his reasoning and his implementation is fine for me as well. But now that you see the same problem, I might rethink my waver on thin.

lua/ttt2/extensions/cvars.lua
```diff
@@ -0,0 +1,174 @@
...
+		net.Start("TTT2ServerConVarGetValue")
+		net.WriteUInt(identifier, identityBitCount)
+
+		local isSuccess = ConVarExists(conVarName) and IsValid(ply) and IsAdmin(ply)
+		net.WriteBool(isSuccess)
+
+		if isSuccess then
+			net.WriteString(conVarName)
+			net.WriteString(GetConVar(conVarName):GetString())
+		end
```
ZenBre4ker, line 170: And you are absolutely right about manual changes, forgot that on the serverside. I will implement a callback entry here, as it's only relevant if someone requested values.

ZenBre4ker (commented):
Generally I rather like an on-demand version. But I will see, if we can either work with a buffered send or make a fullstate kinda update with ttt2net.

lua/ttt2/extensions/cvars.lua
```diff
@@ -0,0 +1,174 @@
...
+
+		local conVarName = net.ReadString()
+		local value = net.ReadString()
+
+		if not IsValid(ply) or not IsAdmin(ply) then return end
+
+		net.Start("TTT2ChangeServerConVar")
+		net.WriteString(conVarName)
+		net.WriteString(value)
+		net.Broadcast()
```
ZenBre4ker, line 150: Changes done by others are synchronized though. Just wanted to point it out, because I changed it to a caching system, that documents every change ^^

lua/ttt2/extensions/cvars.lua
ZenBre4ker: If this is local has hotreloading any effect on it? Otherwise it stated that the local cache is either the global functionCache or empty table. I guess, this would be an error 🤔

TimGoll: if it is local, it can't be persistent while hotreloading

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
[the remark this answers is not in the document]
ZenBre4ker: Okay, the buffered requests should solve the problem.
Now only needed variables are sent to clients, that really need those values. Should minimize traffic I hope.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
```diff
@@ -195,7 +202,33 @@ function PANEL:SetConVar(cvar)
...
+		if wasSuccess and value then
+			self:SetValue(tonumber(value), true)
+		end
+	end)
+
+	local function OnServerConVarChangeCallback(conVarName, oldValue, newValue)
+		if not IsValid(self) then
+			cvars.RemoveChangeCallback(conVarName, "TTT2F1MenuServerConVarChangeCallback")
+			return
```
TimGoll, line 225: empty line in front of return is missing

lua/ttt2/extensions/cvars.lua
```diff
@@ -0,0 +1,243 @@
...
+
+	---
+	-- Get the value of the conVar if it exists on the server or was already cached
+	-- @param string conVarName
+	-- @param function OnReceiveFunc(conVarExists, value) The function that gets called with the results if the conVar exists
+	-- @realm client
+	function cvars.ServerConVarGetValue(conVarName, OnReceiveFunc)
+		if serverConVars[conVarName] then
+			OnReceiveFunc(true, serverConVars[conVarName])
+			return
```
TimGoll, line 114: missing empty line

TimGoll (approved):
just two minor things, nice

saibotk (approved):
I really like those changes!

Thank you very much!

PR #854, opened 2021-08-23
Weapons: Weapon spawn rework Fix 3D-Positioning

ZenBre4ker:
Apparently outside of render Hooks the global EyeAngles() is unreliable. It worked on Tim's Pc but not on mine. Don't really know why.
https://wiki.facepunch.com/gmod/Global.EyeAngles

Also let the traces start a bit outside of the walls to be safe not to hit the same spot again.


ZenBre4ker:
Wait, why are my old commits in changed files?

TimGoll:
you probably used the wrong commit to start your PR

PR #857, opened 2021-09-25
Role function: Changed HasSpecialRole

TheNickSkater:
Changed the "IsSpecial" or "HasSpecialRole" function to also return "false" if the player has "ROLE_NONE".
This also fixes an issue with the Clairvoyant role.


TimGoll (approved):
as dicussed previously, this is a good change and fixes stuff like this: https://github.com/TTT-2/ttt2-role_cv/issues/7

saibotk (approved):
Thank you very very much!

PR #864, opened 2021-10-06
Scoring: Kill distinction and survival penalty

TimGoll:
- As seen in the next two images, I added a destinction between the different scoring types of the kill event, to make it more clear that the -1 stems from suicide
image
image

- Also I added a new variable named `survivePenaltyMultiplier` that is used to give a score penalty for every player in a losing team, the score amount is dependant on the amount of alive players per team. Right now this is used to punish the jester for surviving
image


Histalek (approved):
LGTM

PR #867, opened 2021-10-08
Fix missing ids

ZenBre4ker:
If you register weapons manually before TTT2 is loaded, it can happen, that those weapons dont get any ids, that are necessary for the shop. Those are the minority. As normal weapons are going over the tttweapon_base anyways and need to be initialized in a terrortown folder, therefore wouldnt cause any issues.

This should fix the error of PAC3 
https://steamcommunity.com/sharedfiles/filedetails/?id=104691717
with the file 
PAC3\lua\pac3\extra\shared\hands.lua
Or anything that does the same manual registering in a file-loading with weapons.Register(SWEP).


TimGoll:
@saibotk please rereview

PR #868, opened 2021-10-08
Gamemode: Adding Role Info

DasDarki:
This is adding a small script which shows some information about the role. Following information get displayed:
- The team the role belongs to
- A custom info message (`roleinfo_text_ROLENAME`) or as fallback the `info_popup_ROLENAME` message

The type where the information gets displayed can be changed via a ConVar (0 = Chat, 1 = EPop; default = 1) as well as the time (for EPops).


TimGoll:
First of: Thank you for your contribution. But I have to say I'm not to sure if I like it. Maybe someone from the community wants to contribute to this discussion, but having a role popup each round might be quite annoying.

At some point I (or someone else) want to put roleinfo with gameplay tips and convar Info to the F1 menu. Maybe there is also a different place to put all of it as well. But th epopups might be more annoying than useful.

But I'm open for discussion.

In case we dismiss this PR, it would still be a great addon to TTT2.

TimGoll:
I'm closing this now. Thanks for the contribution anyway

PR #869, opened 2021-10-09
F1 Menu: Added Role Settings

TimGoll:
In this pullrequest I moved the role settings from the ULX menu to the F1 menu. To do so, I added a new menu to the F1 menu: "Role Settings":

image

Inside of this menu, each role has their own submenu with their convars:

image

Additionally there is a new submenu in the administration tab that contains general settings:

image
image

Moreover I added two hook to modify the role submenus. One general hook to add a custom form with custom convars:
```lua
---
-- This hook can be used by role addons to populate the role settings page
-- with custom convars. The parent is the submenu, where a new form has to
-- be added.
-- @param number role The role index of the role whose menu should be changed
-- @param table roleData The role data table
-- @param DPanel parent The parent panel which is the submenu
-- @hook
-- @realm client
function GM:TTT2ModifyRoleSettingsMenu(role, roleData, parent)

end
```

A second one was added to add more convars to the credit section to keep the menu structured:
```lua

---
-- This hook can be used by role addons to populate the role credit settings form
-- with custom convars. The parent is the credits form, where menu elements can be
-- directly added.
-- @param number role The role index of the role whose menu should be changed
-- @param table roleData The role data table
-- @param DPanel parent The parent panel which is the credits form
-- @hook
-- @realm client
function GM:TTT2ModifyRoleSettingsMenu_Credits(role, roleData, parent)

end
```

**The German translation is still missing, but probably unaffected from your review.**

### DIFF
--- a/lua/terrortown/entities/roles/detective/shared.lua
+++ b/lua/terrortown/entities/roles/detective/shared.lua
@@ -34,6 +34,8 @@ function ROLE:PreInitialize()
 end
 
 if CLIENT then
+	---
+	-- @ignore
 	function ROLE:AddToSettingsMenuCreditsForm(parent)
 		parent:MakeSlider({
 			serverConvar = "ttt_det_credits_traitordead",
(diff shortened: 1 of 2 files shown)


TimGoll:
done @saibotk

PR #908, opened 2021-10-31
Revival: Add Block Modes

TimGoll:
Change blocking revival mode from `true`/`false` to
  - `REVIVAL_BLOCK_NONE`: don't block the winning condition during the revival process [default, previously `nil`/`false`]
  - `REVIVAL_BLOCK_COUNT_AS_ALIVE`: only block the winning condition, if the player being alive would change the outcome [previously `true`]
  - `REVIVAL_BLOCK_UNTIL_ALIVE`: block the winning condition until the revival process is ended
  - the old arguments still work, they are automaticvally converted

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
@@ -246,8 +246,8 @@ net.Receive("TTT2RevivalUpdate_IsReviving", function()
 	LocalPlayer().isReviving = net.ReadBool()
 end)
 
-net.Receive("TTT2RevivalUpdate_IsBlockingRevival", function()
-	LocalPlayer().isBlockingRevival = net.ReadBool()
+net.Receive("TTT2RevivalUpdate_RevivalBlockMode", function()
+	LocalPlayer().revivalBlockMode = net.ReadReadUInt(2)
 end)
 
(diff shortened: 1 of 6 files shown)


saibotk (approved):

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -246,8 +246,8 @@ net.Receive("TTT2RevivalUpdate_IsReviving", function()
 	LocalPlayer().isReviving = net.ReadBool()
 end)
 
-net.Receive("TTT2RevivalUpdate_IsBlockingRevival", function()
-	LocalPlayer().isBlockingRevival = net.ReadBool()
+net.Receive("TTT2RevivalUpdate_RevivalBlockMode", function()
+	LocalPlayer().revivalBlockMode = net.ReadReadUInt(2)
```
AaronMcKenney, line 250: Pretty sure this is a typo: "net.ReadReadUInt(2)" should be "net.ReadUInt(2)".

TimGoll: huh, strange that I did not notice an error

Histalek (approved):
LGTM

PR #910, opened 2021-11-07
Role Enhancement: Called Policing Role Hook and Added Docs

TimGoll:
- all public policing roles now appear as detectives in the chat
- fixed that every policing player could be called to a corpse, this is now again restricted to alive only players
- added the hook `GM:TTT2CalledPolicingRole` that is called after all policing role players were called to a corpse
- fixed inconsistency between `.disabledTeamChatRecv` and `.disabledTeamChatRec`
- fixed non-public policing roles having hats and therefore confirming them
- small cole and docs cleanup

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
@@ -76,22 +76,22 @@ end
 
 -- Teamchat
 local function RoleChatMsg(sender, msg)
-	local senderIsSpectator = sender:GetTeam()
+	local senderTeam = sender:GetTeam()
 	local senderRoleData = sender:GetSubRoleData()
 
-	if senderIsSpectator == TEAM_NONE
+	if senderTeam == TEAM_NONE
 		or senderRoleData.unknownTeam
 		or senderRoleData.disabledTeamChat
-		or TEAMS[senderIsSpectator].alone
+		or TEAMS[senderTeam].alone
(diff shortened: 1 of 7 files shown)


gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
```diff
@@ -114,121 +120,132 @@ function GetPlayerFilter(pred)
...
-function GetRoleChatFilter(subrole, alive_only)
+function GetRoleChatFilter(subrole, aliveOnly)
 	if roles.GetByIndex(subrole).disabledTeamChat then
 		return {}
 	end
 
 	return GetPlayerFilter(function(p)
-		return p:IsRole(subrole) and not p:GetSubRoleData().disabledTeamChatRec and (not alive_only or p:IsTerror())
+		return p:IsRole(subrole)
+			and not p:GetSubRoleData().v
```
Histalek, line 215: i suppose you wanted to append the 'v' here?

Histalek (changes requested):

gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
TimGoll: oh man, thanks! I accidentally replaced instead of extending it. Oof

Histalek (approved):

saibotk (approved):

PR #912, opened 2021-11-08
Roleselection: Fix roleselection randomness of additional baseroles

ZenBre4ker:


gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -814,7 +814,7 @@ function roleselection.SelectRoles(plys, maxPlys)
 		}
 
 		-- insert selectable roles into the list. The order doesn't matter, players are chosen randomly and the roles are already filtered and limited
-		for subrole in pairs(selectableRoles) do
```
saibotk, line 817: See the comment above, this does not really fix the issue. 
The cause for this is to be found in the calculation of available roles and their amounts.

saibotk (changes requested):

PR #913, opened 2021-11-09
Roleselection: Fix available baseroles calculation by considering maxPlys when filtering

saibotk:
This patch potentially corrects the filtering of available baseroles.
When the system goes through all layers for each baserole it does not consider the amount of players that can take the role and thus how many players can already be assigned to a role.
With the new condition it will now correctly stop adding baseroles to the final baserole selection table (marked as available), if the amount is equal to the player amount. This ensures that layering actually influences the availability of base roles and thus won't lead to weird issues further down.
The final role assignment also implicitly assumes that the actual selection of what roles appear is already made in the filtering of available roles. This should now work as intended.

This patch still needs testing and verification.

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_roleselection.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_roleselection.lua
@@ -409,7 +409,14 @@ function roleselection.GetSelectableRolesList(maxPlys, rolesAmountList)
 
 	-- first of all, we need to select the baseroles. Otherwise, we would select subroles that never gonna be choosen because if the missing baserole
 	for i = 1, availableBaseRolesAmount do
-		if currentlyAvailableRolesAmount >= maxPlys or maxRoles and maxRoles <= curRoles or maxBaseroles and maxBaseroles <= curBaseroles or #availableBaseRolesTbl < 1 then break end -- if the limit is reached or no available roles left (could happen if removing available roles that weren't already selected in layered "or"-tables), stop selection
+		-- if the limit is reached or no available roles left (could happen if removing available roles that weren't already selected in layered "or"-tables), stop selection
+		if currentlyAvailableRolesAmount >= maxPlys
+		    or maxRoles and maxRoles <= curRoles
+			or maxBaseroles and maxBaseroles <= curBaseroles
+			or #availableBaseRolesTbl < 1
+		then
+			break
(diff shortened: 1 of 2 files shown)


ZenBre4ker (changes requested):
I like this a lot. Didnt think it could be so easy, but yeah considering players in filtering sounds great. ~~Only problem I see is that in one layer roles with a lower index (depending on how pairs works) are always chosen first. Should we maybe at least randomize order per layer?~~

ZenBre4ker (approved):
Well I overlooked the choosing of one layer, it is of course random. LGTM

ZenBre4ker:
Okay, taintedEnergy tested it and I did now too, it seems to work just the way we want layering to work. Good work ❤️

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -405,9 +405,11 @@ function roleselection.GetSelectableRolesList(maxPlys, rolesAmountList)
 		ROLE_INNOCENT
 	}
 
+	local currentlyAvailableRolesAmount = rolesAmountList[ROLE_TRAITOR] + rolesAmountList[ROLE_INNOCENT]
+
 	-- first of all, we need to select the baseroles. Otherwise, we would select subroles that never gonna be choosen because if the missing baserole
 	for i = 1, availableBaseRolesAmount do
-		if maxRoles and maxRoles <= curRoles or maxBaseroles and maxBaseroles <= curBaseroles or #availableBaseRolesTbl < 1 then break end -- if the limit is reached or no available roles left (could happen if removing available roles that weren't already selected in layered "or"-tables), stop selection
+		if currentlyAvailableRolesAmount >= maxPlys or maxRoles and maxRoles <= curRoles or maxBaseroles and maxBaseroles <= curBaseroles or #availableBaseRolesTbl < 1 then break end -- if the limit is reached or no available roles left (could happen if removing available roles that weren't already selected in layered "or"-tables), stop selection
```
TimGoll, line 412: maybe this if would be more readable if it were multi line?

TimGoll (approved):
Just two small remarks. The one I already commented on and in general we are caching stuff like `math.min` etc. But since you didn't touch those parts here, I say it is okay.

PR #914, opened 2021-11-09
Items: Fix hotreload

saibotk:
This patch moves the initialization code specific to items into the items.OnLoaded() function and adds the call to the function to the OnReloaded hook as we do with all other things like huds etc.

This needs discussion, since im not particularly happy about such gamemode stuff / dependencies on those functions inside the items module. But on the other hand it really is a function that is tied to the gamemode.
Another idea would be to only move the ITEM:Initialize call to the items.OnLoaded() function and not touching the ShopEditor related initializations at all.


TimGoll:
> This needs discussion, since im not particularly happy about such gamemode stuff / dependencies on those functions inside the items module. But on the other hand it really is a function that is tied to the gamemode.

I see wha you mean. But this is the same way the other modules do work. In my opinion this isn't a problem.

ZenBre4ker:
It does look a lot cleaner. I guess something similar would also work fine for weapons. The InitPostEntity shouldnt be used for such micromanagement of equipment anyway. 👍

TimGoll (approved):

ZenBre4ker (approved):

PR #917, opened 2021-11-14
Revival: Renamed Revival Flag

TimGoll:
After thinking about it for a while, I thing `REVIVAL_BLOCK_WHILE_REVIVING` is a more fitting name then `REVIVAL_BLOCK_UNTIL_ALIVE`


ZenBre4ker:
If I get this right REVIVAL_BLOCK_COUNT_AS_ALIVE had the problem, that if the reviving person would change the team after revival, the other team would still win.
So REVIVAL_BLOCK_UNTIL_ALIVE/REVIVAL_BLOCK_WHILE_REVIVING means, that you need to wait for the revival and team determination until a win can be certified right?

As it took me some time to get difference between those two states I actually dont like either names. As this usecase is quite specific, the name should probably be too.
So how about actually keeping the old states and instead determine a second block, that determines if team will definitely stay the same or if it can change?

TimGoll:
> If I get this right REVIVAL_BLOCK_COUNT_AS_ALIVE had the problem, that if the reviving person would change the team after revival, the other team would still win.

Yesn't. It counted them as alive. But if a wrath is reviving they are still counted as inno and therefore the round ends. A hack would be to force their role on death. But I don't like that approach. Therefore the new flag blocks the roundend until the revival process is done.

I'm confused. The names are very specific. 
`REVIVAL_BLOCK_COUNT_AS_ALIVE` counts them as alive and `REVIVAL_BLOCK_WHILE_REVIVING` blocks until the revival process is done.

I don't like the idea of having two states that do basically the same.

ZenBre4ker:
But I dont see a difference namewise between, COUNT_AS_ALIVE and while reviving.
This both states that as long the guy is not revived it should both block the win state.
While the true difference is, that one keeps the team check for win and the other blocks the whole win condition as long as the team isnt determined therefore COUNT_AS_ALIVE describes both states and WHILE_REVIVING additionally should say, that team checks are not done, because team/role is undetermined.

And yeah i probably didnt read the other PR or approved it, so i just came across this weird blocking right now ^^

TimGoll:
No, count as alive counts the as alive, but doesn't block the round perse. If only innocent players are alive, counting a further player as innocent doesn't block the round. The innocent win because there are only innocent alive.
The new flag even blocks the round if there is only a single team alive.

Feel free to suggest more fitting names. We discussed this in the programming channel on our discord and had our problems with it as well.

ZenBre4ker:
Yeah your description is correct. So my Ideas would be:
 REVIVAL_BLOCK_COUNT_AS_ALIVE =  REVIVAL_BLOCK_ALIVE_OLD_TEAM
 REVIVAL_BLOCK_UNTIL_ALIVE = REVIVAL_BLOCK_TEAM_UNCERTAIN
 
 or REVIVAL_BLOCK_COUNT_AS_ALIVE
 and REVIVAL_BLOCK_COUNT_AS_ALIVE_TEAM_UNCERTAIN
 
 Cause both old names suggest that the guy is counted as alive or blocks until is revived, which sounds like the same. Im just missing the point point, what they are blocking, not what the states actually do. Im fine with that. But names should make clear, what the difference is 😄

TimGoll:
Your name is way to specific. It says nowhere that a team change will happen, this is only a wrath thing. Therefore your first suggestion isn't really fitting imho.

But the second one is also confusing. Why do you focus on teams so much? It really just blocks until the revival process is done. It has nothing to do with teams. 
I see where you are coming from, but your names don't really describe what happens here. One counts them as alive, one blocks the round completely. I don't see how the teams are fitting here.

saibotk (approved):

ZenBre4ker:
The whole code here is about teams that are alive:
https://github.com/TTT-2/TTT2/pull/917/files#diff-064277ad93b0b71308d5b58a89a4448df6f0476478c8c8aedf979795d351d78aL1517-L1578

If one player has that flag, the win_none is instantly returned. Because you cant determine the team yet. I dont see another use of this. Maybe thats my problem. Only an uncertain team actually could block it. 🤔 Therefore the programmer should know that he uses that block, when he wants to be able to change the team on revival.

In the name I just dont see a difference between COUNT_AS_ALIVE and WHILE_REVIVING. It both states, that there is a revival-block, while one just counts him as alive with his own team and the other counts it with no team, therefore prohibiting that win check.

But if Im the only one that thinks that, change it. 😄 I just find it confusing and wrong. 🤷

TaintedEnergy:
I think REVIVAL_BLOCK_UNTIL_ALIVE is fine.

TimGoll:
> I think REVIVAL_BLOCK_UNTIL_ALIVE is better.

It's wrong though. If the revival process is ended, the round ends as well, even if the player is still dead

ZenBre4ker (approved):
As discussed, feel free to go what you like more. 😄 I was heard and thats enough ^^

PR #919, opened 2021-11-14
Weapon Pickup: Spam Pickup Fix

TimGoll:
Fixed that sometimes you pickup loads of weapons at once. This *should* fix it, but it has to be veriefied


gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1259,12 +1259,8 @@ function plymeta:Give(weaponClassName, bNoAmmo)
 
 	local wep = plymeta_old_Give(self, weaponClassName, bNoAmmo or false)
 
-	timer.Simple(0, function()
-		if not IsValid(self) then return end
-
-		self.forcedPickup = false
-		self.forcedGive = false
-	end)
```
ZenBre4ker, line 1267: Not really comfortable with removing the timer. As you are using the old give-function it seems like the whole code here could be just removed.
Or has forcedPickup and forcedGive and influence on the old give function?

ZenBre4ker (changes requested):
Unsure if this is a fix. Why was this done in the first place, do we even need to overwrite the give-function? Could it be possible that this is some leftover code, that now can be removed alltogether?

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: yes. The Give function internally calls the `GM:PlayerCanPickupWeapon` hook. The flags have to be set in order for it to work.

ZenBre4ker (approved):

TimGoll:
Nick reported back positively, so I'm all for merging it to the master so more people can test it

saibotk (approved):

PR #920, opened 2021-11-14
Weapon Pickup: Fixed Flying Weapons

### LINKED ISSUE #861: Pickup Entites: Problem when picking up entities in close range
## Version

The TTT2 version doesn't seem to matter. This problem occured ever since GMod added its own pickupsystem that we use, maybe even longer.

## Description

This is a pretty vague description for now, because it has to be investigated further. These information were collected together with @TheSuspect2306.

We noticed the bug when trying to pick up [a soda can](https://steamcommunity.com/sharedfiles/filedetails/?id=1815518231). This doesn't cause problems normally, however it breaks if the player is in really close range with the entity. It then tries to teleport the entity to the player. This is no behaviour that is added in the super soda addon.
In fact I noticed something similar when playing TTT. Sometimes, when picking up a weapon, it stays afloat in the air without being added to the inventory. This problem could be caused by the same bug. This has to be investigated more.

While investigating the super soda problem @TheSuspect2306 quickly noticed that returning false in `GM:PlayerCanPickupWeapon` prevents this problem. Now here lies the problem. We now know for a fact that [calling the hook manually](https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/server/sv_player_ext.lua#L1278-L1288) causes entities to move under certain conditions. However we don't know why. And we don't know the conditions. There is nothing in the hook that indicates such a behaviour (we should take a look at the registered hook list to confirm that no other addon initiates this teleporting). This has to be investigated further, maybe the guys over at Facepunch can help more with this.

The [relevant code in the super soda addon](https://github.com/TTT-2/ttt2-super-soda/blob/master/lua/terrortown/autorun/shared/sh_supersoda_handler.lua#L71) shows a usecase of the aforementioned function. While this specific bug could be fixed by moving the check further down, the underlying bug would still persist.

TimGoll:
This PR fixes #861 by moving the weapon teleportation code into a conditional check. It also removes the timer around the weapon drop. After some testing I came to the realization that the timer doesn't seem to make a difference. However it causes problems when switching between two weapons with the same class, because then the pickup fails since the weapon is dropped a tick later.
The solution here was to either add a single tick timer to the pickup function as well, or to remove the existing drop timer. I chose the latter here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed roleselection layering with base roles to ensure layer order is considered correctly when selecting roles
 - Fixed hotreloading items
 - Fixed random playermodel selection on map change not working
+- Fixes weapon pickup sometimes causing floating weapons
+- Fixes weapon pickup sometimes failing if a weapon with the same class as a weapon in the invetory should be picked up
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed hotreloading items
 - Fixed random playermodel selection on map change not working
 - Fixes weapon pickup sometimes causing floating weapons
-- Fixes weapon pickup sometimes failing if a weapon with the same class as a weapon in the invetory should be picked up
+- Fixes weapon pickup sometimes failing if a weapon with the same class as a weapon in the inventory should be picked up
 
 ### Changed
 
--- a/gamemodes/terrortown/gamemode/server/sv_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player_ext.lua
@@ -1255,7 +1255,6 @@ local plymeta_old_Give = plymeta.Give
 -- @realm server
 function plymeta:Give(weaponClassName, bNoAmmo)
 	self.forcedPickup = true
-	self.forcedGive = true
 
 	local wep = plymeta_old_Give(self, weaponClassName, bNoAmmo or false)
 
@@ -1263,7 +1262,6 @@ function plymeta:Give(weaponClassName, bNoAmmo)
 		if not IsValid(self) then return end
 
 		self.forcedPickup = false
-		self.forcedGive = false
 	end)
 
 	return wep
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed roleselection layering with base roles to ensure layer order is considered correctly when selecting roles
 - Fixed hotreloading items
 - Fixed random playermodel selection on map change not working
+- Fixes weapon pickup sometimes causing floating weapons
+- Fixes weapon pickup sometimes failing if a weapon with the same class as a weapon in the invetory should be picked up
```
ZenBre4ker, line 26:
```suggestion
- Fixes weapon pickup sometimes failing if a weapon with the same class as a weapon in the inventory should be picked up
```

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1296,11 +1294,15 @@ end
...
-	if not self:CanSafeDropWeapon(wep) then return end
+	if not self:CanSafeDropWeapon(wep) then
+		return false
+	end
 
 	self:AnimPerformGesture(ACT_GMOD_GESTURE_ITEM_PLACE)
 
 	WEPS.DropNotifiedWeapon(self, wep, false, keepSelection)
+
+	return true
```
ZenBre4ker, line 1305: Missing @return in description

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1313,12 +1315,14 @@ end
 -- @realm server
 function plymeta:CanPickupWeapon(wep, forcePickup, dropBlockingWeapon)
 	self.forcedPickup = forcePickup
+	self.isPickupProbe = true
```
ZenBre4ker, line 1318: As this is saved on the player it reads like the player is a Pickup Probe.
Maybe just add an argument to the hook or some other name?

ZenBre4ker (changes requested):

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: good point, for this variable I could use a hook parameter

ZenBre4ker (approved):
LGTM

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
```diff
@@ -28,7 +28,8 @@ local cv_ttt_detective_hats = CreateConVar("ttt_detective_hats", "0", {FCVAR_NOT
 -- @note Prevent @{Player}s from picking up multiple @{Weapon}s of the same type etc
 -- @param Player ply The @{Player} attempting to pick up the @{Weapon}
 -- @param Weapon wep The @{Weapon} entity in question
--- @param nil|number dropBlockingWeapon should the weapon stored in the same slot be dropped
+-- @param[opt] number dropBlockingWeapon should the weapon stored in the same slot be dropped
+-- @param[opt] boolean isPickupProble Set this to true to mark this hook run as probe
```
saibotk, line 32:
```suggestion
-- @param[opt] boolean isPickupProbe Set this to true to mark this hook run as probe
```

saibotk (changes requested):

saibotk (approved):

PR #921, opened 2021-11-14
Overheadicons: Color icon based on role color

saibotk:
This patch adjusts the color of overhead icons to match the info box role icon. It will now select white / black depending on the role's color.

image


TimGoll (approved):
nice!

ZenBre4ker (approved):
Didnt see that. Looked like a review in discord.

PR #923, opened 2021-11-15
Weapon Pickup: Fixed Pickup if Autopickup is Disabled

TimGoll:


saibotk (approved):

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1358,7 +1362,7 @@ function plymeta:SafePickupWeapon(wep, ammoOnly, forcePickup, dropBlockingWeapon
 	-- drop the weapon safely
 	if not InventorySlotFree(self, wep.Kind) and not self:CanSafeDropWeapon(wep) then return end
 
-	local ret, errCode = self:CanPickupWeapon(wep, forcePickup or true, dropBlockingWeapon)
```
ZenBre4ker, line 1361: forcePickup or true never worked.
should have been a check for nil before that 
if forcePickup == nil then
canPickup = true

but does it default now to true?

Histalek (approved):

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: It defaults to false as intended. At that time I though it would only replace nil which is stupid :D

PR #924, opened 2021-11-15
Overhead Icon: Made modifyable

TimGoll:
Added a hook to add, remove or modify the overhead icon:

```lua
---
-- Can be used to modify, disable or add an overhead icon of a player.
-- @param Player ply The player whose overhead icon should be modified
-- @param boolean shouldDrawDefault Defines if the overhead icon would draw normally
-- @return boolean Return false to prevent overhead icon draw
-- @return Material The material for the overhead icon (normally the role icon)
-- @return Color The color of the overhead icon backdrop
-- @hook
-- @realm client
function GM:TTT2ModifyOverheadIcon(ply, shouldDrawDefault)

end
```

Example usage in the identity desiguiser:
```lua
hook.Add("TTT2ModifyOverheadIcon", "ModifyDisguiserOHI", function(ply, shouldDrawDefault)
	if not ply:IsPlayer() or not ply:HasDisguiserTarget() then return end

	local client = LocalPlayer()
	local target = ply:GetDisguiserTarget()
	local rd = target:GetSubRoleData()

	local shouldShow = target:IsActive()
		and target:HasRole()
		and (not client:IsActive() or target:IsInTeam(client) or rd.isPublicRole)
		and not rd.avoidTeamIcons

	return shouldShow, target:GetSubRoleData().iconMaterial, target:GetRoleColor()
end)
```

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_targetid.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_targetid.lua
@@ -182,7 +182,7 @@ function GM:PostDrawTranslucentRenderables(bDrawingDepth, bDrawingSkybox)
 		local shouldDraw, material, color = hook.Run("TTT2ModifyOverheadIcon", ply, shouldDrawDefault)
 
 		if shouldDraw == false
-			or not shouldDrawDefault and not material and not color
+			or not shouldDrawDefault and not (material and color)
 		then continue end
 
 		DrawOverheadRoleIcon(ply, material or rd.iconMaterial, color or ply:GetRoleColor())
--- a/gamemodes/terrortown/gamemode/client/cl_targetid.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_targetid.lua
@@ -172,13 +172,18 @@ function GM:PostDrawTranslucentRenderables(bDrawingDepth, bDrawingSkybox)
 		local ply = plys[i]
 		local rd = ply:GetSubRoleData()
 
-		if ply:IsActive()
+		local shouldDrawDefault = ply:IsActive()
 			and ply:HasRole()
 			and (not client:IsActive() or ply:IsInTeam(client) or rd.isPublicRole)
 			and not rd.avoidTeamIcons
-		then
-			DrawOverheadRoleIcon(ply, rd.iconMaterial, ply:GetRoleColor())
-		end
+
+		local shouldDraw, material, color = hook.Run("TTT2ModifyOverheadIcon", ply, shouldDrawDefault)
+
+		if shouldDraw == false
+			or not shouldDrawDefault and not material and not color
+		then continue end
+
+		DrawOverheadRoleIcon(ply, material or rd.iconMaterial, color or ply:GetRoleColor())
 	end
 end
 
@@ -434,5 +439,18 @@ end
 -- @hook
 -- @realm client
 function GM:TTTModifyTargetTracedata(startpos, endpos)
+
+end
+
+---
+-- Can be used to modify, disable or add an overhead icon of a player.
+-- @param Player ply The player whose overhead icon should be modified
+-- @param boolean shouldDrawDefault Defines if the overhead icon would draw normally
+-- @return boolean Return false to prevent overhead icon draw
+-- @return Material The material for the overhead icon (normally the role icon)
+-- @return Color The color of the overhead icon backdrop
+-- @hook
+-- @realm client
+function GM:TTT2ModifyOverheadIcon(ply, shouldDrawDefault)
 
 end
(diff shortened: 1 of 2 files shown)


gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -172,13 +172,18 @@ function GM:PostDrawTranslucentRenderables(bDrawingDepth, bDrawingSkybox)
...
 			and not rd.avoidTeamIcons
-		then
-			DrawOverheadRoleIcon(ply, rd.iconMaterial, ply:GetRoleColor())
-		end
+
+		local shouldDraw, material, color = hook.Run("TTT2ModifyOverheadIcon", ply, shouldDrawDefault)
+
+		if shouldDraw == false
+			or not shouldDrawDefault and not material and not color
+		then continue end
```
ZenBre4ker, line 186: Shouldnt this be combined when one or the other is missing?
Not really sure if this works
```suggestion
		if shouldDraw == false
			or not shouldDrawDefault and not (material and color)
		then continue end
```

ZenBre4ker (approved):
Looks good beside that one point im not 100% sure about.

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -172,13 +172,18 @@ function GM:PostDrawTranslucentRenderables(bDrawingDepth, bDrawingSkybox)
...
-		if ply:IsActive()
+		local shouldDrawDefault = ply:IsActive()
 			and ply:HasRole()
 			and (not client:IsActive() or ply:IsInTeam(client) or rd.isPublicRole)
 			and not rd.avoidTeamIcons
-		then
-			DrawOverheadRoleIcon(ply, rd.iconMaterial, ply:GetRoleColor())
-		end
+
+		local shouldDraw, material, color = hook.Run("TTT2ModifyOverheadIcon", ply, shouldDrawDefault)
```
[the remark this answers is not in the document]
TimGoll: this is already done :P

gamemodes/terrortown/gamemode/client/cl_targetid.lua
TimGoll: your suggestion should use an `or`

TimGoll: I designed it in a way that both have to be set, otherwise it could end up in an OHI without icon or the other way around

ZenBre4ker: Okay, then assume this case:
shouldDrawDefault = false
material = MaterialX
color = nil
then it doesnt continue, because:
or not shouldDrawDefault = true
and not material = false
and (not color = true) = false , so it stays false

TimGoll: I think this is okay
if you only return one argument even though you are supposed to return two, this is fine

what should happen instead?

ZenBre4ker: oh okay in this case you'd be okay. Id thought color and material need to be given 😄

TimGoll: DId you mistype? Yes, both have to be returned, just as I said

ZenBre4ker: Well in the current state it is continued aka not drawn, when only both material and icon are missing. When only a material is given or only a color, then it is not continued and it will be drawn.
If that should be like that, then I resolve this conversation 😄

TimGoll: you're absolutely right!

Histalek (approved):

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -172,13 +172,20 @@ function GM:PostDrawTranslucentRenderables(bDrawingDepth, bDrawingSkybox)
...
-		then
-			DrawOverheadRoleIcon(ply, rd.iconMaterial, ply:GetRoleColor())
-		end
+
+		---
+		-- @realm client
+		local shouldDraw, material, color = hook.Run("TTT2ModifyOverheadIcon", ply, shouldDrawDefault)
+
+		if shouldDraw == false
+			or not shouldDrawDefault and not (material and color)
```
saibotk, line 185: Why the additional and check for material and color? Those do not make sense to me

saibotk (changes requested):

gamemodes/terrortown/gamemode/client/cl_targetid.lua
TimGoll: so that it is only used if both are returned. If only one or none are returned, the default OHI is used

PR #930, opened 2021-11-16
Spawneditor: Fixed giving spawneditor when weapons are cached

ZenBre4ker:
We removed a flag, that was used for cached weapons. As the spawneditor is given after caching, we need the forcedGive flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


TimGoll (approved):
Wow, I even forgot removing the old flag in here: https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/server/sv_weaponry.lua#L78

TimGoll:
Also I don't think skipping the changelog here is right, since it is after a release. You should add a changelog entry.

TimGoll:
LGTM

saibotk (approved):

Issue #904, opened 2021-10-23
spawn editor tool causes script errors


mexikoedi:
## Your version of TTT2 (mandatory)

- Workshop version from Steam

## Describe the bug (mandatory)

If you use "Start Spawn Edit" from the F1 menu under "Spawn Editor" from "Administration" and walk/fly around with the "Spawn Editor Tool" then you will get script errors in the client/server console.

This is the error:
[TTT2 (Base) - v0.10.2b] render.DrawSphere: Arguments #3 and/or #4 have invalid value!
  1. fn - gamemodes/terrortown/entities/weapons/weapon_ttt_spawneditor.lua:322
   2. unknown - lua/ulib/shared/hook.lua:109

## Context (please provide as much as you can)

- Collection link of your Server: https://steamcommunity.com/sharedfiles/filedetails/?id=1822673195

ZenBre4ker:
Might be division by zero, is my guess.
https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/entities/weapons/weapon_ttt_spawneditor.lua#L320

TimGoll:
nope, it is that the arguments 3 and 4 of `render.DrawSphere` have to be in a certain range. However there is no documentation on this. So I guessed it has to be greater than one. But probably it has to be greater than 3 or so?

ZenBre4ker:
But isnt (4 + whatever) always greater than 3 or so?

TimGoll:
yes, but my suspicion was that the distance can be negative? But I have no idea. Maybe the minimum is 5. It is hard to test because I could not reproduce it and there is no documentation

ZenBre4ker:
At least lua knows infinity
![grafik](https://user-images.githubusercontent.com/72046466/141976236-c433980b-e301-4d8a-9f38-ac0e09b82e7a.png)

TimGoll:
maybe you're right

ZenBre4ker:
Probably could limit the maxsteps to 50 or 100 anyways. As the quality can increase quite a lot for small distances 😄
Just need to see if gmod math can handle infinity. So max(inf, 50) is 50 ^^

ZenBre4ker:
That could work:
![grafik](https://user-images.githubusercontent.com/72046466/141977480-6f03abf0-7ab2-4ce2-a971-811968fc3a92.png)

TimGoll:
The go ahead and try it!

PR #931, opened 2021-11-16
Spawneditor: Fix too many steps (leading up to infinity)

ZenBre4ker:
It took some time, but after glitching through some walls it happened and I can confirm the error of #904 
I also then forced it to happen and that at least confirms my suspicion that drawSphere cant handle infinity.
```
previewData.distance3d = math.floor(previewData.distance3d)

if previewData.distance3d <= 0 then
	print(previewData.distance3d)
end
```

So this fix should handle that special case and also handle some framedrops, when being close to spheres. As this is an admin tool, I guess we dont need it to be super high fidelity. 🤔 
Opinions on only 10 steps shown in the picture on 2k resolution?
grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Fixed
 
 - Fixed `ply:Give(weapon)` to work again, when weapons are cached, fixing the spawneditor to work again
+- Fixed spawneditor not causing errors, when going through walls due to many steps
 
 ## [v0.11.0b](https://github.com/TTT-2/TTT2/tree/v0.11.0b) (2021-11-15)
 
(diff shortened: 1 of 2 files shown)


TimGoll:
Yes, I designed it in a way that there aren't too many triangles. However the max doesn't matter too much because it only really starts to grow if you're really close. So 10 is fine, 50 would be fine as well.

Also I think you can remove my math.max

saibotk (approved):

TimGoll (approved):

ZenBre4ker:
Please re-review!

5 far away
grafik
vs. 10 close up
grafik

I guess this range is enough for an admin tool, but Fps gain in my scenario maybe a 10% boost, difficult to say as the waterworld scene varies heavily between 30 fps and 60 fps with many spawns, but in general I think the quality should be sufficient.

saibotk (approved):

PR #933, opened 2021-11-16
Damage Overlay: Fix unscaled damage being shown

ZenBre4ker:
As the amount of damage cant be changed in the hook, we use now the info from dmginfo:GetDamage() instead and therefore take everything that is scaled due to hooks and armor and Karma into account.

Warning: This is untested, but should work in theory!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Fixed `ply:Give(weapon)` to work again, when weapons are cached, fixing the spawneditor to work again
 - Fixed spawneditor not causing errors, when going through walls due to many steps
+- Fixed unchanged or unscaled damage being sent to the client, leading to a wrongly working damage-overlay
 
 ## [v0.11.0b](https://github.com/TTT-2/TTT2/tree/v0.11.0b) (2021-11-15)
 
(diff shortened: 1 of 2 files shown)


TimGoll:
Are you sure that works? Imho we should change the hook to the `GM:PostEntityTakeDamage` hook. See: https://wiki.facepunch.com/gmod/GM:PostEntityTakeDamage

A similar fix was recently done to the hitnumbers addon: https://github.com/ief015/hitnumbers/blob/master/lua/autorun/server/sv_hitdamagenumbers.lua#L374

ZenBre4ker:
Yeah I did that for the addon ^^
It was necessary there because he had no access to the end of the playertakedamage-hook, but we get the same result at the end of gm:playertakedamage as you should get in postentitytakedamage.

But as I said, its just what I think. No tests involved, as I dont want to search for a testpartner 🤷

TimGoll:
> It was necessary there because he had no access to the end of the playertakedamage-hook, but we get the same result at the end of gm:playertakedamage as you should get in postentitytakedamage.

Okay. Is scale player damage included as well? Does this happen before this hook?

ZenBre4ker:
ScalePlayerDamage states, that "use @{GM:EntityTakeDamage} instead if you need to detect ALL damage."
https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/server/sv_player.lua#L933

TimGoll (approved):
I accept this PR because it doesn't break anything. If it is still not fixed after that, we can investigate my comments at a later point.

ZenBre4ker:
Sounds good :)

TimGoll:
Oh, you commented at the same time. Alright then :D

PR #938, opened 2021-11-18
ConVars: Fix theoretical indexing error at 0

ZenBre4ker:
First doing this as a draft, as there are some more mistakes I found while copying the code to my new module. Trying to remember them now, but this here could fix a theoretical error of locally cached/stored variables not being affected by serverside changes as they are not correctly addressed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Fixed
+
+- Fix serverConVars indexing 0 at tables (could cause unpredictable issues)
+
 ## [v0.11.2b](https://github.com/TTT-2/TTT2/tree/v0.11.2b) (2021-11-17)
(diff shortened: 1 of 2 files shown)


saibotk (approved):

saibotk:
For sure this needs some testing, but nonetheless if this fixes the issues we are seeing this is still something necessary to fix :)

ZenBre4ker:
Okay, this is the wrong issue to look at. Nonetheless, I think starting at 1 and ending at 255 in lua is still better. Dont know what weird edge cases can exist with 0-indices.

This is gtg for review.

TimGoll:
> Okay, this is the wrong issue to look at. Nonetheless, I think starting at 1 and ending at 255 in lua is still better. Dont know what weird edge cases can exist with 0-indices.
> 
> This is gtg for review.

I guess that it still works as a normal table, just not as an indexed iteratable table.

TimGoll (approved):

PR #939, opened 2021-11-18
Equipment Editor: Fix showing cached variables instead of synced ones

ZenBre4ker:
To have the inherited versions for the AddToSettingsMenu-function, we used the weapons.Get-function which cached an inherited copy of the original weapons table. But as synchronization only affects the original table not the copied cached one, we need to get it as late as possible.

I know that at this point we could also just check if the function is given, but I would rather have the option available to later patch in a default AddToSettingsMenu directly in the ttt_weaponbase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


TimGoll (approved):
LGTM

saibotk (approved):

PR #943, opened 2021-11-24
Panel Dropdown: Added ServerConVar and reworked whole structure

ZenBre4ker:
Okay, Im gonna test those now, but you can already give feedback to the whole restructuring 😄 

Changelog below:
### Added

- Reworked our simplified Dropdowns MakePanel `PANEL:MakeComboBox(data)` version
  - Added possibility to manipulate serverside ConVars with Dropdowns. Just add .serverConvar with the conVarName to the given data similar to .convar
  - .serverConVar and .conVar are also supported
  - data.choices can now be a table containing `{title, value, select, icon, additionalData}`
  - data.selectValue is added, use it instead of data.selectName to choose the value you set
  - data.selectTitle is added and shall replace data.selectName

### Breaking Changes

- Reworked Dropdowns Panel `DComboBoxTTT2` itself
  - `PANEL:AddChoice(title, value, select, icon, data)` now uses the second argument as value string for setting convars, use the fifth argument for special data instead
  - `PANEL:ChooseOption(value, index, ignoreConVar)` no longer chooses the displayed text, but the underlying value, which also sets conVars
- Reworked our simplified Dropdowns MakePanel `PANEL:MakeComboBox(data)` version
  - `data.OnChange(value, additionalData)` is now only called with the two important arguments, the value that e.g. convars are set and the additionalData

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
@@ -318,7 +318,7 @@ end
 
 ---
 -- @realm client
-function PANEl:CloseMenu()
+function PANEL:CloseMenu()
 	if IsValid(self.menu) then
 		self.menu:Remove()
 		self.menu = nil
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
@@ -336,13 +336,16 @@ function PANEL:SetValue(value, ignoreConVar)
 	self:ChooseOptionValue(value, ignoreConVar)
 end
 
+local convarTracker = 0
 ---
 -- @param Panel menu to set the value of
 -- @param string conVar name of the convar
 local function AddConVarChangeCallback(menu, conVar)
+	convarTracker = convarTracker % 1023 + 1
+
 	local function OnConVarChangeCallback(conVarName, oldValue, newValue)
 		if not IsValid(menu) then
-			cvars.RemoveChangeCallback(conVarName, "TTT2F1MenuConVarChangeCallback")
+			cvars.RemoveChangeCallback(conVarName, "TTT2F1MenuConVarChangeCallback" .. tostring(convarTracker))
 
 			return
 		end
@@ -350,7 +353,7 @@ local function AddConVarChangeCallback(menu, conVar)
 		menu:SetValue(newValue, true)
 	end
 
-	cvars.AddChangeCallback(conVar, OnConVarChangeCallback, "TTT2F1MenuConVarChangeCallback")
+	cvars.AddChangeCallback(conVar, OnConVarChangeCallback, "TTT2F1MenuConVarChangeCallback"  .. tostring(convarTracker))
 end
 
 ---
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
@@ -154,7 +154,7 @@ function PANEL:AddChoice(title, value, select, icon, data)
 	self.valueIndices[value] = index
 
 	if select then
-		self:ChooseOption(nil, index, true)
+		self:ChooseOptionID(index, true)
 	end
 
 	return index
@@ -333,7 +333,7 @@ end
 -- @param bool ignoreConVar To avoid endless loops, separated setting of convars and UI values
 -- @realm client
 function PANEL:SetValue(value, ignoreConVar)
-	self:ChooseOption(value, nil, ignoreConVar)
+	self:ChooseOptionValue(value, ignoreConVar)
 end
 
 ---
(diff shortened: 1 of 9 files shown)


TimGoll (approved):
LGTM. However, if you allow .conVar and .serverConVar for drop down menus, you should do the same for all other menu items as well

ZenBre4ker:
Yeah true, but wanted to make changes solely on dropdowns. We can allow them on the others later too.

DONT MERGE! As those are only theoretical changes and Im gonna test all menus before merging.

ZenBre4ker:
LGTM, good to go, when tk or histalek agree

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
```diff
@@ -97,33 +126,94 @@ function PANEL:PerformLayout()
...
+	return index
+end
+
+---
+-- Choose either value or index, title is set 
+-- @param[opt] string/number value should be indexable
+-- @param[opt] number index the option id
+-- @param[default=false] bool ignoreConVar To avoid endless loops, separated setting of convars and UI values
+-- @realm client
+function PANEL:ChooseOption(value, index, ignoreConVar)
```
saibotk, line 169: I don't really like the design on those two parameters here. 
Either use another function specifically for choosing by value which looks up the index and then calls this function with the index or solve it in another way that does not include using two parameters where only one is required, this feels weird.

Just saw that this is the base function for others that already exist.
Please still rewrite them to use eg. the index function as a base and let the others only handle converting the value to the right index.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
```diff
@@ -97,33 +126,94 @@ function PANEL:PerformLayout()
...
+		index = self:GetOptionId(value)
+
+		if not isnumber(index) then return end
+	end
 
 	self.selected = index
-	self:OnSelect(index, value, self.data[index])
+
+	self:SetText(self:GetOptionText(index))
+	self:OnSelect(index, self:GetOptionValue(index), self:GetOptionData(index))
```
saibotk, line 181: why is this using 3 functions to extract the data from the choice instead of just getting the choice and access the data directly?

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
```diff
@@ -252,47 +319,142 @@ end
...
+		if not IsValid(menu) then
+			cvars.RemoveChangeCallback(conVarName, "TTT2F1MenuConVarChangeCallback")
+
+			return
+		end
+
+		menu:SetValue(newValue, true)
+	end
+
+	cvars.AddChangeCallback(conVar, OnConVarChangeCallback, "TTT2F1MenuConVarChangeCallback")
```
saibotk, line 351: "TTT2F1MenuConVarChangeCallback" is not unique enough, eg. when having the same variable multiple times for whatever reason, only one callback is registered. Maybe add a unique number or identifier.

saibotk (changes requested):

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
ZenBre4ker: I mainly tried to stick to the original, can remove all old behaviour nonetheless:
https://wiki.facepunch.com/gmod/DComboBox:ChooseOption

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
ZenBre4ker: Also old structure
https://wiki.facepunch.com/gmod/DComboBox:OnSelect

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
TimGoll: Is this the GMod provided function? You can leave the identifier away if you don't plan on removing the callback again

ZenBre4ker: I need the callback identifier as I remove the function, when the dropdown is inactive.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
ZenBre4ker: ah 3 functions because like this i can change the structure anytime. those are the getters

ZenBre4ker:
Your comments made me remember, that those functions are deprecated.

But the callback doesnt need another identifier. Thats a possibility, thats just weird. I couldnt really track those unless I do something globally.

TimGoll:
> But the callback doesnt need another identifier. Thats a possibility, thats just weird. I couldnt really track those unless I do something globally.

I agree with you there. This is such a rare edge cases. And even *if* it happens, the worst that happens is that the convar isn't updated

ZenBre4ker:
Actually the convar gets updated, but all but one of the dropdowns  of the same convar isnt refreshed onchange

TimGoll:
> Actually the convar gets updated, but all but one of the dropdowns of the same convar isnt refreshed onchange

This is what I meant. All but one convars in the UI aren't updated.

ZenBre4ker:
Forgot that local variables in a file are also "global" to all dropdowns. So here you go 😄

TimGoll:
@saibotk You said we should rerequest your review here to it is on your todo list. But we can't since you already are requested.

saibotk (approved):

PR #956, opened 2021-12-19
Addon Checker: Added Mounted State Checking

Valafi:
This fixes the addon compatibility detection so that only enabled/mounted addons are checked against the incompatible list.


saibotk (approved):

saibotk:
Thank you very much, this always annoyed me when testing locally :D 

:rocket:

Histalek (approved):
Very nice!

PR #961, opened 2022-03-15
F1 Menu: Add check for table of valid usergroups for IsAdminMenu access

Wryyyong:
This change, as the title explains, rewrites `CLGAMEMODEMENU:ShouldShow()` to check if the table `adminMenuAccess` exists globally, when necessary.

`adminMenuAccess` is intended to be a user-created table -- created by the server owner/maintainer/developer/etc. -- containing a list of usergroups permitted to access the gamemode's administration menus, instead of access being restricted solely to `superadmin`.

Said table must exist on the client, so ideally it should be kept in its own Lua script (eg. `lua/autorun/client/adminMenuAccess.lua`).

---

- If true, then the function uses the existing table.
- If false, the function creates the table locally as a fallback, containing only `superadmin`.

---

- If `self:IsAdminMenu() == true`, then iterate through the table of usergroups checking if `LocalPlayer()` is in any of them.
  - Return `self:HasVisibleSubmenus()` succesfully upon finding a valid usergroup.
  - Otherwise return `false`.
- If `self:IsAdminMenu() == false`, then simply return `self:HasVisibleSubmenus()` as normal.

---

Feel free to adjust as necessary, as long as the core purpose remains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
   - data.choices can now be a table containing `{title, value, select, icon, additionalData}`
   - data.selectValue is added, use it instead of data.selectName to choose the value you set
   - data.selectTitle is added and shall replace data.selectName
+- `CLGAMEMODEMENU:ShouldShow()` now checks for global table `adminMenuAccess` when `self:IsAdminMenu() == true`
+  - `adminMenuAccess` should contain a list of usergroups that can access admin menus
+  - This table must exist on the client, so ideally it should be kept in its own Lua script (eg. `lua/autorun/client/adminMenuAccess.lua`)
+  - If `adminMenuAccess` does not exist globally, then the function will fallback on `superadmin` as a default value
 
 ### Breaking Changes
 
--- a/lua/terrortown/menus/gamemode/base_gamemodemenu.lua
+++ b/lua/terrortown/menus/gamemode/base_gamemodemenu.lua
@@ -21,7 +21,13 @@ CLGAMEMODEMENU.submenus = {}
 -- @internal
 -- @realm client
 function CLGAMEMODEMENU:ShouldShow()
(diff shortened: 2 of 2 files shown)


TimGoll:
> to check if the table `adminMenuAccess` exists globally

What's the purpose of this? Could you explain where this comes from? I've never seen such thing in gmod

Wryyyong:
> > to check if the table `adminMenuAccess` exists globally
> 
> What's the purpose of this? Could you explain where this comes from? I've never seen such thing in gmod

Sorry, I should've specified that it's supposed to be a user-created table containing a list of usergroups that the server owner/maintainer/etc. wishes to grant access to admin menus, instead of being restricted to only `superadmin` having access.

In simpler terms, an example would be the server owner wishing to grant a technically lower-ranked usergroup than `superadmin` access to said menus; said usergroup(s) could be named `developer`, `headadmin` or any other staff position they consider trustworthy enough to handle with the gamemode's admin settings.

I'll edit the main post to specify this.

Wryyyong:
> If we really need something like this in the basecode, it should be done with a hook that can be documented, a random table somewhere in the code is not userfriendly.

A hook is definitely not a bad idea. In all honesty, I wanted to keep the commit brief and small-scoped, so I hadn't considered that idea. I'll look into it more when I have time tomorrow.

Wryyyong:
Just gonna close this for now; after further testing, it appears that granting usergroups access via the way I did it doesn't allow reading/writing to cvars (ie. all booleans are false (i think?), all slider values are set dead in the middle, and any changes made from the menus will not be applied or saved).

TimGoll:
Just a hint: All security checks use the `PLAYER:IsSuperAdmin()` check. This is why I suggested that you should overwrite that function (on the server and the client)

Wryyyong:
Shifting over to a new pull req on a new branch, as I feel this req is too sloppy.

Took me a while to get back to this one.

https://github.com/TTT-2/TTT2/pull/969

PR #967, opened 2022-04-21
Base Ammo Entity: Fix collision bounds blocking +use traces by using trigger bounds instead

### LINKED ISSUE #966: Ammo entities block +use near other entites
https://github.com/Facepunch/garrysmod/commit/99afa3c

TTT2 uses an older version of base_ammo_ttt.lua where this bug is not fixed yet

wgetJane:
fixes #966


saibotk (approved):
Looks good to me thx 🙏

TimGoll (approved):

PR #968, opened 2022-04-21
Karma multiplier hook

Rorkh:


TimGoll:
Check the output of the automatic checks as well (changelog etc)

Also (and maybe @Alf21 wants to give his two cents here as well) we changed the style of these multipliers to be modifyable. Your hook has the problem that only one addon is able to change it. We prefer it as a table, so that you have a reference that can be modified in multiple addons. See this hook as an example: https://api-docs.ttt2.neoxult.de/hook/GM/shared/GM:TTTPlayerSpeedModifier

PR #969, opened 2022-04-23
Security: Add new GM:TTT2AdminCheck hook, replace all IsSuperAdmin() checks with hook

Wryyyong:
```lua
---
-- A hook that is called whenever the gamemode needs to check if the player is in the superadmin usergroup.
-- This hook can be used to allow custom usergroups through these checks.
-- @note This hook grants access to powerful functionality, such as the gamemode configuration, damage logs and player role information. Only allow usergroups that absolutely need such access.
-- @param Player ply The player to be checked
-- @return boolean if the player is a valid usergroup
-- @hook
-- @realm shared
```

- Hook was added in `sh_main.lua` as it was the best place I could find for such a hook. If anyone else knows of a better spot for it, please tell me.
- Hook will return `ply:IsSuperAdmin()` by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
   - data.selectValue is added, use it instead of data.selectName to choose the value you set
   - data.selectTitle is added and shall replace data.selectName
 - New setting to disable session limits entirely. (by @Reispfannenfresser)
+- Added `GM:TTT2AdminCheck` hook
+  - Replaced all `IsSuperAdmin()` checks with this hook
+  - This hook can be used to allow custom usergroups through these checks
 
 ### Breaking Changes
 
--- a/gamemodes/terrortown/gamemode/server/sv_admin.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_admin.lua
@@ -47,7 +47,9 @@ end
 -- @param Player ply
 -- @realm server
 function PrintTraitors(ply)
-	if not IsValid(ply) or ply:IsSuperAdmin() then
+	---
+	-- @realm server
+	if not IsValid(ply) or hook.Run("TTT2AdminCheck", ply) then
 		ServerLog(Format("%s used ttt_print_traitors\n", IsValid(ply) and ply:Nick() or "console"))
 
 		local pr = GetPrintFn(ply)
@@ -84,7 +86,9 @@ concommand.Add("ttt_print_usergroups", PrintGroups)
 function PrintReport(ply)
 	local pr = GetPrintFn(ply)
 
-	if not IsValid(ply) or ply:IsSuperAdmin() then
+	---
+	-- @realm server
+	if not IsValid(ply) or hook.Run("TTT2AdminCheck", ply) then
 		ServerLog(Format("%s used ttt_print_adminreport\n", IsValid(ply) and ply:Nick() or "console"))
 
 		for _, e in pairs(SCORE.Events) do
@@ -110,7 +114,9 @@ concommand.Add("ttt_print_adminreport", PrintReport)
 local function PrintKarma(ply)
 	local pr = GetPrintFn(ply)
 
-	if not IsValid(ply) or ply:IsSuperAdmin() then
+	---
+	-- @realm server
+	if not IsValid(ply) or hook.Run("TTT2AdminCheck", ply) then
 		ServerLog(Format("%s used ttt_print_karma\n", IsValid(ply) and ply:Nick() or "console"))
 
 		KARMA.PrintAll(pr)
@@ -178,7 +184,9 @@ local dmglog_save = CreateConVar("ttt_damagelog_save", "0", {FCVAR_NOTIFY, FCVAR
 local function PrintDamageLog(ply)
 	local pr = GetPrintFn(ply)
 
-	if not IsValid(ply) or ply:IsSuperAdmin() or GetRoundState() ~= ROUND_ACTIVE then
+	---
+	-- @realm server
+	if not IsValid(ply) or hook.Run("TTT2AdminCheck", ply) or GetRoundState() ~= ROUND_ACTIVE then
 		ServerLog(Format("%s used ttt_print_damagelog\n", IsValid(ply) and ply:Nick() or "console"))
 		pr("*** Damage log:\n")
 
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -1401,8 +1401,10 @@ hook.Add("PlayerAuthed", "TTT2PlayerAuthedSharedHook", function(ply, steamid, un
 end)
 
 local function ttt_roundrestart(ply, command, args)
+	---
 	-- ply is nil on dedicated server console
-	if not IsValid(ply) or ply:IsAdmin() or ply:IsSuperAdmin() or cvars.Bool("sv_cheats", 0) then
+	-- @realm server
+	if not IsValid(ply) or ply:IsAdmin() or hook.Run("TTT2AdminCheck", ply) or cvars.Bool("sv_cheats", 0) then
 		LANG.Msg("round_restart")
 
 		StopRoundTimers()
--- a/gamemodes/terrortown/gamemode/server/sv_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player_ext.lua
@@ -1466,8 +1466,10 @@ local function SetPlayerReady(_, ply)
 
 	entspawnscript.TransmitToPlayer(ply)
 
+	---
 	-- update playermodels on the client
-	if ply:IsSuperAdmin() then
+	-- @realm server
+	if hook.Run("TTT2AdminCheck", ply) then
 		playermodels.StreamModelStateToSelectedClients(false, ply)
 	end
 
--- a/gamemodes/terrortown/gamemode/shared/sh_main.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_main.lua
@@ -336,6 +336,18 @@ function GM:TTT2PlayerSprintMultiplier(ply, sprintMultiplierModifier)
 
 end
 
+---
+-- A hook that is called whenever the gamemode needs to check if the player is in the superadmin usergroup.
+-- This hook can be used to allow custom usergroups through these checks.
+-- @note This hook grants access to powerful functionality, such as the gamemode configuration, damage logs and player role information. Only allow usergroups that absolutely need such access.
+-- @param Player ply The player to be checked
+-- @return boolean if the player is a valid usergroup
+-- @hook
+-- @realm shared
+function GM:TTT2AdminCheck(ply)
+	return ply:IsSuperAdmin()
+end
+
 -- Drowning and such
 local tm, ply, plys
 
--- a/lua/terrortown/menus/gamemode/base_gamemodemenu.lua
+++ b/lua/terrortown/menus/gamemode/base_gamemodemenu.lua
@@ -21,7 +21,9 @@ CLGAMEMODEMENU.submenus = {}
 -- @internal
 -- @realm client
 function CLGAMEMODEMENU:ShouldShow()
-	if not LocalPlayer():IsSuperAdmin() and self:IsAdminMenu() then
+	---
+	-- @realm client
+	if self:IsAdminMenu() and not hook.Run("TTT2AdminCheck", LocalPlayer()) then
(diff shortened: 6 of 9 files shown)


ZenBre4ker (changes requested):
So our team mostly thinks that this hook is acceptable as every addon could theoretically overwrite the superadmin function anyway. Having a ttt2-own security check doesnt make it unsafe, but makes it easier to appoint ttt-admins, that can only change ttt-specific settings. 👍 
Only thing I'd point out is, that hooks are a weird way to do the check. It's pretty easy for an addon to just return true for a specified user or just everyone everytime. So I'd suggest to at least state at startup to show every addon, that adds a hook to this, enabling a server owner to identify possible malicious addons.
And yes it is possible to add hooks at later points, but maybe we should block those or somehow force the user to add them at a specific time.

Wryyyong:
Aye, true points. I'll see what I can do.

Wryyyong:
Added TTT2AdminCheckWatch module (`sh_admincheckwatch.lua`).

This module takes inspiration from the existing `addonChecker` module.

```lua
---
-- Lists hooks added to GM:TTT2AdminCheck on startup, and locks the hook from further additions after initialization.
-- @author Wryyyong
-- @module TTT2AdminCheckWatch
```

* Copies `hook.Add` to a new function before overwriting it, with the new function passing its params onto the original if it passes the check.
```lua
if not hook._Add then
	hook._Add = hook.Add
end

function hook.Add(event_name, name, func)
	if event_name == "TTT2AdminCheck" and TTT2AdminCheckLocked then return end
	hook._Add(event_name, name, func)
end
```

* `TTT2AdminCheckLocked` is a boolean that gets set to `true` at the end of the gamemode's initialization.
* A report of successful hooks is printed to the server console immediately afterwards, along with the filepaths of each hook's script of origin. This report can be reprinted at anytime via the `admincheckreport` command.

Example output:
```
TTT2ADMINCHECK REPORT
=============================================================

TestHookName | lua/autorun/testhookscript.lua

=============================================================
This is the end of the report output.
```

---
If necessary, I can also rewrite the module to be more, well, *modular*. As in, to watch and report for a table of multiple hookable functions, instead of just a single func.

ZenBre4ker:
I like this a lot, one thing I find rather annoying though is just the filepath. I am not sure if it is possibble to also get the addon-name in case it is from the workshop and not your addonsfolder, but that would be better. 🤔

I roughly have gone over it and find it acceptable, but I will take a more detailed look in the next days ^^
Good work so far 👍

Wryyyong:
> one thing I find rather annoying though is just the filepath. I am not sure if it is possibble to also get the addon-name in case it is from the workshop and not your addonsfolder, but that would be better. 🤔

[debug.getinfo](https://wiki.facepunch.com/gmod/debug.getinfo) isn't that verbose, sadly, and it was the only thing I've find thus far -- in both standard Lua and GLua documentation -- that can even trace the filepath origin of a function.

If there *is* a function or other method that can perform this or somehow trace it to Workshop addons specifically, then it's doing a very good job of eluding me.

TimGoll:
I like this as well. Just one small note: We copy those functions into a local variable, maybe you should do this as well to be consistent to our style

Something like `local oldHookAdd = hook.Add`

Wryyyong:
> I like this as well. Just one small note: We copy those functions into a local variable, maybe you should do this as well to be consistent to our style
> 
> Something like `local oldHookAdd = hook.Add`

Ah, good catch

gamemodes/terrortown/gamemode/shared/sh_admincheckwatch.lua
```diff
@@ -0,0 +1,68 @@
...
+local oldHookAdd = oldHookAdd or hook.Add
+
+---
+-- Checks the contents of event_name and whether or not initialization has been completed before passing the arguments onto the original function.
+-- @param string event_name The name of the event to hook onto
+-- @param string name The unique ID for the hook
+-- @param function func The function to execute when the hook is called
+-- @realm shared
+function hook.Add(event_name, name, func)
+	if event_name == "TTT2AdminCheck" and TTT2AdminCheckLocked then return end
```
ZenBre4ker, line 25: Maybe throw an error, so the user knows why or when he cant add a hook to it.

Also use an empty line after end.

gamemodes/terrortown/gamemode/shared/sh_admincheckwatch.lua
```diff
@@ -0,0 +1,68 @@
...
+-- Copy the existing hook.Add function to oldHookAdd if the latter does not already exist.
+local oldHookAdd = oldHookAdd or hook.Add
+
+---
+-- Checks the contents of event_name and whether or not initialization has been completed before passing the arguments onto the original function.
+-- @param string event_name The name of the event to hook onto
+-- @param string name The unique ID for the hook
+-- @param function func The function to execute when the hook is called
+-- @realm shared
+function hook.Add(event_name, name, func)
```
ZenBre4ker, line 24: Normally we use names without underscore.


```suggestion
function hook.Add(eventName, name, func)
```

ZenBre4ker, line 24: Normally we use names without underscore.


```suggestion
function hook.Add(eventName, name, func)
```

gamemodes/terrortown/gamemode/shared/sh_admincheckwatch.lua
```diff
@@ -0,0 +1,68 @@
...
+if SERVER then
+	---
+	-- Prints all hooks tied to GM:TTT2AdminCheck, along with the filepath of the hook's origin.
+	-- @realm server
+	function TTT2AdminCheckWatch.Report()
+		local tbl
+		tbl = hook.GetTable()
+		tbl = tbl["TTT2AdminCheck"]
+
+		local keys = table.GetKeys(tbl)
```
ZenBre4ker, line 46: Please use full descriptive names, what are the keys representing?

I guess the hookNames here.

ZenBre4ker, line 46: Please use full descriptive names, what are the keys representing?

I guess the hookNames here.

gamemodes/terrortown/gamemode/shared/sh_admincheckwatch.lua
```diff
@@ -0,0 +1,68 @@
...
+		print("TTT2ADMINCHECK REPORT")
+		print("=============================================================")
+		print("")
+
+		for _ = 1, #keys do
+			local k = keys[_]
+			local d = debug.getinfo(tbl[k], S)
+			local v = d["short_src"]
+
+			print(k .. " | " .. v)
```
ZenBre4ker, line 58: Same for those here. Be descriptive or reading code later will be hard.

Why do you use short_src? Would there be any worth in using the full source name? In the wiki I only see an @, so I'm not sure.

ZenBre4ker, line 58: Same for those here. Be descriptive or reading code later will be hard.

Why do you use short_src? Would there be any worth in using the full source name? In the wiki I only see an @, so I'm not sure.

ZenBre4ker (changes requested):
Well beside those nitpicks, I guess this is the best one can do security wise. I couldnt find anything better. Good work. 👍

gamemodes/terrortown/gamemode/shared/sh_admincheckwatch.lua
Wryyyong: `event_name` was used simply to mirror the vanilla `hook.Add` params, nothing more

gamemodes/terrortown/gamemode/shared/sh_admincheckwatch.lua
Wryyyong: https://wiki.facepunch.com/gmod/Structures/DebugInfo

```
Option: S

source [string]

The path to the file where the passed function is defined prepended by an @ (ex. "@lua/autorun/mytestfile.lua"). This will be the CompileString or RunString identifier if the function wasn't defined in a file, also prepended by an @.

Option: S

short_src [string]

The shortened name of the source (without the @). May be truncated if the source path is long.
```

Wryyyong:
Before I make the last commit, should I future-proof the module by allowing it to look for multiple hooks instead of just the one? Should not be difficult at all.

ZenBre4ker:
> Before I make the last commit, should I future-proof the module by allowing it to look for multiple hooks instead of just the one? Should not be difficult at all.

Yeah why not, maybe a hardcoded list at the top 🤷

saibotk:
Maybe im missing something, but what prevents a malicious addon from just adding a hook before all entities are initialized?

Wryyyong:
> Maybe im missing something, but what prevents a malicious addon from just adding a hook before all entities are initialized?

Hence the report that comes paired with the addon checker output. I wouldn't bother making the original AdminCheck function in the first place if I didn't want people to use it, but it also opens up a blatant vulnerability that can grant people unauthorized access to the gamemode configuration (and player role information, which can be used to cheat your way through matches). Yes, bigger and more dangerous vulnerabilites can and likely do exist, but there's no reason we should just have this one lying about unchecked.

The report is intended to assist server owners/maintainers/developers/etc. by listing all additions hooked to the function (and any other functions added to the internal table in the future), giving them a heads-up to double-check the addon's code so they can ensure the hook is being used responsibly.

The only reason the whole "lock" portion exists is so potentially malicious addons can't be sneaky and add the hook after the report is printed at startup. Otherwise, the server owner/etc. would have to reprint the report via the concommand over and over to monitor for any changes.

---

As for the concern about the overhead, I made a point of storing as little as possible in the global environment (both for security *and* memory-conservation reasons), and minimized the value-checking and indexing used in each function (eg. for tables, [storing the value as the key and simply indexing the table for the key, instead of iterating through each value until a match is found](https://wiki.facepunch.com/gmod/Tables:_Bad_Habits)).

Take the aforementioned report function, the largest and likely the most computationally-intensive in the module: even when monitoring multiple functions, executes and prints near-instantly on my end and stores nothing in the global env.

saibotk:
No you got me wrong, i understood what you are trying to solve here and why. But i question the importance. Since it just is not opening up any vulnerabilities. I can just happily overwrite nearly anything in my addon thats just how gmod addons work. And i think this method of locking does not work in the second place since it would only prevent hooks from being added after initialization, but most addons will just register hooks when their code is loaded. Thats what my main question was about.

saibotk:
Im not 100% sure at the moment about the loading order etc, thats why i asked if im missing something. 

Additionally, with overhead i was also talking about code and maintenance in general that is introduced.

ZenBre4ker:
> No you got me wrong, i understood what you are trying to solve here and why. But i question the importance. Since it just is not opening up any vulnerabilities. I can just happily overwrite nearly anything in my addon thats just how gmod addons work. And i think this method of locking does not work in the second place since it would only prevent hooks from being added after initialization, but most addons will just register hooks when their code is loaded. Thats what my main question was about.

I guess he answered that. Every addon can add its hooks before point X in loadup time. But then this feature locks it up, so that there cant be any addons added easily with hook.add after that point in time. Then comes the report.

As the lock is now integrated into the hook.add function with a local variable there should be no easy way to circumvent it after that.

saibotk:
So after some discussion, my point here is, that the locking functionality wont really do much because nearly no one registers hooks after initialization. Thus it does not contribute any real value and is too much maintenance burden to have added with such little value. The command is also not really helpful as nearly no one will know about it and is only very little useful in terms of people having to execute it on purpose which they usually dont tend to do regularly.

Concludingly i think this opens up no real vector, besides this being a minor thing for game addons in general. I am happy to approve the hook itself and if the others insist also the command, but i don't think we need the locking method in here, im sorry for all the work you put in. But all in all it is important to discuss new code changes and decide upon them.

This is not to say we don't appreciate the effort. I really like the code style!

ZenBre4ker:
I also think we can either make the report smaller or remove it entirely, because as tk pointed out to me again, the info about the source-path is not really saying enough. You would still have to download the entire codebase and search for it manually.

I know it was my idea, but it's not as I hoped it could be. So sorry for the inconvenience.

Wryyyong:
Is what it is, I s'pose.

Should I just rollback everything to before the module was commited? Not much point in having the whole setup for it without the aforementioned functionality.

ZenBre4ker:
> Is what it is, I s'pose.
> 
> Should I just rollback everything to before the module was committed? Not much point in having the whole setup for it without the aforementioned functionality.

Yeah I guess that suffices.

ZenBre4ker (approved):

saibotk (approved):

PR #973, opened 2022-06-30
Corrected incorrect translation

sbzlzh:


TimGoll:
Well I guess this is a good change. Therefore I approve it. But please add an entry to the changelog. Something like "updated chinese translation (by @sbzlzh)

PR #976, opened 2022-07-02
Fall Damage: Add ConVars to allow customization for server admins

Wryyyong:
- `ttt2_falldmg_enable (default: 1)` toggles whether or not to apply fall damage at all
- `ttt2_falldmg_min_velocity (default: 450)` sets the minimum velocity threshold for fall damage to occur
- `ttt2_falldmg_exponent (default: 1.75)` sets the exponent to increase fall damage in relation to velocity

All ConVars listed above can also be adjusted in the F1 -> Administration -> Player Settings menu.


TimGoll (approved):
I love it. Thank you! Plase add a changelog entry as well

saibotk (approved):
Amazing 🙌

PR #978, opened 2022-07-04
CD: Add primitive continuous deployment workflow

Histalek:
For now this pretty basic and makes a few assumptions on the environment
target where this should be deployed to:

- correctly configured secrets of the github environment
- already installed and running garrysmod server on the remote host
- correctly configured ssh key on the remote host

Additionally manual intervention on the garrysmod server might be
required if hotreloading breaks.

Besides these points this specific deployment is mainly meant for having
a quick look at a specific pull request on a server and manual testing
the changes on a dedicated resource.


TimGoll (approved):
probably a good thing

.github/workflows/cd.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  deploy-to-dev:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: histalek-dev-env
+      cancel-in-progress: true
+    environment:
+      name: histalek-dev-env
+      #url: https://example.com
```
Histalek, line 17: This could be set to something along the lines of `steam://connect/<domain>:<port>/<password>` to ease connecting to the server (IF it works properly).

Obviously that would allow anyone to join the server. If that's unwanted we could easily omit the password part.

Also regarding that steam:// link: AFAIK that depends on Steam's `Master Server Query Protocol` [1] so the server would have to be started **without** specifying `-nomaster` in the start command. 

[1] https://developer.valvesoftware.com/wiki/Master_Server_Query_Protocol

Histalek:
> Additionally manual intervention on the garrysmod server might be required if hotreloading breaks.

Oh boy does it break :D After re-deploying the server went into a seemingly infinite hotreloading loop. Not sure if that is a genuine bug still or something in this setup is responsible for it.

Histalek:
I'm confident now that this won't break instantly so this can be merged.

only outstanding TODO is what to deploy at the set Deployment URL (the 'View Deployment' button).

I currently see two possibilities:
- a redirect service to the steam connect url (idea by @saibotk )
- a landing page, explaining what that server is meant for, how to connect to it and possibly where to ask for the password :D

saibotk:
depends on your motivation to implement either solution :) 
Number 2 might be the better one, while number 1 is much simpler.

saibotk (approved):

Issue #979, opened 2022-07-22
`plymeta:SetRole` will execute `TTT2UpdateTeam` twice when changing team


NVedsted:
## Your version of TTT2 (mandatory)

Leave one of the following:
- Latest GitHub version

## Describe the bug (mandatory)

The hook `TTT2UpdateTeam` will be executed twice in `plymeta:SetRole` if it has to set a new team either through parameters, default teams for roles, and so on.

## To reproduce

Steps to reproduce the behaviour:

1. Call `plymeta:SetRole` in such a manner that will trigger a team change.
2. Observe that the hook `TTT2UpdateTeam` is called twice.

## Expected behaviour

`TTT2UpdateTeam` should only be called once.

PR #980, opened 2022-07-24
Role Changes: Fix double TTT2UpdateTeam hook call on role change

ZenBre4ker:
When a role change leads to a team change the hook call is made twice.
Due to some roles being able to change teams without changing role (e.g. pirate captain) we need the call in both functions, so we suppress it in the updateTeam call if necessary.

Should solve #979


saibotk (approved):
LGTM

Histalek (approved):
LGTM

PR #982, opened 2022-07-26
Language: Add portuguese translation

ceifa:


lua/terrortown/lang/pt.lua
```diff
@@ -0,0 +1,1422 @@
+-- Portuguese language strings
+
+local L = LANG.CreateLanguage("Português (Brasil)")
```
TimGoll, line 3: I know this is WIP, but this should be `local L = LANG.CreateLanguage("pt")`

TimGoll: also if we differentiate between "normal" portuguese and brasilian portuguese, we should rerflect that in their identifiers. As someone that doesn't speak this language I have to ask if this differentiation is necessary.

Histalek: Language codes/identifiers are a mess and our usage doesn't help either :D

But following our current scheme this should be `pt_br`. This also allows for an european portuguese translation in the future.

```suggestion
local L = LANG.CreateLanguage("pt_br")
```

Histalek: Furthermore the filename should be changed accordingly e.g.: `pt_br.lua`

ceifa: I changed it to `pt_br`. Brazilian portuguese it's a little bit different of Portugal portuguese.

Histalek (approved):
Thank you for your work!

TimGoll (approved):

PR #983, opened 2022-08-05
[Release] v0.11.5b

ZenBre4ker:
Bump version number and add changelog.
Also replace all ' ' with <code></code>.

Waiting for #894 to be merged, which this branch is based on.

PR #985, opened 2022-08-13
cd: Remove 'concurrency' stanza from deploy job

Histalek:
Currently the concurrency leads to a maximum of one active deployment
and one pending deployment.

If a pending deployment already exists and another gets requested the
already existing pending deployment fails. In case of two PRs being
updated this would mean that the PR which is updated first would be
impossible to deploy.

Originally i intended to use the concurrency to prevent a singular PR
from creating multiple pending deployments. While this should still be
possible by using a concurrency group unique to the PR, there is also
the possibility that the default behaviour for deployments on environments
is already doing what i want.
Which is why my current approach is to remove the stanza completely.

Sadly the github documentation on this is severely lacking IMO.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -9,11 +9,6 @@ on:
 jobs:
   deploy-to-dev:
     runs-on: ubuntu-latest
-    # The concurrency group should match the environment name this deploys to.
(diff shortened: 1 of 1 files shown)


saibotk (approved):

Histalek:
Deploying this PR made the last deployment (on #983 ) inactive. So that's still working as intended :D

PR #986, opened 2022-08-20
Roles: Improve hotreloadability of Gamemode

ZenBre4ker:
TTT2 should now be able to be hotreloaded again. At least it doesnt throw errors anymore and seems to work.


ZenBre4ker:
For an unkown reason roleList written as RoleList doesnt work. I couldnt find anything in our or the original terrortown code, that uses that. Maybe I broke a certain addon, that doesnt go over out roles.GetList() function. Not sure.

ZenBre4ker:
Ah found one, the roles debugger changes it. Testing now if removing the addon and local suffices to fix hotreload.

btw, cant change this to a draft or dont I just see the button?

ZenBre4ker:
Okay, the rename is not necessary, that was because of the addon. But the idea with local and global didnt make sense anyways. I used the same pattern we use for other modules and just moved it to an internal table. Therefore if something is colliding with roles, it's more clear than a random global variable RoleList.

It hotreloads now faster, as the roles dont need to be initialized again.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed addon compatibility checker fussing over disabled addons
 - Fixed ammo entities blocking +use traces
 - Fixed double call of `GM:TTT2UpdateTeam`, when a role change leads to a team change
+- Fixed hotreload of TTT2 due to roles library not global
```
saibotk, line 42: Adjust this to the new purpose of the patch

saibotk (changes requested):

Histalek (approved):

saibotk (approved):

TimGoll (approved):
> It hotreloads now faster, as the roles dont need to be initialized again.

This probably also means that a hotreload no longer breaks the role ID sync, doesn't it?

ZenBre4ker:
That I didnt check specifically, but in theory it keeps now the same information on both sides. the game at least still works after hotreload.

PR #987, opened 2022-08-30
Roleselection: Fix not every role could be chosen

ZenBre4ker:
With the help of bud, we identified a problem where not every role could be chosen.
Randomize base role selection and stop after maxRoles was reached or every role chosen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Fixed
 - Fixed the spawn editor tool not having a TargetID in some scenarios by always rendering the 'ttt_spawninfo_ent' (by @NickCloudAT)
+- Roleselection for a lot of roles now considers all possible subroles one after another
 
 ## [v0.11.5b](https://github.com/TTT-2/TTT2/tree/v0.11.5b) (2022-08-05)
 
--- a/gamemodes/terrortown/gamemode/server/sv_roleselection.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_roleselection.lua
@@ -465,13 +465,18 @@ function roleselection.GetSelectableRolesList(maxPlys, rolesAmountList)
 	hook.Run("TTT2ModifyLayeredSubRoles", layeredSubRolesTbl, availableSubRolesTbl)
 
 	-- now we need to select the subroles
-	for cBase = 1, #baseroleLoopTbl do
-		if maxRoles and maxRoles <= curRoles then break end -- if the limit is reached, stop selection
+	while true do
+		if maxRoles and maxRoles <= curRoles or #baseroleLoopTbl < 1 then break end -- if the limit is reached, stop selection
 
+		local cBase = math.random(#baseroleLoopTbl)
 		local currentBaserole = baseroleLoopTbl[cBase]
 
 		local currentSubroleTbl = availableSubRolesTbl[currentBaserole]
-		if not currentSubroleTbl then continue end -- no subroles connected with this baserole
+		if not currentSubroleTbl or #currentSubroleTbl < 1 then
+			table.remove(baseroleLoopTbl, cBase)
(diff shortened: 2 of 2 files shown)


gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -496,6 +501,7 @@ function roleselection.GetSelectableRolesList(maxPlys, rolesAmountList)
 					end
 
 					subrole = cleanedLayerTbl[math.random(#cleanedLayerTbl)]
+					table.RemoveByValue(currentSubroleTbl, subrole)
```
busboybud, line 504: Looking at this now, I'm curious if we're negating the ability to have a sub-role multiple times. For instance, if a particular sub-role has been configured to have up to 2 per round, by only adding it once to selectableRoles and then removing it from the master sub role list, are we preventing it from later being able to assign it twice? It looks like SetSubRoles will remove it in that case.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -512,7 +518,12 @@ function roleselection.GetSelectableRolesList(maxPlys, rolesAmountList)
 			selectableRoles[subrole] = rolesAmountList[subrole]
 
 			curRoles = curRoles + 1
+
+			break
```
busboybud, line 522: Line 483 still has the loop, it looks like you added break here but you could just remove the for loop because you will always hit this break regardless

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -512,7 +518,12 @@ function roleselection.GetSelectableRolesList(maxPlys, rolesAmountList)
...
 		end
+
+		-- If got to this point remove the subrole Table
+		table.remove(availableSubRolesTbl, currentBaserole)
```
busboybud, line 526: You don't want to remove the base role from the available sub roles table, because you want the ability to potentially loop through the sub roles of this base role again. This provides the breadth of the randomness

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: All lists in roleselection.GetSelectableRolesList are local, at line 518 is the important where the number of roles that can be chose get inserted.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: Yes I can, as I said, I tried to change as minimal as possible to see if thats enough. As the for-loop works yet.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: Yes I want to remove it, but not in this case, you are right.

I forgot to put a continue statement over it, thanks :)

ZenBre4ker:
Okay, the code is now longer but hopefully more readable. It also gets now all subroles randomly and still respects layers.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Fixed
+
+- Roleselection for a lot of roles now considers all possible subroles one after another
```
TimGoll, line 9: maybe make sure to mention that the previous issue was that traitor subroles were always assigned first, resulting in only having T-Subroles when the subrole limit was smaller or equal to the amount of traitor subroles installed

TimGoll (approved):
The code looks fine to me. If you (and maybe @busboybud) confirm that it works as intended, then this is fine for me.

saibotk (approved):
LGTM, thank you very much :)

PR #988, opened 2022-08-31
Shop: Rework traitor shop menu

ZenBre4ker:
First Draft by @TimGoll, gonna update with pictures from the real implementation
grafik

Single Shop Icons:
grafik


lua/ttt2/libraries/bind.lua
```diff
@@ -152,6 +152,7 @@ end
 -- @internal
 local function TTT2BindCheckThink()
 	-- Make sure the user is currently not typing anything, to prevent unwanted execution of a binding.
+	-- TODO: Maybe send vguihandler the pressed button to close menus again or enable fast buys
```
ZenBre4ker, line 155: We gotta think about vgui frames being open blocking bindings, it's the reason why we cant close roundendscreen with the binding we opened it with. When we setup a quickbuy or similar it's necessary to get keybindings. Maybe some feedthrough for frames, which button was clicked, but generall block the OnPressed-function?

TimGoll: what? I'm pretty sure the roundendscreen can be closed by pressing a key

Spanospy: > what? I'm pretty sure the roundendscreen can be closed by pressing a key

Just checked: the same key that opens the roundendscreen can't currently close it.

ZenBre4ker:
Closing this. Enough information are on #1062 and this here is just crudely copied code from other UI.

PR #994, opened 2022-09-24
Language: Corrected Simplified Chinese and updated Traditional Chinese

TEGTainFan:
@sbzlzh used Google Translate, which resulted in a large number of translation errors
All Simplified Chinese translation errors have been corrected
Also updated with Traditional Chinese


sbzlzh:
That match was manually changed by me to .

sbzlzh:
It's not Google Translate, Google Translate comes out with a period

TEGTainFan:
Your translation is riddled with errors, For example：你将Player翻译成了球员/播放器，将Spawn翻译成再生/产卵。These are common google translation errors and, as your translation is riddled with grammatical incompatibilities and misinterpretations, I had to re-translate over 70% of the text (not including before September 2021) because of the sheer number of errors.

saibotk:
First of all thank you both, regardless if using google translate or not, for bringing us translations for another huge language.

Since this language is none that anyone of us has learned, we rely on every contribution. So thank you for doing the tedious work of going through all those lines. We really appreciate it, and i think many other players will too.

While this seems good to go, i see you are still changing stuff. Is this ready to be merged?

TEGTainFan:
> First of all thank you both, regardless if using google translate or not, for bringing us translations for another huge language.
> 
> Since this language is none that anyone of us has learned, we rely on every contribution. So thank you for doing the tedious work of going through all those lines. We really appreciate it, and i think many other players will too.
> 
> While this seems good to go, i see you are still changing stuff. Is this ready to be merged?

Yes, this is the final version

saibotk (approved):

TEGTainFan:
> > First of all thank you both, regardless if using google translate or not, for bringing us translations for another huge language.
> > Since this language is none that anyone of us has learned, we rely on every contribution. So thank you for doing the tedious work of going through all those lines. We really appreciate it, and i think many other players will too.
> > While this seems good to go, i see you are still changing stuff. Is this ready to be merged?
> 
> Yes, this is the final version

emmmmm, this is the final final version

Histalek (approved):

PR #998, opened 2022-09-25
cd: Fix `deploy-to-dev` workflow not running on PRs from forks

Histalek:
The key difference here is that `pull_request` only allows PRs originating from branches of the original repository to access secrets while `pull_request_target` also allows this for PRs originating from forked repositories.

While this is potentially unsafe [1] (which i was aware of when i first added the cd workflow with the intentation that this would already work) it is safeguarded by someone having to manually approve the workflow run (currently only myself).

Using the `permissions` keyword [2] should allow to limit the scope further without restricting the access to the secrets. (At least i hope this is the case, because the github documentation is rather unclear on this IMO).

[1] https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

[2] https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -2,6 +2,8 @@ name: CD
 
 # Run on Pull Requests to master and on manual interaction
 on:
+  pull_request_target:
+    branches: [ master ]
   pull_request:
     branches: [ master ]
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


Histalek:
Restricting the permissions did not break the existing workflow, so that's good at least.
Here's to hope that this fixes the fork PR issue :crossed_fingers:

saibotk (approved):

ZenBre4ker (approved):

PR #456, opened 2020-02-03
Allow hold aim instead of toggle

menof36go:
You no longer have to press secondary attack a second time to stop aiming and instead will stop aiming as soon as you let go of the button. This is disabled by default and can be enabled using the clientside convar ttt2_hold_aim or by checking the checkbox in the gameplay menu. I tried to change as little as possible about how the ironsight actually works in order to keep it compatible with old stuff. I also made sure this doesn't mess with prediction, making this change very stable even at high latency like 300ms. I added some missing functions and fields in order to give people an idea of how they should implement ironsights and zoom and added some missing doc.

PR #1002, opened 2022-10-12
SWEPs: Slight fixes to weapon_tttbase

Wryyyong:
- Removed `SWEP.IronSightTime` as it lacked any use within the gamemode, and was redundant as ironsight times are already managed with networked data tables
- Fixed `SWEP.IronSightPos` to `SWEP.IronSightsPos` for both a) consistency with literally every other SWEP and function using the latter variable names, and b) because it was breaking the ironsights of any SWEP trying to use the default values inherited from `weapon_tttbase`.

### DIFF
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -161,10 +161,12 @@ SWEP.ReloadAnim = ACT_VM_RELOAD
 
 SWEP.fingerprints = {}
 
--- The position offset applied when entering the ironsight
-SWEP.IronSightsPos = Vector(0, 0, 0)
--- The rotational offset applied when entering the ironsight
-SWEP.IronSightsAng = Vector(0, 0, 0)
+--[[
+	-- The position offset applied when entering the ironsight
+	SWEP.IronSightsPos = Vector(0, 0, 0)
+	-- The rotational offset applied when entering the ironsight
+	SWEP.IronSightsAng = Vector(0, 0, 0)
+--]]
 
 local skipWeapons = {}
 
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -162,9 +162,9 @@ SWEP.ReloadAnim = ACT_VM_RELOAD
 SWEP.fingerprints = {}
 
 -- Time needed to enter / leave the ironsight in seconds
-SWEP.IronSightTime = 0.25
+SWEP.IronSightsTime = 0.25
 -- The position offset applied when entering the ironsight
-SWEP.IronSightPos = Vector(0, 0, 0)
+SWEP.IronSightsPos = Vector(0, 0, 0)
 -- The rotational offset applied when entering the ironsight
 SWEP.IronSightsAng = Vector(0, 0, 0)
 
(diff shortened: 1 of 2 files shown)


TimGoll:
I'm just wondering if this would break addons that use the old name?

Wryyyong:
> I'm just wondering if this would break addons that use the old name?

It shouldn't -- because if they were also using the old typo'd names, they'd need to copypaste over the functions (or write their own entirely) and edit/write them accordingly to use said typo'd names.

On the off-chance that it somehow does break compatibility, you could just include the typo'd names aliased to the fixed ones:
```lua
SWEP.IronSightsTime = 0.25
SWEP.IronSightsPos = Vector(0, 0, 0)

SWEP.IronSightTime = SWEP.IronSightsTime
SWEP.IronSightPos = SWEP.IronSightsPos
```

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -162,9 +162,9 @@ SWEP.ReloadAnim = ACT_VM_RELOAD
 SWEP.fingerprints = {}
 
 -- Time needed to enter / leave the ironsight in seconds
-SWEP.IronSightTime = 0.25
+SWEP.IronSightsTime = 0.25
 -- The position offset applied when entering the ironsight
-SWEP.IronSightPos = Vector(0, 0, 0)
+SWEP.IronSightsPos = Vector(0, 0, 0)
 -- The rotational offset applied when entering the ironsight
 SWEP.IronSightsAng = Vector(0, 0, 0)
```
saibotk, line 169: After looking at git blame, i think we should remove all those values.

They were introduced in https://github.com/TTT-2/TTT2/pull/456
Also they are not present in normal TTT and i cannot see where it is used in TTT2.

And additionally these would clash with the network var values. eg. GetIronSightstime() should be used instead i think.

saibotk (changes requested):

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
saibotk: Indeed, youre correct for the time.
But i still think we should remove all of them, since i don't think it is sensible to assume 0-vectors to be correct there.
Especially since the code behaves differently when those are not set. (And all calls check if those values even exist).

Wryyyong: Removal from just `weapon_tttbase`, or for all SWEPs? Just so I'm not misunderstanding.

saibotk: Base as before and as in normal TTT but i want to hear someone else about this

ZenBre4ker: Wasnt sure before, but as TTT also doesnt set it, I'd say we just comment it out. So we have the same missing zoom behaviour in TTT2 as in TTT.
So also remove SWEP.IronSightsAng

And definitely remove the sightsTime. 👍

Wryyyong: Roger

saibotk (approved):
Thank you very much and thanks for the elaborate changelog 🙌

ZenBre4ker (approved):
Not a fan of the huge changelog, but it's fine. Thank you <3

Wryyyong:
> Not a fan of the huge changelog, but it's fine. Thank you <3

Ehh, sometimes the extra detail is necessary to avoid misinterpretations

ZenBre4ker:
> > Not a fan of the huge changelog, but it's fine. Thank you <3
> 
> Ehh, sometimes the extra detail is necessary to avoid misinterpretations

Definitely true, but I personally dont think it belongs in a changelog. Thats theoretically documented in the pull requests. But if thats the best spot, I dont know. Just personal preference ^^

PR #1006, opened 2022-11-19
T-Buttons: small efficiency boost

mognify:
tried to make it more legible and efficient
so i made it more complicated =)


mognify:
i just realized i left the recv variable outside the if statement that uses it............. :\

gamemodes/terrortown/entities/entities/ttt_game_text.lua
```diff
@@ -54,28 +54,23 @@ end
...
+				RECEIVE_TRAITOR = GetRoleChatFilter(TEAM_TRAITOR),
+				RECEIVE_INNOCENT = GetRoleChatFilter(TEAM_INNOCENT)
+			}
+			
+			recv = (self.teamReceiver) ? GetTeamChatFilter(self.teamReceiver) : receiver_tbl[self.Receiver]
+			CustomMsg(recv, self.Message, self.Color)
+			return true -- success
+		else
+			ErrorNoHalt("ttt_game_text tried to show message to invalid !activator\n")
+			return false -- failed, either invalid activator or activator was not a player
```
Histalek, line 71: Why did you change the return value? Not saying that this is wrong, but i'd like an explanation for why you return false here when the old code would have returned true instead.

A small nitpick: Please remove the comment, it doesn't add value imo

gamemodes/terrortown/entities/entities/ttt_game_text.lua
```diff
@@ -54,28 +54,23 @@ end
...
 		end
-
-		CustomMsg(recv, self.Message, self.Color)
-
-		return true
 	end
+	
+	return false -- failed, name wasn't "Display"
```
Histalek, line 75: Same as above: why the changed return value?

Histalek (changes requested):
All in all i like this change. Imo it makes the code more readable by using a lookup table and the ternary operator.

In regards to the code comments: I personally would remove them, as they don't really add more to what the code itself already tells us.

Regarding requested changes, please do address your own comments on this pull request, they are definitely valid :)

gamemodes/terrortown/entities/entities/ttt_game_text.lua
mognify: it just seemed more logically correct to return a fail if the operation did not succeed with the provided input

as it stands, returning True really only tells us that name did = "Display"
if it returns False, then it tells us that either the name wasn't "Display", or the Activator was invalid/not a player

although, now that ive written this out... it actually does make more sense troubleshooting-wise to return True... that accompanied with the ErrorNoHalt tells us exactly what happened...

although, ErrorNoHalt has the same effect regardless of return value

**tl;dr imo ErrorNoHalt means we dont have to sacrifice what i believe is logical consistency**
it makes more sense to me that we'd return False if the operation failed, and True if it succeeded

gamemodes/terrortown/entities/entities/ttt_game_text.lua
mognify: i dont understand returning True when an error occurred and the script failed due to invalid activator

is there some trick somewhere else that relies on this one being True?

TimGoll (approved):
I like this, thanks a lot. But please address the points raised by @Histalek

gamemodes/terrortown/entities/entities/ttt_game_text.lua
```diff
@@ -52,30 +52,28 @@ end
...
-		end
-
+		local receiver_tbl = {
+			RECEIVE_ALL = nil,
+			RECEIVE_DETECTIVE = GetRoleChatFilter(ROLE_DETECTIVE),
+			RECEIVE_TRAITOR = GetRoleChatFilter(TEAM_TRAITOR),
+			RECEIVE_INNOCENT = GetRoleChatFilter(TEAM_INNOCENT)
+		}
+			
+		recv = (self.teamReceiver) ? GetTeamChatFilter(self.teamReceiver) : receiver_tbl[self.Receiver]
```
Histalek, line 68: whoops, i just realized that the ternary operator doesn't exist in lua. (no clue why i thought it did ...)

But the following statement should be equivalent:
```suggestion
		recv = (self.teamReceiver) and GetTeamChatFilter(self.teamReceiver) or receiver_tbl[self.Receiver]
```

mognify:
sorry, i had a bunch of things taking up my time and energy
i can refine this and update now (real quick (soon (tm)))

Histalek:
> sorry, i had a bunch of things taking up my time and energy i can refine this and update now (real quick (soon (tm)))

No worries!

(i probably need to reconfigure the stale bot to only annoy us Team-Members or remove it altogether ...)

PR #1025, opened 2023-03-07
fix ttt_quickslot

ruby0b:
ttt_quickslot did not work previously.
QuickSlot wrongly tried to convert the args table to a number.
There was also a mix up between slots and weapon kinds: The current weapon kind is already the same as the displayed number, no need to subtract 1 from user input.


Histalek (approved):
Nice finding! This was seemingly broken for a loooong time :D

Confirmed broken on v0.11.6b and confirmed fixed by this PR

saibotk (approved):
Thank you very much for your efforts ☺️

PR #1026, opened 2023-03-07
ci: Update actions and glualint

Histalek:
Github deprecated support for actions that run on node.js 12 [1]. So we update the affected actions to a version that uses node.js 16.

Glualint was also some versions behind so i've updated it as well, including the changed download url.

[1] https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 
 env:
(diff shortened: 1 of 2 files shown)


saibotk (approved):
LGTM

saibotk (approved):
Nice Thanks

PR #1027, opened 2023-03-09
playermodels: Add option to disable resetting player color via `ttt_enforce_playercolor` convar

Histalek:
This is mostly done to enhance compatibility with colourable
playermodels.

Before this change coloured playermodels would be set to `COLOR_WHITE` or
`GAMEMODE.playercolor` on (re)spawn to prevent players from being
distinguished by different colours.

After this change this behaviour is not changed unless `ttt_enforce_playercolor` is set to 0.
In that case we return early in the `GM:TTTPlayerSetColor` hook and do
nothing.
This does NOT however prevent other addons from hooking into the `TTTPlayerSetColor` hook nor
does it prevent other addons from setting the playercolor somewhere else.

NOTE: This is also not tested yet, hence draft status

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,34 +1,34 @@
-name: CD
-
-# Run on Pull Requests to master and on manual interaction
-on:
-  pull_request_target:
-    branches: [ master ]
-  workflow_dispatch:
-
-jobs:
-  deploy-to-dev:
-    runs-on: ubuntu-latest
-    environment:
-      name: histalek-dev-env
-      # Sadly a 'steam://connect/<server_domain>' url is not allowed by github
-      # Current workaround is a 'redirect webpage' on GitHub Pages
-      # Ref. https://github.com/TTT-2/ttt-2.github.io
-      url: https://ttt-2.github.io/redirect/ttt2-dev-env
-    permissions:
-      contents: read
-
-    steps:
-    - uses: actions/checkout@v3
-
-    - name: rsync deployments
-      uses: burnett01/rsync-deployments@5.2
-      with:
-        switches: -avzr --delete
-        # The path should end in a directory in the garrysmod addons directory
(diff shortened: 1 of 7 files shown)


Histalek:
`ttt_playercolor_mode = 0` seems to be unused [1]. Which looks like a better approach to this.


[1] https://github.com/TTT-2/TTT2/blob/fd45084b5dddcaf8a46f23ecbf825295d2f21f59/gamemodes/terrortown/gamemode/shared/sh_main.lua#L355-L378

SvveetMavis:
I tested this with the latest version and it works good, had to get enhanced playermodel selector to pick a color but like the color got kept between rounds and stuff, i think if there was a way to open the color selector from like the default ttt2 f1 menu then this would be a cool little overhaul to the color system. One thing I noticed with colors though is they aren't applied to the like playermodel display in the search menu so if you ever work on this again that would be a good addition, would also be neat if like there was a setting to make the colors it selects from without the color selector randomized per player rather than 1 color randomly selected for everyone

Histalek:
Thanks for testing this!

I originally got stuck on this because i wasn't able to properly test if this was the only reset for the color and how this would interact with different addons. So i appreciate you testing this!

>  had to get enhanced playermodel selector to pick a color

What happened if you did not do that? Did that break something or was it undesirable in some other way?

SvveetMavis:
oh without playermodel selector I would be stuck as white but once I used playermodel selector to set a color onto myself it would persist and not get replaced by one of the randomized colors, without enforce playercolor thing the color set would reset every round and have to be reapplied, thats why I was talking about like a built in color selector

zenbooreal:
idk if this is necroposting but i cannot find anything about ttt_enforce_playercolor and it comes out as an unknown command in the console using the workshop version v0.14.5b. has there been any development on this command outside this thread? we're using enhanced playermodel selector and i'm currently stuck with this as a problem.

Issue #1028, opened 2023-03-11
ttt_armor_dynamic convar is reversed


TheMarrix:
version: 0.11.6b without any other addons

ttt_armor_dynamic 1 uses the old ttt armor system and ttt_armor_dynamic 0 uses the new ttt2 armor system.

Histalek:
This should now be fixed, release soon :)

PR #1029, opened 2023-03-16
Armor: Hotfix for dynamic armor check being inverted

Wryyyong:
Made to address #1028.


Histalek (commented):
Thanks for looking at #1028 and taking the time to find the issue!

Looking at the changes and git blame, this seems to indeed fix the problem. (also this was broken for 1½ years, whoops)

I'll hold off my approval until Tim has taken a look, as he introduced the change.

saibotk (approved):

Histalek (approved):

PR #1035, opened 2023-04-24
Sprint: Fix prediction & refactor logic

saibotk:
This PR cleans up the whole sprinting logic, to fix the entirely unpredicted previous sprinting system.

## Why was the system previously unpredicted?

It previously used net messages from the client to toggle the sprinting state.
Those messages cannot be combined with a predicted state, which would be needed for the client to properly predict the sprinting speed / movement.

## How is this fixed?

We now use a DTVar on the player for the stamina and the default sprinting key to signal the sprinting intention.
Those are both variables which the client can use in prediction and can reset their values when prediction errors occur. This is for example not (easily) possible with a networked property on the player (eg. NWVars).

## What changed? Why was X removed?

Since the prediction needs a predictable signalling source, we now need to use the builtin keybind (shift usually) instead of our custom keybind. So we had to remove our sprint keybinding, to be able to fix those issues. Sadly there is no other feasible way to add custom binds or key presses etc into a movement command packet.

Additionally, since not only the client starts the signalling, we also needed to drop the double tap sprinting feature, which was not really a good feature to have enabled by default anyways. 
This is mainly due to the need for additional workarounds, syncing & statefulness we would need, to make the server aware of each player's double tap settings. Because now both client and server would need to be able to properly know the sprinting state in a predictable manner (via DTVars and movement keys).

We also removed the unused `ply.preventSprint` flag on the client side, since this was not really used anyways and we already evaluated its usefulness in #455 with the conclusion that we don't know of anyone using this.

Sprinting now also costs stamina while in the air, to prevent players from only sprint jumping around which easily saves stamina and makes the reach much much higher. Now this requires some timing skill, but it is still possible to save stamina when releasing the sprint key while airborne.


wgetJane:
idk if i'm doing something wrong, but it seems like i can't sprint on my end

saibotk:
@wgetJane 
No please dont test this yet xD this is entirely broken and was only pushed so @Histalek and i could test.
We are working on something locally :)

EntranceJew:
I tested this out a little, it's an improvement, although I don't really see why sprint in midair is spent based on move-speed when you're going the same speed as you would be if you let go of sprint right after jumping.

saibotk:
Well sounds alright to me, otherwise you can just abuse this to easily hold sprint and jump to avoid stamina drain.

Now it requires skill 😅

gamemodes/terrortown/gamemode/shared/sh_playerclass.lua
```diff
@@ -0,0 +1,18 @@
+DEFINE_BASECLASS("player_default")
+
+local PLAYER = {}
+
+PLAYER.WalkSpeed = 220
+PLAYER.RunSpeed = 220
+PLAYER.JumpPower = 160
+PLAYER.CrouchedWalkSpeed = 0.3
```
TimGoll, line 8: Just a note for the future: It might be neat to let users change those vairables in the UI? Not a change for this PR though

gamemodes/terrortown/gamemode/shared/sh_speed.lua
```diff
@@ -12,7 +10,9 @@ SPEED = SPEED or {}
 -- @internal
 -- @realm shared
 function SPEED:HandleSpeedCalculation(ply, moveData)
-	if not ply:IsTerror() then return end
+	if not ply:IsTerror() then
+		return
+	end
```
TimGoll, line 15:
```suggestion
	if not ply:IsTerror() then return end
```

I know you're using a formatter, but this breaks our code style

gamemodes/terrortown/gamemode/shared/sh_speed.lua
```diff
@@ -34,7 +34,9 @@ function SPEED:HandleSpeedCalculation(ply, moveData)
 	local oldval = ply:GetSpeedMultiplier()
 	ply.speedModifier = baseMultiplier * returnMultiplier * speedMultiplierModifier[1]
 
-	if SERVER then return end
+	if SERVER then
+		return
+	end
```
TimGoll, line 39:
```suggestion
	if SERVER then return end
```

same here

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -1,188 +1,97 @@
...
+-- Checks if the player is pressing any movement keys and the sprint key at the same time.
+-- @param Player ply
+-- @return boolean
+-- @realm shared
+function SPRINT:PlayerWantsToSprint(ply)
+	local isSprinting = ply:KeyDown(IN_SPEED)
+	local inMovement = ply:KeyDown(IN_FORWARD)
+		or ply:KeyDown(IN_BACK)
+		or ply:KeyDown(IN_MOVERIGHT)
+		or ply:KeyDown(IN_MOVELEFT)
```
TimGoll, line 29: I know we did it like this before, but do we really want the player to be able to sprint in all directions? Because of your rework I paid attention how other games handle this. And as far as I can tell, they only let you sprint forwards. Which kinda makes sense anyway

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -1,188 +1,97 @@
...
-			local time = CurTime()
+	local sprintStamina = ply:GetSprintStamina()
+	local playerWantsToSprint = self:PlayerWantsToSprint(ply)
 
-			if lastPressedMoveKey == key and time - lastPress < 0.4 then
-				PlayerSprint(true, key)
-			end
+	if (sprintStamina == 1 and not playerWantsToSprint) or (sprintStamina == 0 and playerWantsToSprint) then
+		return
+	end
```
TimGoll, line 56: this probably should be a one liner as well I guess?

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -192,9 +101,7 @@ end
 -- @param table modifierTbl The table in which the modifier can be changed
 -- @hook
 -- @realm shared
-function GM:TTT2StaminaRegen(ply, modifierTbl)
-
-end
+function GM:TTT2StaminaRegen(ply, modifierTbl) end
```
TimGoll, line 104: Hm, while this works as well, we always defined empty functions in a different way. I guess it doesn't matter but I don't get why you changed it anyway?

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -203,6 +110,4 @@ end
 -- @param table modifierTbl The table in which the modifier can be changed
 -- @hook
 -- @realm shared
-function GM:TTT2StaminaDrain(ply, modifierTbl)
-
-end
+function GM:TTT2StaminaDrain(ply, modifierTbl) end
```
TimGoll, line 113: same here

TimGoll (approved):
That is a nice change. My comments mostly concern your code formatter that is kinda against our current coding style.
Also I'm not a fan of adding a `,`  after the last entry in a table

TimGoll (changes requested):
oops, misclicked

saibotk:
Honestly the formatting is much better than our original idea. I agree that its not like our current "style" which sadly is also not defined anywhere.

But regarding the trailing commas, i can give you an objective reason:
Less merge conflicts since an addition now only adds a line instead of changing and adding one. Thus it reduces merge conflicts.

Mught want to hear what @Histalek thinks and if we plan on formatting the code anyways in the near future.

gamemodes/terrortown/gamemode/shared/sh_playerclass.lua
saibotk: Nah i would not want anyone messing with these directly.

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
saibotk: Mhh dunno we can change it i have no strong opinion on this

TimGoll:
> Honestly the formatting is much better than our original idea. I agree that its not like our current "style" which sadly is also not defined anywhere.
> 
> But regarding the trailing commas, i can give you an objective reason: Less merge conflicts since an addition now only adds a line instead of changing and adding one. Thus it reduces merge conflicts.
> 
> Mught want to hear what @Histalek thinks and if we plan on formatting the code anyways in the near future.

I think the changes are quite random though. I can get behind your reasoning for the trailing commas, but why make a singleline if to a multiline if, but then change a multiline function to a single line function. Imho these are contradictory

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
TimGoll: I think is is weird that you can sprint sideways because that doesn't work in reality.

In MC, if you hold down the sprint key while moving in any direction but forward, the sprinting is just paused

saibotk:
Those all follow the formatting rules of the formatter i used and in general our one line ifs are rather ugly and less readable. Especially once our conditions are very long, multiline is much better. But we will decide upon this later, where the fact that we just throw a formatter on it and have consistency is the main goal wether or not the formatting is perfect in every case.

TimGoll:
> Those all follow the formatting rules of the formatter i used and in general our one line ifs are rather ugly and less readable. Especially once our conditions are very long, multiline is much better. But we will decide upon this later, where the fact that we just throw a formatter on it and have consistency is the main goal wether or not the formatting is perfect in every case.

Yes, my main reasoning is consitency. If we change the style, it should be done in the whole project, not just for two files

Histalek:
> > Those all follow the formatting rules of the formatter i used and in general our one line ifs are rather ugly and less readable. Especially once our conditions are very long, multiline is much better. But we will decide upon this later, where the fact that we just throw a formatter on it and have consistency is the main goal wether or not the formatting is perfect in every case.
> 
> Yes, my main reasoning is consitency. If we change the style, it should be done in the whole project, not just for two files

consistency is not really a factor if nothing defines what the consistent state should be tbh.

My plan was to get all PRs for a 0.12.0b merged and then run the formatter as the last PR before the Release so that we don't introduce a lot of changes while people are actively working on stuff. So i'd vote for: style wise these contributions can be whatever as they will be fixed anyway.

Histalek:
> > Honestly the formatting is much better than our original idea. I agree that its not like our current "style" which sadly is also not defined anywhere.
> > But regarding the trailing commas, i can give you an objective reason: Less merge conflicts since an addition now only adds a line instead of changing and adding one. Thus it reduces merge conflicts.
> > Mught want to hear what @Histalek thinks and if we plan on formatting the code anyways in the near future.
> 
> I think the changes are quite random though. I can get behind your reasoning for the trailing commas, but why make a singleline if to a multiline if, but then change a multiline function to a single line function. Imho these are contradictory

singleline to multiline is caused by a character limit of 100 (?) and makes absolute sense. E.g. in my personal setup 110-120 characters is the max i can have on half of my horizontal screens or fully on my vertical screens. So 100 is a good breaking point to prevent wordwrapping (bad!) or offscreen text (worse!).

Multiline to singleline is probably because simple statements are generally collapsed if the don't break the character limit.

saibotk:
I feel the same way.
@TimGoll how do you feel about just leaving those right now and reformat all code later anyways?

TimGoll:
> I feel the same way. @TimGoll how do you feel about just leaving those right now and reformat all code later anyways?

Alright, histalek gave me some insight on your plans that I wasn't aware of

TimGoll (approved):

Histalek (commented):
i did comment on something here sometime, but github refuses to show me my own comment ... i'll probably change this review once i see my comment.

Histalek (approved):

PR #1041, opened 2023-08-13
[RTC] Override AllowDrop, mechanism to override Remove() to permit dropping

EntranceJew:
Does what it says on the tin.

A lot of this could have been avoided if the default logic for so many weapons weren't to simply call `:Remove()` in OnDrop or PreDrop but that's a kind of legacy that has to be circumnavigated.

image

image

Some caveats:
- If it were possible to overwrite the metatable of just the weapon of interest being dropped, I'd do so, but I couldn't cause any code to elicit this exact effect at the head of `sv_weaponry`.
- `AllowDrop` doesn't permit things that are capable of being dropped (like the Decoy) from being dropped voluntarily, and I'm not sure where in the chain that happens.
- If I do introduce a new dropdown force/deny/default the voluntary drop behavior, what should be done with `AllowDrop`?
- Could the existing code I've got be shaped better to fit those concerns?
- I'm not sure this code should exist on a shared realm, if it only effectively runs on the server. I'm not even sure what a client-sided `:Remove()` does behind the scenes.
- If an object is not configured to drop then it can produce an invisible point entity, which won't make a difference. This is mostly for things that ordinarily _are_ capable of being dropped.


Histalek (approved):
I do like this change very much. Having this configurable on a per weapon basis is amazing.

I do need however a bit more time to understand this fully and possibly test this myself a bit.

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
```diff
@@ -652,6 +656,13 @@ function WEPS.DropNotifiedWeapon(ply, wep, deathDrop, keepSelection)
 	ply:DropWeapon(wep)
 
 	wep:PhysWake()
+
+	-- Unset this, because we can't use it after this point.
+	wep.IsDroppedBecauseDeath = nil
+
+	if deathDrop and wep.overrideDropOnDeath == DROP_ON_DEATH_TYPE_DENY then
+		wep:Remove()
+	end
```
saibotk, line 665: Can we move this up a bit, to avoid calling dropWeapon at all, if it is removed anyways?

saibotk (changes requested):
LGTM otherwise :) 

Thanks and i think it is okay to do it this way, since i dont see any compatible way around the remove function handling.

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
EntranceJew: it could be possible, i just know some guns such as the defib use OnDrop to cancel revives, or to clean up placement ghosts -- i don't know if i can always count on them to do that during Destroy as well

saibotk: Puh honestly i think this is safe to assume. But ultimately i think its much better not to drop the weapon when deleting it instantly afterwards. This could cause weird behavior. 

What is your opinion @Histalek ?

Histalek: I agree with saibotk on this one. We should simply do what we intend to do and that is simply removing the weapon/equipment.

If the equipment does not correctly clean up after itself in case of a removal, i'd consider that a bug of the equipment. We should not need to work around that.

EntranceJew: makes sense to me, glad to have some valuable perspective on it

saibotk (approved):
Thank you 🙏

PR #1044, opened 2023-08-16
last minute role change hook

EntranceJew:
lets people mess with things without having to blast role distribution and reset models and credits in TTTBeginRound

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added `damageScaling` property to `weapon_tttbase`, adjustable per item via "Balance Settings" under "Edit Equipment" in the F1 Menu (by @EntranceJew)
 - Allow admin spectators to enter "Spawn Edit" mode. (by @EntranceJew)
 - Added cvar `ttt2_bots_lock_on_death` (default: 0) to prevent bots from causing log-spam while wandering as spectators. (by @EntranceJew)
+- Added `TTT2ModifyFinalRoles` hook for last minute opportunity to override role distribution prior to them being announced for the first time (by @EntranceJew)
 
 ### Changed
 
--- a/gamemodes/terrortown/gamemode/server/sv_roleselection.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_roleselection.lua
@@ -907,6 +907,10 @@ function roleselection.SelectRoles(plys, maxPlys)
 
 	GAMEMODE.LastRole = {}
 
+	---
+	-- @realm server
+	hook.Run("TTT2ModifyFinalRoles", roleselection.finalRoles)
+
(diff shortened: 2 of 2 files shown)


AaronMcKenney:
This idea is good. One thing that's been in the back of my head about this is what will happen if multiple roles make calls to this hook and conflict with each other on what the final roles should be? I'm not sure how that kind of conflict resolution should be handled, Admittingly this is somewhat hypothetical and/or rare. Of my roles, only Defective can alter roles at the start of the game. Such a scenario may arise with the Defective + Dunce, but I haven't tested that.

EntranceJew:
> I'm not sure how that kind of conflict resolution should be handled.
IMO this is mostly for things like randomat, which AFAICT forcibly re-assigns roles after the round starts for all-role minigames -- or for server operators that might want to implement "house rule" of how specific roles are distributed. For those purposes, the standard hook priority system should be adequate.

Regular role implementations shouldn't switch to using this and should instead try to find/add a hook earlier in the process to remain compatible with the distribution system. If there needs to be more phases or pockets to store overrides those should be utilized, though I think in order to assess where those sorts of things would go I would need a flowchart for the code annotated with each section's intended purpose.

Personally, I don't think that `roleselection.SelectRoles` should be calling `SendFullStateUpdate()` and that a new intermediary function should be introduced which calls `ply:SetRole` and `ply:SetDefaultCredits` so that this type of hook wouldn't be necessary at all and . This would make the invocation inside `BeginRound()` a little cleaner and more explicit.

Histalek:
> If there needs to be more phases or pockets to store overrides those should be utilized, though I think in order to assess where those sorts of things would go I would need a flowchart for the code annotated with each section's intended purpose.

This is a good point. Currently i don't really see a need for this as the round states are pretty self explanatory (pls tell me if they are not). But if we would ever add another roundstate - either before Preparing or after - we probably will shuffle some intentions and purposes around.

As for now i don't really see much that speaks against this change. Mainly i don't see anybody rethinking the whole round states and purposes any time soon.

EntranceJew:
> As for now i don't really see much that speaks against this change. Mainly i don't see anybody rethinking the whole round states and purposes any time soon.

For clarity, I should specify that I meant `roleselection.SelectRoles` and not the round states. My flattened trace of the logic for SelectRoles is:

1. `GetSelectablePlayers`
2. `GetAllSelectableRolesList`
3. `GetSelectableRolesList`
4. `SelectForcedRoles`
5. `GetHardForcedBaseRoles`
6. `SelectBaseRolePlayers`
7. `UpgradeRoles` (for base roles)
8. `UpgradeRoles` (innos / role-less players)
9. `SetRole` => `SendFullStateUpdate`

Which, is a bit of a process to chew on. I'm probably even neglecting to mention significant hooks within those.

saibotk (approved):

Histalek (approved):

PR #1047, opened 2023-08-17
cvar ttt2_bots_lock_on_death fixes #1043

### LINKED ISSUE #1043: bot spectators tend to wander very far away from the map, creating log noise and instability

- Workshop version from Steam

## Video

image

Basic bots that are added as spectator have a tendency to wander far outside the reasonable realm of the map, and flood consoles with print messages, potentially resulting in long term instability for the server.

## To reproduce

1. Open a game with 3 bots.
2. Kill them.
3. Prevent the round from starting / nextmap from happening.
4. Bots will wander as invisible gheists into parts unknown.

## Expected behaviour

I know a typical Source Bot isn't the direct responsibility of the gamemode, but because they are useful for stress testing other facets of the game mode, they may continue to provide bogus input as a spectator for the purposes of testing other things.

But potentially special handling code be added to warp them back to points of interest, or cycle spectators, rather than let them float away outside the engine's limitations and flood the console or stop simulating inputs entirely.

- see above

EntranceJew:
tested it for a few rounds

admin wallhacks indicate that the spectator ghosts tend to remain where they get warped on death

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added `PANEL:MakeTextEntry(data)` to `DFormTTT2` for strings or string-backed cvars (by @EntranceJew)
 - Added `damageScaling` property to `weapon_tttbase`, adjustable per item via "Balance Settings" under "Edit Equipment" in the F1 Menu (by @EntranceJew)
 - Allow admin spectators to enter "Spawn Edit" mode. (by @EntranceJew)
+- Added cvar `ttt2_bots_lock_on_death` (default: 0) to prevent bots from causing log-spam while wandering as spectators. (by @EntranceJew)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


Histalek (commented):
I'm not sure if this should be a convar at all.

The only usecase for setting this to 0 that comes to my mind is spectator deathmatch with bots. But then again i don't know how exactly that functions.

I really do want to test this change in regards to CPU usage though and if that makes a considerable impact i'd really like to enable this outright. Either by setting the default to 1 or by simply enabling this behaviour without a convar.

Even if spectator deathmatch would be impacted by this, they should be able to simply `:UnLock()` the bots themselves if necessary. (Maybe they use they use our player respawn logic and the bots get unlocked anyway)

EntranceJew:
I left it as a cvar on a whim, but since then I found the [Seance](https://steamcommunity.com/sharedfiles/filedetails/?id=2635618683) role, which would become very difficult to test without the bots driving themselves around.

Histalek (approved):
> I left it as a cvar on a whim, but since then I found the Seance role, which would become very difficult to test without the bots driving themselves around.

Thanks for checking, in that case this seems fine as is :)

saibotk (approved):

PR #1048, opened 2023-08-21
make {walkkey} + {usekey} modifier consistent

EntranceJew:
it simply doesn't work if you press them in the opposite order


saibotk (approved):
Thank you so much 🙏

PR #1050, opened 2023-08-26
passes name of panel down for comparing with GetName()

EntranceJew:
image
image
image
after, before, the code in question

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
@@ -640,7 +640,7 @@ derma.DefineControl("DFormTTT2", "", PANEL, "DCollapsibleCategoryTTT2")
 -- @return Panel The created collapsable form
 -- @realm client
 function vgui.CreateTTT2Form(parent, name)
-	local form = vgui.Create("DFormTTT2", parent)
+	local form = vgui.Create("DFormTTT2", parent, name)
 
 	form:SetName(name)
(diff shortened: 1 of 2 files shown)


saibotk (approved):
Looks good, is this a normal thing to pass to the vgui function? Sorry currently traveling so cant look that up quickly otherwise looks sane 👌

EntranceJew:
I believe so, I don't really understand why SetName is too late to make the similar change for GetName and not for display/title purposes.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
```diff
@@ -640,7 +640,7 @@ derma.DefineControl("DFormTTT2", "", PANEL, "DCollapsibleCategoryTTT2")
 -- @return Panel The created collapsable form
 -- @realm client
 function vgui.CreateTTT2Form(parent, name)
-	local form = vgui.Create("DFormTTT2", parent)
+	local form = vgui.Create("DFormTTT2", parent, name)
 
 	form:SetName(name)
```
Histalek, line 645: This would be redundant then, right?

Definitely weird that this seemingly does not set the name of the panel ...

EntranceJew: i think SetName changes the display string somewhere down the line, but I need to pass it when vguiCreate is run for then to not be named dpanel, i tested without the SetName and that is what I observed

Histalek (approved):

PR #1051, opened 2023-08-26
override weapon `Kind` to allow changing equipment slots

EntranceJew:
a lot of throwables out there that don't go to the nade slot

### DIFF
--- a/lua/terrortown/lang/en.lua
+++ b/lua/terrortown/lang/en.lua
@@ -1864,6 +1864,7 @@ L.tbut_adminarea = "ADMIN AREA:"
 -- 2023-08-10
 L.equipmenteditor_name_damage_scaling = "Damage Scaling"
 
+
 -- 2023-08-11
 L.equipmenteditor_name_allow_drop = "Allow Drop"
 L.equipmenteditor_desc_allow_drop = "If enabled, the equipment can be dropped freely by the player."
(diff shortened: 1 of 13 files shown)


lua/terrortown/lang/en.lua
```diff
@@ -1874,3 +1875,17 @@ L.equipmenteditor_desc_drop_on_death_type = "Attempt overriding the action taken
 L.drop_on_death_type_default = "Default (weapon-defined)"
 L.drop_on_death_type_force = "Force Drop on Death"
 L.drop_on_death_type_deny = "Deny Drop on Death"
+
+-- 2023-08-26
+L.equipmenteditor_name_kind = "Equipment Slot"
```
saibotk, line 1880: Can you add those commented out to all other languages?

saibotk (changes requested):

lua/terrortown/lang/en.lua
```diff
@@ -1864,6 +1864,7 @@ L.tbut_adminarea = "ADMIN AREA:"
 -- 2023-08-10
```
saibotk, line 1864: Could you do us a favor and add those too to all other languages?

saibotk (changes requested):

lua/terrortown/lang/en.lua
EntranceJew: i didn't understand at first, but I think I got it now

saibotk (approved):

Histalek (approved):

PR #1052, opened 2023-08-27
allow extending item/role menus without detouring as much

EntranceJew:
goes in hand with #1050

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
   - `overrideDropOnDeath` now permits forcing weapons to be dropped instead of removed on death (by @EntranceJew)
   - "Damage Scaling" editable under "Balance Settings" (by @EntranceJew)
 - `vgui.CreateTTT2Form` passes the name on so that it can be accessed via `Panel:GetName()`  (by @EntranceJew)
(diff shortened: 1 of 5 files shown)


saibotk (approved):
LGTM

Histalek (approved):

TimGoll:
Just for my understanding - this is a solution so that you don't have to use the filebased menu creation?

Not that big of a fan of this change. But then again, TTT is already a mess, so who cares I guess :D

EntranceJew:
For stuff that people make, it'll still be significant for them to add their own menus and such, but this was mostly to facilitate putting relevant options next to things that isn't within code an addon developer controls [like I did with SMORE](https://steamcommunity.com/sharedfiles/filedetails/?id=3032725825), which I've also been using to prototype porting some stray ULX-dependent options into the relevant role/item menus to alleviate the number of server addons that don't affect global behavior.

TimGoll:
> For stuff that people make, it'll still be significant for them to add their own menus and such, but this was mostly to facilitate putting relevant options next to things that isn't within code an addon developer controls [like I did with SMORE](https://steamcommunity.com/sharedfiles/filedetails/?id=3032725825), which I've also been using to prototype porting some stray ULX-dependent options into the relevant role/item menus to alleviate the number of server addons that don't affect global behavior.

Thanks. Understandable :)

PR #1053, opened 2023-08-27
Random Ammo Spawn Issue, fixes #1049

### LINKED ISSUE #1049: Random Ammo Spawn inconsistency

Leave one of the following:

- Latest GitHub version

Using the function "entspawn.SpawnRandomAmmo" seems to always spawn the ammo "item_ammo_revolver_ttt", though in map reset random ammo spawns work as they should

## To reproduce

Just try to luarun the following: "entspawn.SpawnRandomAmmo(player.GetAll()[1])"
It will spawn a ammo pack at your position if you're the first player.

## Expected behaviour

It should spawn a random ammo, not always deagle/revolver ammo.

Things I found out so far:

In the code, there is "SPAWN_TYPE_WEAPON" and "SPAWN_TYPE_AMMO" which have the numbers 1 and 2 respectively.

In the function "entspawn.SpawnRandomAmmo" TTT2 uses "SPAWN_TYPE_AMMO" as "entType" for the SpawnTable, which results in a 2 in "entspawn.SpawnEntities" for "entType". But because we also set "AMMO_TYPE_RANDOM", which has the value 1, things don't match up in "entspawn.SpawnEntities" (line 147 in entspawn.lua).

This bug however does not occur on map reset, because the spawnTable for the ammo seems to not use "SPAWN_TYPE_AMMO" but just a "1":

image

In this image I printed the table from line 270 in "entspawn.lua".

Because in this table, the "entType" is 1, things again match up in "entspawn.SpawnEntities".

This just seems inconsistent to me and more like an oversight that probably resulted in "normal" behaviour as it was never noticed.

I'm not that deep into the code and can't find where the "1" is set for the table seen in the image.. Maybe someone else has an Idea.

EntranceJew:
The results are similar, sometimes but the semantics differ. In this case, the final random parameter was so future-proofed that it failed to distinguish itself. Maybe a jargon file of some form would help?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
   - This fixes errors in the F1 Menu language selection
 - Fixed two unmatched ConVars in performance menu (by @NickCloudAT)
 - Fixed Round End Scoreboard (Round Begin) error if a player disconnected while round with no score events (by @NickCloudAT)
+- Fixed behavior of `entspawn.SpawnRandomAmmo` to produce non-deagle ammo. (by @NickCloudAT, mostly)
 
 ## [v0.11.6b](https://github.com/TTT-2/TTT2/tree/v0.11.6b) (2022-09-25)
 
--- a/lua/ttt2/libraries/entspawn.lua
+++ b/lua/ttt2/libraries/entspawn.lua
@@ -45,7 +45,7 @@ end
 -- @realm server
 function entspawn.SpawnRandomWeapon(ent)
 	local spawns = {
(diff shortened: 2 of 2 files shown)


Histalek (approved):
Good catch!

Histalek:
> The results are similar, sometimes but the semantics differ. In this case, the final random parameter was so future-proofed that it failed to distinguish itself. Maybe a jargon file of some form would help?

I don't think a jargon file would be the best approach, we should probably improve our documentation regarding these constants in `sh_init`. I'm not sure if/how our api documentation tool is able to reflect these things.

saibotk (approved):
Thank you very much!

PR #1056, opened 2023-09-20
Add the missing "L.c4_disarm_t" translation in C4 and improve Simplified Chinese and Traditional Chinese.

sbzlzh:
The discovery of the missing L.c4_disarm_t was found when I had an error in the screen after setting the time after using C4.The translation of missing L.c4_disarm_t is modified from the original TTT.


saibotk (approved):
Good catch thanks 🙏

sbzlzh:
:D

Histalek (approved):

PR #1057, opened 2023-09-28
allow reading files from data_static

EntranceJew:
not tested extensively, feature still in dev for gmod itself

### DIFF
--- a/gamemodes/terrortown/entities/entities/ttt_traitor_button/init.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_traitor_button/init.lua
@@ -16,14 +16,15 @@ local path = dir .. "/" .. game.GetMap() .. ".json"
 local function ReadMapConfig()
 	file.CreateDir(dir)
 
-	local modTime = (not file.Exists(path, "DATA") and (lastRead + 1)) or file.Time(path, "DATA")
+	local modTime = (not file.Exists("data_static/" .. path, "GAME") and (lastRead + 1)) or file.Time("data_static/" .. path, "GAME")
+	modTime = modTime <= lastRead and (not file.Exists(path, "DATA") and (lastRead + 1)) or file.Time("data_static/" .. path, "DATA")
 
 	if modTime <= lastRead then
 		return TButtonMapConfig
 	end
 
 	lastRead = modTime
-	local content = file.Read(path, "DATA")
+	local content = (file.Exists(path, "DATA") and file.Read(path, "DATA")) or file.Read("data_static/" .. path, "GAME")
 
 	if not content then
 		return TButtonMapConfig
--- a/gamemodes/terrortown/entities/entities/ttt_traitor_button/init.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_traitor_button/init.lua
@@ -17,7 +17,7 @@ local function ReadMapConfig()
 	file.CreateDir(dir)
 
 	local modTime = (not file.Exists("data_static/" .. path, "GAME") and (lastRead + 1)) or file.Time("data_static/" .. path, "GAME")
-	modTime = modTime <= lastRead and (not file.Exists(path, "DATA") and (lastRead + 1)) or file.Time("data_static/" .. path, "DATA")
+	modTime = modTime <= lastRead and (not file.Exists(path, "DATA") and (lastRead + 1)) or file.Time(path, "DATA")
 
 	if modTime <= lastRead then
 		return TButtonMapConfig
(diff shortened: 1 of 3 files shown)


saibotk:
Nice!
Is there any documentation on what is allowed in the folder or how did you come up with this?
E.g. is there any standard folder for the map entspawnscript json to follow the same path in the data_static folder?

EntranceJew:
I did some testing on the `Dev` branch to confirm that there was no auto-overlay behavior, I was prompted to look into it after I saw this commit in the [gmod discord](https://commits.facepunch.com/461197)

There was some discussion about it but ultimately Rubat said:
> it's just a folder you file.Read from, omg you guys
which revealed it was likely just going to exist along-side `data/`, which makes sense with my testing.
(no `DATA_STATIC` context, no reading `data/file.txt` in place of `data_static/file.txt` or vice-verse, etc)

Ultimately means there is no convention and the fall-through behavior is up to us to enforce if we desire it. Otherwise it's just another folder we can't write to like `maps/`. It probably only obeys the filetype restrictions that `gmad.exe` enforces globally.

I use gmpublisher so I haven't actually experimented with what `gmad.exe` will or won't reject just yet.

saibotk:
Okay, so there is nothing really special about it?

Then another question, what benefits do we get from adding the new folder?

And are these two the only parts that actually read from the data folder? I dont remember currently but i could imagine there being more.

EntranceJew:
The only thing special about it is that it isn't ignored by GMA like data is, so people can create addons out of their weapon placements similar to all the rearm script addons that are out there for TTT.

Now people can distribute weapon spawn rebalance packs or button fix configs for maps specifically using TTT2's files, because even with the `ttt2.fgd` you can't specify a random grenade spawn which is a small thing but could be huge for how someone's got their configurations laid out.

image
These are all the places I see file.Read being run, so if there's something else reading files somewhere, I'm imagining it's going through one of these locations.

The only one I didn't change is reading the TTT1 text files included inside `maps/` which, since it's legacy behavior, is something better suited for that step in the process before it gets converted to the TTT2 spawn file.

saibotk:
Alright sounds sensible, thanks for explaining this in detail!

saibotk (approved):

lua/ttt2/libraries/entspawnscript.lua
```diff
@@ -353,7 +354,12 @@ if SERVER then
 	-- @internal
 	-- @realm server
 	function entspawnscript.ReadFile(dir)
-		return utilJSONToTable(fileRead(dir .. gameGetMap() .. ".json", "DATA"))
+		local fullDir = dir .. gameGetMap() .. ".json"
+		if fileExists(fullDir, "DATA") then
+			return utilJSONToTable( fileRead(fullDir, "GAME") )
```
Histalek, line 359: This should be `DATA` instead of `GAME`, right?

```suggestion
			return utilJSONToTable( fileRead(fullDir, "DATA") )
```

gamemodes/terrortown/entities/entities/ttt_traitor_button/init.lua
```diff
@@ -16,14 +16,15 @@ local path = dir .. "/" .. game.GetMap() .. ".json"
 local function ReadMapConfig()
 	file.CreateDir(dir)
 
-	local modTime = (not file.Exists(path, "DATA") and (lastRead + 1)) or file.Time(path, "DATA")
+	local modTime = (not file.Exists("data_static/" .. path, "GAME") and (lastRead + 1)) or file.Time("data_static/" .. path, "GAME")
+	modTime = modTime <= lastRead and (not file.Exists(path, "DATA") and (lastRead + 1)) or file.Time("data_static/" .. path, "DATA")
```
Histalek, line 20: There should be no `data_static/` here to retain the old behaviour, correct?

```suggestion
	modTime = modTime <= lastRead and (not file.Exists(path, "DATA") and (lastRead + 1)) or file.Time(path, "DATA")
```

Histalek (changes requested):
So does this mean that server owners can now setup entspawnscripts that only affect TTT2(or rather terrortown) and don't mess with other game modes?
That's neat.

EntranceJew:
> So does this mean that server owners can now setup entspawnscripts that only affect TTT2(or rather terrortown) and don't mess with other game modes? That's neat.

Yes, totally. And excellent catches. Subconsciously I think I wanted to just force it to read from "data/" as "GAME" and fallback to "data_static/" as "GAME" but that wouldn't work for writes, and got distracted. I appreciate the oversight for situations like this.

image
I was also able to pack the files with gmad in the dev branch, although I'm not able to test reading a packed gma from the addons folder by itself at the moment, but I do know that you can always read anything inside `garrysmod/` as "GAME" so it isn't that big of a deal.

image


At least, I was able to prove that `data/` is still not whitelisted:
image

Histalek (approved):
Definitely a nice addition. Thank you very much!

PR #1058, opened 2023-10-05
voice volume scaling / spectator ducking

EntranceJew:
I think there should be hooks in here somewhere so equipment like the Mute Radio can continue to exist, but I'm not sure where.

I don't know if I make note of it specifically but bots report distinct values for SteamID64 so that workaround isn't necessary anymore.

problems:
 - if someone dies while talking, once their audio becomes non-3D their volume should be changed prior to that, but presently I don't know where that happens, because the name-plate stays team-colored until they finish talking.
 - also the dropdown selector raises errors for me if changed several times which I believe is just an issue that already existed, the convar still functions

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_voice.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_voice.lua
@@ -26,6 +26,27 @@ local MutedState
 
 g_VoicePanelList = nil
 
+---
+-- @realm client
+local duck_spectator = CreateConVar("ttt2_voice_duck_spectator", "0", {FCVAR_ARCHIVE})
+
+---
+-- @realm client
+local duck_spectator_amount = CreateConVar("ttt2_voice_duck_spectator_amount", "0", {FCVAR_ARCHIVE})
+
+---
+-- @realm client
+local scaling_mode = CreateConVar("ttt2_voice_scaling", "linear", {FCVAR_ARCHIVE})
+
+local function CreateVoiceTable()
+	if not sql.TableExists("ttt2_voice") then
+		local query = "CREATE TABLE ttt2_voice (guid TEXT PRIMARY KEY, mute INTEGER DEFAULT 0, volume REAL DEFAULT 1)"
+		sql.Query(query)
+	end
+end
+
+CreateVoiceTable()
+
 local function VoiceTryEnable()
 	local client = LocalPlayer()
 
@@ -135,6 +156,8 @@ function GM:PlayerStartVoice(ply)
 	if not IsValid(ply) then return end
 
 	local client = LocalPlayer()
+
+	VOICE.UpdatePlayerVoiceVolume(ply)
 
 	if not IsValid(g_VoicePanelList) or not IsValid(client) then return end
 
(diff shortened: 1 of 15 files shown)


gamemodes/terrortown/gamemode/client/cl_voice.lua
```diff
@@ -346,6 +371,116 @@ end
...
+VOICE.ScalingFunctions = {
+	power4 = VOICE.LinearToPower4,
+	log = VOICE.LinearToLog,
+	linear = VOICE.LinearToLinear,
+}
+
+VOICE.GetScalingFunctions = function()
+	local opts = {}
+	for mode in pairs(VOICE.ScalingFunctions) do
+		table.insert(opts, {
```
TimGoll, line 417: at some point we decided to use absolute indexing when inserting into a table because it is waay faster. I guess this should be possible here as well?

Something like
```lua
table[#table + 1] = <new entry>
```

gamemodes/terrortown/gamemode/client/cl_voice.lua
```diff
@@ -346,6 +371,116 @@ end
...
+	if duck_spectator:GetBool() and ply:IsSpec() then
+		vol = vol * (1 - duck_spectator_amount:GetFloat())
+	end
+	local func = VOICE.ScalingFunctions[scaling_mode:GetString()]
+	local out_vol = vol
+	if isfunction(func) then
+		out_vol = func( vol )
+	end
+	ply:SetVoiceVolumeScale( out_vol )
+	return out_vol, mute
```
TimGoll, line 481: our conding style looks a bit different. Something like `if ply.SetMute then ply:SetMute(mute) end` should only be a one-liner if it is a simple fencing check

Also there are a few new lines missing

TimGoll (changes requested):
Besides the things I commented you sometimes add spaces around brackets. I don't know if we don't care or if you should change this. Maybe someone else could chime in as well.

I can't say much about the feature itself as it is something I never worked on. I guess it is alright.

Thanks for your contribution as always!

TimGoll:
> if someone dies while talking, once their audio becomes non-3D their volume should be changed prior to that, but presently I don't know where that happens, because the name-plate stays team-colored until they finish talking.

I'm pretty sure that is something that we investigated two years or so ago while working on the UI. I think this was engine handled and not something that can be influenced by us

EntranceJew:
I updated the PR with some changes, is that more like what you had in mind?

> I'm pretty sure that is something that we investigated two years or so ago while working on the UI. I think this was engine handled and not something that can be influenced by us

For this one, the issue I think is more in the timing.
1. The panel doesn't change because it only thinks about changing color when `PlayerStartVoice` and `PlayerEndVoice` are run.
2. On the server, when PlayerCanHearPlayersVoice, the first argument remains true due to despite a team change. The player is still audible, the sudden jump in volume has to do with how the is3D argument changes without an explicit event to the client.

If there were an event that told the observer (a player already spectating) that another player had died (gone from team innocent to spectator) prior to the server returning the is3D, then the observer could run `VOICE.UpdatePlayerVoiceVolume` and not have that happen.

TimGoll (approved):

saibotk:
What about the old functions? Can we remove them? Are those builtin functions? 
And did you make sure they are not used anywhere else?

EntranceJew:
The old functions are builtin, and they're not called anywhere else, which is why this new API provides a centralized location for further adjusting player volume -- because this one function drives it globally, so the fewer things using it and the more things using this API should prevent the player volumes form being knocked when the user doesn't want them to be.

gamemodes/terrortown/gamemode/client/cl_voice.lua
```diff
@@ -279,6 +302,8 @@ function GM:PlayerEndVoice(ply, no_reset)
 	end
 
 	if IsValid(ply) and not no_reset then
+		VOICE.UpdatePlayerVoiceVolume(ply)
+
```
saibotk, line 306: I don't think we need to update the volume on end, do we?

saibotk (changes requested):

gamemodes/terrortown/gamemode/client/cl_voice.lua
EntranceJew: Originally when testing with bots it became useful. I was assuming due to issues where the speaker tags in the top left show the wrong color, that these signals weren't 100% reliable, so having them be set on both edges might prevent any jarring audio volume jumps when someone begins speaking.

saibotk: I think its safe to assume that the Begin function is always called.
So we can remove this, since the volume adjustment is not needed here. 
Also i dont think this should cause any jarring.

saibotk (approved):
Amazing, thanks! 
Love the spectator ducking!

PR #1066, opened 2023-10-06
[WIP] Added: Slider text area for values

NickCloudAT:
Changed how the textbox for sliders work. You can now klick on them to make changes directly via input. WIP

Demonstration:
https://cloud.nickcloud.at/index.php/s/NEXjr8SCfnC5jdc

Currently the textbox does not allow to go below the minimum, but above the maximum. Should I change it so both is disallowed? Or maybe that both is allowed to have more fine tuning?

Also the function "getHighestParent": Maybe this should move to a global function?

Currently the text box is not very pretty, but I actually have to idea right now on how to make it prettier..
Open for all ideas.

Changelog will follow at last.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
@@ -850,7 +850,7 @@ function SKIN:PaintSliderTextAreaTTT2(panel, w, h)
 		)
 	else
 		self:PaintTextEntryTTT2(panel, w, h)
-		local vguiColor = util.GetActiveColor(util.GetChangedColor(util.GetDefaultColor(vskin.GetBackgroundColor()), 25))
+		local vguiColor = utilGetActiveColor(utilGetChangedColor(utilGetDefaultColor(vskinGetBackgroundColor()), 25))
 		panel:DrawTextEntryText(vguiColor, vguiColor, vguiColor)
 	end
 end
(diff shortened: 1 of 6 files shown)


NickCloudAT:
WIth the last commit I prettified the Textarea by using the TTT2 Textentry.
Also made the code a bit simpler I guess and moved the function "getHighestParent" to the utils file and renamed it to "getHighestParentUtil".

Values entered can now also go below or above the min/max value of the slider.

Dark theme:
image

Light theme:
image

gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
```diff
@@ -836,16 +836,23 @@ function SKIN:PaintSliderTextAreaTTT2(panel, w, h)
...
+			panel:GetFont(),
+			0.5 * w,
+			0.5 * h,
+			colorText,
+			TEXT_ALIGN_CENTER,
+			TEXT_ALIGN_CENTER
+		)
+	else
+		self:PaintTextEntryTTT2(panel, w, h)
+		local vguiColor = util.GetActiveColor(util.GetChangedColor(util.GetDefaultColor(vskin.GetBackgroundColor()), 25))
```
TimGoll, line 853: check the other functions, there are cached versions of this (most of the time only the `.` is removed)

TimGoll (changes requested):
very nice, only a minor nitpick

gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
NickCloudAT: Should now be correct :)

TimGoll (approved):

saibotk (approved):

Issue #1068, opened 2023-10-07
Language selection error


mexikoedi:
## Your version of TTT2 (mandatory)

- Workshop version from Steam

## Describe the bug (mandatory)

If you switch the language in the F1 menu to another language and then switch back to another language you will receive the following error in the console:
[TTT2 (Base) - v0.11.7b] lua/includes/modules/cvars.lua:48: attempt to call local 'callback' (a nil value)
  1. unknown - lua/includes/modules/cvars.lua:48
   2. SetString - [C]:-1
    3. SetConVarValues - gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua:398
     4. ChooseOptionID - gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua:188
      5. ChooseOptionValue - gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua:197
       6. SetValue - gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua:336
        7. DoClick - gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua:307
         8. OnMouseReleased - lua/vgui/dlabel.lua:237
          9. unknown - lua/vgui/dmenuoption.lua:76

The language completion rate is also bugged.
For example English is 100 % complete then you change to another language and back to English and you will see something different like 81.7 % or 45.8 % ... .

## To reproduce

Steps to reproduce the behaviour:

1. Go to the F1 menu
2. Click on Language
3. Select a different language
4. Switch back to another language
5. See error and weird completion rate

## Expected behaviour

It should not create an error and the completion rate should be correct.

## Context (please provide as much as you can)

Tested with TTT2 Base, ULX, ULIB, ULX for TTT2 and some TTT2 roles installed.

TimGoll:
yes, the language selection has a few minor bugs that we have to fix

NickCloudAT:
Not sure if I'm correct, but it seems like the game can't keep track of the callback function created here: https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua#L343

If I take out those two line here:
https://github.com/TTT-2/TTT2/blob/master/lua/terrortown/menus/gamemode/language/language.lua#L31
[...]

PR #984, opened 2022-08-05
Shopeditor: Replace syncing with database module

ZenBre4ker:
Building on the database module I ripped out the entire custom equipment editor syncing.

PR #1073, opened 2023-10-16
Fix/Changes: Combobox callback fixes/changes

NickCloudAT:
With this commit comboboxes won't throw an error for a "nil" callback anymore. For some reason the game can not keep track of callbacks correctly anymore.

Also added a "oldValue" to the onSelect/onChange function of the combobox to ensure the previous set value is still sent.

I would happily have some people test this to see if I broke something or not and give their 2 cent on this.

Some things I'm not yet sure and maybe somebody has a better idea:
- Should the "oldValue" be nil if "self.selected" is not present or should we set it to the new value?
- I put "oldValue" as the last argument as to not break compatibility with perhaps already existing comboboxes.

Related issue: https://github.com/TTT-2/TTT2/issues/1068

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Updated file code to read from `data_static` as fallback in new location allowed in .gma (by @EntranceJew)
 - Scoreboard now sets preferred player volume and mute state in client's new `ttt2_voice` table (by @EntranceJew)
   - Keyed by steamid64, making it more reliable than UniqueID or the per-session mute and volume levels.
+- Added "olValue" to the onSelect/onChange function of comboboxes (by @NickCloudAT)
+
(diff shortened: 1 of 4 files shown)


ZenBre4ker:
Didnt look at your changes but I addressed similar stuff in #984, I believe. Gonna check that tomorrow.

saibotk:
I have investigated this on my own and found the root cause for the issue.

Will open a PR to fix this without extra clean up functions, but why exactly are you adding the oldValue here? Do we need this?

saibotk:
@ZenBre4ker actually you made the same mistake:
https://github.com/TTT-2/TTT2/pull/984/files#diff-a910efa517f7a2fbab8e7128c8091e74f3e0af9cc28e617e969c31612c317c60R104

Will write up the explaination in my PR

saibotk:
Closed in favor of #1075 

Thanks for investigating and writing this up!

PR #1074, opened 2023-10-16
Chore/multiple minor(?) fixes

Histalek:
Ok while playing around with the lua language server [1] and its glua-api-snippets [2] addon it made me aware of multiple issues.

Not gonna lie there is a lot and it's probably not all as our current docs don't integrate perfectly with the language server.

i've added the `.luarc.json.example` as a reference with my used settings. There are a few more globals that are not recognized, but frankly i couldn't be arsed to add them all. (mostly `EVENT_*` globals). See my first commit message for a bit more info on using the .luarc config with the glua definitions. Or see the documentation of the linked Repositories.

Also NOTE: Everything in this PR is currently untested. All `chore:` commits should be fine and not actually change functionality. I'm not so sure with the `fix:` commits, those could potentially be breaking (or hopefully **fixing** :D )

[1] https://github.com/LuaLS/lua-language-server
[2] https://github.com/luttje/glua-api-snippets/tree/main


TimGoll:
I'm a bit confused. There are some changes that look like they were previously bugs. Why were they never noticed? How did you find them?

saibotk:
The magic of type checking and language server diagnostics :sparkles:

Histalek:
> I'm a bit confused. There are some changes that look like they were previously bugs. Why were they never noticed? How did you find them?

yes exactly :D there are a lot of things that should not have happened!

I noticed this because @saibotk made me aware that someone integrated the functions and symbols documented in the GMOD wiki with the lua-language-server. Which means intellisense and autocompletion for gmod-lua functions and symbols which in turn also enabled better linting and type-checking.

(The longterm goal is to generate such Definitions for our TTT2 Codebase so that we can have auto-completion for our functions as well and also publish those so that potential addon developers can benefit from those as well :) )

TimGoll:
@Histalek wow

EntranceJew:
I had been using this on my devstreams, and I agree, the notation once it's all set up is really great especially since the types are already documented so thoroughly, that with literally one more `-` per line it could be used for autocomplete and typechecking.

Histalek:
> I had been using this on my devstreams, and I agree, the notation once it's all set up is really great especially since the types are already documented so thoroughly, that with literally one more `-` per line it could be used for autocomplete and typechecking.

We would need to switch the name and type for @params, but yes most of the work is already done for the documentation side of things.

lua/ttt2/extensions/draw.lua
```diff
@@ -112,7 +112,7 @@ function draw.ShadowedBox(x, y, w, h, color, scale)
 	drawBox(x + shift2, y + shift2, w, h, tmpCol)
 	drawBox(x + shift1, y + shift1, w, h, tmpCol)
 	drawBox(x + shift1, y + shift1, w, h, tmpCol)
-	drawBox(x, y, w, h, t, color)
+	drawBox(x, y, w, h, color)
```
saibotk, line 115: wtf

.luarc.json.example
```diff
@@ -0,0 +1,34 @@
...
+	"diagnostics.ignoredFiles": "Enable",
+	"format.enable": false,
+	"runtime.special": {
+		"include": "require",
+		"includeCS": "require"
+	},
+	"runtime.version": "Lua 5.1",
+	"runtime.nonstandardSymbol": [
+		"//",
+		"/**/",
```
saibotk, line 30: be gone!

saibotk (changes requested):

.luarc.json.example
Histalek: done

TimGoll (approved):
Like I said yesterday on discord, I'm really impressed with this. I don't see anything that would break, so it LGTM

saibotk (approved):

EntranceJew:
I checked the changelist but it doesn't seem to show this, when removing `AccessorFuncDT` (something I agree with doing), did we appropriately migrate the variables in question to the appropriate `SetupDataTables()` calls as `DTVar()` calls? Without doing that, these changes could cause some difficult to diagnose errors.

This is on top of bringing down and migrating the files for:
```
weapon_ttt_phammer
weapon_ttt_push
ttt_flame
ttt_hat_deerstalker
```
Which hasn't happened yet.

Histalek:
> I checked the changelist but it doesn't seem to show this, when removing `AccessorFuncDT` (something I agree with doing), did we appropriately migrate the variables in question to the appropriate `SetupDataTables()` calls as `DTVar()` calls? Without doing that, these changes could cause some difficult to diagnose errors.

Excellent point. Both the health_station [1] and c4 [2] create the self.dt.<var> alias' with `DTVar()` in their respective `SetupDataTables()` call. Although now with the AccessorFunc we do not access the self.dt.<var> (DataTable alias) anymore but instead self.<var> which is most likely not tied to the DataTable.

As far as i can tell, using `Entity:NetworkVar`[3] instead of `DTVar()` in the `SetupDataTables()` call and removing the `AccessorFunc()` call for these vars completely, should get us back to where we were before the `AccessorFuncDT()` removal.

I'll probably create a PR for this tomorrow. Thanks a lot for noticing! 

[1] https://github.com/TTT-2/TTT2/blob/94f158b6024cc9c23e7dc67d75ca810b7d02b008/gamemodes/terrortown/entities/entities/ttt_health_station.lua#L40
[2] https://github.com/TTT-2/TTT2/blob/94f158b6024cc9c23e7dc67d75ca810b7d02b008/gamemodes/terrortown/entities/entities/ttt_c4/shared.lua#L92
[3] https://wiki.facepunch.com/gmod/Entity:NetworkVar

PR #1075, opened 2023-10-16
fix: Combobox ConVar change callback deletion

### LINKED ISSUE #1068: Language selection error

saibotk:
Fixes #1068

We need to remove the callback in a timer, because otherwise the ConVar change callback code
will throw an error while looping over the callbacks.
This happens, because the callback is removed from the same table that is iterated over.
Thus, the table size changes while iterating over it and leads to a nil callback as the last entry.

See https://github.com/Facepunch/garrysmod/blob/1b512930d1f8fb1acf6235e584c4ec1ff84e9362/garrysmod/lua/includes/modules/cvars.lua#L44
and
https://github.com/Facepunch/garrysmod/blob/1b512930d1f8fb1acf6235e584c4ec1ff84e9362/garrysmod/lua/includes/modules/cvars.lua#L97

here the code uses table.Remove, which causes the table to be reindexed and resized, this is the root cause for the problem.
We should thus avoid removing callbacks in the callback itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Scoreboard now sets preferred player volume and mute state in client's new `ttt2_voice` table (by @EntranceJew)
   - Keyed by steamid64, making it more reliable than UniqueID or the per-session mute and volume levels.
 
+### Fixed
+
+- Fixed removing the convar change callback in `DComboboxTTT2` (by @saibotk)
+
 ## [v0.11.7b](https://github.com/TTT-2/TTT2/tree/v0.11.7b) (2022-08-27)
 
 ### Added
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
@@ -338,23 +338,29 @@ end
 
 local convarTracker = 0
 ---
--- @param Panel menu to set the value of
+-- @param Panel panel to set the value of
 -- @param string conVar name of the convar
-local function AddConVarChangeCallback(menu, conVar)
-	convarTracker = convarTracker % 1023 + 1
-	local myIdentifierString = "TTT2F1MenuConVarChangeCallback" .. tostring(convarTracker)
+local function AddConVarChangeCallback(panel, conVar)
+	convarTracker = convarTracker + 1
+	local myIdentifierString = "TTT2F1MenuComboboxConVarChangeCallback" .. tostring(convarTracker)
 
-	local function OnConVarChangeCallback(conVarName, oldValue, newValue)
-		if not IsValid(menu) then
-			cvars.RemoveChangeCallback(conVarName, myIdentifierString)
+	local callback = function(conVarName, oldValue, newValue)
+		if not IsValid(panel) then
+			-- We need to remove the callback in a timer, because otherwise the ConVar change callback code
+			-- will throw an error while looping over the callbacks.
+			-- This happens, because the callback is removed from the same table that is iterated over.
+			-- Thus, the table size changes while iterating over it and leads to a nil callback as the last entry.
+			timer.Simple(0, function()
+				cvars.RemoveChangeCallback(conVarName, myIdentifierString)
+			end)
 
 			return
 		end
 
-		menu:SetValue(newValue, true)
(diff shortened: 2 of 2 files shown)


TimGoll (approved):

Histalek (approved):
LGTM

ZenBre4ker:
Interesting and makes sense.
Gonna approve tomorrow and take a closer look.
Thanks :D

saibotk:
Too late tim was here already

TimGoll:
> Interesting and makes sense. Gonna approve tomorrow and take a closer look. Thanks :D

well it is already merged :D

NickCloudAT:
I would still like to comment on this:

I did think about using a timer, but I decided against it as I though 0 tick timers are an "ugly" solution.
Also: This PR does not fix the language % calculations in the F1 menu. It's still takes % of the old language set.

The "oldValue" was merely a nice to have in my PR I guess as you may want the previous value directly from the combobox and not have to get the conVar first, which is more CPU intensive as stated in the wiki.

saibotk:
Yes, this is only a fix for the callbacks themselves. 
I do not think that this is particularly ugly in this context, but rather i think of this whole functionality as conceptually flawed and would rather replace this sooner than later with something that does not need to remove itself.

But please open up a new PR for the language fix, if you can spare the time! Id love to have that one in too.

NickCloudAT:
> Yes, this is only a fix for the callbacks themselves. I do not think that this is particularly ugly in this context, but rather i think of this whole functionality as conceptually flawed and would rather replace this sooner than later with something that does not need to remove itself.
> 
> But please open up a new PR for the language fix, if you can spare the time! Id love to have that one in too.

I give you that the concept is probably flawed, yes. Maybe it's just my concept of "ugly" that makes it weird in my mind.

If I get to fix the lang thingy: Should I maybe still add the "oldValue" too? This still gives advantages I think.

NickCloudAT:
> What advantage exactly? I did not really see any need for it, but maybe i just did not get it correctly.

Maybe I'm just dumb, but I did not see another way of getting the old value except for reading out the conVar value itself. Which would be with "GetConVar". The GMOD-Wiki to that states:
"Due to this function using [GetConVar_Internal](https://wiki.facepunch.com/gmod/Global.GetConVar_Internal) internally it tends to be relatively slow. Please attempt to 'cache' the return of what you used to make it instead of using this function."

I mean in daily use this would probably not make that much of a difference, but IMO giving back the "oldValue" in the function itself is cleaner then needing to get it from the convar.
Also in my PR I put the "onSelect" AFTER the convar gets set, which in turn makes it impossible to get the old value of the convar but in turn would automatically fix the language % error as the VGUI refresh (at least IMO) should happen after a convar is changed, not before.

EntranceJew:
The wiki's comment is out of date, in the code itself on the FP repo you can see it is caching the reference to the convar

saibotk:
"Performance" is super negligible here as its not called often, but sure go ahead and fix the language percentage.

PR #1081, opened 2023-10-18
[Rendering] SVG to Material

TimGoll:
[Based on this project](https://github.com/noaccessl/glua-SVG) I created a functions that creates GMod materials with mipmapping from SVGs. This can be used to replace all role icons etc.

**Completely untested!**


TimGoll:
Current roadblock mentioned here: https://github.com/Facepunch/garrysmod-issues/issues/5595

PR #1085, opened 2023-10-20
docs: Fix partial docstrings

Histalek:
Also 'Doublicated' was bugging me, so i fixed that too.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_shopeditor.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_shopeditor.lua
@@ -42,7 +42,7 @@ function ShopEditor.GetEquipmentForRoleAll()
 		local name = WEPS.GetClass(eq)
 
 		if name
-			and not eq.Doublicated
+			and not eq.Duplicated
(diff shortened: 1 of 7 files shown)


saibotk (approved):
Holy cow finally

TimGoll:
I think it is `@param type name`, you did it the other way around.

And regarding your changed variable name: I think if you change it anyway, you should change it to lowercase as well

saibotk:
Actually that would be somewhat inconsistent with the other vars on an equipment like CanBuy for example

Histalek:
> I think it is `@param type name`, you did it the other way around.

True i mixed it up with how it's done for the lua language server, will fix it.

> And regarding your changed variable name: I think if you change it anyway, you should change it to lowercase as well

AFAIK all SWEP vars are Capitalized and i don't think we should change that

TimGoll (approved):
I understand your reasoning behind the decision to keep it upper case. I just want to mention that there are already incosistencies. I'm pretty sure there are some variables in the table that are lowercase

PR #1087, opened 2023-10-21
fix: More undefined var fixes

Histalek:
Still needs some work. Inline comments where i wasn't sure what to do.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_targetid.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_targetid.lua
@@ -8,7 +8,6 @@ local surface = surface
 local draw = draw
 local GetPlayers = player.GetAll
 local math = math
-local table = table
 local IsValid = IsValid
 local hook = hook
 local targetid = targetid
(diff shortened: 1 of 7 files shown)


gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -58,7 +58,7 @@ local materialBaseOverlay = Material("vgui/ttt/dynamic/sprite_base_overlay")
 -- @hook
 -- @realm client
 function GM:GetClassHints()
-	return ClassHint
+	return ClassHint -- this is most definitely not localized
```
Histalek, line 61: i have literally no idea what this would like to reference.

@TimGoll do you have an idea?

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenubutton_ttt2.lua
```diff
@@ -123,7 +123,7 @@ end
 ---
 -- @ignore
 function PANEL:PerformLayout()
-	DLabel.PerformLayout(self)
+	DLabel.PerformLayout(self) -- where does DLabel come from?!?!
```
Histalek, line 126: Maybe this is actually correct, but most of the GUI stuff just goes right above my head :/

Histalek (commented):
For the sake of getting this somewhere i commented on the two places where i need help.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenubutton_ttt2.lua
TimGoll: I spotted some elements like this in the base gmod code. I guess this is either a bug or something "localized" as a global

gamemodes/terrortown/gamemode/client/cl_targetid.lua
TimGoll: I recently read a report (on discord I think) that hints aren't working anymore in TTT2 targetID. This might be the reason why it is broken. I propably removed this unintenionally:

https://github.com/Facepunch/garrysmod/blob/e189f14c088298ca800136fcfcfaf5d8535b6648/garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua#L11-L23

TimGoll: Howver, looking at it I think there is more missing. Might have to be reworked a bit to fix it.

Histalek: i can't find any other reference to `ClassHint` in the original ttt code. Anyways i'll just throw it in again then we can see how much of it works again.

TimGoll: I think there is more missing to make it work, because right now it is used nowhere

Histalek: `ClassHint` was removed in this commit 9cef0ced0edb102255a5377db5c610f2c2364d4f

so it's definitely not the recent reason for TargetID not working.

For this function we should probably return something else entirely, i have not found out what.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenubutton_ttt2.lua
Histalek: You are right. Maybe this is correct then. I'll take my chances and leave this as is.

gamemodes/terrortown/gamemode/client/cl_targetid.lua
TimGoll: It might be broken since then without anyone noticing though

saibotk:
Can we get this one ready anytime soon? @Histalek do you need any help here?

Histalek:
> Can we get this one ready anytime soon? @Histalek do you need any help here?

This is done except for the `ClassHint` stuff. i didn't find time to investigate that one.

I could just leave it for now and we investigate this some other time or i would need some time to play around with reintroducing it

saibotk (approved):

TimGoll (approved):
As discussed earlier, this is a nice change. If at some point we get problems with ClassHint, then it is because of this PR

PR #1105, opened 2023-10-26
Added: Pulsating healthbar

NickCloudAT:
This PR adds a pulsating effect for the healthbar when a player is at or below 25% of his max health. The puslating gets faster the lower the health gets.

Demonstrational Video: https://cloud.nickcloud.at/index.php/s/mNg4WjWACo8b6xA

I also added a new util function "mapToValue" to map one value to one in another range. I think that's pretty useful.

I added the convar to the Miscellaneous settings as I did not find other sections more fitting. But because I'm not yet sure, and not yet sure about the language strings, I have not yet added the language strings to all language files. I will do so later of course.

Currently the frequency is fixed. I myself think the frequency is perfectly balanced to not be to fast or to slow.

Any suggestions/changes are more than welcome :)


gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
```diff
@@ -31,6 +31,8 @@ if CLIENT then
 		minsize = {w = 240, h = 146}
 	}
 
+	local healthPulsate = CreateConVar("ttt2_hud_pulsate_health", "1", FCVAR_ARCHIVE)
```
saibotk, line 34: This should be defined somewhere outside the Hudelement, and since this is a generic feature not directly tied to the design, we might want to also add this to the old HUD).

lua/terrortown/lang/en.lua
```diff
@@ -1899,3 +1899,7 @@ L.label_voice_scaling = "Voice Volume Scaling Mode"
 L.label_voice_scaling_mode_linear = "Linear"
 L.label_voice_scaling_mode_power4 = "Power 4"
 L.label_voice_scaling_mode_log = "Logarithmic"
+
+-- 2023-10-23
+L.header_miscellaneous_settings = "Miscellaneous Settings"
+L.label_hud_pulsate_health_enable = "Pulsate healthbar when below 25% health"
```
saibotk, line 1905: add a newline at the end & add these to the other languages

lua/ttt2/extensions/util.lua
```diff
@@ -483,6 +483,27 @@ function util.VectorInBounds(vec, lowerBound, upperBound)
...
+-- @param number minValue The minimum value that 'value' can have
+-- @param number maxValue The maximum value that 'value' can have
+-- @param number minTargetValue The minimum value that the mapped value can have
+-- @param number maxTargetValue The maximum value that the mapped value can have
+-- @realm shared
+function util.mapToValue(value, minValue, maxValue, minTargetValue, maxTargetValue)
+	local minValue = minValue
+	local maxValue = maxValue
+	local minTargetValue = minTargetValue
+	local maxTargetValue = maxTargetValue
```
saibotk, line 498: what are these for? 

I'd suggest removing those, as they do not contribute anything.

lua/ttt2/extensions/util.lua
```diff
@@ -483,6 +483,27 @@ function util.VectorInBounds(vec, lowerBound, upperBound)
 		and vec.z > lowerBound.z and vec.z < upperBound.z
 end
 
+---
+-- Maps one integer value to another in a different range
```
saibotk, line 487:
```suggestion
-- Adjusts a numeric value from one range to a different one in a relative transformation.
```

lua/ttt2/extensions/util.lua
```diff
@@ -483,6 +483,27 @@ function util.VectorInBounds(vec, lowerBound, upperBound)
...
+-- @param number value The value that should be mapped
+-- @param number minValue The minimum value that 'value' can have
+-- @param number maxValue The maximum value that 'value' can have
+-- @param number minTargetValue The minimum value that the mapped value can have
+-- @param number maxTargetValue The maximum value that the mapped value can have
+-- @realm shared
+function util.mapToValue(value, minValue, maxValue, minTargetValue, maxTargetValue)
```
saibotk, line 494: Let's rename this to something like `util.transformToRange`

saibotk (changes requested):

lua/terrortown/lang/en.lua
```diff
@@ -1051,7 +1051,7 @@ L.submenu_appearance_crosshair_title = "Crosshair"
 L.submenu_appearance_dmgindicator_title = "Damage Indicator"
 L.submenu_appearance_performance_title = "Performance"
 L.submenu_appearance_interface_title = "Interface"
-L.submenu_appearance_miscellaneous_title = "Misellaneous"
+L.submenu_appearance_miscellaneous_title = "Miscellaneous"
```
TimGoll, line 1054: thanks!

lua/ttt2/extensions/util.lua
```diff
@@ -483,6 +483,27 @@ function util.VectorInBounds(vec, lowerBound, upperBound)
...
+	local minValue = minValue
+	local maxValue = maxValue
+	local minTargetValue = minTargetValue
+	local maxTargetValue = maxTargetValue
+
+	value = mathMax(minValue, mathMin(maxValue, value))
+
+	local targetValue = minTargetValue + (maxTargetValue - minTargetValue) * (value - minValue) / (maxValue - minValue)
+
+	return targetValue
```
TimGoll, line 504: why the local variable? you could return this without storing it in a variable

TimGoll (changes requested):

lua/ttt2/extensions/util.lua
TimGoll: also functions in lua are upper case. Hence `util.TransformToRange`

lua/ttt2/extensions/util.lua
NickCloudAT: Ah, yeah that and the other local variables are just bad habbits from debbuging.. Will remove/change

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
NickCloudAT: I already thought that this would come up :)
I defined it in the hud element as I really had no freakin idea where I should put it else.

So.. where should I put it?
Also, not in misc but under a different menu? I mean sure possible, I just felt it would fit more under misc.

Also, no I'm actually strictly against adding this to the old hud. The old hud is, imo, legacy and should not be altered in such a way.

TimGoll: I'm pretty sure you can remnove the convar and add the setting here, then it will appear in the HUD switcher:

https://github.com/TTT-2/TTT2/blob/42f5bb613bcdeb948dd19ea3b0370b5ec18d45c9/gamemodes/terrortown/gamemode/shared/huds/base_huds/scalable_hud/cl_init.lua#L32-L46

Altough you should probably only do this in the cl_init of the pure_skin hud.

NickCloudAT:
I pushed a new commit which moves the setting to the Hudswitcher menu as a savingKey.
Also this commit fixes broken bool checkboxes in the hudswitcher section.

Added the lang strings to all files.

lua/terrortown/menus/gamemode/appearance/miscellaneous.lua
```diff
@@ -6,5 +6,5 @@ CLGAMEMODESUBMENU.priority = 91
 CLGAMEMODESUBMENU.title = "submenu_appearance_miscellaneous_title"
 
 function CLGAMEMODESUBMENU:Populate(parent)
-
+    --local form = vgui.CreateTTT2Form(parent, "header_miscellaneous_settings")
```
TimGoll, line 9: either keep it in there or remove it completely

TimGoll (approved):

lua/terrortown/menus/gamemode/appearance/miscellaneous.lua
NickCloudAT: Guess it's better to have it at least there. I left it in and uncommented it.

gamemodes/terrortown/gamemode/shared/huds/pure_skin/cl_init.lua
```diff
@@ -50,6 +52,28 @@ function HUD:Initialize()
 	BaseClass.Initialize(self)
 end
 
+---
+-- This function will return a table containing all keys that will be stored by
+-- the @{HUD:SaveData} function.
+-- @return table
+-- @realm client
+function HUD:GetSavingKeys()
+	if not savingKeys then
```
saibotk, line 61: there is no need to store this / cache this

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
```diff
@@ -38,6 +38,7 @@ if CLIENT then
 		self.lpw = lpw
 		self.sri_text_width_padding = sri_text_width_padding
 		--self.secondaryRoleInformationFunc = nil
+		self.healthpulsate = huds.GetStored(HUDManager.GetHUD()).healthPulsate
```
saibotk, line 41:
```suggestion
		self.healthpulsate = huds.GetStored("pure_skin").healthPulsate
```

So this is actually something we are missing, because usually you would just add this variable to the saving keys of the element instead of the HUD.

But we currently do not have an extra UI being generated in the F1 settings for those, which we should!
Dunno if you want to walk the extra mile or maybe someone else volunteers to do this in this PR :D 

Because if you use the current hud this might actually return a different one than pure_skin, this is due to HUDs only being a collection of hudelements. For example incomplete HUDs may use this element as a fallback.

saibotk (changes requested):

gamemodes/terrortown/gamemode/shared/huds/pure_skin/cl_init.lua
saibotk: just saw, that we also do this in the scalable hud, dunno why though :D i think this can be removed too :D

NickCloudAT: I think I understand what you mean and will change that.

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
NickCloudAT: Okay.. well.. I do understand the thing about HUDManager.GetHUD returning a different hud it it's a fallback.
But what menu do you have in mind? I actually don't 100% get what you mean.

NickCloudAT:
I have now added a Menu that works for HudElements.
I spoke with Tim and he also finds no better solution to this other than rewriting the saveKey system.

Maybe if you have time @saibotk we can talk about this later.

NickCloudAT:
Pushed new commit which should now be what was requested.
It now uses per HudElement saveKeys for the new section in the F1 menu.

gamemodes/terrortown/gamemode/shared/hud_elements/base_elements/hud_element_base/cl_init.lua
```diff
@@ -730,6 +730,6 @@ function HUDELEMENT:LoadData()
 	end
 
 	for k, v in pairs(loadedData) do
-		self[k] = v or self[k]
```
saibotk, line 733: I don't think this is correct, since now any defined key will just be kept and never be overridden by the loaded data

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
   - Added a convar `ttt2_voice_scaling` to control voice volume scaling, options like "power4" or "log" cause the volume scaling to have a greater perceptual impact between discrete volume settings.
   - Added convars `ttt2_voice_duck_spectator` and `ttt2_voice_duck_spectator_amount` to lower spectator voice volume automatically.
     - A value of `0.13` ducks someone's volume at 90% down to effectively 78%, according to the client's scaling mode.
+- Translated all strings still needed to german (by @NickCloudAT)
```
saibotk, line 18: remove

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
```diff
@@ -38,10 +38,31 @@ if CLIENT then
 		self.lpw = lpw
 		self.sri_text_width_padding = sri_text_width_padding
 		--self.secondaryRoleInformationFunc = nil
+		self.healthpulsate = hudelements.GetStored(self.id).healthPulsate
```
saibotk, line 41: Remove this, as it is not needed

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
```diff
@@ -65,6 +86,7 @@ if CLIENT then
 		self.lpw = lpw * self.scale
 		self.pad = pad * self.scale
 		self.sri_text_width_padding = sri_text_width_padding * self.scale
+		self.healthpulsate = hudelements.GetStored(self.id).healthPulsate
```
saibotk, line 89: remove this too

gamemodes/terrortown/gamemode/shared/huds/pure_skin/cl_init.lua
```diff
@@ -50,6 +50,17 @@ function HUD:Initialize()
...
+-- This function will return a table containing all keys that will be stored by
+-- the @{HUD:SaveData} function.
+-- @return table
+-- @realm client
+function HUD:GetSavingKeys()
+	local savingKeys = BaseClass.GetSavingKeys(self) or {}
+
+	return table.Copy(savingKeys)
+end
+
```
saibotk, line 63: this is not needed anymore

saibotk (changes requested):

saibotk (approved):

PR #1108, opened 2023-10-27
use global vectors, especially inside those loops

EntranceJew:


Histalek (approved):
Nice

TimGoll:
I think our recent coding style in TTT2 is more like `vectorOrigin`. So frist the type, then what it is. Some parts of the old code still use the old style though. I think this should be changed?

Histalek:
> I think our recent coding style in TTT2 is more like `vectorOrigin`. So frist the type, then what it is. Some parts of the old code still use the old style though. I think this should be changed?

These globals are not controlled by us ;)

TimGoll:
> > I think our recent coding style in TTT2 is more like `vectorOrigin`. So frist the type, then what it is. Some parts of the old code still use the old style though. I think this should be changed?
> 
> These globals are not controlled by us ;)

nevermind then

TimGoll (approved):

PR #1109, opened 2023-10-27
[Language] Cleanup

TimGoll:
Cleaned up language file with new auto cleanup tool: https://github.com/TTT-2/ttt2-language_parser

### DIFF
--- a/lua/terrortown/lang/de.lua
+++ b/lua/terrortown/lang/de.lua
@@ -1,7 +1,7 @@
 -- German language strings
--- 
+
 local L = LANG.CreateLanguage("de")
--- 
+
 -- Compatibility language name that might be removed soon.
 -- the alias name is based on the original TTT language name:
 -- https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/lang/german.lua
--- a/lua/terrortown/lang/de.lua
+++ b/lua/terrortown/lang/de.lua
@@ -1,7 +1,7 @@
 -- German language strings
-
+-- 
 local L = LANG.CreateLanguage("de")
-
+-- 
 -- Compatibility language name that might be removed soon.
 -- the alias name is based on the original TTT language name:
 -- https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/lang/german.lua
@@ -288,7 +288,7 @@ L.c4_remove_destroy2 = "Bestätigen: Vernichten"
 L.c4_disarm = "C4 entschärfen"
 L.c4_disarm_cut = "Klicke zum Durchschneiden von Kabel {num}"
 
-L.c4_disarm_t     = "Durchschneide ein Kabel zum Entschärfen der Bombe. Wenn du Verräter bist, ist jedes Kabel sicher. Unschuldige haben es da nicht so einfach!"
+L.c4_disarm_t = "Durchschneide ein Kabel zum Entschärfen der Bombe. Wenn du Verräter bist, ist jedes Kabel sicher. Unschuldige haben es da nicht so einfach!"
 L.c4_disarm_owned = "Durchschneide ein Kabel zum Entschärfen der Bombe. Es ist deine Bombe, also wird jedes Kabel sie sicher entschärfen."
 L.c4_disarm_other = "Durchschneide das richtige Kabel, um die Bombe zu entschärfen. Sie explodiert, wenn du das falsche triffst!"
 
@@ -475,7 +475,7 @@ L.hp_wounded = "Verwundet"
 L.hp_badwnd = "Schwer Verwundet"
 L.hp_death = "Dem Tode nah"
 
--- TargetID karma status
+-- TargetID Karma status
 L.karma_max = "Verlässlich"
 L.karma_high = "Grob"
 L.karma_med = "Schießwütig"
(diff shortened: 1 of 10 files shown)


lua/terrortown/lang/zh_hans.lua
```diff
@@ -1458,7 +1454,7 @@ L.button_delete_all_spawns = "删除所有生成点位置"
 L.label_dynamic_spawns_enable = "为该地图启用动态生成"
 L.label_dynamic_spawns_global_enable = "为所有地图启用自定义生成"
 
-L.header_equipment_weapon_spawn_setup = "武器生成设定"
+L.header_equipment_weapon_spawn_setup = "武器生成设置"
```
Histalek, line 1457: there where multiple translations for this key and your tool let the first one win, although ingame the last won probably and this would change the current behaviour.

IMO this should be changed back.

lua/terrortown/lang/zh_hans.lua
```diff
@@ -1492,8 +1488,8 @@ L.help_spawn_editor_spawn_amount = [[
 
 {playerrandom} x 玩家随机位置生成]]
 
-L.equipmenteditor_name_auto_spawnable = "设备在地图中随机产生"
-L.equipmenteditor_name_spawn_type = "选择生成类型"
+L.equipmenteditor_name_auto_spawnable = "装备随机生成"
+L.equipmenteditor_name_spawn_type = "生成类型"
```
Histalek, line 1492: same here, should be reverted IMO

lua/terrortown/lang/zh_tw.lua
```diff
@@ -34,8 +34,8 @@ L.round_traitors_one = "叛徒，你得自己頂住了。"
 L.round_traitors_more = "叛徒，你的隊友是：{names}"
 
 L.win_time = "時間用盡，叛徒失敗了。"
-L.win_traitor = "叛徒取得了勝利！"
-L.win_innocent = "叛徒們被擊敗了！"
+--L.win_traitors = "The Traitors have won!"
+--L.win_innocents = "The Innos have won!"
```
Histalek, line 38:
```suggestion
L.win_traitors = "叛徒取得了勝利！"
L.win_innocents = "叛徒們被擊敗了！"
```
missing `s` again

Histalek (changes requested):
good stuff.

The tool probably needs changing in regards to duplicates so that the last translation wins. This would prevent the tool from introducing changes.

lua/terrortown/lang/de.lua
```diff
@@ -1,7 +1,7 @@
 -- German language strings
-
+-- 
 local L = LANG.CreateLanguage("de")
-
+--
```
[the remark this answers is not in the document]
TimGoll: Because it is easier than to handle this special case

TimGoll:
> good stuff.
> 
> The tool probably needs changing in regards to duplicates so that the last translation wins. This would prevent the tool from introducing ingame changes.

This only happens if there are multiple. This should not happen once it is cleaned up. The tool checks the english file and then the other language. If there is a string available, then it uses it. Otherwise a new comment is added.

lua/terrortown/lang/zh_hans.lua
TimGoll: I can do that, yes. I could also change the tool to iterate backwards over the file

TimGoll:
I adressed all requests @Histalek

lua/terrortown/lang/de.lua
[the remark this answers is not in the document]
TimGoll: fixed it here manually and fixed it in the tool as well

Histalek (approved):

saibotk:
Can we just add such small one file tools into this codebase instead? I think that would make it much more accessible to use. And we can also exclude those from the addon within the addon.json
Maybe some folder like `/scripts/`

lua/terrortown/lang/ru.lua
```diff
@@ -1,13 +1,8 @@
 -- Russian language strings
 
--- This localization file is being moderated and constantly updated by Satton(RU).
--- Please, in case you are making some changes ping @Satton2 on GitHub, contact him on Steam (STEAM_0:0:85981967) or Discord (Satton(RU)#5794).
-
```
saibotk, line 5: I would like the tool to not change such sections, can we introduce something to prevent it from changing certain sections like these? Maybe something like `-- %keep%`

saibotk (changes requested):
Otherwise i really like this!

TimGoll:
> Can we just add such small one file tools into this codebase instead? I think that would make it much more accessible to use. And we can also exclude those from the addon within the addon.json Maybe some folder like `/scripts/`

I can put it in the code base if so desired, yes

lua/terrortown/lang/ru.lua
TimGoll: In theory it keeps the whole header (until L.__alias is reached), I guess it made a mistake here, hm

TimGoll: I think I removed this manually by accident because I thought it was twice the same

TimGoll:
done @saibotk

lua/terrortown/lang/ru.lua
```diff
@@ -1,13 +1,9 @@
...
 local L = LANG.CreateLanguage("ru")
 
--- Compatibility language name that might be removed soon.
--- the alias name is based on the original TTT language name:
--- https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/lang/russian.lua
```
saibotk, line 10: i think a comment like this is actually helpful in the main english translation, sorry to bother you again, but can we add this to every file as a comment?

TimGoll: Hmm? Isn't this in every file anyway?

saibotk: No, there is no comment.
I'd just like to keep a small comment explaining that this is the alias to the TTT language name to provide a fallback for when addons register translations for languages referencing those names.

TimGoll: ah, I removed it in the russian file but nowhere else, hence my confusion. Dunno why I did that. Probably by mistake

saibotk (approved):

PR #1111, opened 2023-10-27
[Language] forked from TTT and cleaned up

TimGoll:
Pulled ukrainian and swedish language from base TTT and let our new tool clean up the files:

https://github.com/Facepunch/garrysmod/tree/master/garrysmod/gamemodes/terrortown/gamemode/lang


saibotk:
Looks good, but can you quickly confirm that the language system can handle missing translation params? 
At least in a non-erroring way? :D Because i don't know if all of the params stayed the same over the years.

Also something we could check with the language tool actually!

TimGoll:
> Looks good, but can you quickly confirm that the language system can handle missing translation params? At least in a non-erroring way? :D Because i don't know if all of the params stayed the same over the years.
> 
> Also something we could check with the language tool actually!

yes, english is the fallback language. As you can see most of our languages are not 100%. So this works

The tool only added the strings that have identifiers in the up-to-date english language file

saibotk:
No I am talking about parameterized translations. With placeholders in them.

TimGoll:
> No I am talking about parameterized translations. With placeholders in them.

ah, that's an issue, hm. I don't know how I should check this. Welp

saibotk:
Well the parameters can be extracted using a pattern, and then we need to compare them with the base translation entry.
I think this would be a great addition, and we might either want to add this to the language tool now and check these or at least test, that the system does not error when some params are missing or renamed and in the worst case display a wrong translation.

saibotk (approved):

Histalek (approved):

PR #1112, opened 2023-10-27
fix dynamic armor leftover damage calculation

ruby0b:


TimGoll:
could you explain what you did here? and add a changelog?

ruby0b:
Sorry for not providing a description, I used the new GitHub VSCode thing and it immediately created a pull request after committing.

This PR addresses an issue I've noticed when a player takes damage that is too strong for their armor to fully absorb.
The current calculation is wrong and can even lead to situations where you take *more* damage if you have just a little bit of armor left.

## Example
Let's choose unconventional armor and health factors to demonstrate the issue:
```
damage = 100
armor = 10
armor_factor = 0.8
health_factor = 0.5
```
We take 80 armor damage but only have 10 armor left:
```
armorDamage = armor_factor * damage
            = 80
```
### Old calculation
The old calculation would now end up with 120 damage, which is *higher* than the base damage:
```
SetDamage: health_factor * damage - min(0, armor - armorDamage)
         = 0.5*100 - (-70)
         = 120
```
### New calculation
First we calculate the leftover damage that our armor couldn't take.
It's important that we divide by the  `armor_factor` afterwards because it should only apply to the damage taken by our armor.
```
leftoverDamage = max(0, armorDamage - armor) / armor_factor
               = (80 - 10) / 0.8
               = 87.5
```
Now we calculate the damage. We only apply the `health_factor` to the part of the damage that isn't leftover damage. The leftover damage gets added without any modifier because it didn't get weakened by any armor.
```
SetDamage: health_factor*(damage-leftoverDamage) + leftoverDamage
         = 0.5*(100-87.5) + 87.5
         = 93.75
```

sbzlzh:
You need to write the changes in CHANGELOG.md Oh!

ruby0b:
Ah, I should mention that I've tested this change by overriding `ARMOR:HandlePlayerTakeDamage` in a `TTT2Initialize` hook and it works as expected

TimGoll:
This is a great catch, looks good to me!

TimGoll (approved):

gamemodes/terrortown/gamemode/server/sv_armor.lua
```diff
@@ -141,10 +141,11 @@ function ARMOR:HandlePlayerTakeDamage(ply, infl, att, amount, dmginfo)
 
 	ply:DecreaseArmorValue(armorDamage)
 
-	-- The armor offset is used to catch the damage that should be taken,
+	-- The leftover damage is used to catch the damage that should be taken,
 	-- if the armor is not strong enough to take that many hitpoints.
 	-- It is zero as long as the armor is able to take the damage.
-	dmginfo:SetDamage(self.cv.health_factor * damage - math.min(armor - armorDamage, 0))
+	local leftoverDamage = math.max(0, armorDamage - armor) / self.cv.armor_factor
```
saibotk, line 147: I think this is still incorrect because this uses the armor damage as a calculation instead of calculating the total damage - the blocked damage.

I am talking about the difference between armor damage, which is used to calculate how much armor you lose and the actual damage you get reduced by the health modifier.

I think we would rather need to calculate the blocked amount based on the armor modifier and then subtract that from the total damage.

saibotk (changes requested):

gamemodes/terrortown/gamemode/server/sv_armor.lua
ruby0b: Let's say you take 150 damage, have 50 armor, 100 hp, and health_factor and armor_factor are both 0.5.
As far as I understand, this should work the same way as taking 100 damage and then another 50 damage.

The 100 damage would fully deplete your armor to 0 and you would take 50 damage.
The next 50 damage would then be directly subtracted from your health because you have no armor left.
So the code would have to arrive at a total of 100 damage in this case, which mine does:
```
armorDamage = 150 * 0.5 = 75
leftoverDamage = (armorDamage - armor) / armor_factor = (75 - 50) / 0.5 = 50
healthDamage = health_factor*(damage-leftoverDamage) + leftoverDamage
             = 0.5 * (150 - 50) + 50 = 100
```

Would you agree that this is how it should generally work? I wasn't quite able to understand your issue.

saibotk: The correct solution would be:
```lua
local leftoverDamage = damage * (armorDamage - armor) / armorDamage
```
which can be transformed to:
```lua
local leftoverDamage = damage - armor / armor_factor
```
which then needs to be capped to 0 with `math.max(0, leftoverDamage)`.


But we can do it more easily by calculating it the other way around:
So let's calculate the amount of damage that will be reduced based on our current armor value:
```lua
local damageReduced = damage * (armor / armorDamage) -- <- This is basically the percentage of the armor damage that we can supply.
```
Then this can be simplified to just:
```lua
local damageReduced = armor / armor_factor
```
which equals the total amount of damage we can reduce. Note: This might exceed the damage value wastly, so we need to also cap this to the max of the damage below.

The final damage calculation would then be:
```lua
local newDamageValue = math.min(damage, damageReduced) * health_factor + math.max(0, damage - damageReduced)
```

saibotk: Yes your assumptions are correct, yes.

I just did try to solve this too and came up with the (to me understandable) solution above.

Now i will just need to check if yours is equivalent or if one of us did something wrong

saibotk: nvm. in my first part i made a stupid mistake when simplifying 🤦 

Seems like the solution is equivalent.

ruby0b: I ran them through some examples in python and in these few cases they were the same:
```py
def mine(damage, armor, armor_factor, health_factor):
    armorDamage = damage * armor_factor
    print(f"{armorDamage=}")
    leftoverDamage = max(0, (armorDamage - armor) / armor_factor)
    print(f"{leftoverDamage=}")
    healthDamage = health_factor * (damage - leftoverDamage) + leftoverDamage
    print(f"{healthDamage=}")
    return healthDamage


def yours(damage, armor, armor_factor, health_factor):
    damageReduced = armor / armor_factor
    print(f"{damageReduced=}")
    healthDamage = min(damage, damageReduced) * health_factor + max(
        0, damage - damageReduced
    )
    print(f"{healthDamage=}")
    return healthDamage


test_cases = [
    (150, 50, 0.5, 0.5),
    (100, 50, 1, 1),
    (125, 50, 0.2, 0.7),
    (50, 100, .5, .5),
    (23, 2, .7, .1),
]
for damage, armor, armor_factor, health_factor in test_cases:
    print(f"Damage: {damage}, Armor: {armor}, Armor Factor: {armor_factor}, Health Factor: {health_factor}")
    print(">>> Mine:")
    a = mine(damage, armor, armor_factor, health_factor)
    print(">>> Yours:")
    b = yours(damage, armor, armor_factor, health_factor)
    print("CORRECT" if a == b else "\n\n\n\n DIFFERENT \n\n\n")
    print()

```

saibotk: Okay let me correct my self again xD
I did not make any mistake in my first section.
I just checked it and yours is just one step before my simplification.

```lua
local yours = (damage * armor_factor - armor) / armor_factor
-- can be simplified to 
local mine = damage - armor / armor_factor
```

ruby0b: Ah yeah, seems like we ended up with basically the same solution :)
Do you think I should edit something for clarity or just leave it as is?

saibotk: But we should definitely add some comments to explain the meaning of the temporary result.

For my way, it would be along the lines of:
```lua
-- Describes the maximum amount of damage, that our current armor can endure. This might exceed the damage, so we need to limit this.
```

saibotk (approved):
Thank you very much for catching this!

And also thank you very much for your nice and elaborate explanations!
Appreciate such effort!

PR #1114, opened 2023-10-27
Corpse: Added hook for preventing transfer of found credits

Spanospy:
Intended for addons that wish to prevent picking up credits from bodies during a round, such as custom roles that are designed to not pickup credits from bodies until a criteria is met.

This has NOT been tested yet, but it's copying the structure of TTT2CheckCreditAward and GiveFoundCredits doesn't return anything, so I don't think much can go wrong.


saibotk:
Codewise this looks sensible!

But I'm unsure about the behavior and response for players once you disable this. 
Players will still see the hint to pickup credits and can try to do so and it will look as if it works clientside, but they will not actually receive the credits. Sadly this is not implemented in a way where the server actively responds to a client and can answer with some kind of message or confirmation.

Maybe @TimGoll has some ideas for solving this given his recent work on the frontend for this.

Spanospy:
> But I'm unsure about the behavior and response for players once you disable this. Players will still see the hint to pickup credits and can try to do so and it will look as if it works clientside, but they will not actually receive the credits. Sadly this is not implemented in a way where the server actively responds to a client and can answer with some kind of message or confirmation.

Yeah, I've noticed this behaviour when I had a role that took the appearance of a shopper role clientside, but was actually non-shopper serverside. Perhaps there should be an 'override' property on the ragdoll/corpse that the client can read.

TimGoll:
I'd say merge it and add a note to the bodysearch rework so I can make this hook shared. But then addons have to use this hook shared as well. What's your opinion on this @Spanospy @saibotk?

saibotk:
Sounds good to me

TimGoll:
Okay, I'd rename the hook because I had to check the code to see what it does. What about `GM:GiveFoundCredits`?

Spanospy:
More sensible name I suppose! I would suggest `TTT2ShouldGiveFoundCredits` but perhaps that's too long.

TimGoll:
> More sensible name I suppose! I would suggest `TTT2ShouldGiveFoundCredits` but perhaps that's too long.

I'd prefer something like `OnXYZ` for events like this, but there is no consistency in Garrys Mod, TTT or TTT2. So whatever

Spanospy:
Just checked the body search rework you're doing @TimGoll , I see you've [updated](https://github.com/TTT-2/TTT2/pull/1067/files#diff-b3cb386207438f4819b734529e8233c5dacd3ce6a3f11b34d724ee9f62a6aaeeR540) the targetid code responsible for adding the credits hint on bodies to also check if the viewer's subrole isn't prevented from finding credits. Would it be sensible to have a hook there so addons can override whether the hint should be displayed, or should addons just instead modify the SubRoleData on the client?

TimGoll:
> Just checked the body search rework you're doing @TimGoll , I see you've [updated](https://github.com/TTT-2/TTT2/pull/1067/files#diff-b3cb386207438f4819b734529e8233c5dacd3ce6a3f11b34d724ee9f62a6aaeeR540) the targetid code responsible for adding the credits hint on bodies to also check if the viewer's subrole isn't prevented from finding credits. Would it be sensible to have a hook there so addons can override whether the hint should be displayed, or should addons just instead modify the SubRoleData on the client?

We can talk about this once it is merged, otherwise this will get out of hand. But yes this could be useful. But in general I try to limit hooks that are executed on every frame

Spanospy:
Good call.

TimGoll (approved):

saibotk (approved):

TimGoll:
I edited your PR and I merge it into the body search rework so I can move it into the new system

PR #1117, opened 2023-10-27
backs out Vector/Angle changes with safe copying

EntranceJew:
still better than using a zeroed constructor

### DIFF
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -163,9 +163,9 @@ SWEP.fingerprints = {}
 
 --[[
 	-- The position offset applied when entering the ironsight
-	SWEP.IronSightsPos = vector_origin
+	SWEP.IronSightsPos = Vector(vector_origin)
 	-- The rotational offset applied when entering the ironsight
-	SWEP.IronSightsAng = vector_origin
+	SWEP.IronSightsAng = Vector(vector_origin)
 --]]
 
 local skipWeapons = {}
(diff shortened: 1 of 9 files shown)


TimGoll:
Does this fix the issue? Did you test it? If so, it looks good to me

saibotk:
Is there really any benefit in this now that we copy those again?

Histalek:
> Is there really any benefit in this now that we copy those again?

> If this is a Vector, this function will return a copy of the given vector.
- https://wiki.facepunch.com/gmod/Global.Vector

i suppose this might still be better? i'm not sure tho

saibotk:
Honestly, i don't think we even use any of those in a performance relevant region, so this would only be a micro micro optimization when initializing the entities / vectors.

EntranceJew:
Yeah, the docs for Angle and Vector both indicate that a copy is being done, seemingly with less work.
In the very least, by deliberately copying the globals we prevent trapping a reference to a global that someone might think is available to modify.

saibotk:
I don't think this is actually really less, but i would still vote against using this.

1. It does not provide real value at any of those places, because none are called every frame.
2. Makes it harder to read
3. Relies on global variables being defined in gmod
4. Can still be messed with, eg. when overriding or mutating one of these globals (addons may screw this up in the same way we forgot about this here)

EntranceJew:
Vectors were showing up as a hot point in the code, but those changes are better suited for its own PR. I'm fine with just closing this and doing a revert on the last commit; I'm not sure how it didn't become a bigger problem in my testing environment, but I think a recent change that made some silent errors not-so-silent masked it.

There are material perf optimizations I found with profiling but some of them are contingent on more upstream changes being released, like player/ent iterator.

saibotk:
Yeah, that sounds reasonable, can you open up the PR with the revert?

And thank you for looking into this! As always happy to have you as a contributor!

Issue #1093, opened 2023-10-24
[Prop Possession] Inputs reversed


TimGoll:
Depending on the view angle, the left and right move keys are sometimes inversed, front and back work as expected for any viewing angle

PR #1118, opened 2023-10-27
Prop Possesion: Made input directions more intuitive

TimGoll:
Reworked the propspec inputs to be more predictable. The old implementation was really janky and did not take into account in which way the player is looking, see #1093 

I also removed the angular force, because I'm pretty sure it was only added to "fix" the broken implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed blurred box rendering for boxes not starting at `0,0` (by @TimGoll)
 - Optimized allocations by using global Vector / Angle when possible.
 - Fixed the dynamic armor damage calculation being wrong when damage can only get partially reduced
-- Fixed propspec inputs being sometimes unpredictable (by @TimGoll)
+- Fixed propspec inputs behaving sometimes unexpectedly (by @TimGoll)
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed blurred box rendering for boxes not starting at `0,0` (by @TimGoll)
 - Optimized allocations by using global Vector / Angle when possible.
 - Fixed the dynamic armor damage calculation being wrong when damage can only get partially reduced
+- Fixed propspec inputs being sometimes unpredictable (by @TimGoll)
```
saibotk, line 50: Can we use another terminology, as prediction here is usually referred to networking.

saibotk (changes requested):

saibotk (approved):
Very nice!

gamemodes/terrortown/gamemode/server/sv_propspec.lua
```diff
@@ -147,42 +147,44 @@ function PROPSPEC.Key(ply, key)
...
 
-		pr.t = CurTime() + 0.05
+		pSpec.t = CurTime() + 0.05
 	elseif key == IN_FORWARD then
-		-- bump away from player
-		phys:ApplyForceCenter(aim * mf)
+		phys:ApplyForceCenter(vectorAimPlanar * mf)
 	elseif key == IN_BACK then
-		phys:ApplyForceCenter(aim * (mf * -1))
+		phys:ApplyForceCenter(vectorAimPlanar * (mf * -1))
```
Alf21, line 176:
```suggestion
		phys:ApplyForceCenter(-vectorAimPlanar * mf)
```

Alf21 (approved):
Nice!

PR #1120, opened 2023-10-28
[Lang] Cleaned up Language with Param Translation

TimGoll:
Over the time some language strings changed. But this change was not applied to all languages. Since our new language tool is able to spot these, it was easy to fix them.

I also noticed a few errors that I could fix without removing the strings (missing brackets, wrong var names etc)


saibotk (approved):
Very nice!

Histalek (approved):

PR #1121, opened 2023-10-28
[Feature] voice and volume menu

EntranceJew:
image
it looks like this


TimGoll (approved):
please add a changelog. Otherwise it looks nice

saibotk (approved):

saibotk:
@TimGoll i don't really think we need a changelog for this

TimGoll:
not part of this PR, but I think we should explain the last three convars. Because I have no clue what they do

PR #1129, opened 2023-11-06
reimplemented @wgetJane's dropclip/reserve

EntranceJew:


gamemodes/terrortown/gamemode/server/sv_weaponry.lua
```diff
@@ -455,67 +455,16 @@ local function DropActiveWeapon(ply)
...
-		box:SetOwner(nil)
-	end)
-end
-concommand.Add("ttt_dropammo", DropActiveAmmo)
+	ply:SafeDropAmmo(wep, false)
+end)
+concommand.Add("ttt_dropclip", function(ply)
+	if not IsValid(ply) then return end
+	local wep = ply:GetActiveWeapon()
+	ply:SafeDropAmmo(wep, true)
```
TimGoll, line 466:
```suggestion
	if not IsValid(ply) then return end

	ply:SafeDropAmmo(ply:GetActiveWeapon(), true)
```

TimGoll (approved):
This doesn't have to be a part of your PR, but the "drop ammo" and "drop weapon" bind should both be shown in the "extra" category imho

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1311,6 +1311,133 @@ function plymeta:SafeDropWeapon(wep, keepSelection)
...
+
+		return false
+	end
+	amt = hook_data[1]
+	--TODO: revisit here with convar for ammo quarter thing
+	if amt < 1 or amt <= wep.Primary.ClipSize * 0.25 then
+		LANG.Msg(self, useClip and "drop_no_ammo" or "drop_no_reserve", nil, MSG_MSTACK_WARN)
+
+		return false
+	end
```
Histalek, line 1386:
```suggestion
	if amt < 1 or amt <= wep.Primary.ClipSize * 0.25 then
		LANG.Msg(self, useClip and "drop_no_ammo" or "drop_no_reserve", nil, MSG_MSTACK_WARN)

		return false
	end
```
This should probably be an issue instead of a TODO comment. That would allow us to discuss if this restriction should be configurable at all or maybe even removed outright.

Histalek (approved):
LGTM

EntranceJew:
> This doesn't have to be a part of your PR, but the "drop ammo" and "drop weapon" bind should both be shown in the "extra" category imho

I tried to start on this after the suggestion, but since it's one of those binds that is tangled up in a default bind, and I don't think it's elegant for the next _last time this code would ever reasonably be modified_ to change it to use `RunConsoleCommand` calls (seems kind of like something we ought to be eliminating) -- it means writing a netmsg to reach the SV code, and that's a bit complicated and more nuanced

I do definitely agree, though; a PR that blanket factors out the default binds and makes all the controls rebindable via the bind menu would really be something.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1311,6 +1311,132 @@ function plymeta:SafeDropWeapon(wep, keepSelection)
...
+		wep:SetClip1( math.max(wep:Clip1() - amt, 0) )
+	else
+		self:RemoveAmmo(amt, wep.Primary.Ammo)
+	end
+
+	return true
+end
+
+---
+-- Called to drop a weapon in a safe manner (e.g. preparing and space-check).
```
saibotk, line 1401: This comment is wrong.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1311,6 +1311,132 @@ function plymeta:SafeDropWeapon(wep, keepSelection)
...
+-- @param Weapon wep The weapon that should be dropped
+-- @param boolean keepSelection If set to true the current selection is kept if not dropped
+-- @return boolean Returns if this weapon is dropped
```
saibotk, line 1404: These need to be adjusted

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1311,6 +1311,132 @@ function plymeta:SafeDropWeapon(wep, keepSelection)
...
+-- @realm server
+function plymeta:DropAmmo(wep, amt)
```
saibotk, line 1406: Why do we want this function to be exposed to developers? 
Shouldn't we make sure that they use the safe drop all the time?

Im bringing this up, because this function is missing some stuff that the safedrop function currently has eg. The animation etc. if we would just differentiate between safe and non safe drops, i would expect the normal drop function to take care of most of the logic except the safe check call.

saibotk (changes requested):

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: Either I'm missing something, or you're talking about `plymeta:SafeDrioWeapon`, but this is about ammo

EntranceJew: I wrote it this way because it is meant to be a parallel to [Player:DropWeapon](https://wiki.facepunch.com/gmod/Player:DropWeapon), as a majority of this code exists to mirror the architecture of similar functions for dropping weapons. The reason this one stands out, is that it has no primitive equivalent.

saibotk: I don't see a reason to keep this extra function and would rather merge both together. Because we would not want a user to actually use the DropAmmo function, do we?

TimGoll: I think the whole point of this PR is to add an option to drop ammo in a reasonable way. It was possible before, but it took the ammo out of the weapon instead of your storage afaik

saibotk: I understand the PR and the reason im only questioning the existence of two functions where the latter is rather incomplete and does not make sense to call on its own

EntranceJew: My only personal justification is that sometimes addons have called DropWeapon, when SafeDropWeapon would have stopped them, and it was beneficial because the collision check is too tight in some circumstances (especially if you're crouching, looking at a weapon on the floor) -- or in situations where the grenade disarms should always be able to force drop a weapon rather than letting people keep it if they keep their face snug to a corner. Some sort of effect which forcibly spills a player's ammo regardless of the room to do it wouldn't be unreasonable to imagine.

saibotk: Thats why I asked. If you only want to do that, then move the calculation and all other logic from the safe function to the drop function and only do the safety check in the safedrop function. In this way their responsibilities are much more clear and are closer to what someone would think they do.

saibotk (changes requested):

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
EntranceJew: which line ranges in particular do you think should be moved? i'll need to change some signatures potentially

saibotk: Everything except for the candrop check

EntranceJew: I think that did it? The only thing now is that ply:DropAmmo could still be interrupted, and show MSTACK messages even if it doesn't originate as a user-driven intention.

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
```diff
@@ -666,6 +615,19 @@ function WEPS.DropNotifiedWeapon(ply, wep, deathDrop, keepSelection)
...
+-- @param Player ply The player whose ammo is about to be dropped
+-- @param Weapon wep The weapon to reference when dropping ammo
+-- @param boolean useClip If set to true the ammo is dropped from the clip, otherwise, from reserve ammo.
+-- @param number amt The amount of ammo to drop
+-- @realm server
+function WEPS.DropNotifiedAmmo(ply, wep, useClip, amt)
+	if not IsValid(ply) or not IsValid(wep) then return end
+
+	ply:DropAmmo(wep, useClip, amt)
+end
```
saibotk, line 629: This function does not contribute anything, does it? I think we should remove it

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1311,6 +1311,135 @@ function plymeta:SafeDropWeapon(wep, keepSelection)
...
+-- @param Weapon wep The weapon that should be referenced when dropping ammo.
+-- @param boolean useClip If set to true the ammo is dropped from the clip, otherwise, from reserve ammo.
+-- @return boolean Returns if ammo is dropped
+-- @realm server
+function plymeta:SafeDropAmmo(wep, useClip)
+	if not self:CanSafeDropAmmo(wep) then
+		return false
+	end
+
+	self:AnimPerformGesture(ACT_GMOD_GESTURE_ITEM_GIVE)
```
saibotk, line 1361: I think you might still want to play the animation in the weapon drop function. What did you think @TimGoll ?

saibotk (changes requested):

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: Uhm, weapon drop function? Are you talking about `WEPS.DropNotifiedAmmo`? Honestly, I don't care. But @EntranceJew did it the same way we(?) did it for weapon drops:

https://github.com/TTT-2/TTT2/blob/660d5151a1da7dd2272a154f998cb3080e1db0c2/gamemodes/terrortown/gamemode/server/sv_player_ext.lua#L1302-L1312

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
TimGoll: I get why @EntranceJew added it. He wanted to reflect the way dropping weapons work. But I agree, it is confusing and adds no benefit. Especially since the current call-stack looks like this: `ply:SafeDropWeapon` -> `WEPS.DropNotifiedAmmo` -> `ply:DropWeapon`

Which is honestly kinda weird. But I guess it is done in the same way for weapon drops.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1311,6 +1311,135 @@ function plymeta:SafeDropWeapon(wep, keepSelection)
...
+-- @param number amt The quantity of ammo to drop.
+-- @return boolean Returns if this weapon's ammo is dropped.
+-- @realm server
+function plymeta:DropAmmo(wep, useClip, amt)
+	local box = ents.Create(wep.AmmoEnt)
+
+	if not IsValid(box) then return end
+
+	-- most of this goes into computing ammo amount
+	amt = amt or 0
```
TimGoll, line 1381:
```suggestion
	amt = amt or 0

	local box = ents.Create(wep.AmmoEnt)

	if not IsValid(box) then return end

	-- most of this goes into computing ammo amount
```

validation of parameters should be on the top of the function, as these represent the default values and should therefore be visible at first glance

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1311,6 +1311,135 @@ function plymeta:SafeDropWeapon(wep, keepSelection)
...
+	WEPS.DropNotifiedAmmo(self, wep, useClip)
+
+	return true
+end
+
+---
+-- Called to drop a weapon's ammo. Does no safety checks.
+-- @param Weapon wep The weapon's ammo that should be dropped
+-- @param boolean useClip If set to true the ammo is dropped from the clip, otherwise, from reserve ammo.
+-- @param number amt The quantity of ammo to drop.
```
TimGoll, line 1372:
```suggestion
-- @param[default=0] number amt The quantity of ammo to drop.
```

If you add validation, add the default value to the docs as it now is an optional parameter

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1311,6 +1311,135 @@ function plymeta:SafeDropWeapon(wep, keepSelection)
...
+
+	WEPS.DropNotifiedAmmo(self, wep, useClip)
+
+	return true
+end
+
+---
+-- Called to drop a weapon's ammo. Does no safety checks.
+-- @param Weapon wep The weapon's ammo that should be dropped
+-- @param boolean useClip If set to true the ammo is dropped from the clip, otherwise, from reserve ammo.
```
TimGoll, line 1371:
```suggestion
-- @param[default=false] boolean useClip If set to true the ammo is dropped from the clip, otherwise, from reserve ammo.
```

Since your code accepts both values, this should be marked as an optional parameter with a default. Even though the default is `nil` it is handled as if it is a `false`

TimGoll (changes requested):
In general it looks good to me. I have on question that is open to discussion though. Should we add bindings for ammo drop? or should it be limited to a command?

I guess this boils down to the question: Is this a debug feature that you'd like to use or is this something you expect people to use when playing TTT2?

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
saibotk: Nah i was talking about the ammo drop, where the safe method should only do a safety check and the base drop function should do everything else.

gamemodes/terrortown/gamemode/server/sv_weaponry.lua
saibotk: Yes i feel the same way but there is no reason to do it this way just because the weapon drop also calls a function here

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: Yes, but the implementation here is fine imho, because the animation is done in the safeDrop for the weapon as well

EntranceJew:
There should be bindings, they just happen to overlap with existing buttons and would require a net message to be more than concommands so I didn't do that part.

EntranceJew:
What's still in the way? Is it just factoring out `WEPS.DropNotifiedAmmo`?

saibotk:
Yes from my side it would be just removing the function and we are good to go

saibotk:
@TimGoll you still need to approve

TimGoll (approved):

PR #1134, opened 2023-11-15
Net: Enhance for 256KB tables

ZenBre4ker:
This enhancement splits longer messages into separate smaller ones and awaits the clients approval before sending more of them.

It solves the problem of a limited buffer, when 256KB are reached.
As it's just an edge-case it shouldnt be too important, but you never know.

Especially helpful if #984 gets merged.

### DIFF
--- a/lua/ttt2/extensions/net.lua
+++ b/lua/ttt2/extensions/net.lua
@@ -27,7 +27,7 @@ net.STREAM_FRAGMENTATION_SIZE = 65400
 
 -- Stream cache variables
 net.send_stream_cache  = {}
-net.eligible_players  = {}
+net.receiving_players  = {}
 net.receive_stream_cache = {}
 net.stream_callbacks = {}
 
@@ -64,7 +64,7 @@ local function SendNextStream(messageId, streamId, split, plys)
 	end
 
 	if split <= 1 then
-		local eligiblePlayerList = net.eligible_players[messageId][streamId]
+		local eligiblePlayerList = net.receiving_players[messageId][streamId]
 
 		if eligiblePlayerList then
 			plys = istable(plys) and plys or {plys}
@@ -75,7 +75,7 @@ local function SendNextStream(messageId, streamId, split, plys)
 
 			if next(eligiblePlayerList) ~= nil then return end
 
-			net.eligible_players[messageId][streamId] = nil
+			net.receiving_players[messageId][streamId] = nil
 		end
 
 		net.send_stream_cache[messageId][streamId] = nil
@@ -93,7 +93,7 @@ local function SendNextSplit(len, ply)
 	local streamId = net.ReadUInt(32)
 	local nextSplit = net.ReadUInt(8)
 
-	local eligiblePlayerList = net.eligible_players[messageId][streamId]
+	local eligiblePlayerList = net.receiving_players[messageId][streamId]
 
 	if nextSplit < 1 or eligiblePlayerList and not eligiblePlayerList[ply:SteamID64()] then return end
 
@@ -115,18 +115,18 @@ function net.SendStream(messageId, data, plys)
 	local splits = string.SplitAtSize(encodedString, net.STREAM_FRAGMENTATION_SIZE)
 
 	net.send_stream_cache[messageId] = net.send_stream_cache[messageId] or {}
-	net.eligible_players[messageId] = net.eligible_players[messageId] or {}
+	net.receiving_players[messageId] = net.receiving_players[messageId] or {}
 
 	local streamId = #net.send_stream_cache[messageId] + 1
 
 	net.send_stream_cache[messageId][streamId] = splits
 
 	if SERVER and plys and #splits > 1 then
-		net.eligible_players[messageId][streamId] = {}
+		net.receiving_players[messageId][streamId] = {}
 		plys = istable(plys) and plys or {plys}
 
 		for i = 1, #plys do
-			net.eligible_players[messageId][streamId][plys[i]:SteamID64()] = true
+			net.receiving_players[messageId][streamId][plys[i]:SteamID64()] = true
 		end
 	end
 
--- a/lua/ttt2/extensions/net.lua
+++ b/lua/ttt2/extensions/net.lua
@@ -17,7 +17,7 @@ if SERVER then
 
 	-- Add the network string for streaming data
 	util.AddNetworkString(NETMSG_STREAM)
-	-- Add the network string for confirming single splits
+	-- Add the network string for requesting single splits
 	util.AddNetworkString(NETMSG_REQUEST_NEXT_SPLIT)
 end
 
@@ -99,7 +99,7 @@ net.Receive(NETMSG_REQUEST_NEXT_SPLIT, SendNextSplit)
 -- smaller fragments. The data will be converted (with sPON) to an encoded string during this process.
 --
 -- @param string messageId A unique message id similar to the network strings
--- @param table data The data table to send, this will be reconstructed at the player.
+-- @param table data The data table to send, this will be reconstructed at the destination.
 -- @param[opt] table|player plys SERVERSIDE only! Optional, use it to send a stream to a single player or a group of players otherwise it's broadcasted.
 -- @realm shared
 function net.SendStream(messageId, data, plys)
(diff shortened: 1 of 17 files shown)


lua/ttt2/extensions/net.lua
```diff
@@ -6,58 +6,157 @@
...
+-- @param Player ply player that the message has received, `nil` on the client
+-- @realm shared
+-- @internal
+local function SendNextSplit(len, ply)
+	local messageId = net.ReadString()
+	local streamId = net.ReadUInt(32)
+	local nextSplit = net.ReadUInt(8)
+
+	local eligiblePlayerList = eligible_Players[messageId][streamId]
+	if nextSplit < 1 or eligiblePlayerList and not eligiblePlayerList[ply:SteamID64()] then return end
```
TimGoll, line 98:
```suggestion
	local eligiblePlayerList = eligible_Players[messageId][streamId]

	if nextSplit < 1 or eligiblePlayerList and not eligiblePlayerList[ply:SteamID64()] then return end
```

lua/ttt2/extensions/net.lua
```diff
@@ -6,58 +6,157 @@
...
 ---
 -- Initiates a stream message, usually for data that can be longer than
 -- the 64kb limit of a single net message. This will split up the data and send them in
 -- smaller fragments. The data will be converted (with sPON) to an encoded string during this process.
 --
 -- @param string messageId A unique message id similar to the network strings
--- @param table data The data table to send, this will be reconstructed at the client.
--- @param[opt] table|player client SERVERSIDE only! Optional, use it to send a stream to a single client or a group of clients.
+-- @param table data The data table to send, this will be reconstructed at the player.
+-- @param[opt] table|player plys SERVERSIDE only! Optional, use it to send a stream to a single player or a group of players otherwise it's broadcasted.
```
TimGoll, line 111: Is there a reason for you to change that? I think `client` made it pretty clear, that it is serverside only

lua/ttt2/extensions/net.lua
```diff
@@ -6,58 +6,157 @@
...
+	util.AddNetworkString(NETMSG_REQUEST_NEXT_SPLIT)
 end
 
 -- Size to split the network stream at (currently a bit lower than the max value, just to have some buffer)
+-- Can be up to 65.533KB see: https://wiki.facepunch.com/gmod/net.Start
 net.STREAM_FRAGMENTATION_SIZE = 65400
 
 -- Stream cache variables
+local wait_stream_cache = {}
+local eligible_Players  = {}
```
TimGoll, line 30: is there a reason they are local and not part of net like the two below?

TimGoll (approved):
Looks good to me. However this is something @saibotk has to comment on as I'm not entirely sure how his system works.

lua/ttt2/extensions/net.lua
ZenBre4ker: Well the warning was there before, because you could always think as a client you can send it to another client. But I believe my main intention was to make it uniform with other syncing libraries, where we use plys.

lua/ttt2/extensions/net.lua
ZenBre4ker: Only Idea that comes to mind is, that those are handled in another way on a reload. 
Maybe I did that, because saving them in net is unnecessary.

Not sure whats better or if there was a reason.

saibotk: First, lowercase player in the variable name.

Second, id say it would be better to also keep those one the table just because the others are too.

lua/ttt2/extensions/net.lua
```diff
@@ -68,11 +165,11 @@ end
 -- the data is reconstructed from all fragments.
 --
 -- @param string messageId a unique message id similar to the network strings
--- @param function callback This is the function that is called after the data was received.
+-- @param function callback(receivedTable, ply) This is the function that is called with the received table.
 -- @realm shared
 function net.ReceiveStream(messageId, callback)
 	-- has to be saved as string, otherwise the key lookups will fail on the table
-	local msg = tostring(util.CRC(messageId))
+	local msg = messageId
```
saibotk, line 172:
```suggestion
	local msg = tostring(messageId)
```

lua/ttt2/extensions/net.lua
```diff
@@ -6,58 +6,155 @@
...
+-- @realm shared
+-- @internal
+local function SendNextSplit(len, ply)
+	local messageId = net.ReadString()
+	local streamId = net.ReadUInt(32)
+	local nextSplit = net.ReadUInt(8)
+
+	local eligiblePlayerList = net.receiving_players[messageId][streamId]
+
+	if nextSplit < 1 or eligiblePlayerList and not eligiblePlayerList[ply:SteamID64()] then return end
```
saibotk, line 96:
```suggestion
	local receivingPlayerList = net.receiving_players[messageId][streamId]

	if nextSplit < 1 or receivingPlayerList and not receivingPlayerList[ply:SteamID64()] then return end
```

lua/ttt2/extensions/net.lua
```diff
@@ -6,58 +6,155 @@
...
+			net.Send(plys)
+		else
+			net.Broadcast()
+		end
+	else
+		net.SendToServer()
+	end
+
+	if split <= 1 then
+		local eligiblePlayerList = net.receiving_players[messageId][streamId]
```
saibotk, line 65: same here

lua/ttt2/extensions/net.lua
```diff
@@ -6,58 +6,155 @@
...
 -- Stream network message name constant
 local NETMSG_STREAM = "TTT2_NET_STREAM"
+local NETMSG_REQUEST_NEXT_SPLIT = "TTT2_NET_REQUEST_NEXT_SPLIT"
 
 if SERVER then
 	AddCSLuaFile()
 
 	-- Add the network string for streaming data
 	util.AddNetworkString(NETMSG_STREAM)
+	-- Add the network string for confirming single splits
```
saibotk, line 20:
```suggestion
	-- Add the network string for requesting single splits
```

lua/ttt2/extensions/net.lua
```diff
@@ -6,58 +6,155 @@
...
+
 ---
 -- Initiates a stream message, usually for data that can be longer than
 -- the 64kb limit of a single net message. This will split up the data and send them in
 -- smaller fragments. The data will be converted (with sPON) to an encoded string during this process.
 --
 -- @param string messageId A unique message id similar to the network strings
--- @param table data The data table to send, this will be reconstructed at the client.
--- @param[opt] table|player client SERVERSIDE only! Optional, use it to send a stream to a single client or a group of clients.
+-- @param table data The data table to send, this will be reconstructed at the player.
```
saibotk, line 108:
```suggestion
-- @param table data The data table to send, this will be reconstructed at the destination.
```

This could also be the server

saibotk (changes requested):

saibotk (approved):

lua/ttt2/extensions/net.lua
ZenBre4ker: We changed that to the net-table

TimGoll (approved):
I don't know why I have to approve again, but here we go!

ZenBre4ker:
Actually you didnt have to, just had to close your comment ^^ We cant merge with open comments apparently.

TimGoll:
> Actually you didnt have to, just had to close your comment ^^ We cant merge with open comments apparently.

No, there was also a huge banner on top that said I should add my review even though I already did. And it was there after I closed the convos

ZenBre4ker:
Oh weird. Okay

Histalek:
> > Actually you didnt have to, just had to close your comment ^^ We cant merge with open comments apparently.
> 
> No, there was also a huge banner on top that said I should add my review even though I already did. And it was there after I closed the convos

That is because saibotk re-requested your review, probably with the intention that you resolve your started conversations. GitHub's UX definitely isn't the best

TimGoll:
> > > Actually you didnt have to, just had to close your comment ^^ We cant merge with open comments apparently.
> > 
> > 
> > No, there was also a huge banner on top that said I should add my review even though I already did. And it was there after I closed the convos
> 
> That is because saibotk re-requested your review, probably with the intention that you resolve your started conversations. GitHub's UX definitely isn't the best

Ahhh, that's probably it, thanks

PR #1135, opened 2023-11-15
Database: Rework library to use net.SendStream

ZenBre4ker:
This reduces code size a lot by just streaming the tables instead of using my own custom syncing methods.
That should make the code more readable and less error-prone.


lua/ttt2/libraries/database.lua
```diff
@@ -102,14 +106,28 @@ end
...
+		return false
+	end
+
+	local isReceived = true
+	if not isstring(key) then
+		local keys = registeredDatabases[index].keys
+		for checkKey in pairs(keys) do
+			if not receivedValues[index][itemName][checkKey] then
+				isReceived = false
+				break
```
TimGoll, line 123:
```suggestion
				isReceived = false

				break
```

lua/ttt2/libraries/database.lua
```diff
@@ -393,310 +411,165 @@ end
...
 local clientReceiveFunctions = {}
-
--- Server send and receive functions
-local serverSendFunctions = {}
 local serverReceiveFunctions = {}
 
 ---
--- Send query for registered databases to server
--- @param table data contains only the messageIdentifier here
+-- An Identifier was received, checks if all sent identifiers were received and resets the identifiers
```
TimGoll, line 419:
```suggestion
-- An identifier was received, checks if all sent identifiers were received and resets the identifiers
```

But honestly I'm struggling to understand your sentence here. I don't get what you're trying to say

lua/ttt2/libraries/database.lua
```diff
@@ -393,310 +411,165 @@ end
...
+	if identifiersReceived < messageIdentifier then return end
+
+	identifiersReceived = 0
+	messageIdentifier = 0
 end
 
 ---
 -- Receive query for registered databases from client
 -- and sync them
+-- @param table data = {identifier}
```
TimGoll, line 435: I'm not sure what this doc string means? The description text is `= {identifier}`. That doesn't seem helpful

lua/ttt2/libraries/database.lua
```diff
@@ -393,310 +411,165 @@ end
...
 	}
 
-	-- If more than one database will be send, then sentAdditionalInfo is true
-	local sentAdditionalInfo = net.ReadBool()
-
-	if sentAdditionalInfo then
-		local identifier = net.ReadUInt(uIntBits)
-		local tableCount = net.ReadUInt(uIntBits)
-
+	if data.identifier and table.Count(registeredDatabases) == data.tableCount then
```
TimGoll, line 459: Just a note: Maybe the variable name `data.tableCount` is a bit confusing. At least I first thought you were misusing a function reference.

maybe something like `data.sizeTable`?

lua/ttt2/libraries/database.lua
```diff
@@ -393,310 +411,165 @@ end
...
-	net.WriteUInt(request.index, uIntBits)
-
-	net.WriteString(request.itemName)
-	net.WriteString(request.key)
-end
-
 ---
 -- Receive query for getting a value from client
 -- and returning it
+-- @param table data = {index, accessName, itemName, key, identifier}
```
TimGoll, line 469: same here, the docstring looks weird to me

lua/ttt2/libraries/database.lua
```diff
@@ -393,310 +411,165 @@ end
...
-	local itemName = net.ReadString()
-	local key = net.ReadString()
-	local value = net.ReadString()
-
-	value = ConvertValueWithKey(value, registeredDatabases[index].accessName, key)
-	OnChange(index, itemName, key, value)
-	ValueReceived(index, itemName, key)
+serverReceiveFunctions[MESSAGE_SET_VALUE] = function(data, plyID64)
+	local accessName = registeredDatabases[data.index].accessName
+	database.SetValue(accessName, data.itemName, data.key, data.value, plyID64)
```
TimGoll, line 512:
```suggestion
	local accessName = registeredDatabases[data.index].accessName

	database.SetValue(accessName, data.itemName, data.key, data.value, plyID64)
```

lua/ttt2/libraries/database.lua
```diff
@@ -393,310 +411,165 @@ end
...
 end
 
 ---
--- Send query for resetting the database to server
--- @param table data contains index
+-- Receives changed value from server
+-- Mark value as received and call OnChange-function
+-- @param table data = {index, itemName, key, value}
```
TimGoll, line 518: again

lua/ttt2/libraries/database.lua
```diff
@@ -393,310 +411,165 @@ end
...
-clientSendFunctions[MESSAGE_RESET] = function(data)
-	net.WriteUInt(data.index, uIntBits)
+clientReceiveFunctions[MESSAGE_SET_VALUE] = function(data)
+	OnChange(data.index, data.itemName, data.key, data.value)
+	ValueReceived(data.index, data.itemName, data.key)
 end
 
 ---
 -- Receive query for resetting the database from client
+-- @param table data = {index}
```
TimGoll, line 528: and again

lua/ttt2/libraries/database.lua
```diff
@@ -393,310 +411,165 @@ end
...
-serverSendFunctions[MESSAGE_RESET] = function(data)
-	net.WriteUInt(data.index, uIntBits)
+serverReceiveFunctions[MESSAGE_RESET] = function(data, plyID64)
+	database.Reset(registeredDatabases[data.index].accessName, plyID64)
 end
 
 ---
 -- Receive reset command from server
 -- Reset all stored values back to the defaults
+-- @param table data = {index}
```
TimGoll, line 539: and again!

lua/ttt2/libraries/database.lua
```diff
@@ -393,310 +411,165 @@ end
...
-		net.WriteString(tostring(data.value))
-	end
+clientReceiveFunctions[MESSAGE_RESET] = function(data)
+	ResetDatabase(data.index)
 end
 
 ---
 -- Receive requested default values from server
 -- and cache it
+-- @param table data = {index, itemName, key, value, sentTable,}
```
TimGoll, line 549: huh

lua/ttt2/libraries/database.lua
```diff
@@ -1139,14 +974,14 @@ if SERVER then
 	---
 	-- This is called upon receiving a get request from a player to send a value back
 	-- @warning Dont use this function if you want to get a value from the database, this is meant to be used internally for a client request
-	-- @param table requestData = {plyID64, identifier, index, itemName, key} contains player and the data they requested
+	-- @param table requestData = {identifier, index, itemName, key} contains player and the data they requested
```
TimGoll, line 977: as an example, this looks better to me because there is actual explanation there

TimGoll (changes requested):
As a general note, I think we should nail down a variable naming style. Because you use two different styles in this PR. I always try to put the type first. Like `dataSent` or `sizePlayer`. I don't know if this is the best style, but I'm not a fan of having different styles in the same project.

This is no reason to reject this PR, I only want to note it for discussion so we make sure that we adhere to our style in the future

lua/ttt2/libraries/database.lua
ZenBre4ker: the general idea was to give a c-style class explanation whats inside that variable. THey are mostly internal and shall help to understand what you can use.

lua/ttt2/libraries/database.lua
ZenBre4ker: Well then I always prefer real words. So either tablesize or size I'd like better than sizeTable. Because to mention its a table would be kinda unneccessary.

lua/ttt2/libraries/database.lua
ZenBre4ker: But yeah, could make sense to describe single entries as well

lua/ttt2/libraries/database.lua
ZenBre4ker: This describes 2 functions. Overall description is, that an identifier was received, that is used to identify messages.
If the last identifier was received, this resets it.

TimGoll: You don't have to erxplain it to me, but probably imporve the comment so that it reflects your thinking

lua/ttt2/libraries/database.lua
TimGoll: Maybe someone else wants to chime in as well, but if a variable is named `tableSize` I expect a table with the size. Hence my suggestion of `sizeTable`. I don't know if this is just me or a general coding style. However datatypes are generally put in front like `boolState` for example, sometimes just `bState`

lua/ttt2/libraries/database.lua
TimGoll: probably @Histalek should explain how it works best with his tool

Histalek: this is fine, if/when we switch to the luals annotations we will need to figure out alot of stuff anyways.

right now this is fine as every key in the table is documented and it will just be part of the parameter description. Best case would be a short description of each key, but i think that is to much to put into the parameter description as of now.

lua/ttt2/libraries/database.lua
Histalek: do we really need a table here for a single parameter? This makes it hard to understand what type `identifier` is expected by the function for no real gain?

lua/ttt2/libraries/database.lua
Histalek: additionally most of these (all?) are in `@internal` functions. so only people working on the database library would be exposed to these docs. And we have worse docs for internal stuff tbh

lua/ttt2/libraries/database.lua
ZenBre4ker: Not for this particular function, but for the general workflow, we do. All receiveFunctions follow the same signature of (table, ID64)
And because some functions only need a few parameters, I documented them. In this case its just one 😄

lua/ttt2/libraries/database.lua
ZenBre4ker: I can see what you mean but many variables are called:
screenSize, iconSize, (ClipSize, scope_size), itemSize, queueSize, scrollSize
and I only find:
sizeIcon, which contradicts iconSize you use inside the drawing functions
heightButton, widthButton
and SizeVel which is also not camelCase.

So Im personally no fan of combined words, that sound strange.
and bState, well is a boolean, but we seldom actually use it.
In c++ we also rather have a diffrence between private m_state and public variables state.

Normally I would call it like
numberOfTables or databaseSize could be fine if you really dont like tableSize.

ZenBre4ker: We actually use tableSize in table.lua and sh_rolelayering.lua.

TimGoll: yes, I started that change just recently, because I stubled over the variable names like the one in this comment. However it is fine for me either way, just a preference thing

ZenBre4ker:
Okay, so what is the general opinion?
Shall I rename all variables to a new standard? Ergo adjectiveNoun / typeDescription (/ sizeTable / sizeDatabase ...)

Do I need to explain the table contents? I would rather wait for the next iteration of our Docs and change it then appropriately.

I'd like to merge this and work on the removal of old shop stuff 😄

Histalek (approved):

TimGoll (approved):
Since it seems like the only one that is confused by the unexplained parameter-tables I will approve it

PR #1141, opened 2023-11-19
HUD: Added ammo icons to the PlayerInfo

EntranceJew:
https://github.com/TTT-2/TTT2/assets/5711436/255cccd3-3712-449c-8013-406455e6c69d


gamemodes/terrortown/entities/weapons/weapon_zm_mac10.lua
```diff
@@ -28,7 +28,7 @@ SWEP.Primary.ClipSize = 30
 SWEP.Primary.ClipMax = 60
 SWEP.Primary.DefaultClip = 30
 SWEP.Primary.Automatic = true
-SWEP.Primary.Ammo = "smg1"
+SWEP.Primary.Ammo = "SMG1"
```
Histalek, line 31: why was this changed?

gamemodes/terrortown/gamemode/shared/hud_elements/tttwswitch/pure_skin_wswitch.lua
```diff
@@ -101,7 +109,7 @@ if CLIENT then
 		local empty_color = active and color_empty or color_empty_dark
 
 		local name = TryTranslation(wep:GetPrintName() or wep.PrintName or "...")
-		local cl1, am1 = wep:Clip1(), (wep.Ammo1 and wep:Ammo1() or false)
+		local cl1, am1, ty1 = wep:Clip1(), (wep.Ammo1 and wep:Ammo1() or false), wep.Primary.Ammo or "Pistol"
```
Histalek, line 112: In line 130 we make sure that we have a valid fallback and `ty1` isn't used anywhere else. Is there a specific reason to fallback to the "Pistol" ammo type here?

Histalek (commented):
overall a nice addition. I've just got two questions before i approve

gamemodes/terrortown/entities/weapons/weapon_zm_mac10.lua
EntranceJew: because it [was wrong](https://wiki.facepunch.com/gmod/Default_Ammo_Types) and running a tolower to cover the edge case is unpleasant

Histalek: ah great catch! thanks!

Histalek (approved):

saibotk (changes requested):
Code wise this is fine, but im not convinced that we should add this.

The icon is a bit too big in the player info and looks off.

I'd like to hear @TimGoll on this and feel like the UI gets more cluttered with more icons etc, without big benefit. Since i think it is not really easy to understand the icons, so players are still just confused what it tries to say.

EntranceJew:
There's a few reasons why I made this:
1. Being able to see the ammo type of the equipped gun at a glance in the quick-switch, to know whether or not your primary/secondary/equipment are drawing from the same ammo pool and you'd be better off diversifying your weapon selection.
2. Being able to see what ammo your presently equipped weapon is drawing from. Some addons provide this information for weapons laying on the ground, I put in a matching PR with [weaponstats](https://github.com/TTT-2/ttt2-weaponstats/commit/dd5a1cef0189a46e16fb7131bdc7ea45fc92a95b) to permit seeing the icons of loaded ammo for weapons also featured in the demonstration video, in addition to the names of the ammo being looked at.
3. Knowing whether or not your equipment uses a replenish-able ammo supply. A lot of equipment weapons don't allow for this, but that's hard to know without throwing it on the ground and looking at it, making it even harder to discern if you're wielding an equipment item that can utilize additional ammo.
4. Matching the pickup of ammo you have equipped with weaponstats at a glance is pretty handy, especially since deagle and sniper ammo models are just different color swaps and a variety of conditions can cause it to be extremely difficult to tell the difference.

I did prototype plain icons for each round type but the fact is without crisper SVG rendering implemented the icons of the bullets I drafted are too narrow to fill up a square area, and the only other substitute I could think of was using the gun type icons directly. That suffers from the connotation of dictating the shape of the weapon it should be associated with to some extent.

TimGoll:
> Code wise this is fine, but im not convinced that we should add this.
> 
> The icon is a bit too big in the player info and looks off.
> 
> I'd like to hear @TimGoll on this and feel like the UI gets more cluttered with more icons etc, without big benefit. Since i think it is not really easy to understand the icons, so players are still just confused what it tries to say.

I have to agree with you, at least party. I think it works in the weapon selection, maybe it should be decreased by 10% though. But it doesn't work in the playerinfo panel. It is also not aligned with the armor icon, which makes it look off. Also the size is different than the armor icon size.

If you want to merge that, it should be behind a convar that is disabled by default. At least for the playerinfo panel. You should also retouch the icon sizes

TimGoll:
> There's a few reasons why I made this:
> 
> 1. Being able to see the ammo type of the equipped gun at a glance in the quick-switch, to know whether or not your primary/secondary/equipment are drawing from the same ammo pool and you'd be better off diversifying your weapon selection.
> 2. Being able to see what ammo your presently equipped weapon is drawing from. Some addons provide this information for weapons laying on the ground, I put in a matching PR with [weaponstats](https://github.com/TTT-2/ttt2-weaponstats/commit/dd5a1cef0189a46e16fb7131bdc7ea45fc92a95b) to permit seeing the icons of loaded ammo for weapons also featured in the demonstration video, in addition to the names of the ammo being looked at.
> 3. Knowing whether or not your equipment uses a replenish-able ammo supply. A lot of equipment weapons don't allow for this, but that's hard to know without throwing it on the ground and looking at it, making it even harder to discern if you're wielding an equipment item that can utilize additional ammo.
> 4. Matching the pickup of ammo you have equipped with weaponstats at a glance is pretty handy, especially since deagle and sniper ammo models are just different color swaps and a variety of conditions can cause it to be extremely difficult to tell the difference.
> 
> I did prototype plain icons for each round type but the fact is without crisper SVG rendering implemented the icons of the bullets I drafted are too narrow to fill up a square area, and the only other substitute I could think of was using the gun type icons directly. That suffers from the connotation of dictating the shape of the weapon it should be associated with to some extent.

I get all your points. If you remove the icon from the playerinfo panel and slightly decrease the size in the weaponswitcher, would this work for you? Becase I think it would stil satisfy your request?
What's your opinion on this @saibotk?

saibotk:
Mhh the more i think about it, the more i come to the point where I see this as a theoretical improvement where no real need is satisfied.
It does add more UI, which is not a good thing, because our UI is already pretty hard to understand.
And I think displaying the icons does not really help players in understanding which ammo they are looking for, especially since ammo models are still something i need to look out for instead of the icons.

I honestly think we should focus more on trying to be concise and rather minimalistic with sane defaults, instead of the gmod way of introducing 3000 settings for everything, which i really dislike.

Also regarding your other points @EntranceJew, i think you still display the fallback ammo icon even if a weapon has ammo which cannot be found, correct me if i am wrong.

But all those points are something I think is safe to have the player explore by themselves. After some rounds you just get to know which ammo models / colors are for which weapon type. 

If we are talking about external weapons etc. I can understand that it is often difficult to know which ammo they use, but if we want to solve that, i think we should rather try to solve this differently. Without extra icons and maybe add something to the ammo target ID to indicate that this ammo can be used with the current weapon or maybe showing if the ammo can be used with any weapon currently in your inventory.

This feels more intuitive to me than having ammo icons on display all the time even if i do not need the information at all once i played some rounds.

That are my rough thoughts, 
please do not see this in a disrespecting way towards your work @EntranceJew, i really appreciate that, but i still want to maintain a constructive discussion about potential improvements and their broader need in the game instead of just adding them because they seem like an improvement.

Thanks!

EntranceJew:
image
I agree that the weight of the box icons really throw off the whole appearance of things, but do you think all the same things when using just the gun's icon by itself?

EntranceJew:
image
there's also this alternative one i was considering that more generally just shows the weapon kind like the pickup hints

ZenBre4ker:
> image I agree that the weight of the box icons really throw off the whole appearance of things, but do you think all the same things when using just the gun's icon by itself?

WIthout the box I personally like it a lot more. The alternative is too much though.

EntranceJew:
> But it doesn't work in the playerinfo panel. It is also not aligned with the armor icon, which makes it look off. Also the size is different than the armor icon size.

image

when I do that it looks goofy and all by itself, was the major reason I right-aligned it
if the "HEALTH:" label were a heart icon, then the alignment would be less alien:

image

TimGoll:
> hen I do that it looks goofy and all by itself, was the major reason I right-aligned it

My call is still: remove it from the playerinfo panel completely. I kinda like the solution in the weapon selector in the last few images

EntranceJew:
> My call is still: remove it from the playerinfo panel completely. I kinda like the solution in the weapon selector in the last few images

image
image
image

I changed it to use the full bullet shape which is a lot more solid of a shape for the playerinfo HUD. Is dropping it still the call?

TimGoll:
Okay, that does look good imho, but I'm not sure if these bullet icons are understandable?

TimGoll:
More Icon Feedback:
The magneto Stick slot always has this one item. So it should look tike the stick in my opinion. Also holstered should show empty hands, shouldn't it?

Slot 9 is class weapons. Maybe we should use the C from TTTC as an icon here?

saibotk:
I still would vote against using any icons all together, since i think they do not solve this problem in a good way.
And using the ammo icons in the weapon switcher is as confusing as using the "ammo-named weapon" as an icon. 

It's all just very confusing and i still think the best way would just be to not try to educate the player about what ammo types are and just tell them that they can use the particular ammo that they are looking at. In the end they dont care if it is "Shotgun ammo" or "SMG ammo" they just need the information if they can use it and this will also intuitively let them learn about the different ammo models / colors and what weapon can use which.

ZenBre4ker:
I guess this comes down to personal preferences. I also think it is of no benefit at all. The ammo system stays weird. You would rather need the picture of the actual ammo box to make it clear. So rather a little hotkey to shortly show a 3D Version of the ammo you need.

But if we want those icons, this is my take on the examples:
The icons in the weapon selector next to the numbers are way too much. Absolute no!

Next to the ammo number and not for every slot, as well as the icon under the health bar look good and dont insert too much noise. Probably yes!

EntranceJew:
i deleted the weapon switch
image
but with more things using fewer ammo types it's going to become even more significant to know whether your primary is using rifle ammo or 9mm ammo, because of the limited max ammo, you're still going to not want to have a fully automatic primary that chews through your secondary's ammo

TimGoll:
> i deleted the weapon switch image but with more things using fewer ammo types it's going to become even more significant to know whether your primary is using rifle ammo or 9mm ammo, because of the limited max ammo, you're still going to not want to have a fully automatic primary that chews through your secondary's ammo

In my opinion this is OK now. Could you please remove the `health: ` string from the language files as well?

You might be right with your comment. Once we start that rework, we should discuss our goals beforehand

gamemodes/terrortown/gamemode/shared/hud_elements/base_elements/pure_skin_element.lua
```diff
@@ -82,4 +82,32 @@ if CLIENT then
...
+		[WEAPON_MELEE] = Material("vgui/ttt/slot/slot_weapon_melee"),
+		[WEAPON_PISTOL] = Material("vgui/ttt/slot/slot_weapon_pistol"),
+		[WEAPON_HEAVY] = Material("vgui/ttt/slot/slot_weapon_heavy"),
+		[WEAPON_NADE] = Material("vgui/ttt/slot/slot_weapon_nade"),
+		[WEAPON_CARRY] = Material("vgui/ttt/slot/slot_weapon_carry"),
+		[WEAPON_UNARMED] = Material("vgui/ttt/slot/slot_weapon_unarmed"),
+		[WEAPON_SPECIAL] = Material("vgui/ttt/slot/slot_weapon_special"),
+		[WEAPON_EXTRA] = Material("vgui/ttt/slot/slot_weapon_extra"),
+		[WEAPON_CLASS] = Material("vgui/ttt/slot/slot_weapon_class"),
+	}
```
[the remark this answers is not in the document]
EntranceJew: those were relocated from pure_skin_pickup and changed to not use the png versions of those images; they're here so there more generally available for any pure_skin_element descendant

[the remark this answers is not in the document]
TimGoll: oh, then nevermind. Thanks for letting me know

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
```diff
@@ -278,15 +293,30 @@ if CLIENT then
...
 					local text = string.format("%i + %02i", ammo_clip, ammo_inv)
 
-					self:DrawBar(nx, ty, bw, bh, color_ammoBar, ammo_clip / ammo_max, t_scale, text)
+					self:DrawBar(nx, ty, bw, bh, color_ammoBar, ammo_clip / ammo_max, t_scale)
+
+					local icon_mat = BaseClass.BulletIcons[ammo_type] or mat_tid_ammo
+
+					a_pos_x = nx + bw - math.Round(a_size * t_scale)
+
+					if true then
```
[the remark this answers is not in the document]
EntranceJew: scoped something that didn't need to be conditional while experimenting, deleted this on my end now

gamemodes/terrortown/gamemode/shared/hud_elements/tttpickup/pure_skin_pickup.lua
```diff
@@ -84,7 +76,7 @@ if CLIENT then
 
 		if item.type == PICKUP_WEAPON then
 			tipColor = LocalPlayer():GetRoleColor()
-			icon = self.SlotIcons[item.kind] or self.icon_item
+			icon = BaseClass.SlotIcons[item.kind] or self.icon_item
```
saibotk, line 79:
```suggestion
			icon = self.SlotIcons[item.kind] or self.icon_item
```

This should work since it is inherited from the base

materials/vgui/ttt/tid/tid_big_weapon_deagle.vtf
saibotk, line 1: this is not used anywhere anymore i think?

saibotk (changes requested):
Looks good!

But i think you will also need to add those new icons to the force download file.

materials/vgui/ttt/tid/tid_big_weapon_deagle.vtf
EntranceJew: I made it, because it didn't exist, and seemed like it should. The spawn editor still has the ammo crate with the deagle icon on it, but there was no standalone deagle icon in the repository, even though setting deagle spawns should be a functionality if you can spawn their ammo.

saibotk: Sorry, i dont understand what you are trying to say. 
But if this is not actually being used anywhere then it should be removed.

EntranceJew: ok

gamemodes/terrortown/gamemode/shared/hud_elements/base_elements/pure_skin_base_stacking_element.lua
saibotk, line 1: We know this is a limitation of the current system to not allow some kind of trait / multi inheritance, but i suggest we solve this in the same way as we did in the pure skin weapon switcher, where we grab the pure_skin_element in PreInitialize to access it. 

And we should maybe think about a better solution later on. But i think this is better for now than copying this code here again.

saibotk (changes requested):

PR #1142, opened 2023-11-21
Feature/role preference options

EntranceJew:
made dead code not dead


gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -33,7 +33,15 @@ roleselection.cv = {
 
 	---
 	-- @realm server
-	ttt_max_baseroles_pct = CreateConVar("ttt_max_baseroles_pct", "0", {FCVAR_NOTIFY, FCVAR_ARCHIVE}, "Maximum amount of different baseroles based on player amount. ttt_max_baseroles needs to be 0")
+	ttt_max_baseroles_pct = CreateConVar("ttt_max_baseroles_pct", "0", {FCVAR_NOTIFY, FCVAR_ARCHIVE}, "Maximum amount of different baseroles based on player amount. ttt_max_baseroles needs to be 0"),
+
+	---
+	-- @realm server
+	ttt2_roles_shuffle_players = CreateConVar("ttt2_roles_shuffle_players", "0", {FCVAR_NOTIFY, FCVAR_ARCHIVE}, "If not enabled, players will get priority for role selection based on the order they connected."),
```
TimGoll, line 40: I wouldn't add a convar for this, just code it as always enabled. As your solution is an improvement you don't need a convar to renable the "bug"

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -33,7 +33,15 @@ roleselection.cv = {
...
+
+	---
+	-- @realm server
+	ttt2_roles_trickle_down_odds = CreateConVar("ttt2_roles_trickle_down_odds", "3", {FCVAR_NOTIFY, FCVAR_ARCHIVE}, "The 1/N odds of a player receiving a role and not being passed up."),
```
TimGoll, line 44: I'm also not 100% sure we need this. If you add this, you have to explain in detail in the UI how changing this convar affects the selection. Because I think it is hard to understand.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -782,30 +782,22 @@ end
...
-		-- give this player the role if
-		if subrole == ROLE_INNOCENT -- this role is an innocent role
-			or #plys <= roleAmount -- or there aren't enough players anymore to have a greater role variety
-			or ply:GetBaseKarma() > min_karmas -- or the player has enough karma
-				and not ply:GetAvoidRole(subrole) -- and the player doesn't avoid this role
-			or math.random(3) == 2 -- or if the randomness decides
-		then
-			table.remove(plys, pick)
+		local ply = plys[i]
+		if not (curRoles < roleAmount and #plys > 0) then break end
```
TimGoll, line 791:
```suggestion
		if not curRoles < roleAmount and #plys > 0 then break end
```

TimGoll (changes requested):
Nice contribution thank you. I also like the way you named your variables. I'm hesitating a bit because of the convars though

TimGoll:
I also have to run the TTT2LanguageParser on this one

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
EntranceJew: Personally, I don't even like having the math.random call in there -- I think the shuffle is doing all the work the RNG really needs to do to consider things.

Should we just only focus on conditions that prevent the role from being selected and pass true by default?

TimGoll: That's hard to make a call on as I never worked on the roleselection. Maybe @saibotk has more insight into this. Or @ZenBre4ker?

ZenBre4ker: The system is broken in a way anyway. As @saibotk mentioned in another PR, I wouldnt give so many options. I would either always allow or forbid avoiding roles alltogether. I dont like that system anyway, so I would suggest to remove role avoiding alltogether.

About the random call, I believe they tried to make it more random as in the past often the same players in a row got the same role over and over again.

And the trickle system I need to take a look again.

saibotk: I feel the same way, all this role avoidance stuff is rarely used by players and should rather be something a server owner decides upon. This just adds unnecessary complexity and confusion.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -826,6 +813,11 @@ function roleselection.SelectRoles(plys, maxPlys)
 
 	plys = roleselection.GetSelectablePlayers(plys or player.GetAll())
 
+	-- we should give the player list an inital shuffle,
+	-- because if we're populating from player.GetAll(),
+	-- it heavily biases players that connected first for roles
+	table.Shuffle(plys)
```
saibotk, line 819: Could you please explain this?
As i can currently not make out any part of the code below where this is the case since the players are selected at random for each role. The table order is in no way considered for the role selection, at least as far as i can see this currently.

saibotk (changes requested):

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -892,8 +884,8 @@ function roleselection.SelectRoles(plys, maxPlys)
 
 		-- last but not least, upgrade the innos and players without any role to special/normal innos
 		local innos = {}
-		for i = 1, #plysSecondPass do
-			local ply = plysSecondPass[i]
+		for j = 1, #plysSecondPass do
+			local ply = plysSecondPass[j]
```
ZenBre4ker, line 888: This should stay as "i". I dont see where this would be a problem.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: I think its fine, it was just moved up. They removed the math.random pick in SelectBaseRolePlayers and shuffled the list instead.

So my suggestion was to simplify the comment. "Randomize role assignment by shuffling the list early"

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: Yeah checked again, I would just set the tricklerate as before to 3 and check if it is 2, so 1/3 chance, that role gets assigned without the other conditions.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -782,30 +778,21 @@ end
 local function SelectBaseRolePlayers(plys, subrole, roleAmount)
 	local curRoles = 0
 	local plysList = {}
+	local roleData = roles.GetByIndex(subrole)
 
-	local minKarmaCVar = GetConVar("ttt_" .. roles.GetByIndex(subrole).name .. "_karma_min")
-	local min_karmas = minKarmaCVar and minKarmaCVar:GetInt() or 0
-
-	while curRoles < roleAmount and #plys > 0 do
```
ZenBre4ker, line 789: Doesnt removing this while loop here means, that players might not get a role in this pass?

As far as I can see, the "CanSelectRole" method can most often just return false.

ZenBre4ker (changes requested):
You definitely revived the dead code 😄 

But I believe, that you implemented SelectBaseRolePlayers wrongly. Removing the random aspect and the while loop means, that this pass could be false for many players.
I am unsure about the consequences, but it does not have the same behaviour as before, if I am not wrong.

Also in my opinion please dont insert two more convars. The trickle rate is just there to randomize the other states.

TimGoll:
I see you worked again on this PR. We discussed this one as well las sunday and we had the uniform opinion, that the trickle down convar should be removed. It might be a good idea to move it to a library variagble though. Something like `roleselection.trickleDownRate`. That way addons can easily interact with it, if they so desire.

@ZenBre4ker Wasn't sure if you broke any of the roleselection logic, so he has to confirm your changes.

Last but not least we'd like to let you know that we tend to remove the role avoidance completely as it doesn't work that well anyway. Most players seem really confused by it.

EntranceJew:
is the intention to have this commit remove the role avoidance settings instead? or will that be later?

TimGoll:
> is the intention to have this commit remove the role avoidance settings instead? or will that be later?

Later. If you want to add to the role avoidance that is fine. Just be aware that it will probably be removed at some point

saibotk (changes requested):

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
ZenBre4ker: (After requested Review)
Isnt this still having the same problem here?
Exchanging the while loop with a for loop seems like different behaviour?

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -16,6 +16,7 @@ roleselection.finalRoles = {}
 roleselection.selectableRoles = nil
 roleselection.baseroleLayers = {}
 roleselection.subroleLayers = {}
+roleselection.trickleDownRate = 0
```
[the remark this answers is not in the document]
ZenBre4ker: This is still 0, should be at least 3.

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -572,10 +573,14 @@ local function SetSubRoles(plys, availableRoles, selectableRoles, selectedForced
 	local plysAmount = #plys
 	local availableRolesAmount = #availableRoles
 	local tmpSelectableRoles = table.Copy(selectableRoles)
+	-- plys are already shuffled, continue with existing priority
+	local newPlys = table.Copy(plys)
 
-	while plysAmount > 0 and availableRolesAmount > 0 do
-		local pick = math.random(plysAmount)
-		local ply = plys[pick]
+	for i = #newPlys, 1, -1 do
```
ZenBre4ker, line 579: The for loop here also has the wrong behaviour. It only goes through every player once. But it could be that some players are not assigned.

ZenBre4ker (changes requested):

CHANGELOG.md
```diff
@@ -108,6 +108,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Moved reset buttons onto the left (by @a7f3)
 - Added ammo icons to the weapon switch HUD and player status HUD elements (by @EntranceJew)
 - Changed the disguiser icon to be more fitting (by @TimGoll)
+- Role selection RNG improvements (by @EntranceJew):
```
ZenBre4ker, line 111: Also in my opinion there are no RNG improvements, as you said in this PR, it generally only "made dead code not dead"

PR #1145, opened 2023-11-25
Fix nil values for "dcombobox", "playermodels", "dcombobox_ttt2

sbzlzh:
@ZenBre4ker Is that what you said, and is there anywhere else?

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcheckboxlabel_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcheckboxlabel_ttt2.lua
@@ -91,14 +91,12 @@ function PANEL:SetServerConVar(cvar)
 	self.serverConVar = cvar
 
 	-- Check if self is valid before calling SetValue.
-	if IsValid(self) then
-		cvars.ServerConVarGetValue(cvar, function(wasSuccess, value, default)
-			if wasSuccessn then
-				self:SetValue(tobool(value), true)
-				self:SetDefaultValue(tobool(default))
-			end
-		end)
-	end
+	cvars.ServerConVarGetValue(cvar, function(wasSuccess, value, default)
+		if wasSuccess and IsValid(self) then
+			self:SetValue(tobool(value), true)
+			self:SetDefaultValue(tobool(default))
+		end
+	end)
 
 	callbackEnabledVarTracker = callbackEnabledVarTracker + 1
 	local myIdentifierString = "TTT2CheckBoxConVarChangeCallback" .. tostring(callbackEnabledVarTracker)
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcheckboxlabel_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcheckboxlabel_ttt2.lua
@@ -90,8 +90,9 @@ function PANEL:SetServerConVar(cvar)
 
 	self.serverConVar = cvar
 
+	-- Check if self is valid before calling SetValue.
 	cvars.ServerConVarGetValue(cvar, function(wasSuccess, value, default)
-		if wasSuccess then
+		if wasSuccessn then
 			self:SetValue(tobool(value), true)
 			self:SetDefaultValue(tobool(default))
 		end
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcheckboxlabel_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcheckboxlabel_ttt2.lua
@@ -91,12 +91,14 @@ function PANEL:SetServerConVar(cvar)
 	self.serverConVar = cvar
 
 	-- Check if self is valid before calling SetValue.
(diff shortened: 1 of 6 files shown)


sbzlzh:
Can I just remove the local box = boxCache[name] from the deletion on lines 80 and 94 and use the local on line 67 instead?

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcheckboxlabel_ttt2.lua
```diff
@@ -90,12 +90,15 @@ function PANEL:SetServerConVar(cvar)
...
-	end)
+	-- Check if self is valid before calling SetValue.
+	if IsValid(self) then
+		cvars.ServerConVarGetValue(cvar, function(wasSuccess, value, default)
+			if wasSuccessn then
+				self:SetValue(tobool(value), true)
+				self:SetDefaultValue(tobool(default))
+			end
+		end)
+	end
```
ZenBre4ker, line 101:
```suggestion
	cvars.ServerConVarGetValue(cvar, function(wasSuccess, value, default)
		if wasSuccess and IsValid(self) then
			self:SetValue(tobool(value), true)
			self:SetDefaultValue(tobool(default))
		end
	end)
```

This was correct in your previous version already. Please change it in the other call as well.

lua/terrortown/menus/gamemode/administration/playermodels.lua
```diff
@@ -63,13 +63,17 @@ function CLGAMEMODESUBMENU:Populate(parent)
...
+		local outerBox = boxCache[name]
+		if outerBox then
+			playermodels.IsSelectedModel(name, function(value)
+				boxCache[name]:SetModelSelected(value, false)
+			end)
+
+			playermodels.IsHattableModel(name, function(value)
+				boxCache[name]:SetModelHattable(value, false)
+			end)
+		end
```
ZenBre4ker, line 76:
```suggestion
		playermodels.IsSelectedModel(name, function(value)
			if not boxCache[name] then return end
			
			boxCache[name]:SetModelSelected(value, false)
		end)

		playermodels.IsHattableModel(name, function(value)
			if not boxCache[name] then return end
			
			boxCache[name]:SetModelHattable(value, false)
		end)
```

What I meant was around the call of the function you dont need to check it. But inside the anonymous function you do. Because that function may be called later when the f1 menu is not valid anymore for example due to server lag or any other causes.

ZenBre4ker (changes requested):
Thanks for the separate PR. Your changes on the serverconvar before were correct. Now they arent anymore. I gave you two examples you could copy or directly commit if you want.

sbzlzh:
It's done.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dnumslider_ttt2.lua
```diff
@@ -312,12 +312,15 @@ function PANEL:SetServerConVar(cvar)
...
-	end)
+	-- Check if self is valid before calling SetValue.
+	if IsValid(self) then
+		cvars.ServerConVarGetValue(cvar, function(wasSuccess, value, default)
+			if wasSuccess then
+				self:SetValue(tonumber(value), true)
+				self:SetDefaultValue(tonumber(default))
+			end
+		end)
+	end
```
ZenBre4ker, line 323: this here needs to be changed correctly as well

ZenBre4ker: and maybe remove the comments, should be self-explanatory

sbzlzh: You want to be like those two up there, don't you.

ZenBre4ker: Yes. Because they have the same structure, they should be the same ^^

sbzlzh: Does it look like this when it's done?

ZenBre4ker (approved):
thanks a lot. As I mentioned, if you want you can use your fix in combobox and textentry as well. Otherwise this is fine 👍

sbzlzh:
> thanks a lot. As I mentioned, if you want you can use your fix in combobox and textentry as well. Otherwise this is fine 👍

Okay. I'm new to GitHub and not very familiar with this stuff. I was wondering what this combobox and textentry is?

ZenBre4ker:
@sbzlzh Those are different files I mean, that stand for the ingame VGUI Elements:
grafik
Here you changed the dnumslider and dcheckboxlabel. But there are the combox and textentry files/elements as well, which have the same behaviour. If you want you can fix those as well.

sbzlzh:
> @sbzlzh Those are different files I mean, that stand for the ingame VGUI Elements: grafik Here you changed the dnumslider and dcheckboxlabel. But there are the combox and textentry files/elements as well, which have the same behaviour. If you want you can fix those as well.

This means that using this in other lua can also be solved using this method. I see :)

ZenBre4ker (approved):

Histalek (approved):

PR #1152, opened 2023-12-07
fix: prevent nil error in database module

Histalek:
In cases where `itemName` is not in the `accessName` database table a nil error would occur while trying to index `sqlData`.

`sqlData` is populated via the ORM library `:find(itemName)` function which returns the database entry with `itemName` as its primary key. This returns nil if no such entry is found.

This behaviour was not checked here and results in breaking the equipment shop.

The minimal working fix would be to if-clause only the `sqlData[_key]` access. I've scoped the if-clause over a few more items which should all be NOOPs in case of `sqlData` being nil.

### DIFF
--- a/lua/ttt2/libraries/database.lua
+++ b/lua/ttt2/libraries/database.lua
@@ -1089,15 +1089,18 @@ if SERVER then
 				return true, value
 			end
 
-			local newTable = {}
-			for _key in pairs(dataTable.keys) do
-				newTable[_key] = sqlData[_key]
+			if sqlData then
+				local newTable = {}
+				for _key in pairs(dataTable.keys) do
+					newTable[_key] = sqlData[_key]
+				end
+
+				sqlData = newTable
+				ConvertTable(sqlData, accessName)
+
+				dataTable.storedData[itemName] = sqlData
 			end
 
-			sqlData = newTable
-			ConvertTable(sqlData, accessName)
-
-			dataTable.storedData[itemName] = sqlData
 		else
 			-- Get all data, convert and return it
 			sqlData = dataTable.orm:All()


TimGoll (approved):
good work, thanks for finding the issue here!

saibotk (approved):
Thanks!

ZenBre4ker (commented):
This fix seems to be fine. But I believe that whatever used that call is probably not working anymore. The description itself says, that normally the default keys are inserted, but not if you leave name and key `nil`. So I guess, that there needs to be a change somewhere else as well.

TimGoll:
> This fix seems to be fine. But I believe that whatever used that call is probably not working anymore. The description itself says, that normally the default keys are inserted, but not if you leave name and key `nil`. So I guess, that there needs to be a change somewhere else as well.

@saibotk this probably means we can merge it?

saibotk:
No it means this has another root cause that needs to be investigated

ZenBre4ker:
But the root cause here should be, that orm:Find doesnt return an empty table when there is no database.
Could it be that it only faily on a new installment?

TimGoll:
> But the root cause here should be, that orm:Find doesnt return an empty table when there is no database. Could it be that it only fails on a new installment?

My installalation is at least three years old

Histalek:
> But the root cause here should be, that orm:Find doesnt return an empty table when there is no database.

The database is there, the specific entry is not.

I'm fine with changing the return value of `orm:Find` if that would help. There aren't many consumers anyway and we should be able to quickly adapt those as well.

> Could it be that it only fails on a new installment?

No it fails if any weapon is not already in the database. Presumably because it wasn't installed on the server prior to your change.

Histalek:
> I just remembered that I did this rework here last minute [a415025](https://github.com/TTT-2/TTT2/commit/a415025d3efe14efc71b8e851a9913f6f6e6a45e)
> 
> Maybe I didnt fully did it right? But thats probably causing the error. If thats handled correctly with this fix it should be fine.

I'll look at this in a few hours

Histalek:
i've changed it to a way simpler fix, which just makes me believe that i should change the orm:Find return value ...

i'll still need to fully test this, but this fix should additionally allow the equipment editor to work without a map-reload which my former fix did not allow.

ZenBre4ker (approved):
Yeah that should be intended behavior.

Histalek:
Ok this seems good to go.

The only thing i've realized while testing is that there seems to be a visual bug in the Equipment editor. e.g. an item that is not already in my server's database will show as not `limited` in the editor but the default and actual behaviour is that it is indeed limited.
As far as i could tell this will not be saved to the database if i change any other value, so it seems to be purely visual.
This still leads to a confusing UX and should be investigated and fixed.

TimGoll:
> Ok this seems good to go.
> 
> The only thing i've realized while testing is that there seems to be a visual bug in the Equipment editor. e.g. an item that is not already in my server's database will show as not `limited` in the editor but the default and actual behaviour is that it is indeed limited. As far as i could tell this will not be saved to the database if i change any other value, so it seems to be purely visual. This still leads to a confusing UX and should be investigated and fixed.

agreed. But keep in mind that the reworked shop UI is our next big update anyway

PR #1157, opened 2023-12-10
Mini Scoreboard Confirmed Spectator Fix

EntranceJew:
mini scoreboard kept showing confirmed players that became spectators as revived, which is wrong

video of before / after fix hotreloads:
https://github.com/TTT-2/TTT2/assets/5711436/baf1a083-4165-4a17-9395-550735f2e903

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
   - `DNumSliderTTT2`, `DCheckBoxLabelTTT2`, `DComboBoxTTT2`
 - Added dashing to propspec (by @TimGoll)
 - Added new functions to database module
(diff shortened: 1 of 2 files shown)


TimGoll (approved):
huh, nice. Good catch!

EntranceJew:
I caught it because there was a similar classification of bug with my role outlines that took a really long time to discern the cause of, but a T-Posing silhouette of a player visible from anywhere on the map was a lot easier to track down.

TimGoll:
> I caught it because there was a similar classification of bug with my role outlines that took a really long time to discern the cause of, but a T-Posing silhouette of a player visible from anywhere on the map was a lot easier to track down.

You know, I had the feeling something was off with the miniscoreboard for a while no. I just wasn't able to narrow it down. Thanks!

Histalek (approved):

Issue #1146, opened 2023-12-03
Sprinting: Wrong initial value


TimGoll:
After the sprint rework https://github.com/TTT-2/TTT2/pull/1035 the initial value of the sprint bar is set to 0. This is unnoticable in most cases but it should have the max value as initial value. This is noticable when you set the sprinting to "always sprint" without recharging because now the stamina is always 0 even though the consumption is zero as well

PR #1159, opened 2023-12-10
Sprint: Fix stamina start value

ZenBre4ker:
#1146

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
@@ -24,6 +24,10 @@ function plymeta:SetupDataTables()
 	-- This has to be transferred, because we need the value when predicting the player movement
 	-- It turned out that this is the only reliable way to fix all prediction errors.
 	self:NetworkVar("Float", 0, "SprintStamina")
+
+	if SERVER then
+		self:SetSprintStamina(1)
+	end
 end
 
 ---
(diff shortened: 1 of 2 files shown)


gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -23,7 +23,7 @@ end
 function plymeta:SetupDataTables()
 	-- This has to be transferred, because we need the value when predicting the player movement
 	-- It turned out that this is the only reliable way to fix all prediction errors.
-	self:NetworkVar("Float", 0, "SprintStamina")
+	self:NetworkVar("Float", 1, "SprintStamina")
```
Histalek, line 26: This only changes the slot of the network var, not the default/starting value

Histalek (changes requested):

ZenBre4ker:
Yeah now I actually tested it 😄

TimGoll:
Is 1 the max value? Aren't you able to set the stamina in the UI to something greater than 1?

ZenBre4ker:
https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/shared/sh_sprint.lua#L62-L74

As far as I understqand the code, 1 is max. The rest is just appropriately scaled I guess.

TimGoll:
I agree:

https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/shared/sh_sprint.lua#L54

TimGoll (approved):

Histalek (approved):

PR #1172, opened 2023-12-12
Fix missing fastutf8

sbzlzh:
image
Fixed missing fastutf8, found that there seems to be something missing to introduce it correctly causing the error in the picture to occur when pressing F1 or pressing TAB after already using Chinese.

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/sh_init.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_init.lua
@@ -551,6 +551,7 @@ include("ttt2/extensions/input.lua")
 include("ttt2/extensions/cvars.lua")
 
 -- include libraries
+include("ttt2/libraries/fastutf8.lua")
 include("ttt2/libraries/huds.lua")
 include("ttt2/libraries/hudelements.lua")
 include("ttt2/libraries/items.lua")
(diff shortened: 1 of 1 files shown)


TimGoll:
I guess @saibotk didn't acutally test his PR 👀 :D

TimGoll (approved):

gamemodes/terrortown/gamemode/shared/sh_init.lua
```diff
@@ -577,6 +577,7 @@ include("ttt2/libraries/playermodels.lua")
 include("ttt2/libraries/entspawnscript.lua")
 include("ttt2/libraries/bodysearch.lua")
 include("ttt2/libraries/keyhelp.lua")
+include("ttt2/libraries/fastutf8.lua")
```
saibotk, line 580: This should be the first in this block

saibotk (changes requested):

saibotk:
I did but seems like i missed selecting this change for the commit

gamemodes/terrortown/gamemode/shared/sh_init.lua
sbzlzh: all right

TimGoll:
@saibotk done, you can approve it

saibotk (approved):

PR #1173, opened 2023-12-12
revert most of the change from #1006

TimGoll:
PR #1006 tried to simplify the gametext entity. While doing so it changed many things that break the functionality. This one is on us for being to lax while reviewing it. Most of these changes should not have been done.


TimGoll:
image

The error was caused by maps trying to send messages to MStack and failing

saibotk (approved):

ZenBre4ker (approved):

Issue #1171, opened 2023-12-12
NULL Entity Client Error


thebigsleepjoe:
## TTT2 Version

- Workshop version from Steam (v0.12.0b)

## Describe the bug (mandatory)

A clientside error is spammed in the console, often at seemingly random intervals. It does not occur outside of an active round, and only happens when the UI is rendering.

The stack trace is as follows:
```
[TTT2 (Base) - v0.12.0b] gamemodes/terrortown/gamemode/shared/sh_player_ext.lua:46: Tried to use a NULL entity!
  1. Team - [C]:-1
   2. IsSpec - gamemodes/terrortown/gamemode/shared/sh_player_ext.lua:46
    3. Draw - gamemodes/terrortown/gamemode/shared/hud_elements/tttminiscoreboard/pure_skin_miniscoreboard.lua:182
     4. DrawElemAndChildren - gamemodes/terrortown/gamemode/shared/huds/base_huds/hud_base/cl_init.lua:306
      5. DrawElemAndChildren - gamemodes/terrortown/gamemode/shared/huds/base_huds/hud_base/cl_init.lua:303
       6. Draw - gamemodes/terrortown/gamemode/shared/huds/base_huds/hud_base/cl_init.lua:332
        7. DrawHUD - gamemodes/terrortown/gamemode/client/cl_hud_manager.lua:18
         8. unknown - game modes/terror town/gamemode/client/cl_hud_manager.lua:45
```
## To reproduce

Steps to reproduce the behaviour:

1. Play the game with any significant number of players
2. Have a player leave the game at any point during the match. This seems to be best replicated when a player in the match that dies disconnects.
3. Sometimes the error will occur, depending on your luck.

## Expected behaviour

There should be no errors in console.

## Context (please provide as much as you can)

- I am playing with my TTT bots that I have been actively developing, and this error only began as of this recent TTT2 update. I have triple-and-quadruple-checked that my bots are not causing this error.
- I am not playing with any other addons.

## Anticipated source/resolution

It appears there is an unsafe player cache being maintained in pure_skin_miniscoreboard.lua at HUDELEMENT:Draw()

Within this code, the line 182 (`if not ply:IsSpec() and ply:WasRevivedAndConfirmed() then`) is throwing the error, presumably because ply is NULL. This error propagates up the chain until the Player meta's Team function finally throws this entity-invalid error.

[...]

PR #1174, opened 2023-12-12
Fix invalid entity client error

sbzlzh:
It does solve the problem with #1171. could the code in line 46 also be used in line 38.

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
@@ -43,7 +43,7 @@ end
 -- @return boolean
 -- @realm shared
 function plymeta:IsSpec()
-    if not IsValid(self) then return false end
+	if not IsValid(self) then return false end
 	return self:Team() == TEAM_SPEC
 end
 
--- a/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
@@ -43,6 +43,7 @@ end
 -- @return boolean
 -- @realm shared
 function plymeta:IsSpec()
+    if not IsValid(self) then return false end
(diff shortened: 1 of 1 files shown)


sbzlzh:
This is before it's used.
image

This is after use
image

TimGoll:
@saibotk @Histalek What is your opinion on this?

My first thought was "oh, no, this shouldn't be in there. But on a second though this might be alright? Because a player, that is disconnected, is no spec.

Might cause problems with checks that are `if not ply:IsSpec` though

sbzlzh:
> @saibotk @Histalek What is your opinion on this?
> 
> My first thought was "oh, no, this shouldn't be in there. But on a second though this might be alright? Because a player, that is disconnected, is no spec.
> 
> Might cause problems with checks that are `if not ply:IsSpec` though

There may be more than one way to do this, let's see if you have any others.

TimGoll:
> > @saibotk @Histalek What is your opinion on this?
> > My first thought was "oh, no, this shouldn't be in there. But on a second though this might be alright? Because a player, that is disconnected, is no spec.
> > Might cause problems with checks that are `if not ply:IsSpec` though
> 
> There may be more than one way to do this, let's see if you have any others.

I'd have fixed this where the issue occurs. In the lines where IsSpec is called.

Histalek:
> > > @saibotk @Histalek What is your opinion on this?
> > > My first thought was "oh, no, this shouldn't be in there. But on a second though this might be alright? Because a player, that is disconnected, is no spec.
> > > Might cause problems with checks that are `if not ply:IsSpec` though
> > 
> > 
> > There may be more than one way to do this, let's see if you have any others.
> 
> I'd have fixed this where the issue occurs. In the lines where IsSpec is called.

Yes that should be the way. Whoever calls this should make sure they do so on a valid entity. I'm against shifting the responsibility to the called function.

Histalek:
will be fixed by #1180

PR #1180, opened 2023-12-12
Fix #1171, undefined entity after a player disconnected

### LINKED ISSUE #1171: NULL Entity Client Error

TimGoll:
@EntranceJew added a spec check in #1157. This causes issues if the player entity is null that were previously ignored. I added a validity check here.

I also moved the Spec check into the existing function, so all HUDs (e.g. octagonal) have this fix as well

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed ttt_game_text not working due to a refactor
 - Fixed dete call HUD being invisible
 - Fixed edgecase where undefined killer angle or pos were accessed
-- Fixed a null entitiy error in the miniscoreboard
+- Fixed a null entity error in the miniscoreboard
 
 ## [v0.12.0b](https://github.com/TTT-2/TTT2/tree/v0.12.0b) (2023-12-11)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed ttt_game_text not working due to a refactor
 - Fixed dete call HUD being invisible
 - Fixed edgecase where undefined killer angle or pos were accessed
+- Fixed a null entitiy error in the miniscoreboard
 
 ## [v0.12.0b](https://github.com/TTT-2/TTT2/tree/v0.12.0b) (2023-12-11)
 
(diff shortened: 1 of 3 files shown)


TimGoll:
Due to @saibotks valid concerns, I changed it a bit. Should be safe now

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed ttt_game_text not working due to a refactor
 - Fixed dete call HUD being invisible
 - Fixed edgecase where undefined killer angle or pos were accessed
+- Fixed a null entitiy error in the miniscoreboard
```
Histalek, line 17:
```suggestion
- Fixed a null entity error in the miniscoreboard
```

Histalek (approved):

saibotk (approved):

PR #1194, opened 2023-12-12
Lua-Server: Fix hidden bugs

ZenBre4ker:
The lua Server mentioned many globals that were unused, but it seems that they are just misspelled variables in most cases.
THis is untested and is just a theoretical fix. So gotta look what could be wrong now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - DNA Tester works now with more than one fingerprint on a weapon
 - TraitorButton config files should now actually work
-- Map Setting Entities are now triggered on roundstart with an actual value of ROUND_ACTIVE
 
 ## [v0.12.1b](https://github.com/TTT-2/TTT2/tree/v0.12.1b) (2023-12-12)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Fixed
 
-- Fixed DNA Tester now actually adding new samples
+- DNA Tester works now with more than one fingerprint on a weapon
 - TraitorButton config files should now actually work
-- Entities on roundstart are now triggered with the actual value of ROUND_ACTIVE
+- Map Setting Entities are now triggered on roundstart with an actual value of ROUND_ACTIVE
 
 ## [v0.12.1b](https://github.com/TTT-2/TTT2/tree/v0.12.1b) (2023-12-12)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Fixed
+
+- Fixed DNA Tester now actually adding new samples
+- TraitorButton config files should now actually work
+- Entities on roundstart are now triggered with the actual value of ROUND_ACTIVE
+
 ## [v0.12.1b](https://github.com/TTT-2/TTT2/tree/v0.12.1b) (2023-12-12)
 
 ### Added
--- a/gamemodes/terrortown/gamemode/client/cl_voice.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_voice.lua
@@ -178,7 +178,7 @@ function GM:PlayerStartVoice(ply)
 	if not IsValid(g_VoicePanelList) or not IsValid(client) then return end
 
 	-- There'd be an extra one if voice_loopback is on, so remove it.
-	GAMEMODE:PlayerEndVoice(ply, true)
+	GM:PlayerEndVoice(ply, true)
 
 	-- Tell server this is global
 	if client == ply then
@@ -297,7 +297,7 @@ local function VoiceClean()
 	for ply, pnl in pairs(PlayerVoicePanels) do
 		if IsValid(pnl) and IsValid(ply) then continue end
 
-		GAMEMODE:PlayerEndVoice(ply)
+		GM:PlayerEndVoice(ply)
 	end
 end
 timer.Create("VoiceClean", 10, 0, VoiceClean)
--- a/gamemodes/terrortown/gamemode/client/cl_voice.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_voice.lua
@@ -178,7 +178,7 @@ function GM:PlayerStartVoice(ply)
 	if not IsValid(g_VoicePanelList) or not IsValid(client) then return end
 
 	-- There'd be an extra one if voice_loopback is on, so remove it.
-	GM:PlayerEndVoice(ply, true)
+	GAMEMODE:PlayerEndVoice(ply, true)
 
 	-- Tell server this is global
 	if client == ply then
@@ -297,7 +297,7 @@ local function VoiceClean()
 	for ply, pnl in pairs(PlayerVoicePanels) do
 		if IsValid(pnl) and IsValid(ply) then continue end
 
-		GM:PlayerEndVoice(ply)
+		GAMEMODE:PlayerEndVoice(ply)
 	end
 end
 timer.Create("VoiceClean", 10, 0, VoiceClean)
--- a/gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
@@ -515,32 +515,7 @@ addonChecker.curatedList = {
 		reason = "Less bottles, no integration in TTT2 systems, generally buggy.",
 		type = ADDON_OUTDATED
 	},
-	["284419411"] = { -- Minifier by Lykrast
-		alternative = "1896918348",
-		reason = "TTT2 integration, updated hitbox, updated camera position, etc.",
-		type = ADDON_OUTDATED
-	},
-	["1338887971"] = { -- Minifier by Hudson
-		alternative = "1896918348",
-		reason = "TTT2 integration, updated hitbox, updated camera position, etc.",
-		type = ADDON_OUTDATED
-	},
-	["1354031183"] = { -- Minifier by SnowSoulAngel
-		alternative = "1896918348",
-		reason = "TTT2 integration, updated hitbox, updated camera position, etc.",
-		type = ADDON_OUTDATED
-	},
 	["2672799157"] = { -- Minifier by The Stig
-		alternative = "1896918348",
-		reason = "TTT2 integration.",
-		type = ADDON_OUTDATED
-	},
-	["1551396306"] = { -- Minifier by FaBe2$tRonK
-		alternative = "1896918348",
-		reason = "TTT2 integration.",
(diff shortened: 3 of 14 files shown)


gamemodes/terrortown/gamemode/client/cl_voice.lua
```diff
@@ -178,7 +178,7 @@ function GM:PlayerStartVoice(ply)
 	if not IsValid(g_VoicePanelList) or not IsValid(client) then return end
 
 	-- There'd be an extra one if voice_loopback is on, so remove it.
-	GAMEMODE:PlayerEndVoice(ply, true)
```
TimGoll, line 181: This change breaks the gamemode

gamemodes/terrortown/gamemode/client/cl_voice.lua
```diff
@@ -297,7 +297,7 @@ local function VoiceClean()
 	for ply, pnl in pairs(PlayerVoicePanels) do
 		if IsValid(pnl) and IsValid(ply) then continue end
 
-		GAMEMODE:PlayerEndVoice(ply)
+		GM:PlayerEndVoice(ply)
```
TimGoll, line 300: This one as well

gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
TimGoll, line 1: Why did you remove addons here?

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -1210,7 +1210,7 @@ function BeginRound()
 	-- @realm server
 	hook.Run("TTTBeginRound")
 
-	ents.TTT.TriggerRoundStateOutputs(ROUND_BEGIN)
+	ents.TTT.TriggerRoundStateOutputs(ROUND_ACTIVE)
```
TimGoll, line 1213: Why?

gamemodes/terrortown/gamemode/shared/sh_door.lua
```diff
@@ -232,7 +232,7 @@ if SERVER then
 	function entmeta:SetDoorCanTouchOpen(state, surpressPair)
 		door.SetPlayerCanTouch(self, state)
 
-		self:SetNWBool("ttt2_door_player_touch", PlayerCanTouchDoor(self))
+		self:SetNWBool("ttt2_door_player_touch", door.PlayerCanTouch(self))
```
TimGoll, line 235: I have to check the code, but I'm pretty sure that change is wrong

TimGoll (changes requested):

gamemodes/terrortown/gamemode/server/sv_main.lua
ZenBre4ker: There is no ROUND_BEGIN defined

gamemodes/terrortown/gamemode/client/cl_voice.lua
ZenBre4ker: Why? There is no GAMEMODE:PlayerENdVoice, I checked.

gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
ZenBre4ker: They were duplicates

gamemodes/terrortown/gamemode/shared/sh_door.lua
ZenBre4ker: couldnt find anything better, but PLayerCanTouchDoor is not a function anywhere

gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
Histalek: Maybe we should order these by ID?

gamemodes/terrortown/gamemode/server/sv_main.lua
Histalek: `ROUND_BEGIN` was **never** defined (at least not in sh_init.lua)

gamemodes/terrortown/gamemode/client/cl_voice.lua
EntranceJew: image

ZenBre4ker: Thats GM:PlayerEndVoice for which I exchanged this.

gamemodes/terrortown/gamemode/client/cl_voice.lua
```diff
@@ -178,7 +178,7 @@ function GM:PlayerStartVoice(ply)
 	if not IsValid(g_VoicePanelList) or not IsValid(client) then return end
 
 	-- There'd be an extra one if voice_loopback is on, so remove it.
-	GAMEMODE:PlayerEndVoice(ply, true)
+	GM:PlayerEndVoice(ply, true)
```
ZenBre4ker, line 181: See one line below please.

gamemodes/terrortown/gamemode/client/cl_voice.lua
ZenBre4ker: WHile I dont get it, GAMEMODE is the actual working hook...... I reseted it

EntranceJew: GAMEMODE is defined at the first pass when the gamemode is loaded, GM is only available after that

ZenBre4ker: > GAMEMODE is defined at the first pass when the gamemode is loaded, GM is only available after that

Is this anywhere documented?

TimGoll: I'm pretty sure it is tge other was around. GM is defined on first pass, GAMEMODE, when the game is running

gamemodes/terrortown/gamemode/shared/sh_door.lua
TimGoll: I used git blame a bit and I guess you're right. At some point we moved these local functions into global functions.

old:
image

new:
image

ZenBre4ker: Good to know, I tested a lot and those functions are never called. They are not even referenced inside TTT2, might they have become useless?

gamemodes/terrortown/gamemode/server/sv_main.lua
TimGoll: You're right. These are the options:

```
ROUND_WAIT = 1
ROUND_PREP = 2
ROUND_ACTIVE = 3
ROUND_POST = 4
```

`ROUND_ACTIVE` should be the correct choice. This bug existed (and still exists(!)) in base TTT: 
https://github.com/Facepunch/garrysmod/blob/89b23437b5075e7efd5c379e74ccbf534679bcea/garrysmod/gamemodes/terrortown/gamemode/init.lua#L695

This means that the entity output is never triggered:
https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/terrortown/entities/entities/ttt_map_settings.lua#L59-L68

This fix might trigger some things on maps that have *never* been triggered. We should test this, even though it is a bugfix

gamemodes/terrortown/gamemode/shared/sh_door.lua
TimGoll: such as `door.PlayerCanTouch`? These are used for targetID by setting the NW bool on the server. They are also an external API that can be used by addons - if they want to. So I don't see why we shouldn't keep them.

ZenBre4ker: grafik
These 3 here are not called anywhere, might they be deprecated?

ZenBre4ker: I meant entmeta:SetDoorCanTouchOpen and CanUseOpen and AutoCloses

TimGoll: I don't think I intended on using them in TTT2, I added them as part of the API so addons can use this on entities. I don't know if any addon uses this, but I wouldn't say that it is deprecated.

I know for a fact that i use `entmeta:ToggleDoor` in one of my addons for example. As well as `entmeta:LockDoor`. I therefore exposed all settings to the API, instead of just those I needed at the time.

ZenBre4ker: > I don't think I intended on using them in TTT2, I added them as part of the API so addons can use this on entities. I don't know if any addon uses this, but I wouldn't say that it is deprecated.
> 
> I know for a fact that i use `entmeta:ToggleDoor` in one of my addons for example. As well as `entmeta:LockDoor`. I therefore exposed all settings to the API, instead of just those I needed at the time.

Yeah those I tested. Okay well, thanks for the input ❤️

TimGoll: > I meant entmeta:SetDoorCanTouchOpen and CanUseOpen and AutoCloses

It is probably not uses right now, yes. But because this is a general API for doors, I don't see how it is deprecated.

One could make an addon that is an admin tools to setup doors for example

TimGoll:
> I tested these changes on a single per-change basis. At least I didnt break anything. Should be good to go.

I'm mostly fine with your changes. But why did you remove items from the addon checker?

gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
ZenBre4ker: @TimGoll I removed items that were duplicated by steamID

TimGoll: I missed the replies hier. Then I'm fine with it

TimGoll (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Fixed
+
+- Fixed DNA Tester now actually adding new samples
+- TraitorButton config files should now actually work
+- Entities on roundstart are now triggered with the actual value of ROUND_ACTIVE
```
TimGoll, line 11: It's not about general map entities, the entity in question is the `ttt_map_settings` entity

TimGoll:
Still a bit unprecise. The map_settings entity (the actual entity on the map) was never triggered before because the check did not pass

ZenBre4ker:
> Still a bit unprecise. The map_settings entity (the actual entity on the map) was never triggered before because the check did not pass

ACtually I just saw the implementation, this change does nothing at all. If the state is nil it just gets the current round_state

TimGoll:
> > Still a bit unprecise. The map_settings entity (the actual entity on the map) was never triggered before because the check did not pass
> 
> ACtually I just saw the implementation, this change does nothing at all. If the state is nil it just gets the current round_state

Ahh, well, then it didn't really change anything I guess

saibotk (approved):
Looks good but you did not mention the other fixes in the changelog

ZenBre4ker:
yeah the rest are not really worth mentioning, some minor changes.

PR #1201, opened 2023-12-14
Beacon: Rework

TimGoll:
TTT has a disabled item called "Beacon". Its use is pretty limited and was therefore disabled. While working on the entity bringdown vol2 I got an idea on how to add some fun to this weapon:

The beacon is now a "live-radar" that only works in a small-ish sphere. It uses the marks module for this:
image
Once a player walks out of this sphere, the wallhack is removed again.

The owner of the beacon is also informed about two events:
- a player died next to their beacon
- one of their beacons was destroyed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Added
 
-- Added the beacon back into TTT2, a equipment weapon that was disabled long ago in base TTT
-  - A weapon that can only be bought by policing roles
-  - Adds a wallhack in a radius around them that is visible to anyone
+- Added the beacon back into TTT2, an equipment that was disabled long ago in base TTT
+  - Can only be bought by policing roles
+  - Creates a wallhack in a sphere around it, which is visible to everyone
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Added
+
+- Added the beacon back into TTT2, a equipment weapon that was disabled long ago in base TTT
+  - A weapon that can only be bought by policing roles
+  - Adds a wallhack in a radius around them that is visible to anyone
+
 ### Fixed
 
 - Fixed missing water level icon breaking scoreboard
--- a/gamemodes/terrortown/entities/entities/ttt_beacon.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_beacon.lua
@@ -16,6 +16,11 @@ ENT.Model = Model("models/props_lab/reciever01b.mdl")
 ENT.CanHavePrints = true
 ENT.CanUseKey = true
 
+local beaconDetectionRange = 150
+
+local soundZap = Sound("npc/assassin/ball_zap1.wav")
+local soundBeep = Sound("weapons/c4/c4_soundBeep1.wav")
+
 function ENT:Initialize()
 	self:SetModel(self.Model)
 
@@ -24,8 +29,8 @@ function ENT:Initialize()
 	end
 
 	self:SetMoveType(MOVETYPE_VPHYSICS)
-	self:SetSolid(SOLID_VPHYSICS)
-	self:SetCollisionGroup(COLLISION_GROUP_INTERACTIVE)
+	self:SetSolid(SOLID_BBOX)
+	self:SetCollisionGroup(COLLISION_GROUP_WEAPON)
 
 	if SERVER then
 		self:SetMaxHealth(100)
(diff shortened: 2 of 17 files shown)


TimGoll:
image

owners are now able to see the sphere of influence if they are in close proximity

a7f3:
Will the spheres be able to targets through walls?

TimGoll:
image
they render in the 3d world as you'd expect

TimGoll:
I've got two bots here. One in the sphere and one outside of it:
image

I'm only able to see one of them through the wall:
image

a7f3:
I meant like this, which it seems yes

TimGoll:
yes, this is my intention here. To create a locally limited wallhack

Spanospy:
Like a distant cousin of the decoy. Liking the idea!

Any chance for an option that the victim gets notified about them being detected, or able to see the owner as well whilst in the radius? (Could make this an equipment item as well)

TimGoll:
> Like a distant cousin of the decoy. Liking the idea!
> 
> Any chance for an option that the victim gets notified about them being detected, or able to see the owner as well whilst in the radius? (Could make this an equipment item as well)

Right now everybody can see everybody that is in the radius. I don't know if this is the correct decision though. And I'll keep the other suggestion in mind

TimGoll:
> Oh right, so its not just a personal pseudo-camera?
> 
> You could make it configurable by the owner whether the beacon is shared to everyone, himself or just their team (though I can see this being abused by the innocent team if they use it and ask a suspected traitor if they can see the wallhack)

I want to get an basic version out there first and then act on the feedback

a7f3:
What if it's only active while the person who put it down is holding an item form of it?

TimGoll:
> What if it's only active while the person who put it down is holding an item form of it?

maybe, but we should first playtest it, before adding balancing

TimGoll:
image

gamemodes/terrortown/entities/weapons/weapon_ttt_beacon.lua
```diff
@@ -0,0 +1,245 @@
+---
+-- @class SWEP
+-- @section weapon_ttt_beacon
+
+if SERVER then
+	AddCSLuaFile()
+end
+
+DEFINE_BASECLASS "weapon_tttbase"
```
Histalek, line 9:
```suggestion
DEFINE_BASECLASS("weapon_tttbase")
```

lua/terrortown/lang/en.lua
```diff
@@ -2073,3 +2073,18 @@ L.label_spec_prop_dash = "Dash force multiplier"
 L.label_keyhelper_possession_dash = "prop: dash in view direction"
 L.label_keyhelper_weapon_drop = "drop selected weapon if possible"
 L.label_keyhelper_ammo_drop = "drop ammo from selected weapon out of clip"
+
+-- 2023-12-14
+L.visualizer_help_pri = "Throw Beacon on the ground"
+L.visualizer_help_sec = "Stick Beacon to surface"
```
Histalek, line 2079:
```suggestion
L.beacon_help_pri = "Throw Beacon on the ground"
L.beacon_help_sec = "Stick Beacon to surface"
```

gamemodes/terrortown/entities/weapons/weapon_ttt_beacon.lua
```diff
@@ -0,0 +1,245 @@
...
+
+		self:PlacedBeacon()
+	end
+end
+
+if CLIENT then
+	---
+	-- @realm client
+	function SWEP:Initialize()
+		self:AddTTT2HUDHelp("visualizer_help_pri", "visualizer_help_sec")
```
Histalek, line 240:
```suggestion
		self:AddTTT2HUDHelp("beacon_help_pri", "beacon_help_sec")
```

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Added
+
+- Added the beacon back into TTT2, a equipment weapon that was disabled long ago in base TTT
+  - A weapon that can only be bought by policing roles
+  - Adds a wallhack in a radius around them that is visible to anyone
```
```suggestion
- Added the beacon back into TTT2, an equipment that was disabled long ago in base TTT
  - Can only be bought by policing roles
  - Creates a wallhack in a sphere around it, which is visible to everyone
```

Histalek (changes requested):

gamemodes/terrortown/entities/entities/ttt_beacon.lua
```diff
@@ -0,0 +1,222 @@
...
+		end
+
+		marks.Remove(self.lastPlysFound or {})
+		marks.Add(plysFound, roles.DETECTIVE.color)
+
+		self.lastPlysFound = plysFound
+	end
+
+	-- only sets the next think on the server
+	self:NextThink(CurTime() + 5)
```
saibotk, line 114: Doesn't this mean we effectively render people that just touch the range of the beacon for 5 seconds until they are removed again and not in the radius anymore?

How about adjusting this to be every tick instead and adding some kind of "this frame only"-function to the marks library?

gamemodes/terrortown/entities/weapons/weapon_ttt_beacon.lua
saibotk: I think this can be removed as the other entities also do not use this anymore.

But dont know if this actually changes anything. Would have to look that up

gamemodes/terrortown/entities/entities/ttt_beacon.lua
```diff
@@ -0,0 +1,222 @@
...
+	if not IsValid(activator) or self:GetOwner() ~= activator then return end
+
+	local wep = activator:GetWeapon("weapon_ttt_beacon")
+
+	if IsValid(wep) then
+		local pickup = wep:PickupBeacon()
+
+		if not pickup then return end
+	else
+		wep = activator:GiveEquipmentWeapon("weapon_ttt_beacon")
```
saibotk, line 61: We should make sure that this then only has 1 ammo in it, otherwise i would be able to pickup the last beacon and reset my ammo to place infinite beacons.

saibotk (changes requested):

gamemodes/terrortown/entities/entities/ttt_beacon.lua
TimGoll: no, as the command says this is only sets the next tick for the server. On the client it is updated every tick. It is a bit confusing and it took me a while to understand that part of the code as well. Therefore I added the comment

gamemodes/terrortown/entities/entities/ttt_beacon.lua
TimGoll: We can reduce this to one if you like, yes. But I don't get your exact problem here.

You have three beacons. every time you drop one, the ammo is reduced by one. If you drop all three, you have non left. If you pick up one again, then you have one back in your inventory. You can only plant three, not an infinite ammunt.

The marked line gives you a beacon with 1 ammo. It is only set to three in the `WasBought` callback: https://github.com/TTT-2/TTT2/blob/0cb5472df06884a48f005a7e0978a4d8bb01c91c/gamemodes/terrortown/entities/weapons/weapon_ttt_beacon.lua#L105-L107

gamemodes/terrortown/entities/weapons/weapon_ttt_beacon.lua
TimGoll: Right now all our entities and weapons do this:

https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/entities/weapons/weapon_ttt_wtester.lua#L12

Histalek: my main concern is twofold:
	- it's not used how the wiki tells us to e.g. `DEFINE_BASECLASS( "my_weapon" )` [1]
	- it breaks github's syntax highlighting (it doesn't break treesitter nor vim's default highlighting; so maybe this is just a github problem. Still not ideal)

if we can eliminate this usage completely that would be fine by me as well.

[1] https://wiki.facepunch.com/gmod/Global.DEFINE_BASECLASS

saibotk: This should be safe to remove, since we are not using it in here

TimGoll: ok, but we still have this line in all (or most) other entities. Maybe someone else can remove them then

TimGoll:
Note: I kept bomb Vision and this PR two seperate PRs. As soon as both are merged, I will overhaul the beacon rendering a bit, right now it is pretty bare bones

saibotk (approved):

TimGoll:
if @Histalek or @ZenBre4ker can approve this, I can integrate the new bomb vision into this

Histalek (approved):

PR #1203, opened 2023-12-14
MarkerVision: A Bomb Radar Rework

TimGoll:
Added a new markerVision system that is similar to targetID. It handles wallhack and UI. It supports teamchange as well

https://www.youtube.com/watch?v=hdo0waT3cOw


TimGoll:
I probably have to handle players being added to / removed from the team as well

Histalek (changes requested):
Not a full review yet, IMHO this should be a draft.

i wanted to Veto calling this `bombvision` or `radarvision`.

Neither of them properly reflect what exactly this is and it is confusing as hell. Even more so as both names are still used throughout the PR.

Naming a library after a builtin item (that does not even use this) is beyond confusing.

The name for this i know from other games would be along the lines of "POI hints" (POI = Point of Interest) maybe we can find a better descriptive word even, but naming this after a specific item/equipment is out of the question for me.

gamemodes/terrortown/entities/entities/ttt_radio.lua
```diff
@@ -326,6 +331,28 @@ if CLIENT then
...
+
+		rData:EnableText()
+
+		rData:SetTitle(TryT(ent.PrintName))
+		rData:AddIcon(materialRadio)
+
+		rData:AddDescriptionLine(ParT("bombvision_owner", {owner = nick}))
+		rData:AddDescriptionLine(ParT("bombvision_distance", {distance = distance}))
+
+		rData:SetCollapsedLine(ParT("bombvision_collapsed", {distance = distance}))
```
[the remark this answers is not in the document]
TimGoll: yes, so that the owner sees where their radio is

TimGoll:
> Not a full review yet, IMHO this should be a draft.

I don't get why this should be a draft as I'm more or less done with this

Histalek:
> > Not a full review yet, IMHO this should be a draft.
> 
> I don't get why this should be a draft as I'm **more or less done** with this

"Even more so as both names are still used throughout the PR." -> the PR itself does not refer to its introduced library in a consistent way which seems to be because of a recent change which in turn means it's still actively worked on -> should be a draft

EntranceJew:
fwiw i think POI / InterestPoint would be a decent name for it

Spanospy:
> i wanted to Veto calling this `bombvision` or `radarvision`.
> 
> Neither of them properly reflect what exactly this is and it is confusing as hell. Even more so as both names are still used throughout the PR.

Seconding. While I think radarvision is a better name than bombvision, it can still be confused with the radar item that is completely unrelated to this HUD feature.

HUD Markers might be a better name?

TimGoll:
`markerVision` maybe?

TimGoll:
> > > Not a full review yet, IMHO this should be a draft.
> > 
> > 
> > I don't get why this should be a draft as I'm **more or less done** with this
> 
> "Even more so as both names are still used throughout the PR." -> the PR itself does not refer to its introduced library in a consistent way which seems to be because of a recent change which in turn means it's still actively worked on -> should be a draft

But it was finished from my point of view. I missed the name problem but that's what a review is for. Besides a fitting name, this PR is done from my side, here's a short demonstration video:
https://www.youtube.com/watch?v=hdo0waT3cOw

TimGoll:
I kept the net message `TTT_C4Warn` in there because I'm not sure if other addons use it. Maybe @EntranceJew has some insight as he talked about other addons using the old TTT radar

gamemodes/terrortown/entities/entities/ttt_radio.lua
```diff
@@ -286,7 +289,7 @@ local nextplay = 0
 ---
 -- @realm shared
 function ENT:Think()
-	if CurTime() <= nextplay or #self.SoundQueue <= 0 then return end
+	if CurTime() <= nextplay or (istable(self.SoundQueue) and #self.SoundQueue <= 0) then return end
```
TimGoll, line 292: not sure if this change is needed. At some point I had an issue while hotreloading that was fixed with this change

Histalek:
> image

i have to admit this screenshot confused me really bad. It just looked like a second sidebar of status-icons.
Speaking of status icons, how do they interact with this? do they clash in some way?

regarding the name: `markerVision` is definitely better, but my brain is reading `vision` and is thinking of stuff that affects every pixel of what a player sees e.g. nightvision or heatvision.

Would be nice to find an alternative name, but that's not a blocker. Just wanted to throw it out there.

TimGoll:
> > 
> 
> i have to admit this screenshot confused me really bad. It just looked like a second sidebar of status-icons. Speaking of status icons, how do they interact with this? do they clash in some way?
> 
> regarding the name: `markerVision` is definitely better, but my brain is reading `vision` and is thinking of stuff that affects every pixel of what a player sees e.g. nightvision or heatvision.
> 
> Would be nice to find an alternative name, but that's not a blocker. Just wanted to throw it out there.

Ah, this was just to show how it looks off-screen.

Regarding the name: I'm open to suggestions. It should be something that works as a module name. So `POI` won't work

ZenBre4ker:
So what does it do?
It is associated with an entity, enables icons and description. Handles events like being offscreen or behind walls. And renders them or not?

TimGoll:
> So what does it do? It is associated with an entity, enables icons and description. Handles events like being offscreen or behind walls. And renders them or not?

Best would be to take a look at the linked video, but to give you a summary:

- you register an entity on the server or client that should be tracked
  - if you register it on the server, it is automatically synced to everyone specified by the type
  - if you register it on the client, it is only visible to the one player
 - it adds a wallhack with the marks module
 - it calls a hook that is used (in the same style as targetID) to add an icon, text etc

ZenBre4ker:
EntityTrackerData?

TimGoll:
> EntityTrackerData?

Is entityTracker really more fitting than markerVision?

EntranceJew:
> I kept the net message `TTT_C4Warn` in there because I'm not sure if other addons use it. Maybe @EntranceJew has some insight as he talked about other addons using the old TTT radar

AFAICT, anything that still uses TTT_C4Warn would have a plain bomb indicator image, even the ones that want to pipe in extra data all reimplement it. I haven't seen anything use the C4Warn but I think the worst that could happen is something doesn't redeclare a netmessage string, but that should only make it easier to smoke out things that needed to be updated or marked as incompatible anyway, because something that isn't C4 acting like C4 is bad information to work off of.

EntranceJew:
> Regarding the name: I'm open to suggestions. It should be something that works as a module name. So `POI` won't work

`traitor_sense`, `ESP`, `spatial_notifications`

food for thought

TimGoll:
@EntranceJew noticed a few minor issues with this that I will polish

TimGoll:
Applied review changes from @EntranceJew

TimGoll:
Is there any input on this PR?

Histalek:
> Is there any input on this PR?

Sorry i'll probably review it this weekend.

Until then i'll try one last time to propose another name. If that doesn't find support i'll just shut up about it.

Firstly i'll try to explain why i don't particularly like `markerVision` as a name:
- `Vision` imo implies something that affects the whole vision of a player (like `nightvision`, `thermalvision` etc.) which this certainly does not do. This simply places pointers (or markers :D )
- `marker` i would have liked this as a standalone name, but we also have a `Marker` role and this would just lead to unnecessary confusion.

My proposal therefore would be either **waypoint** or **signpost**.

- `waypoint` is rather straightforward i think, a bit plain but maybe that's for the better
- `signpost` a bit more corny, but would convey that this isn't only a simple pointer but can have more information (like distance etc.) attached

NovaDiablox:
How about ("entesp" = entity + esp)?

TimGoll:
> How about ("entesp" = entity + esp)?

We don't want to use abbreviations, but thanks for the suggestion!

saibotk:
What about `entityMarks` or just `marks`?

TimGoll:
> What about `entityMarks` or just `marks`?

We already have a marks module

ZenBre4ker:
> > What about `entityMarks` or just `marks`?
> 
> We already have a marks module

Maybe integrate it into the marks module?
The marks module currently only optimizes the halo library. Maybe it has some common ground?

ZenBre4ker (approved):
Beside the name, the code looks fine. I also dont have a better idea. 😄

TimGoll:
> > > What about `entityMarks` or just `marks`?
> > 
> > 
> > We already have a marks module
> 
> Maybe integrate it into the marks module? The marks module currently only optimizes the halo library. Maybe it has some common ground?

nah, this has nothing to to with the marks module. The marks module just adds a colored stencil that is rendered on top, this adds networking, role change tracking, death tracking, icons, text and much more. Imho, they are not related

gamemodes/terrortown/gamemode/client/cl_hud_manager.lua
```diff
@@ -51,20 +57,20 @@ function GM:HUDPaint()
 		keyhelp.Draw()
 	end
 
-	if not client:Alive() or client:Team() == TEAM_SPEC then return end
-
 	---
 	-- @realm client
-	if hook.Run("HUDShouldDraw", "TTTRadar") then
-		RADAR:Draw(client)
+	if hook.Run("HUDShouldDraw", "TTTBombVision") then
```
Histalek, line 62: is this for the C4? if so it should be named accordingly

lua/ttt2/libraries/marker_vision.lua
```diff
@@ -0,0 +1,402 @@
...
+		-- note: when VISIBLE_FOR_ALL receiver is nil, therefore SendStream uses net.Boradcast
+		net.SendStream("ttt2_register_entity_added", streamTable, receiverList or receiver)
+	end
+
+	if CLIENT then
+		if visibleFor == VISIBLE_FOR_ALL then
+			color = color or TEAMS[TEAM_INNOCENT].color
+		else
+			color = color or TEAMS[LocalPlayer():GetTeam()].color
+		end
```
Histalek, line 67: We do have translation strings for all 3 cases, so we should reflect these in color as well.

would the `TEAM_NONE.color` work for the `VISIBLE_FOR_PLAYER` case, wdyt?

Histalek (changes requested):
a few doc suggestions and a small nit. Otherwise looks good

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
saibotk, line 1: Is the Draw function with the timer here still needed now that this has the timer in targetID?

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
```diff
@@ -476,29 +474,10 @@ end
 -- Timer configuration handling
 
 if SERVER then
-	---
-	-- Inform traitors about us
-	-- @param boolean armed
-	-- @realm server
-	function ENT:SendWarn(armed)
-		net.Start("TTT_C4Warn")
```
saibotk, line 484: Remove this from sv_main
Also remove the net receive in cl_init

We could discuss if we keep this for compatibility but i do not think anyone is abusing this for displaying their own bombs in the radar currently

saibotk (changes requested):

lua/ttt2/extensions/util.lua
```diff
@@ -532,6 +533,15 @@ function util.EditingModeActive(ply)
 	return (HUDEditor and HUDEditor.IsEditing) or entspawnscript.IsEditing(ply)
 end
 
+---
+-- Converts hammer units to meters.
+-- @param number value The value in hammer units
+-- @return number The converted value in meters
+-- @realm shared
+function util.HammerUnitsToMeters(value)
```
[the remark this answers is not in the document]
TimGoll: [according to wikipedia](https://en.wikipedia.org/wiki/List_of_unusual_units_of_measurement), a hammer unit is 3/4 of an inch

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
TimGoll: are you sure? aren't addons such as the turtlenade using this?

lua/ttt2/libraries/marker_vision.lua
TimGoll: I think that would be confusing, wouldn't it? If something is visible only to you, then you'd expect it in your team color I think.

Also this would still clash with the unknown role etc

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
saibotk: Im not sure, thats why we should discuss this

lua/ttt2/libraries/marker_vision.lua
Histalek: no it's the other way around IMO. If i see something in my team's color i expect my whole team to see it.

Tbh the 'clash' with the unknown role is non-existent as the unknown doesn't have a team in gameplay terms and therefore does not care if a marker is only for them or for their "Team".

TimGoll: I see your point now, thanks for raising the issue

lua/ttt2/libraries/marker_vision.lua
```diff
@@ -0,0 +1,402 @@
...
+		marks.Remove({ent})
+	end
+end
+
+---
+-- Updates the entity on the server and all related clients.
+-- @param Entity ent The entity that should be updated
+-- @param Player owner The owner of the wallhack that takes their ownershipo with them on team change
+-- @param[opt] table receiverList A list of players that should receive the netmessage, overwrites the default
+-- @note Call on server to add entity on server and all defined clients.
```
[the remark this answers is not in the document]
TimGoll: It should be "update", not "add". But why remove it entirely?

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
TimGoll: You didn't mention a line, so I guess this one?
image

If so, yes. This renders the time onto the model itself. It populates the display

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
TimGoll: as far as I can tell other addons add their own custom syncing, but use the old TTT rendering: 
image

lua/ttt2/libraries/marker_vision.lua
```diff
@@ -0,0 +1,404 @@
+---
+-- A module that let's devs create bomb visions for any entity that is automatically synced
```
Histalek, line 2:
```suggestion
-- A module that let's devs create marker visions for any entity that is automatically synced
```

TimGoll:
Everything should be done @Histalek and @saibotk

lua/ttt2/libraries/marker_vision.lua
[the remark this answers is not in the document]
Histalek: whoops my bad. it's fine now

Histalek (approved):

saibotk (approved):

PR #1204, opened 2023-12-14
Removed restriction that only team can pickup Radio

### LINKED ISSUE #1125: Equipment/Radio: Traitors cannot pick up innocent radios?
- Workshop version from Steam

Radio functions as portable tester, because Ts cannot pick up one a detective dropped for others.

## To reproduce

1. drop radio as inno
2. attempt pick up radio as T

## Expected behaviour

T picks up a radio as normal

a7f3:
Possibility of having this as a convar, in the small cases where this is a wanted feature?

Haven't been able to get `-multirun` launch variable working to be able to test this in game, so if someone could test would be great.

Would close #1125

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Fixed
 
+- Removed the Restriction that only teammates can pickup a radio
 - Fixed missing water level icon breaking scoreboard
 - DNA Tester works now with more than one fingerprint on a weapon
 - TraitorButton config files should now actually work
--- a/gamemodes/terrortown/entities/entities/ttt_radio.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_radio.lua
@@ -57,7 +57,7 @@ end
 -- @param Player activator
 -- @realm shared
 function ENT:UseOverride(activator)
-	if IsValid(activator) and activator:IsPlayer() and activator:IsInTeam(self:GetOwner()) then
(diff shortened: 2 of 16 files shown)


xopez:
You should also cleanup the language vars since they are don't used anywhere else:
image

xopez:
I can confirm that this fixes it:
https://youtu.be/9DFrVb2G7Nc

Histalek:
i'm somewhat torn on this. while this would solve the issue of #1125 this is also quite a change in behaviour.

With this change Radios can now be 'stolen' from T's as well. Normally a T will already try to hide a radio so it's not that big of a deal, but still worth mentioning.

Personally i would be fine with this change. I would appreciate more opinions on this though.

Maybe we even want to put this under the `Changed` section in the changelog. Thoughts?

TimGoll:
I had the same thoughts as you had. This might be something that is worth a convar actually

TimGoll:
That's a good compromise. The old owner could also be notified if someone from a different Team stellt their Radio.

Or what about only the deployer being able to pick it up again? That way no one can be tested

TimGoll:
> > Or what about only the deployer being able to pick it up again? That way no one can be tested
> 
> That's probably the best way of doing this

C4, Beacon etc also already work that way

TimGoll:
okay @a7f3 here's what you should do:

- make sure only the owner can pick it up
- add a generic string to the language file that says "only the owner is able to pick up this entity" that is displayed as targetID subtitle if the focusing player is not the owner
- replace the key in that case with a radio icon, feel free to hit me up for an icon

Issue #1095, opened 2023-10-24
[Feature] Rework Radio


TimGoll:
- Remove the radio UI from the shop
- add a custom menu with a custom keybind
- the keybind should be listed in the equipment section if a player owns a radio

@EntranceJew had a few feature ideas that could be added here as well

EntranceJew:
Yes, the major thing being:

- items become able to register their main functionality sound effects, so that other weapons (such as the radio) can intentionally invoke a noise that is the same as it
- concommand ttt_radio plays using the format `ttt_radio c4beep [n]` instead of using the arbitrary entity ID (a netmessage workaround hack atm), where `[n]` is a radio in the player or team's control, but defaults to their first owned one

TimGoll:
RE point two: Should we add a keybind for this? Something like "play last sound"? Then we can remove the command that is annoying

EntranceJew:
repeat last sound bind would be a worthwhile, but the ability to quickly tell a "story" with radio commands IMO such as "footsteps => death scream" is still the priority.
on a custom menu, if it were a radial then "repeat last command" would probably be the center option.
![image](https://github.com/TTT-2/TTT2/assets/5711436/943a8640-751b-4a3e-befc-9bcd04142395)
horrible diagram warranted

there should still be the annoying concommand, to call the sounds individually, just change the structure of it, since it's being abused, and people with other radial bind menus might like to manually remap them the way I do with the radio callouts

TimGoll:
This was done in #1437 

We might add a radial menu in the future, but the main task of this issue is done

PR #1206, opened 2023-12-15
Equipment Menu: command for opening a specific tab

Spanospy:
Mainly throwing this pull request to get some feedback/opinions on adding this.

My use case is that I want my addon to be able to display a tab it adds, for example when a user presses a keybind for it. For now I just have it open the tab as soon as it's brought: [link](https://github.com/Spanospy/ttt2-item_beacon_hacker/blob/2c38a74c62646fd67bc4e79c3ff0283214e50d42/lua/terrortown/entities/items/item_beacon_hacker.lua#L107)

The way this is implemented also allows users to make their own bindings to quickly open the transfer, radio or disguise tabs, if they so wish.

I am aware that this half-a-decade old equipment menu is probably due for a overhaul, even if just UI wise. Still, I feel like the idea of this functionality could be useful to have.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Added
+
+- Added `ttt_cl_traitorpopup_tab` to allow players or addons to open a specified tab in the Equipment Menu (by @Spanospy):
+  - Example: `ttt_cl_traitorpopup_tab "Transfer"` for opening the credit transfer selection tab.
+
 ### Fixed
 
 - Binoculars scan no longer gets interrupted when changing zoom level
--- a/gamemodes/terrortown/gamemode/client/cl_equip.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_equip.lua
@@ -957,6 +957,45 @@ end
 concommand.Add("ttt_cl_traitorpopup_close", ForceCloseTraitorMenu)
 
 --
+-- Opens the menu to a specified tab, if it exists.
+--
+
+local function SwitchTraitorMenuTab(ply, cmd, args)
+
+	--Open the menu if it's not already open
+	if not IsValid(eqframe) then
+		TraitorMenuPopup()
+	end
+
+	if IsValid(eqframe) then
+
+		---@cast eqframe -nil
+		local dsheet = eqframe:Find("DPropertySheet") --Or eqframe:GetChild(4) ?
+
+		local tabs = dsheet:GetItems()
+		local tabindex = nil
+
+		local targettab = args[1]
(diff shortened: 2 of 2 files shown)


TimGoll:
First of, yes, a Shop UI and backend overhaul is in the works right now!

Regarding your additions. Why did you chose commands for this? 

I understand your reasoning that some equipment items want a possibility to directly open a tab, but this would be also possible with a function call.

If you want bindings, I'd like to vote against commands as we do not want to make the user use the console.

I'd appreciate an API that allows you to open tabs (or whatever it is that we will have after the rework)

That way you can

- open the menu with function calls in addons
- create an addon that adds these as bindings to the bindings menu

Spanospy:
Makes sense. I'll be honest, the way I did it was just the quickest and easiest as I based it off of how the disguiser does bindings.

I'll see if I can make sense of your proposed way of doing it when I get home from holiday gatherings!

EntranceJew (changes requested):
I think the concommand implementation works out great, 

https://github.com/TTT-2/TTT2/assets/5711436/08c5e643-b69a-438e-aa8d-7d1971f44523

The only thing I would do is if it is invoked while it's already open on that tab, then it should perhaps instead close the window so when used in this radial, it's easy to close the menu without having to use a separate keypress (pressing the bound buy menu button will close any page)

Spanospy:
> If you want bindings, I'd like to vote against commands as we do not want to make the user use the console.
> 
> I'd appreciate an API that allows you to open tabs (or whatever it is that we will have after the rework)
> 
> That way you can
> 
> * open the menu with function calls in addons
> * create an addon that adds these as bindings to the bindings menu

Do you have anything I can refer to as a "good example" of what you're thinking?

Spanospy:
> First of, yes, a Shop UI and backend overhaul is in the works right now!

Are you referring to https://github.com/TTT-2/TTT2/pull/988 ?

Spanospy:
Also noticed a posted issue that implies it's not desired for the equipment / traitor menu to have equipment-specific tabs: https://github.com/TTT-2/TTT2/issues/1095

With this in mind, I'm not sure how useful it is to have this functionality if only two tabs should live in that menu.

TimGoll:
> Also noticed a posted issue that implies it's not desired for the equipment / traitor menu to have equipment-specific tabs: https://github.com/TTT-2/TTT2/issues/1095
> 
> With this in mind, I'm not sure how useful it is to have this functionality if only two tabs should live in that menu.

Yes, I'm not a big fan of those tabs. But the radio still needs an UI. The new UI can still be opened with your binding. We have to find a good solution for it

ZenBre4ker:
I think we can close this, as we dont want to use the concommands for that and are working on a better solution for the tabs in general.

Issue #1150, opened 2023-12-06
Ammo: Simplify ammo system


TimGoll:
Right now there are five kinds of ammo in the game:
![image](https://github.com/TTT-2/TTT2/assets/13639408/5da6b4de-6104-48c4-9450-832445e98d83)

Reducing these to three doesn't remove any tactics from the game, but improves visual distinctiveness between those (when using 2, 4, 5). It also simplifies many things such as distinct icons for each type. It also makes it clear to players that they need ammo X for pistol-like weapons for example.

This means we'd reduce it to three types:
- shotgun ammo
- pistol/9mm ammo
- rifle ammo

there might be more descriptive names for these three types,

Related to this change is a rebalance of the spawning, as there are now fewer spawns than before and some are unused. There are already addons that tackle this where we could take a look at: https://steamcommunity.com/sharedfiles/filedetails/?id=2990353959

This might also be slightly related to #1107

EntranceJew:
## Clustered Ammo / Weapon Spawns
Mentioned in #1210, also have a way to get a clear point near ammo spawns to spawn a circle of ammo / weapons, or even sodas.. Configurable by the spawn editor?

EntranceJew:
I was looking through the backlog and spotted some overlap between this and #1004

PR #1210, opened 2023-12-18
hooks for amending spawn positions, because all the data was localed

EntranceJew:
example usage:
```lua
local test_spawns = {
	weapon = {
		["neo_ghostspawnpoint"] = WEAPON_TYPE_SPECIAL,
		["neo_ghost_retrieval_point"] = WEAPON_TYPE_SPECIAL,
	},
	player = {
		["info_player_attacker"] = PLAYER_TYPE_RANDOM,
		["info_player_defender"] = PLAYER_TYPE_RANDOM,
	}
}

hook.Add("TTT2MapRegisterWeaponSpawns", "EJ_NEOTOKYO_TTT2MapRegisterWeaponSpawns", function(spawns)
	for k, v in pairs(test_spawns.weapon) do
		spawns[k] = v
	end
end)

hook.Add("TTT2MapRegisterPlayerSpawns", "EJ_NEOTOKYO_TTT2MapRegisterPlayerSpawns", function(spawns)
	for k, v in pairs(test_spawns.player) do
		spawns[k] = v
	end
end)

hook.Add("Initialize", "EJ_NEOTOKYO_Initialize", function()
	for cat, data in pairs(test_spawns) do
		for ent_name, ent_type in pairs(data) do
			local stored = scripted_ents.GetStored(ent_name)
			if stored == nil then
				scriped_ents.Register({
					Type = "point",
					Base = "base_point",
				}, ent_name)
			end
		end
	end
end)
```


EntranceJew:
Because it's a situation where we have to dummy those entities if they don't exist, theoretically someone might want to add more appropriate point entities as an addon, and then they'd be getting dummied by the basegame.

the problem with my example of neotokyo is based a bit out of the problems i [referenced here](https://github.com/TTT-2/TTT2/issues/1150#issuecomment-1843698552):
- random special weapon spawns still behave strangely with random weapon spawns
- there aren't enough weapon or ammo spawn points (within neotokyo maps) to claim to support those maps natively, they'd still need hand-placed spawn points. there's only like six spawn points that aren't player related, and they're as spread out as CS:S' bomb sites.
- _unless_ we had an entity that functioned like a "cluster" spawn that spread out weapons in an area like theStig's floorspawners or older TTT1 solutions tended to have

So it still wouldn't be super ideal to try and support it out of the box because the experience would be nearly unplayable IMO.

If I were to ignore all that, then what I'd probably prefer to do is:
1. guard DummifyFallbackWeaponEnts to check if a scripted_ent is Stored first to prevent unnecessarily overwriting entities.
2. extend the Dummyify to also apply to ammo/players.
3. restructure the local tables onto a modifiable structure on the `map` library.
4. leave just one hook to augment `map` with all the info to support a game.

TimGoll:
> older TTT1 solutions tended to have

We also have that for players and I'd really like somthing like that for outher spawns as well so I don't have to hack the soda spawns.

But I'm fine with those hooks, they don't degrade the experience and you seem to have a valid use case for them. I just wanted to understand your point

EntranceJew:
i believe i have, now

lua/ttt2/libraries/map.lua
```diff
@@ -269,6 +269,12 @@ function map.GetWeaponSpawnEntities()
 
 	FindSpawnEntities(spawns, ttt_weapon_spawns)
 	FindSpawnEntities(spawns, hl2_weapon_spawns)
+	local hook_weapon_spawns = {}
```
TimGoll, line 272:
```suggestion

	local hook_weapon_spawns = {}
```

lua/ttt2/libraries/map.lua
```diff
@@ -288,6 +294,12 @@ function map.GetAmmoSpawnEntities()
 
 	FindSpawnEntities(spawns, ttt_ammo_spawns)
 	FindSpawnEntities(spawns, hl2_ammo_spawns)
+	local hook_ammo_spawns = {}
```
TimGoll, line 297:
```suggestion

	local hook_ammo_spawns = {}
```

lua/ttt2/libraries/map.lua
```diff
@@ -299,6 +311,12 @@ end
 function map.GetPlayerSpawnEntities()
 	local spawns = {}
 
+	local hook_player_spawns = {}
```
TimGoll, line 314:
```suggestion
	local spawns = {}
	local hook_player_spawns = {}
```

lua/ttt2/libraries/map.lua
```diff
@@ -363,7 +391,10 @@ function map.GetSpawnsFromClassTable(spawns)
 		end
 
 		-- next check if it is a weapon spawn
-		local wepType = ttt_weapon_spawns[cls] or hl2_weapon_spawns[cls] or css_weapon_spawns[cls] or tf2_weapon_spawns[cls]
+		local hook_weapon_spawns = {}
+		-- @realm shared
+		hook.Run("TTT2MapRegisterWeaponSpawns", hook_weapon_spawns)
+		local wepType = ttt_weapon_spawns[cls] or hl2_weapon_spawns[cls] or css_weapon_spawns[cls] or tf2_weapon_spawns[cls] or hook_weapon_spawns[cls]
```
TimGoll, line 397:
```suggestion
		hook.Run("TTT2MapRegisterWeaponSpawns", hook_weapon_spawns)

		local wepType = ttt_weapon_spawns[cls] or hl2_weapon_spawns[cls] or css_weapon_spawns[cls] or tf2_weapon_spawns[cls] or hook_weapon_spawns[cls]
```

lua/ttt2/libraries/map.lua
```diff
@@ -424,14 +455,56 @@ function map.GetDataFromSpawnEntity(ent, spawnType)
 	}
 
 	if spawnType == SPAWN_TYPE_WEAPON then
-		return ttt_weapon_spawns[cls] or hl2_weapon_spawns[cls] or css_weapon_spawns[cls] or tf2_weapon_spawns[cls], data
+		local hook_weapon_spawns = {}
+		-- @realm shared
+		hook.Run("TTT2MapRegisterWeaponSpawns", hook_weapon_spawns)
+		return ttt_weapon_spawns[cls] or hl2_weapon_spawns[cls] or css_weapon_spawns[cls] or tf2_weapon_spawns[cls] or hook_weapon_spawns[cls], data
```
TimGoll, line 461:
```suggestion
		hook.Run("TTT2MapRegisterWeaponSpawns", hook_weapon_spawns)

		return ttt_weapon_spawns[cls] or hl2_weapon_spawns[cls] or css_weapon_spawns[cls] or tf2_weapon_spawns[cls] or hook_weapon_spawns[cls], data
```

TimGoll (approved):

TimGoll:
approved, but please commit these requests

EntranceJew:
@saibotk @Histalek

Histalek (approved):

PR #1216, opened 2023-12-18
body_confirm spam go away

### LINKED ISSUE #903: Option in a Convar to allow Body identification messages to appear before "Confirmed Death" lists
**Is your feature request related to a problem? Please describe.**
I currently run TTT2 on a public server, and a somewhat common issue is the ordering of UI messages when a dead body is identified. It'll display the "Confirmed the death of" list before displaying the identified body's name and role. This has caused problems when people stumble onto the aftermath of a heated gunfight, and instead of being able to quickly get information on what body is what, they have to sit through a (sometimes 3 second long) list of "confirmed the death of" messages before being told the identity and roles of bodies. This has resulted in people getting killed for "Not iding a body", or for assuming they were even evil though they did ID the body, due to the large amount of time before their role is confirmed.

**Describe the solution you'd like**
A good solution would be to have a server option to decide which messages take priority, the "Confirmed Deaths" list or the body name + role popup. I know some people like the way it is currently set up, as it works fine with smaller player counts as "confirmed death" lists never get lenghty.

**Describe alternatives you've considered**
A secondary solution could be to simply speed up the messages, but that seems less ideal than a convar option to swap the order around as it could cause readability issues. 

**Additional context**
It feels weird that by default the body information popup is considered lesser priority than the confirmed the death of popups, as the body identifity popup is the more urgently important information, while the "Confirmed death" information is usually less urgent information.

EntranceJew:
it is noisy up there

also fixes #903, not a convar because behavior is no longer optional, ez clap


gamemodes/terrortown/gamemode/server/sv_corpse.lua
```diff
@@ -181,6 +161,51 @@ function CORPSE.IdentifyBody(ply, rag, searchUID)
...
+			--local vicrag = vic.server_ragdoll
+			--CORPSE.SetFound(vicrag, true)
+		end
+
+		if #killnicks == 1 then
+			LANG.Msg("body_confirm_one", {finder = finder, victim = killnicks[1]})
+		elseif #killnicks > 1 then
+			if #killnicks >= 3 then
+				table.sort(killnicks, CorpseSorting)
+			end
```
saibotk, line 194:
```suggestion
			table.sort(killnicks, CorpseSorting)
```
Can we also inline the sorting function here, since we do not need it elsewhere and it technically does sort nicknames and not corpses

gamemodes/terrortown/gamemode/server/sv_corpse.lua
```diff
@@ -181,6 +161,51 @@ function CORPSE.IdentifyBody(ply, rag, searchUID)
...
+
+			local names = ""
+			for k = 1, #killnicks do
+				local name = killnicks[k]
+				if name == ply:Nick() then continue end
+
+				names = names .. name .. ", "
+			end
+
+			names = string.sub(names, 1, -3)
```
saibotk, line 204:
```suggestion
			local names = killnicks[1]
			for k = 2, #killnicks do
				names = names .. "," .. killnicks[k]
			end
```

saibotk (changes requested):

gamemodes/terrortown/gamemode/server/sv_corpse.lua
EntranceJew: redeclaring a function within the loop scope would hurt perf pretty unnecessarily

saibotk: there is no loop and this is negligible

saibotk: Please still inline the function

EntranceJew:
image
image
changes eliminate spacing, and these changes should probably be propagated to where the traitor announcement is made

saibotk:
yes just forgot the space, please add it back

Histalek (approved):

saibotk (approved):

TimGoll:
It is already merged, but shouldn't it be `"{finder} confirmed the {count} death(s) of: {victims}."`?

EntranceJew:
It uses the pluralization pattern of `_one` or `_more` so it should always be correctly pluralized for the singular using the old existing string

TimGoll:
Ahh, well than this works!

PR #1217, opened 2023-12-18
edit equipment accent stock equipment (fixes #964)

### LINKED ISSUE #964: Make vanilla weapons obvious in the equipment editor
example of how this can be confusing:

vanilla weapons in the editor could be named like "_MAC10 (Built-in)_" or have a little icon or border that signifies it

EntranceJew:
image
tooltip also, for admins


gamemodes/terrortown/entities/weapons/weapon_ttt_binoculars.lua
```diff
@@ -46,6 +46,7 @@ SWEP.Secondary.Delay = 0.2
 SWEP.Kind = WEAPON_EQUIP2
 SWEP.CanBuy = {ROLE_DETECTIVE} -- only detectives can buy
 SWEP.WeaponID = AMMO_BINOCULARS
+SWEP.StockEquipment = true
```
saibotk, line 49: first of, i'd rename this to `builtin`

saibotk (changes requested):
But also i am not sure, if an outline is any good solution for this. Because no one knows the meaning of the outline and its in no way obvious to the average user. 
We might need to think about a more easy way maybe same symbol with a tooltip or like a grouping in the list for this?

EntranceJew:
the yellow `(c)` icon exists for custom equipment, but penalizing everything except for the builtin stuff with an ugly badge seems like a negative way to do things; is there an icon for `builtin`?

gamemodes/terrortown/entities/weapons/weapon_ttt_binoculars.lua
TimGoll: I agree with this point because we already use this flag for other purposes, such as roles: https://github.com/TTT-2/TTT2/blob/master/lua/terrortown/entities/roles/innocent/shared.lua#L15

TimGoll:
Could we add a blue? (or accent color colored) b in a circle (similar to the C icon) in-line in front of the name?

And then in the editor is a new form on top that only contains a helpbox that says that this weapon in builtin. That way admins leran the meaning of the icon

EntranceJew:
image
this is an example of an 8px badge like the way the custom equipment menu is set, we could add a b instead i just don't have art for that

TimGoll:
I can create the art for you, but I'd put it in-line in front of the name so it is more visible. Don't you think?

EntranceJew:
I think it'd be a little disorienting to have a puck offsetting some of the text when I'm scanning the list alphabetically, unless they were all aligned further to the right to leave space for that

EntranceJew:
image
this is 16x16, i can't overhang it because the submenu clips the overhang if i tried to put it perfectly over the bottom right corner

Histalek:
Can we document `.builtin` somehow? e.g. that it's not meant to be set by other addons? otherwise i'm afraid that it will just be copied or misused to highlight other addons. which would lead to this change  being somewhat useless.

EntranceJew:
I don't know where we'd document it, I specifically excluded it from the weapon bases themselves.

TimGoll (changes requested):
please add the material to the force download list and add an empty line to the end of the vmt file

TimGoll:
Additionally I wonder if builtin weapons should have an extra form on top that explains that they are part of the TTT2 base. That way the admins can learn the meaning of the badge

EntranceJew:
image
i remember last time i added a disclaimer banner ...

TimGoll:
> image i remember last time i added a disclaimer banner ...

I guess you should wait for more feedback then. I'm most of the time a fan of adding more help text, but not everyone is

materials/vgui/ttt/vskin/markers/builtin.vmt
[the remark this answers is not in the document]
EntranceJew: image
no, the existing shop image is blurry, and applies to the exact opposite selection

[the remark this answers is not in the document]
saibotk: Oh only saw the screenshot above and quite liked the idea of them having the same icon. This would let users recognize the icon much more. I don't like to introduce so many different things on how we explain stuff via UI elements (e.g. different icons) 
@TimGoll what do you think?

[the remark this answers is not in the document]
TimGoll: It was my idea to use this icon here and ditch the old one in the show rework. As most addons are not builtin anyway, I don't see the point of marking them. Also this builtin icon can be used in the role settings menu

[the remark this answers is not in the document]
saibotk: Then we should use the icon in the shop too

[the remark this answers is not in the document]
TimGoll: yes, but Sven just started with the shop rework, so I think it should be part of the rework instead of doing the work twice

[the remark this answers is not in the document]
EntranceJew: for reference, the display of the `(c)` shop icon is optional and can be disabled. we'd need probably a higher res version of the c icon if we were showing them side by side.

TimGoll:
> Additionally I wonder if builtin weapons should have an extra form on top that explains that they are part of the TTT2 base. That way the admins can learn the meaning of the badge

@EntranceJew This got two upvotes so I guess it is a good idea. Besides this, your PR is fine for me. I quite like the changes introduced. Good work!

materials/vgui/ttt/vskin/markers/builtin.vmt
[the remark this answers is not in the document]
TimGoll: Adding to that: For most users it doesn't matter anyway. So I don't see the benefit of the extra work here. If @EntranceJew wants to integrate it into the old shop it is fine for me, but I think their time is used more efficient on other projects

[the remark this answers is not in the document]
saibotk: I just think the shop rework will still take quite some time and would like users to learn what this icon is by using it in the same way.

So id like this to also do:
- Use the icon in the role settings
- Remove the C icon entirely
- Use the new icon in the shop instead based on the same logic as in the equipment editor

EntranceJew:
image
image
i think this fits the criteria @TimGoll and @saibotk specified

TimGoll (approved):
LGTM

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Added a new `fastutf8` library that provides faster utf8 functions (added by @saibotk, created by @blitmap)
 - Added new hooks: `TTT2MapRegisterWeaponSpawns`, `TTT2MapRegisterAmmoSpawns`, `TTT2MapRegisterPlayerSpawns` to allow converting a wider variety of source map ports (by @EntranceJew)
+- Added tooltip to `Edit Equipment` menu with the equipment's class name and outline equipment with `builtin` set (by @EntranceJew)
```
saibotk, line 43:
```suggestion
- Added tooltip to `Edit Equipment` menu with the equipment's class name and icon for equipment with `builtin` set (by @EntranceJew)
```
Maybe also mention the icon changes in the role menu and shop

lua/terrortown/lang/zh_tw.lua
```diff
@@ -2090,5 +2090,11 @@ L.search_eyes = "透過你的探查技能，你確信他臨死前見到的最後
 --L.entity_pickup_owner_only = "Only the owner can pick this up"
 
 -- 2023-12-18
-L.body_confirm = "{finder} 確認了 {victim} 的死。"
+--L.body_confirm_one = "{finder} confirmed the death of {victim}."
```
saibotk, line 2093: This should stay and only be renamed

gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
```diff
@@ -342,34 +355,43 @@ function SKIN:PaintSubMenuButtonTTT2(panel, w, h)
...
 	end
 
 	drawBox(0, 0, sizes.border, h, colorBar)
 	drawBox(sizes.border, 0, w - sizes.border, h, colorBackground)
 
 	if hasIcon then
-		drawFilteredShadowedTexture(pad + sizes.border, padIcon + shift, sizeIcon, sizeIcon, panel:GetIcon(), iconAlpha, colorText)
+		drawFilteredShadowedTexture(pad + sizes.border, padIcon + shift, sizeIcon, sizeIcon, panel:GetIcon(), iconAlpha, colorIcon)
+		if iconBadge then
+			local miniSize = nearestPowerOf2(sizeIcon / 4)
```
saibotk, line 392: I think this is nice but overengineered. When the icon size is not a power of 2 it is also blurry and if it is a power of 2 so also will the mini version. So i dont think we need to ensure this just for the badge. This should be something the UI author has to take care of

gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -412,14 +412,15 @@ local function CreateEquipmentList(t)
 			if item.iconMaterial then
 				ic = vgui.Create("LayeredIcon", dlist)
 
-				if item.custom and showCustomVar:GetBool() then
+				if item.builtin and showCustomVar:GetBool() then
 					-- Custom marker icon
 					local marker = vgui.Create("DImage")
-					marker:SetImage("vgui/ttt/custom_marker")
```
saibotk, line 418: Can we remove this asset or is this referenced elsewhere?

saibotk (changes requested):

gamemodes/terrortown/gamemode/client/cl_equip.lua
EntranceJew: I think other things that might interact with BEM may still expect us to provide it, but, I don't know what they are, because they're not in this codebase.

lua/terrortown/lang/zh_tw.lua
EntranceJew: i had a fear that this would happen in the middle of one of them, corrected now

gamemodes/terrortown/gamemode/client/cl_equip.lua
TimGoll: Fair point, I think there is no harm in leaving the icon in there.

Isn't it a vanilla TTT icon anyway?

saibotk (approved):

PR #1219, opened 2023-12-18
make tooltips wider (fixes #1011)

### LINKED ISSUE #1011: [Round Report] Not enough space for localization
Version: v0.11.6b - **Unreleased - 2022-01-04**
Round report screen has some tooltips that are very limited in space:
```
L.tooltip_karma_gained = "Karma changes for this round:"
L.tooltip_score_gained = "Score changes for this round:"
L.tooltip_roles_time = "Role changes for this round:"
```
Desktop_Screenshot_2022 12 26_-_00 15 10 14

In the Russian language, there is not enough width to fit all possible texts, especially when it comes to role changes. So, please increase the width of these tooltips (for about 15% or 20%).

EntranceJew:
image
unfortunately these functions are so far removed from any possible GetTextSize implementation that it's virtually impossible to be accurate to any greater degree

and if we were more accurate, then we should make this tooltip functionality more generic or at least make a derma version


TimGoll:
we plan on doing a UI rework that stops using translations on every frame and reverts back to translating on creation / when the UI is updated. That way stuff like this should be easier to fix

lua/terrortown/menus/score/info.lua
```diff
@@ -29,11 +29,16 @@ local function MakePlayerRoleTooltip(parent, width, ply)
...
-
+		local length = fastutf8.len(plyRoleBox.GetTitle())
+		if length > shortest then
+			shortest = length
+		end
 		height = height + 20
 	end
 
-	return height
+	width = (shortest * 8)
```
TimGoll, line 39: brackets are not needed here

lua/terrortown/menus/score/info.lua
```diff
@@ -29,11 +29,16 @@ local function MakePlayerRoleTooltip(parent, width, ply)
 		plyRoleBox.GetTitle = function()
 			return tostring(i) .. ". " .. TryT(roleData.name) .. " (" .. TryT(plyRole.team) .. ")"
 		end
-
+		local length = fastutf8.len(plyRoleBox.GetTitle())
+		if length > shortest then
```
TimGoll, line 33: should it really be shortest then? not longest?

Also probably somehing like `lengthLongestLine`

lua/terrortown/menus/score/info.lua
TimGoll: also where is the magic 8 coming from?

TimGoll (changes requested):

lua/terrortown/menus/score/info.lua
TimGoll: This happens three more times, maybe even put it in a function?

EntranceJew: getting the font for finding the string length is too far out since the paint function controls its presentation and not dlabel stuff, so it's ballparking it

the three functions are three functions, i don't know how much i can really combine them though

EntranceJew: actually, found a way that i might be able to deduplicate and work toward the stated goal of unraveling things at the same time

EntranceJew:
@Histalek can you double check https://github.com/TTT-2/TTT2/pull/1219/commits/668ed96345df20186648db457bc49ba77447c6dd 's [test runs](https://github.com/TTT-2/TTT2/actions/runs/7256905937) and see that it was the best thing to do there

vscodium doesn't have glualint so i forgot about it

lua/terrortown/menus/score/info.lua
```diff
@@ -3,128 +3,62 @@
...
+	-- add some padding to the right margin
+	lengthLongestLine = lengthLongestLine + iconOffset
 
-		plyRoleBox.GetTitle = function()
-			return "- " .. TryT(karmaText) .. ": " .. karma
-		end
-
-		height = height + 20
+	-- we found the longest line, now make all entries that long
+	for _, v in ipairs( boxLayout:GetChildren() ) do
```
TimGoll, line 57: please do not use ipairs, `for i = 1, #xxx do` is much faster

TimGoll (changes requested):
Slight nitpick, but looks overall pretty solid

lua/terrortown/menus/score/info.lua
EntranceJew: idk if it is still the case with the GetChild / ChildCount overhead but i never benched it

TimGoll: you could do:

```lua
local children = boxLayout:GetChildren()

for i = 1, #children do
    local child = children[i]

end
```

This is how we do it based on this: https://springrts.com/wiki/Lua_Performance

EntranceJew: image
unless i bricked the measure of this script, including # in for structure measures worse in glua than it does in 5.1 that that site measures (it doesn't mention luajit anywhere either, which has mildly better perf of ipairs in general)

saibotk: Please use @TimGoll's approach, as it is close to our code style. 
Also in your test, ipairs seems slower, so i don't get what you are trying to say in your comment. Even though this is really niche and not noticable anyways.

(also regarding your latest commits, please use it the way Tim described above, because extra function calls in the loop to GetChild is more of a performance overhead)

EntranceJew: image
ipairs is faster across longer consecutive iterations is what i mean, the performance profile is different. idk that we ever have a 10000 item long table, but it is still a characteristic difference

TimGoll (approved):
LGTM

EntranceJew:
@Histalek @saibotk

TimGoll:
> @Histalek @saibotk

at least tobi is busy right now, so you don't have to ping them

EntranceJew:
i can't assign or request reviewers normally,

is there a preferred way to distinguish something that's ready-to-go vs a clotted one that is in the list here still?

saibotk:
You can usually rerequest reviews but not if the request is still pending. 
And you should not have to ping us because we all tend to take good care of our notification inbox. It mostly is just that we are lacking time to review all of the changes but we know that it is still to do.
So please only ping for urgency reasons.

EntranceJew:
image
the UX for requesting reviewers (or even self-assigning) is unavailable to me

saibotk:
Yes that is okay, i don't really see an issue with that. We are mostly all subscribed to these prs anyways and get a notification for them. 

Also i meant to ask you to please not merge main into your branches so often. Because i get the PR as a notification and always think something has actually changed and need to take a look.
You might want to stop using git town for this. As the tool is also rather intended for stacking prs

CHANGELOG.md
```diff
@@ -77,6 +78,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed missing bodysearch information if victim was killed without leaving a trace caused by a weapon hit
 - Fixed "body_confirm" MSTACK noise by batching all the kills from a body into one message. (by @EntranceJew)
 - Fixed "body_confirm" message sending before corpse confirmation message.
+- Fixed round scoreboard tooltips not being wide enough for their strings. (by @EntranceJew)
```
Histalek, line 82: this needs to be in the `unreleased` section

lua/ttt2/libraries/database.lua
```diff
@@ -332,7 +332,7 @@ function database.RemoveChangeCallback(accessName, itemName, key, identifier)
 	end
 
 	for i = #callbacks, 1, -1  do
-		callback = callbacks[i]
+		local callback = callbacks[i]
```
Histalek: this is already fixed in master is it not? why does github show this in the diff? :confused:

Histalek (changes requested):
please disregard my second comment ...

lua/ttt2/libraries/database.lua
TimGoll: I was confused as well, I even checked the master branch and this makes no sense to me

EntranceJew: idk why this is showing up this way, but locally the callback is assigned as local

CHANGELOG.md
EntranceJew: ✔ automatic merge must have missed, fixed

Histalek: i don't see it fixed yet.

Might just be my github dying idk

EntranceJew: turns out making a commit works best after staging files 🙃

CHANGELOG.md
```diff
@@ -62,6 +62,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed corpses not listing their kills
 - Comboboxes now show correct values for database driven entries
 - Database-Callbacks are now called with the correct valuetype
+- Fixed round scoreboard tooltips not being wide enough for their strings (by @EntranceJew)
```
Histalek, line 65: still wrong section :D

Histalek (approved):

Issue #1071, opened 2023-10-10
[UI] Combobox sensitive to value type


ZenBre4ker:
As the Combox doesnt work if it is used with other values than strings, there should be a check.

This doesnt work:
![grafik](https://github.com/TTT-2/TTT2/assets/72046466/fe84e968-b8dd-4add-9cf1-0cca0a1df381)

This does:
![grafik](https://github.com/TTT-2/TTT2/assets/72046466/b5c6b614-20f3-4b8f-9c79-fb6046f4f3f2)

Note to self:
Check why value has to be given as string.
Use a tostring( ) to make sure its a string.

EntranceJew:
I feel bad that you had to retrace this, because I discovered I had found this out at some point prior -- but I wound up not even utilizing the combobox and in the end and I thought I was using it incorrectly to begin with. 

I *believe* the issue has more to do with the OnSelect* callback, more-over any of the internals themselves.

AaronMcKenney:
For better or worse, most of my roles make use of integer enums, so this one is pretty important to me.

NickCloudAT:
@ZenBre4ker have you already looked at this? I just did and found a "tostring" for "choice.value" in the following line will fix it:
https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua#L322C7-L322C7

If you want I can open a PR for this.

ZenBre4ker:
Sounds good 👍

NickCloudAT:
This can be closed now

PR #1222, opened 2023-12-19
VGUI: Fix comboboxes by converting strings and numbers

ZenBre4ker:
The fix #1122 for #1071 was generally working but denied the use of numbers.

This fix now fully integrates numbers into the combobox and makes sure, that Callbacks are called with the correct value type.
Main problem was, that Convars are always strings, but with the addtion of the database, which **should** output correct types, the combox wasnt working.

As addition, numbers work now as defaults, therefore, we finally can reset all comboboxes again.

Note: The database itself still needs a security conversion in case any other addon uses the wrong type inside for setting values.
This is done in another PR.


gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
```diff
@@ -322,10 +322,10 @@ function PANEL:MakeComboBox(data)
 			local choice = data.choices[i]
 
 			if istable(choice) then
-				right:AddChoice(choice.title, tostring(choice.value), choice.select, choice.icon, choice.data)
+				right:AddChoice(choice.title, choice.value, choice.select, choice.icon, choice.data)
 			else
 				-- Support old simple structure
-				choice = tostring(choice)
+				choice = choice
```
saibotk, line 328: this seems wrong :D

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
```diff
@@ -56,10 +58,37 @@ function PANEL:Clear()
...
 	self:CloseMenu()
 end
 
+---
+-- Stores the type of value that is currently used
+-- @param any value the value to store the type of
+-- @realm client
+function PANEL:StoreValueType(value)
+	-- Save value type of first entry for conversion
+	if self.isValueTypeString == nil then
```
saibotk, line 72: I don't think such a function should only allow storing the type once.

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
```diff
@@ -37,6 +37,8 @@ function PANEL:Init()
 	self.choices = {}
 	self.titleIndices = {}
 	self.valueIndices = {}
+
+	self.isValueTypeString = nil
```
saibotk, line 41: just name this valueType and write a string like "string", "number" "undefined" or something into it.

saibotk (changes requested):

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
TimGoll: Use these type definitions: https://wiki.facepunch.com/gmod/Global.type

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
ZenBre4ker: But in the combobox case its the only thing that makes sense.
The first value that is stored determines the type until the combobox is cleared again.

A combobox should only support one type anyways.

TimGoll (approved):
LGTM. Please discuss the one open issue with @saibotk but everything else is good imho

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcombobox_ttt2.lua
saibotk: I don't agree here, because i think this should be able to handle every combination of types in different entries but i get your reasoning and why this is a fix. So im fine for now

saibotk (approved):

PR #1224, opened 2023-12-19
Debug: Add quotation marks to strings in print statements

ZenBre4ker:
While working on #1222 and #1223 I realized how hard it is to find a problem with valuetypes, that are string.

This should allow to faster see when a value is actually a string and not another valuetype.

If this is useful is still outstanding for discussion.
Helped me to find the problems more easily.


TimGoll (approved):
but add an empty line at the file end

lua/ttt2/extensions/debug.lua
```diff
@@ -0,0 +1,42 @@
...
+
+---
+-- Adds quotation marks to strings otherwise just converts them to strings
+-- @param any object The object to convert
+-- @realm shared
+-- @internal
+local function ConvertToString(object)
+	if isstring(object) then
+		return "\"" .. object .. "\""
+	else
```
Histalek, line 20:
```suggestion
	else
```

Histalek (approved):
Can't hurt having this :)

For the future a `debug.type()` function that can actually distinguish between a `Color` and a `table` might be another possible addition.

ZenBre4ker:
> Can't hurt having this :)
> 
> For the future a `debug.type()` function that can actually distinguish between a `Color` and a `table` might be another possible addition.

Most definitely. Not having to.think about types would be nice, I dislike that you have to convert nil values or use PrintTable for tables. Maybe even allow functions, if possible.
As this will only be used during debugging sessions nobody cares about speed here.

PR #1228, opened 2023-12-19
make the c4 actually suggest the right button (_really_ fixes #1161)

### LINKED ISSUE #1161: C4 hover tooltip wrong when holding a Defuser
- Latest GitHub version

While holding the Defuser over a C4 the hover tooltip says "Press [E] to use defuser" but you actually have to Primary click on the C4 to use the Defuser

## To reproduce
Hold Defuser pointing at C4

## Expected behaviour
Change the "[E]" to be whatever Primary click is.

EntranceJew:
i hope
image


TimGoll (approved):

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - DNA Tester works now with more than one fingerprint on a weapon
 - TraitorButton config files should now actually work
 - Translation strings not rendering on detective's body search mode combobox
+- C4 defusal prompt suggesting the wrong key
```
ZenBre4ker, line 25: We should do our changelog in the same way.
Either we say:
Fixed:
-"Whats wasnt working"
or
-"How it should now behave correctly"

Here the first one is used.
In my opinion the second should be better.
Therefore I'd suggest changing it to
```suggestion
- C4 defusal prompt now suggesting the right key
```

ZenBre4ker (approved):
I still approve, but would be interest about your opinions @TimGoll @saibotk @Histalek

CHANGELOG.md
TimGoll: Since added and changed show the new state, this should probably be handled the same

Histalek: having this be consistent would be nice, but i don't care too much about this. From context it is pretty clear what was done here:

* The old version would read: 'Fixed C4 defusal prompt suggesting the wrong key'
* your version would read: 'As of this change (the) C4 defusal prompt (is) now suggesting the right key'

Histalek (approved):

CHANGELOG.md
Histalek: However i do like Tim's approach. Simple and to the point.

PR #1231, opened 2023-12-20
ttt_game_text mode 4 correction (fixes #972)

### LINKED ISSUE #972: ttt_game_text does not work at all when set to "All except traitors"

Latest GitHub version

`GetTeamChatFilter(TEAM_INNOCENT)` will always return an empty table, and so do `GetTeamFilter(TEAM_INNOCENT)` and `GetInnocentFilter()`

because ttt_game_text uses it, this breaks the "All except traitors" option
https://github.com/TTT-2/TTT2/blob/b18611f3b4fa33c258b011de60d12e5141464fb9/gamemodes/terrortown/entities/entities/ttt_game_text.lua#L66

## To reproduce

load into a map with a ttt_game_text entity that is set to send a message to "All except traitors"

no message will be displayed to non-traitors by the ttt_game_text entity

## Expected behaviour

the message should be displayed to non-traitors, which is what i would assume should be expected by the phrase "All except traitors"

also, the `GetInnocentFilter()` function should just give a list of every player on the innocent team, it doesn't seem to make sense that it does all these extra checks when the function just means "give me every innocent player"
https://github.com/TTT-2/TTT2/blob/d76aa1b313d3766ef279cba252a233317e948e26/gamemodes/terrortown/gamemode/server/sv_gamemsg.lua#L141
it should just check for `p:GetTeam() == TEAM_INNOCENT and (not alive_only or p:IsTerror())` and that's it

EntranceJew:
logically, it doesn't address innocents

it has always referred to everyone but traitors

refer to:
https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/terrortown/entities/entities/ttt_game_text.lua#L48-L49

https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/gamemsg.lua#L83-L85

https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/player_ext_shd.lua#L14-L17

### DIFF
--- a/gamemodes/terrortown/entities/entities/ttt_game_text.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_game_text.lua
@@ -66,6 +66,9 @@ function ENT:AcceptInput(name, activator)
 	elseif inputReceiver == RECEIVE_TRAITOR then
 		messageReceiver = GetTeamChatFilter(TEAM_TRAITOR)
 	elseif inputReceiver == RECEIVE_INNOCENT then
+		-- TTT originally defined this as "All except traitors" even though it is labeled as "RECEIVE_INNOCENT",
+		-- but the implementation literally only checked that a player was not a traitor, therefore the intent is
+		-- preserved here since maps aren't likely to be updated
 		messageReceiver = GetPlayerFilter(function(p)
 			local plyRoleData = ply:GetSubRoleData()
 			return p:GetTeam() ~= TEAM_TRAITOR and not plyRoleData.disabledTeamChatRecv
--- a/gamemodes/terrortown/entities/entities/ttt_game_text.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_game_text.lua
@@ -66,7 +66,10 @@ function ENT:AcceptInput(name, activator)
 	elseif inputReceiver == RECEIVE_TRAITOR then
 		messageReceiver = GetTeamChatFilter(TEAM_TRAITOR)
 	elseif inputReceiver == RECEIVE_INNOCENT then
-		messageReceiver = GetTeamChatFilter(TEAM_INNOCENT)
+		messageReceiver = GetPlayerFilter(function(p)
+			local plyRoleData = ply:GetSubRoleData()
+			return p:GetTeam() ~= TEAM_TRAITOR and not plyRoleData.disabledTeamChatRecv
+		end)
 	elseif inputReceiver == RECEIVE_ACTIVATOR then
 		if not IsValid(activator) or not activator:IsPlayer() then
 			ErrorNoHalt("ttt_game_text tried to show message to invalid !activator\n")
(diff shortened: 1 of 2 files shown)


EntranceJew:
fixes #972

TimGoll:
I tend to accept this. But do we want to use this as "everything but traitor" instead of "team innocent"? I see your reasoning behind this, I'm just not sure if this is the best decision

gamemodes/terrortown/entities/entities/ttt_game_text.lua
```diff
@@ -66,7 +66,10 @@ function ENT:AcceptInput(name, activator)
 	elseif inputReceiver == RECEIVE_TRAITOR then
 		messageReceiver = GetTeamChatFilter(TEAM_TRAITOR)
 	elseif inputReceiver == RECEIVE_INNOCENT then
-		messageReceiver = GetTeamChatFilter(TEAM_INNOCENT)
+		messageReceiver = GetPlayerFilter(function(p)
+			local plyRoleData = ply:GetSubRoleData()
```
saibotk, line 70: Can you please add a small comment here explaining that reasoning? So it clears up confusion about why this does not only filter innocents

saibotk (changes requested):

ZenBre4ker:
Im also unclear about what it does, because generally nothing should be innocent only, as they could identify themselves with that. So where is this used?

EntranceJew:
TEAM_INNOCENT covers detectives and innocents, it exists solely to withold info from traitors and only traitors

I think preserving what is practical and observed makes the most sense, IDK what solution TTTCR or other offshoots had. I definitely think authors should have the ability to target just the innocents, and they can, with "ROLE_INNOCENT" as a string (if the fgd is to be believed)

TimGoll:
> TEAM_INNOCENT covers detectives and innocents, it exists solely to withold info from traitors and only traitors
> 
> I think preserving what is practical and observed makes the most sense, IDK what solution TTTCR or other offshoots had. I definitely think authors should have the ability to target just the innocents, and they can, with "ROLE_INNOCENT" as a string (if the fgd is to be believed)

This is how CR solved it

1. https://github.com/NoxxFlame/TTT-Custom-Roles/blob/23f3363aeb0b0cd8d2c695ee16ce8cde81bcae83/gamemodes/terrortown/entities/entities/ttt_game_text.lua#L52-L53

2. https://github.com/NoxxFlame/TTT-Custom-Roles/blob/23f3363aeb0b0cd8d2c695ee16ce8cde81bcae83/gamemodes/terrortown/gamemode/gamemsg.lua#L126

They have a good solution in case mappers update their maps to support the other teams. We know that this won't happen. Therefore I think the solution by @EntranceJew is probably the best

saibotk (approved):
Thx 👍

TimGoll (approved):

Issue #1230, opened 2023-12-19
Can't open bodysearch


ChrisBkt:
# Version of TTT2
Steam Workshop, Version from 12.12.2023
"Update: 12. Dec. at 12:54"

# Console Error
[TTT2 (Base) - v0.12.1b] lua/ttt2/libraries/bodysearch.lua:271: Tried to use a NULL CTakeDamageInfo!
  1. IsBulletDamage - [C]:-1
   2. AssimilateSceneData - lua/ttt2/libraries/bodysearch.lua:271
    3. ShowSearch - gamemodes/terrortown/gamemode/server/sv_corpse.lua:207
     4. unknown - gamemodes/terrortown/gamemode/server/sv_player.lua:544
      5. unknown - lua/includes/modules/concommand.lua:54
      
# Info
The error is caused when you kill someone with a weapon and try to search the body with "E"
When your cause of death is Fall Damage, there's no error and the body search will open

# Workshop List
https://steamcommunity.com/sharedfiles/filedetails/?id=2682937664

TimGoll:
Do you know which weapon was used? Because with the weapons I tested this wasn't a problem

ChrisBkt:
Actually currently every weapon on our server.
Just if you don't die with a weapon, you don't get this error.

ChrisBkt:
Could you try it with the Crescent Rose?
https://steamcommunity.com/sharedfiles/filedetails/?id=684592258

There's also the error

And also every weapon we tried with the upper Workshop list

ChrisBkt:
The only addons but these from the workshop is the Pointshop2 with it's libs

TimGoll:
I will try to see if I can reproduce this. Because I tested it with so many different weapons without a problem

TimGoll:
Ahhh, all weapons that overwrite stuff like the crowbar are likely to cause problems. Could you try to reproduce the issue if you remove that addon from the server?

ChrisBkt:
Still the same problem with the default crowbar

TimGoll:
Hmm, weird. I will take a look at the weapons in your collection tomorrow

ChrisBkt:
Okay.
Also tried once again killing someone with crowbar and there's also the error. So it's also with the default weapons.

EntranceJew:
The default crowbar raises this problem after the addon is removed, and the server and clients restart? If the problem is specifically an overwritten file the server needs to refresh without that.

TimGoll:
[...]

PR #1232, opened 2023-12-20
fix invalid dmginfo being invalid (mitigates #1230)

EntranceJew:
i don't like it but it works

image

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_corpse.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_corpse.lua
@@ -2,6 +2,22 @@
 -- Corpse functions
 -- @module CORPSE
 
+---@class DamageInfoData
+---@field ammoType number The ammo type which was used, referring to game.GetAmmoTypes()
+---@field attacker Entity The attacker (character who originated the attack), for example a player or an NPC that shot the weapon. Or an Entity.
+---@field baseDamage number The initial unmodified by skill level ( game.GetSkillLevel() ) damage
+---@field damage number The total damage.
+---@field damageBonus number The amount of bonus damage.
+---@field damageCustom number Custom damage type. This is used by Day of Defeat: Source and Team Fortress 2 for extended damage info, but isn't used in Garry's Mod by default.
+---@field damageForce Vector The force taken from the damage, sometimes used for knockback.
+---@field damagePosition Vector The position where the damage was or is going to be applied to.
+---@field damageType number A bitflag which indicates the damage type(s) of the damage.
+---@field inflictor Entity The inflictor of the damage, a projectile, a weapon, or an ordinariy Entity.
+---@field reportedPosition Vector The initial, unmodified position where the damage occured
+---@field isBulletDamage boolean Whether the DamageInfo contained DMG_BULLET
+---@field isExplosionDamage boolean Whether the DamageInfo contained DMG_BLAST
+---@field isFallDamage boolean Whether the DamageInfo contained DMG_FALL
+
 -- namespaced because we have no ragdoll metatable
 CORPSE = {}
 
@@ -318,7 +334,14 @@ local function GetSceneDataFromPlayer(ply)
 	return data
 end
 
-local function CloneDamageInfo(dmginfo)
+---
+
+---
+-- Clones a CTakeDamageInfo into a table called DamageInfoData
+---@param CTakeDamageInfo dmginfo
+-- @return table the DamageInfoData
+-- @realm server
+local function CreateDamageInfoData(dmginfo)
 	return {
 		ammoType = dmginfo:GetAmmoType(),
 		attacker = dmginfo:GetAttacker(),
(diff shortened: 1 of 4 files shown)


saibotk:
Is this the only occurrence of the issue in the codebase?

EntranceJew:
the only one that's come up in my bot testing, it's hard to get past 'cause not being able to confirm a body deadlocks their AI

saibotk:
Yes but i meant codewise did you search the codebase for DmgInfo usages?

EntranceJew:
i searched the codebase for references, and none of them appear to have kept it for outside the lifecycle of the hook (which is the main problem) -- i can't promise it is 100% thorough but it does seem like it's isolated to just this

saibotk (approved):

gamemodes/terrortown/gamemode/server/sv_corpse.lua
```diff
@@ -318,10 +318,30 @@ local function GetSceneDataFromPlayer(ply)
...
+		damageType = dmginfo:GetDamageType(),
+		inflictor = dmginfo:GetInflictor(),
+		maxDamage = dmginfo:GetMaxDamage(),
+		reportedPosition = dmginfo:GetReportedPosition(),
+		isBulletDamage = dmginfo:IsBulletDamage(),
+		isExplosionDamage = dmginfo:IsExplosionDamage(),
+		isFallDamage = dmginfo:IsFallDamage(),
+	}
+end
+
```
ZenBre4ker, line 340: Is it possible to keep the dmgInfo function structure, so we dont have to revert other code than this cloning?
Like
isBulletDamage = dmgInfo:IsBulletDamage()
IsBulletDamage = function (dmgInfo) return dmgInfo.isBulletDamage end

lua/ttt2/libraries/bodysearch.lua
```diff
@@ -268,7 +268,7 @@ net.Receive("ttt2_client_reports_corpse", function(_, ply)
 		end
 
 		sceneData.killOrientation = CORPSE_KILL_NO_DATA
-		if rag.scene.hit_trace and isangle(rag.scene.hit_trace.StartAng) and rag.scene.dmginfo:IsBulletDamage() then
+		if rag.scene.hit_trace and isangle(rag.scene.hit_trace.StartAng) and rag.scene.dmginfo.isBulletDamage then
```
ZenBre4ker, line 271: So this can stay as it is?

ZenBre4ker (commented):
Just an idea, so we dont forget to properly revert it in the future, if we ever need to.

gamemodes/terrortown/gamemode/server/sv_corpse.lua
EntranceJew: i thought about it, but it's a lot of extra work -- IsBulletDamage is the same as `dmginfo:GetDamageType() == DMG_BULLET`, same with the rest of the `Is` functions

ZenBre4ker: Yeah I get that, but what I mean is, store our own table of dmginfo with private variables and then keep the functions referencing those cloned states.

TimGoll: if you do not implement @ZenBre4ker s suggestion, you shuld probably rename the gmginfo table to something like `damageInfoData` or so because otherwise people might think this is a real damageinfo object. Which it isn't

EntranceJew: this makes sense, most of the existing consumers of the .dmginfo are probably still going to break with the properties being different so no sense in keeping it as dmginfo

ZenBre4ker:
Well, if this is really the only use case of dmgInfo, I'd say just copy IsBulletDamage directly into scene data and leave the cloning out of it. I dont like that we copy only those entries. Either we fully clone DamageInfo or not.

gamemodes/terrortown/gamemode/server/sv_corpse.lua
```diff
@@ -318,10 +334,37 @@ local function GetSceneDataFromPlayer(ply)
...
 end
 
+---
+
+---
+-- Clones a CTakeDamageInfo into a table called DamageInfoData
+---@param CTakeDamageInfo dmginfo
+-- @return table the DamageInfoData
+-- @realm server
+local function CreateDamageInfoData(dmginfo)
```
TimGoll, line 344:
```suggestion
---
-- Clones a CTakeDamageInfo into a table called DamageInfoData
-- @param CTakeDamageInfo dmginfo
-- @return DamageInfoData The damage info data table
-- @realm server
function CreateDamageInfoData(dmginfo)
```

you should probably make this a global function if you take your time to add that much documentation

TimGoll (changes requested):

gamemodes/terrortown/gamemode/server/sv_corpse.lua
TimGoll: or what are your thoughts on this @saibotk and @ZenBre4ker?

TimGoll (approved):

gamemodes/terrortown/gamemode/server/sv_corpse.lua
TimGoll: @ZenBre4ker what's your current stance on this? can we merge it?

PR #1237, opened 2023-12-20
AFK/Idle timer now reads inputs

### LINKED ISSUE #1015: [Idle mode] Regular bypass method by a few console commands
Some not-so-good people remind about a method to bypass the Idle Mode, so they will not become spectators automatically.

To do that, you have to open your console, write something like `+walk; +left` and that's it. Now you will walk around the circle and will not be forced to the spectator team by the system.

My suggestion here is to use a method similar to Arizard's Deathrun (https://github.com/Arizard/deathrun/blob/master/gamemode/init.lua#L665). There is check about pressing keys in some period of time, which actually a bit harder to avoid and will resolve the issue.

EntranceJew:
benefits:
 - doing an unsquared distance check is expensive, this is not
 - getting crowbarred around no longer un-idles you
 - cannot be circumvented with console commands to rotate or walk in one direction

additional thoughts:
 - if people want crowbar pushes to un-idle then that should be a deliberate feature done elsewhere

fixes #1015


gamemodes/terrortown/gamemode/client/cl_main.lua
```diff
@@ -746,13 +746,23 @@ end
...
+---
+-- Sniff for any button inputs to prevent scamming the AFK timer.
+-- @realm client
+hook.Add("SetupMove", "DeathrunIdleCheck", function( ply, mv )
+	if not GetGlobalBool("ttt_idle", false) then return end
+	if mv:GetButtons() ~= idle.btn then
+		idle.t = CurTime()
+	end
+	idle.btn = mv:GetButtons()
+end)
```
TimGoll, line 764:
```suggestion
hook.Add("SetupMove", "DeathrunIdleCheck", function(_, mv)
	if not GetGlobalBool("ttt_idle", false) then return end

	if mv:GetButtons() ~= idle.btn then
		idle.t = CurTime()
	end

	idle.btn = mv:GetButtons()
end)
```

TimGoll (approved):

gamemodes/terrortown/gamemode/client/cl_main.lua
```diff
@@ -746,13 +746,25 @@ end
...
 	mx = 0,
 	my = 0,
-	t = 0
+	t = 0,
 }
 
+---
+-- Sniff for any button inputs to prevent scamming the AFK timer.
+-- @realm client
+hook.Add("SetupMove", "DeathrunIdleCheck", function(_, mv)
```
Histalek, line 758:
```suggestion
hook.Add("SetupMove", "TTT2IdleCheck", function(_, mv)
```

EntranceJew: i've been awake too long, i got the angles to stop interfering and suddenly everything else slips right past

Histalek (approved):

PR #1241, opened 2023-12-20
Rework: Sprinting, Crosshair, FOV/ViewBobbing

TimGoll:
The goal is to make sprinting feel more immersive.

- [x] prevent sprinting when in iron sight
- [x] change FOV based on speed
- [x] add convar to disable this
- [x] decrease accuracy while sprinting
- [x] increase crosshair size while sprinting 
- [x] Remove sprinting hide crosshair convar
- [x] Remove crosshair static convar
- [x] Decrease weapon accuracy while sprinting / in air

https://www.youtube.com/watch?v=Gd3NsF5VRpM

I also fixed a typo in the docs (identifer to identifier)


gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -22,13 +20,13 @@ SPRINT = {
 -- @return boolean
 -- @realm shared
 function SPRINT:PlayerWantsToSprint(ply)
-	local isSprinting = ply:KeyDown(IN_SPEED)
+	local inSprinting = ply:KeyDown(IN_SPEED)
```
Histalek, line 23: why? if you want to keep this in line with `inMovement` at least make it `inSprint`

lua/terrortown/lang/en.lua
```diff
@@ -2098,3 +2096,15 @@ L.builtin_marker = "Built-in."
...
+-- 2023-12-22
+L.submenu_gameplay_accessibility_title = "Accessibility"
+
+L.header_accessibility_settings = "Accessibility Settings"
+
+L.label_enable_dynamic_fov = "Enable dynamic FOV change"
+L.label_enable_bobbing = "Enable view bobbing"
+L.label_enable_bobbing_strafe = "Enable view bobbing when strafing"
+
+L.help_enable_dynamic_fov = "Dynamic FOV is applied depending on the payer's speed. When a player is sprinting for example, the FOV is increased to visualize the speed."
```
Histalek, line 2109:
```suggestion
L.help_enable_dynamic_fov = "Dynamic FOV is applied depending on the player's speed. When a player is sprinting for example, the FOV is increased to visualize the speed."
```

Histalek (changes requested):
Did this PR change behavior of weapon recoil while not sprinting? There are a lot of magic numbers which i can't really relate. If so that should be at least mentioned in the changelog

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
TimGoll: check the line below, it looked like a typo to me

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -751,8 +775,8 @@ function SWEP:SecondaryAttack()
 
 	local bNotIronsights = not self:GetIronsights()
 
-	self:SetIronsights(bNotIronsights)
 	self:SetZoom(bNotIronsights)
+	self:SetIronsights(bNotIronsights)
```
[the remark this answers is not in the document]
TimGoll: I have to test this again. Might be irrelevant by now. But at some point it was relevant because of the way I have to hack into the FOV stuff. I will test this and report back

TimGoll:
> Did this PR change behavior of weapon recoil while not sprinting? There are a lot of magic numbers which i can't really relate. If so that should be at least mentioned in the changelog

no, this stayed the same. I only set the fallback value in the function to the same value that is used in the weapon base - but this fallback is probably never used anyway. This was just a change to clean it up slightly

lua/terrortown/lang/tr.lua
```diff
@@ -2096,3 +2096,15 @@ L.builtin_marker = "Bütünleşik."
...
+-- 2023-12-22
+--L.submenu_gameplay_accessibility_title = "Accessibility"
+
+--L.header_accessibility_settings = "Accessibility Settings"
+
+--L.label_enable_dynamic_fov = "Enable dynamic FOV change"
+--L.label_enable_bobbing = "Enable view bobbing"
+--L.label_enable_bobbing_strafe = "Enable view bobbing when strafing"
+
+--L.help_enable_dynamic_fov = "Dynamic FOV is applied depending on the payer's speed. When a player is sprinting for example, the FOV is increased to visualize the speed."
```
NovaDiablox, line 2109: Did you mean "player's speed" ?

NovaDiablox (changes requested):
"payer's speed" should be "player's speed"

NovaDiablox (approved):

TimGoll:
@EntranceJew noticed a few issues with this and I will have to do some polishing on this one

TimGoll:
Just noticed that there is this:

image

I will therefore remove my custom syncing as this already is part of the game

TimGoll:
> Just noticed that there is this:
> 
> image
> 
> I will therefore remove my custom syncing as this already is part of the game

I decided that I will keep using my custom system because that way I have a hook that can trigger on a value change. Additionally my system supports more data types

TimGoll:
@Histalek RE the call order of `SetZoom` and ``SetIronsights`: It doesn't change anything. If you prefer it, I can revert that change, but it works both ways. Therefore I'll keep it that way

Histalek:
> @Histalek RE the call order of `SetZoom` and `SetIronsights`: It doesn't change anything. If you prefer it, I can revert that change, but it works both ways. Therefore I'll keep it that way

Please revert the change then.

TimGoll:
> > @Histalek RE the call order of `SetZoom` and `SetIronsights`: It doesn't change anything. If you prefer it, I can revert that change, but it works both ways. Therefore I'll keep it that way
> 
> Please revert the change then.

Done

gamemodes/terrortown/entities/weapons/weapon_ttt_m16.lua
```diff
@@ -68,14 +68,15 @@ function SWEP:SecondaryAttack()
...
+
 	self:SetNextSecondaryFire(CurTime() + 0.3)
 end
 
 ---
 -- @ignore
 function SWEP:PreDrop()
-	self:SetZoom(false)
 	self:SetIronsights(false)
+	self:SetZoom(false)
```
Histalek, line 79:
```suggestion
	self:SetZoom(false)
	self:SetIronsights(false)
```

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -1056,8 +1122,8 @@ function SWEP:Initialize()
 	elseif SERVER then
 		self.fingerprints = {}
 
-		self:SetIronsights(false)
 		self:SetZoom(false)
+		self:SetIronsights(false)
```
Histalek, line 1126:
```suggestion
		self:SetIronsights(false)
		self:SetZoom(false)
```

gamemodes/terrortown/entities/weapons/weapon_zm_rifle.lua
```diff
@@ -91,8 +91,8 @@ end
 ---
 -- @ignore
 function SWEP:PreDrop()
-	self:SetZoom(false)
 	self:SetIronsights(false)
+	self:SetZoom(false)
```
Histalek, line 95:
```suggestion
	self:SetZoom(false)
	self:SetIronsights(false)
```

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -7,6 +7,14 @@ if not plymeta then
...
 end
 
+---
+-- @realm client
+local cvEnableBobbing = CreateConVar("ttt2_enable_bobbing", "1", FCVAR_ARCHIVE)
+
+---
+-- @realm client
+local cvEnableBobbingStrafe = CreateConVar("ttt2_enable_bobbing_strafe", "1", FCVAR_ARCHIVE)
+
```
Histalek, line 17: I'm not a fan of enabling these by default.

Having a setting, that is generally known to cause motion sickness for some people, turned on by default just doesn't seem like a good idea to me.

Histalek (changes requested):

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: I can't think of any game that has them disabled by default. In all games that I play you have to disable them in the accessability settings if you don't like it

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -282,3 +290,105 @@ end
...
+-- heavily inspired from V92's "Head Bobbing": https://steamcommunity.com/sharedfiles/filedetails/?id=572928034
+hook.Add("CalcView", "TTT2ViewBobbingHook", function(ply, origin, angles, fov)
+	if not cvEnableBobbing:GetBool() then return end
+
+	local observerTarget = ply:GetObserverTarget()
+
+	-- handle observing players
+	if not ply:IsTerror() and IsValid(obersverTarget) and observerTarget:IsPlayer() then
+		ply = observerTarget
+	end
```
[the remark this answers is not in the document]
TimGoll: huh, probably not. I have to test this

gamemodes/terrortown/gamemode/shared/sh_sprint.lua
```diff
@@ -11,8 +11,6 @@ SPRINT = {
 		consumption = CreateConVar("ttt2_sprint_stamina_consumption", "0.6", { FCVAR_ARCHIVE, FCVAR_NOTIFY, FCVAR_REPLICATED }, "The speed of the stamina consumption (per second; Def: 0.6)"),
 		-- @realm shared
 		regeneration = CreateConVar("ttt2_sprint_stamina_regeneration", "0.3", { FCVAR_ARCHIVE, FCVAR_NOTIFY, FCVAR_REPLICATED }, "The regeneration time of the stamina (per second; Def: 0.3)"),
-		-- @realm shared
-		showCrosshair = CreateConVar("ttt2_sprint_crosshair", "0", { FCVAR_ARCHIVE, FCVAR_NOTIFY, FCVAR_REPLICATED }, "Should the Crosshair be visible while sprinting? (Def: 0)"),
```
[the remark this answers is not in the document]
TimGoll: I can't find any reference anywhere in code

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -282,3 +290,105 @@ end
...
+function plymeta:SetSettingOnServer(identifier, value)
+	if self.playerSettings[identifier] == value then return end
+
+	self.playerSettings[identifier] = value
+
+	net.Start("ttt2_set_player_setting")
+	net.WriteString(identifier)
+	net.WriteTable({value}, true) -- use table to support any data type
+	net.SendToServer()
+end
```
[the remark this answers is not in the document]
TimGoll: I see what you mean. I'm not a huge fan of your suggestion though. Is this really a concern? This can only happen when the server has cheats enabled and a player wants to abuse this by sending lots of data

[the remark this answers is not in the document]
TimGoll: ok, it might overcomplicate this whole ordeal, but what about the following:
only identifiers that are registered on the server and the client can be set, that way the only way to abuse is to overwrite them with a table, this is finite.

If that's not enough, one could go one step forward and only allow strings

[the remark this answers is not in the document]
TimGoll: Another solution would be to use `GetInfoNum` and add a custom change detection (because you cannot attach a on change callback to a dsynced convar for some reason):

image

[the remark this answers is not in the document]
TimGoll: @saibotk could you take a look at this suggestion so I can finish this PR?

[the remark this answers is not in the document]
saibotk: Nah this not really nice and also not really a difference to the "only send strings" approach.

Maybe lets just use integers for this, so they can be used for 0 = false, 1 = true or could be used as some sort of index value for selects etc. 
I think we do not have use cases for strings or entire tables currently and in the near future.
Also limit the settings a user can set by having a serverside list of possible settings.

I think this would be an okayish solution to this.

I am not 100% sure how severe this really is, but you should never trust the client in anything. A malicious client does not care about sv_cheats and anyone can easily execute custom lua on the client. But id rather like to be a bit careful with such generic network functions though.

Maybe @Histalek has some opinion on this?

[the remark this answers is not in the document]
TimGoll: I need floats though. Should I parse everything as a float then, even bools? or send everything as a string and define on the serverside the type of that variable?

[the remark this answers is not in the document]
saibotk: yes lets do that, then it is basically like convars

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -682,6 +682,10 @@ function plymeta:InitialSpawn()
 
 	self:SetPos(spawnPoint.pos)
 	self:SetAngles(spawnPoint.ang)
+
+	-- register synced vars
+	self:RegisterSettingOnServer("enable_dynamic_fov", "bool")
+	self:RegisterSettingOnServer("fov_desired", "float")
```
saibotk, line 688: I think it would be easier to not do that for each player and maybe handle this globally?

TimGoll: Maybe, not so sure. But as the setting is ok the player it might be interesting to allow different kinds of settings for different players.

E.g. an admin could have more synced settings

saibotk: Yes, but since this is also a case of "could/might" and it is not necessary for now we should not make it more complicated, as it can easily be adjusted again in the future to do so. 

Just feels more hacky having to set this on initial spawn on every player and for other people trying to use this it is also weird because they need to initialize it on every player.

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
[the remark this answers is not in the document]
TimGoll: It is not disabled - but should it be disabled? I just checked how it is done in Minecraft (my go to reference for this PR) and there it is enabled in 3rd person. So I'd vote to keep it that way

TimGoll:
@saibotk done!

saibotk (approved):

PR #1242, opened 2023-12-21
Mark the old Death Note addons as imcompatible

xopez:
- Roles are hardcoded
- In TTT2 all addons print only detective roles
- The new addon prints all roles except the own teamrole no matter for which team they play
- Hidden roles for the user are still hidden for him so there is no reveal


TimGoll:
LET THE CODE BEAUTYFIERS FIGHT!

gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
```diff
@@ -184,25 +184,40 @@ addonChecker.curatedList = {
 	},
 	["1729451064"] = { -- TTT - More Melee Weapons by Solrob
 		reason = "Overwrites stock TTT2 crowbar which chauses problems with doors and ttt_map_settings entity.",
```
TimGoll, line 186: not your fault, but could you fix this typo while you're at it? Thanks!

xopez: fixed, had to remerge from origin, hope it didn't break anyting

TimGoll:
looks good! thank you

TimGoll:
FYI, in our workflow the reviewer closes the conversations. Doesn't matter in this instance, but just so you are aware

xopez:
Thanks, good to know. I will follow that in future

ZenBre4ker (approved):

PR #1243, opened 2023-12-21
Changelog: Move entry to correct position

ZenBre4ker:
Well, I made a mistake with the last merge. I resolved the conflict, but the entry was still for the old release. Added the other categories as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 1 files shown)


TimGoll (approved):
happened to me on all PRs that are open right now. Really annoying

Histalek (approved):
Not sure if whatever tool you used to resolve the conflict supports this, but i like git's `merge.conflictStyle = diff3` [1]. that has helped me in many cases to prevent things like this.

[1] https://git-scm.com/docs/git-merge#_how_conflicts_are_presented

Issue #1191, opened 2023-12-12
Tracking Issue: Improve our CI


Histalek:
Our CI can be improved in many aspects.
This issue should track all **ideas** we can come up with and their respective progress.

## Pull Requests

- [ ] Check all files against the [workshop addon whitelist](https://github.com/Facepunch/gmad/blob/master/include/AddonWhiteList.h)
- [ ] Check all referenced files (e.g. `Material()` calls) for inclusion in the project and if they are added to the force-download
- [ ] Make extensive use of the [Lua Language Server](https://github.com/LuaLS/lua-language-server) and use it to find common errors
	- this is blocked until we switch our code comments to the LLS annotations
- [ ] Keep language files tidy with our [language parser](https://github.com/TTT-2/ttt2-language_parser)
- [ ] Keep a consistent code style with [stylua](https://github.com/JohnnyMorganz/StyLua)
	- evaluate how this would interact with our current glualint ci 

## Release

(Maybe automate the whole release process on the github side of things?)
- [ ] Generate the html in `cl_changes.lua` from the Changelog
- [ ] Draft a Pre-Release with the Changes from the Changelog

## Other

- [ ] Use [stale action](https://github.com/actions/stale) instead of the old unmaintained project
	- this should also be reconfigured to only mark issues with the label `status / needs info` as stale
	- if possible: it should ping project members instead of marking something as stale, in all other cases

TimGoll:
> * Check all referenced files (e.g. `Material()` calls) for inclusion in the project and if they are added to the force-download

we should probably have a table with the file names that can be iterated by a loop that calls AddFile

xopez:
> * [ ]  Keep a consistent code style with [stylua](https://github.com/JohnnyMorganz/StyLua)

If stylua works great maybe we should recommend the VS Code Extension in this repository.
Maybe you wanna think about that. 
Extension: https://marketplace.visualstudio.com/items?itemName=JohnnyMorganz.stylua
Guide: https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

Histalek:
> > * [ ]   Keep a consistent code style with [stylua](https://github.com/JohnnyMorganz/StyLua)
> 
[...]

PR #1245, opened 2023-12-21
CI: Add job to run stylua for style checking

Histalek:
The last commit styles our whole codebase with stylua and a maximum line length of 100 characters.
A few commits back there is a version with 120 character limits.

For now this is a draft only to gather feedback.

The goal here was to have defaults mostly and to not add another config file (stylua respects .editorconfig if possible).

## TODOs

- [x] fix glualint issues
	- done by setting `quote_style = "ForceDouble"` in stylua.toml
- [x] figure out NeoDoc failure
	- [x] use `-- stylua: ignore` for `CreateConVar()` calls as NeoDoc cannot handle multiline calls
	- [x] use `-- stylua: ignore` for `hook.Run` calls as NeoDoc cannot handle multiline calls
- [x] ~~decide on whether to use `collapse_simple_statement = "ConditionalOnly"` ( 0d6adc357bd583df81cb0f0425e8c2570ccb00ab shows the changes done by this)~~
- [x] ignore the formatting commit from git blame view

- [ ] create Follow-up PR which amends CONTRIBUTING.md, possibly describing VSCode and "vanilla" stylua usage

Ref #1191

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,4 @@
+# .git-blame-ignore-revs
+#
+# Run stylua formatter on entire codebase
+77b89186415f188ff465ee3e2c43ef6632aa37a9
--- a/gamemodes/terrortown/entities/entities/ttt_beacon.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_beacon.lua
@@ -98,6 +98,7 @@ if SERVER then
 
 			---
 			-- @realm server
+			-- stylua: ignore
 			if hook.Run("TTT2BeaconDetectPlayer", ent, self) == false then continue end
 
 			plysFound[ent] = true
@@ -164,6 +165,7 @@ if SERVER then
 
 			---
 			-- @realm server			
+			-- stylua: ignore
 			if hook.Run("TTT2BeaconDeathNotify", victim, beacon) == false then continue end
 
 			LANG.Msg(beaconOwner, "msg_beacon_death", nil, MSG_MSTACK_WARN)
--- a/gamemodes/terrortown/entities/entities/ttt_beacon.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_beacon.lua
@@ -3,12 +3,12 @@
 -- @section ttt_beacon
 
 if SERVER then
-	AddCSLuaFile()
+    AddCSLuaFile()
 end
 
 if CLIENT then
-	ENT.Icon = "vgui/ttt/icon_beacon"
-	ENT.PrintName = "Beacon"
+    ENT.Icon = "vgui/ttt/icon_beacon"
+    ENT.PrintName = "Beacon"
 end
 
 ENT.Base = "ttt_base_placeable"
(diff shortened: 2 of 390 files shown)


gamemodes/terrortown/entities/entities/ttt_beacon.lua
```diff
@@ -191,11 +208,15 @@ if CLIENT then
 	end)
 
 	hook.Add("PostDrawTranslucentRenderables", "BeaconRenderRadius", function(_, bSkybox)
-		if bSkybox then return end
+		if bSkybox then
+			return
+		end
```
ZenBre4ker, line 213: This looked fine before, is it not in the limit or is the style generally enforce this structure?

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
```diff
@@ -619,7 +633,10 @@ if SERVER then
 				return
 			elseif time < C4_MINIMUM_TIME or time > C4_MAXIMUM_TIME then
 				return
-			elseif IsValid(bomb:GetPhysicsObject()) and bomb:GetPhysicsObject():HasGameFlag(FVPHYSICS_PLAYER_HELD) then
+			elseif
+				IsValid(bomb:GetPhysicsObject())
```
ZenBre4ker, line 637: And why is this not in line with else if?

Histalek:
hmm okay glualint is understandably not happy. Will need to see how i can accommodate that one

Not sure what happened with NeoDoc though ...

lua/ttt2/libraries/database.lua
```diff
@@ -198,37 +207,47 @@ local function OnChange(index, itemName, key, newValue)
...
 	local cache = callbackCache[accessName]
 
-	if not cache then return end
+	if not cache then
+		return
+	end
 
 	-- Call all callbacks
-	local itemNameTable = {itemName, allCallbackString}
+	local itemNameTable = { itemName, allCallbackString }
```
ZenBre4ker, line 223: Is it really nicer to have spaces between brackets?

lua/ttt2/libraries/database.lua
```diff
@@ -465,7 +519,7 @@ clientReceiveFunctions[MESSAGE_REGISTER] = function(data)
 		keys = data.savingKeys,
 		data = data.additionalData,
 		storedData = {},
-		defaultData = {}
+		defaultData = {},
```
ZenBre4ker, line 522: Is this correct? A comma after the last entry?

ZenBre4ker (commented):
Not so sure about all those changes. It makes it more readable in most cases, but also larger, making it harder to get a quick overview.

TimGoll:
NIce that you mention it @ZenBre4ker because I raised the same concerns. All those style changes add nothing in my opinion, they only add unnecessary characters like whitespaces, newlines and commas

lua/ttt2/libraries/database.lua
Histalek: lua does allow this, yes.

We don't need to cripple ourselves only because of JSON being stupid :D

as for the reason why this is a good change: it removes the last-entry special case which causes unneeded friction when moving entries around or appending entries in the source. Both would otherwise lead to more changes than necessary (adding an removing `,`)

gamemodes/terrortown/entities/entities/ttt_beacon.lua
Histalek: i would need to check this. But i agree that this is a questionable style.

lua/ttt2/libraries/database.lua
EntranceJew: imo, lua should always leave the last `,` because it decreases line noise in commits and make merges go cleaner

TimGoll: fair point, you two convinced me with your arguments

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
Histalek: this styling puts every literal on a new line so that it is easier to grasp what literals exist and which logical operators combine them.
This also separates the logical struct from its statements/conditions, which makes it easier to grasp that only the statement has changed and not the logical struct.

lua/ttt2/libraries/database.lua
Histalek: i don't care too much tbh, but it does look nicer to me.

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
TimGoll: imho it would look nicer like this:

```lua
if something then

elseif IsValid(bomb:GetPhysicsObject())
	and bomb:GetPhysicsObject():HasGameFlag(FVPHYSICS_PLAYER_HELD)
then return end
```

Histalek:
> NIce that you mention it @ZenBre4ker because I raised the same concerns. All those style changes add nothing in my opinion, they only add unnecessary characters like whitespaces, newlines and commas

Those are not exactly unnecessary, most of these changes aim to generate less diff 'noise' and allow for diffs to show changes more selectively.

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
Histalek:
```lua
if -- <-- logical control structure
	something() -- <-- expression
then -- <-- logical control structure
	do_something()
elseif -- <-- logical control structure
	IsValid(bomb:GetPhysicsObject()) -- <-- expression
	and bomb:GetPhysicsObject():HasGameFlag(FVPHYSICS_PLAYER_HELD) -- <-- expression
then -- <-- logical control structure
	return -- <-- expression
end -- <-- logical control structure
```

This way you don't mix the control structure of the if statement with the actual expressions that are used.
Not only do you get a more readable diff out of these bacause if the control structure doesn't change you won't have the control structure in the line diff (only as context), additionally having each literal on a new line starting with the logical operator allows you to easier grasp under which conditions a statement can terminate early and what is not evaluated in that case.


Without comments because github's syntax highlighting gets overly confused by comments ...
```lua
if
	something()
then
	do_something()
elseif
	IsValid(bomb:GetPhysicsObject())
	and bomb:GetPhysicsObject():HasGameFlag(FVPHYSICS_PLAYER_HELD)
then
	return
end
```

TimGoll: I **really** dislike this structure because on first glance you can differntiate between something that is checked and something that is done when the check passes

Histalek: > allows you to easier grasp under which conditions a statement can terminate early and what is not evaluated in that case.

This is especially important for us because need so damn many nil checks ...

Histalek: > I _**really**_ dislike this structure because on first glance you can't differentiate between something that is checked and something that is done when the check passes

i think (?) i know what you mean, but instead of going to the start of the line going up to the first non-indented line is hardly more effort.

TimGoll: but that's like defining a variable like this

```lua
local
	varname
	= value
```

it's a really weird way to write things

saibotk: I really like this style and dislike the harder to read inline style. 
Especially since this is only the case for multi line statements.

It is soo freaking hard to differentiate multiline conditions and the control structure of an `elseif` in the example above.

I like this consistent style much more and honestly i think it is something we get used too quite quickly.

ZenBre4ker:
I can mostly agree with the changes, but especially one-liners, that already look nice, should stay imho.
Github diff wise, the new style is definitely way better.

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
TimGoll: Honestly, I don't think Histaleks suggestion makes it easier to read, at all. 

This
```lua
if something() then
	do_something()
elseif IsValid(bomb:GetPhysicsObject())
	and bomb:GetPhysicsObject():HasGameFlag(FVPHYSICS_PLAYER_HELD)
then
	return
end
```

or
```lua
if something() then
	do_something()
elseif IsValid(bomb:GetPhysicsObject())
	and bomb:GetPhysicsObject():HasGameFlag(FVPHYSICS_PLAYER_HELD)
then return end
```
is so much cleaner looking

Histalek: I disagree, besides my points above those two examples have inconsistent `then` placements for no particular reasons other than "must use less lines". These make it worse to see at a glance what is "done" and not part of the condition.

Second point, the `elseif` is terrible to read, because the conditions don't start at the same indentation, which again makes it hard to grasp what is checked and what is done.

The style stylua uses, places the different "do" and "checked" parts into distinct groups separated by the keywords of the control structure.

I'm not advocating for a "cleaner" look, that's highly subjective. I'm advocating for a clearer and consistent code style that is next to impossible to misread and removes unnecessary noise in diffs. This will make reviewing changes much, much easier.

saibotk: Also i don't think we should try to argue about looks too much its much much more important that this is not a bad style and rather is something that is consistent and auto fixed.

EntranceJew:
something to keep in mind: let's try to close existing PRs and WIPs before this goes through so they don't get delayed any further by the upstream codebase getting formatted

saibotk (approved):
I like this very much and its a welcome improvement to a consistent style.
Dont really care too much about how it is styled same as with prettier, i dont agree with every decision they made but overall its just sooo good to have an auto formatting tool. 

Note i did not check all files, only Ci and config. I trust you that all those were only automated changes.

Histalek:
> something to keep in mind: let's try to close existing PRs and WIPs before this goes through so they don't get delayed any further by the upstream codebase getting formatted

Don't worry that was the plan. I'll try to get this once there are no open PRs (from non-Teammembers at least) :)

i just wanted to get this out for feedback.

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
TimGoll: I guess I'm the only one who dislikes the new solution so the decision is on you

gamemodes/terrortown/entities/weapons/weapon_ttt_m16.lua
```diff
@@ -83,7 +87,12 @@ end
 ---
 -- @ignore
 function SWEP:Reload()
-	if self:Clip1() == self.Primary.ClipSize or self:GetOwner():GetAmmoCount(self.Primary.Ammo) <= 0 then return end
+	if
+		self:Clip1() == self.Primary.ClipSize
+		or self:GetOwner():GetAmmoCount(self.Primary.Ammo) <= 0
+	then
+		return
+	end
```
TimGoll, line 95: I think I know why this looks weird to me. The third line here starts with an operator (`or`). The first line as well (`if`). But the second line looks kinda lost without an operator. This is why I prefer
```
if ThisIsTheFirstCheck()
	or ThisIsTheSecondCheck()
	and ThisIsTheThirdCheck()
then
	-- do some stuff
and
```

gamemodes/terrortown/entities/weapons/weapon_ttt_phammer.lua
```diff
@@ -4,60 +4,65 @@
...
 -- Returns if an entity is a valid physhammer punching target. Does not take
 -- distance into account.
 local function ValidTarget(ent)
-	return IsValid(ent) and ent:GetMoveType() == MOVETYPE_VPHYSICS and ent:GetPhysicsObject() and (not ent:IsWeapon()) and (not ent:GetNWBool("punched", false)) and (not ent:IsPlayer())
+	return IsValid(ent)
+		and ent:GetMoveType() == MOVETYPE_VPHYSICS
+		and ent:GetPhysicsObject()
+		and (not ent:IsWeapon())
+		and (not ent:GetNWBool("punched", false))
+		and (not ent:IsPlayer())
```
TimGoll, line 65: these breackets should be removed here

gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
```diff
@@ -65,40 +65,44 @@ SWEP.PrevOwner = nil
 
 -- ConVar syncing
 if SERVER then
-
 	---
 	-- @realm server
-	local cvAllowRagCarry = CreateConVar("ttt_ragdoll_carrying", "1", {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+	local cvAllowRagCarry =
+		CreateConVar("ttt_ragdoll_carrying", "1", { FCVAR_NOTIFY, FCVAR_ARCHIVE })
```
TimGoll, line 71: In another file your CreateConVar looked like this:

```lua
local cvAllowRagCarry = CreateConVar(
	"ttt_ragdoll_carrying",
	"1",
	{ FCVAR_NOTIFY, FCVAR_ARCHIVE }
)
```

Or this way, but at least it should be a single style in the whole project

gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
```diff
@@ -65,40 +65,44 @@ SWEP.PrevOwner = nil
...
 	---
 	-- @realm server
-	local cvAllowRagPin = CreateConVar("ttt_ragdoll_pinning", "1", {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+	local cvAllowRagPin = CreateConVar("ttt_ragdoll_pinning", "1", { FCVAR_NOTIFY, FCVAR_ARCHIVE })
 
 	---
 	-- @realm server
-	local cvAllowRagPinInno = CreateConVar("ttt_ragdoll_pinning_innocents", "0", {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+	local cvAllowRagPinInno =
+		CreateConVar("ttt_ragdoll_pinning_innocents", "0", { FCVAR_NOTIFY, FCVAR_ARCHIVE })
```
TimGoll, line 89: I mean it is even differnt in the same file. Is this because of the line length? If so, I think this should be handled differently. Because your argument for this is that wordwrap is bad. But is this really any better?

gamemodes/terrortown/gamemode/client/cl_help.lua
```diff
@@ -275,11 +289,23 @@ end
...
 	-- @realm client
-	if hook.Run("TTT2OnHelpSubmenuClear", parent, self.currentMenuId, self.lastMenuData, self.submenuClass) == false then return end
+	if
+		hook.Run(
+			"TTT2OnHelpSubmenuClear",
+			parent,
+			self.currentMenuId,
+			self.lastMenuData,
+			self.submenuClass
+		) == false
```
TimGoll, line 305: Just as a reference, here is the hook in the style I suggested earlier

gamemodes/terrortown/gamemode/client/cl_hud_editor.lua
```diff
@@ -62,10 +64,15 @@ local function TryResizeLocalHUD(elem, trans_data, dif_x, dif_y, shift_pressed)
...
+		if
+			dif_x * trans_data.direction_x * client.size.h
+			> dif_y * trans_data.direction_y * client.size.w
+		then
+			dif_x = mathRound(dif_y * trans_data.direction_y * client.aspect)
+				* trans_data.direction_x
 		else
-			dif_y = mathRound(dif_x * trans_data.direction_x / client.aspect) * trans_data.direction_y
+			dif_y = mathRound(dif_x * trans_data.direction_x / client.aspect)
+				* trans_data.direction_y
```
TimGoll, line 75: I think this shouldn't be done

gamemodes/terrortown/gamemode/client/cl_hud_editor.lua
```diff
@@ -135,7 +144,9 @@ local function TryResizeLocalHUD(elem, trans_data, dif_x, dif_y, shift_pressed)
 end
 
 local function TryMoveLocalHUD(elem, dif_x, dif_y, shift_pressed)
-	if not elem:ShouldDraw() then return end -- restrict movement when element is hidden
+	if not elem:ShouldDraw() then
+		return
+	end -- restrict movement when element is hidden
```
TimGoll, line 149: comment should be before the if

gamemodes/terrortown/gamemode/client/cl_hud_editor.lua
```diff
@@ -169,7 +180,9 @@ local function Think_EditLocalHUD()
 	local mouse_down = input.IsMouseDown(MOUSE_LEFT)
 
 	-- open context menu when rightclicked
-	if input.IsMouseDown(MOUSE_RIGHT) and (client.editOptionsX ~= x or client.editOptionsY ~= y) then
+	if
+		input.IsMouseDown(MOUSE_RIGHT) and (client.editOptionsX ~= x or client.editOptionsY ~= y)
+	then
```
TimGoll, line 185: why this and not in a single line? It isn't even shorter that way

gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -285,23 +302,40 @@ function CLSCORE:SaveLog()
...
 
-		log = log .. string.format("[%02d:%02d] | %-15s | %s\n", minutes, seconds, event.type, event:Serialize() or "")
+		log = log
+			.. string.format(
+				"[%02d:%02d] | %-15s | %s\n",
+				minutes,
+				seconds,
+				event.type,
+				event:Serialize() or ""
+			)
```
TimGoll, line 328: This should be

```lua
log = log .. string.format(
	"[%02d:%02d] | %-15s | %s\n",
	minutes,
	seconds,
	event.type,
	event:Serialize() or ""
)
```

TimGoll (commented):
If everyone but me likes these changes, I will approve them. But I'm really not a fan of most of them. The added comma in tables is fine and I see the benefit. The added spaces in tables (`{ 0, 0, 0 }`) but not in function parameters (`Vector(0, 0, 0)`) is weird and kinda inconsitent. The change to stop using `if not a then return end` for fencing is not really understandable to me because it adds two lines without any reasonable benefit.

I already talked lengths about me not being a fan of the if/else structure changes. But I took the time to review most files and I think a few things should still be changed

gamemodes/terrortown/entities/entities/ttt_traitor_button/init.lua
```diff
@@ -282,9 +293,11 @@ function ENT:TraitorUse(ply)
...
 
-	if not self:PlayerRoleCanUse(ply)
-	or not self:IsUsable()
-	or self:GetPos():Distance(ply:GetPos()) > self:GetUsableRange() then
+	if
+		not self:PlayerRoleCanUse(ply)
+		or not self:IsUsable()
+		or self:GetPos():Distance(ply:GetPos()) > self:GetUsableRange()
+	then
 		return false
```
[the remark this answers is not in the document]
Histalek: if_condition_alignment

gamemodes/terrortown/entities/weapons/weapon_ttt_binoculars.lua
```diff
@@ -260,23 +272,34 @@ if CLIENT then
...
+		if not c_wep:GetProcessing() then
+			return
+		end
 
-		tData:AddDescriptionLine(
-			GetPT("binoc_progress", {progress = progress}),
-			hud_color
+		local progress = mathRound(
+			mathClamp((CurTime() - c_wep:GetStartTime()) / c_wep.ProcessingDelay * 100, 0, 100)
 		)
```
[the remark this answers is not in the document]
Histalek: there is no consistent way of styling this as these are wildly different cases of chaining wildly different functions and their parameters.

In these cases a cutoff point is chosen to stay below 100 characters.

gamemodes/terrortown/entities/weapons/weapon_ttt_knife.lua
```diff
@@ -270,7 +275,8 @@ function SWEP:SecondaryAttack()
 	local vfw = ang:Forward()
 	local vrt = ang:Right()
 
-	local src = ply:GetPos() + (ply:Crouching() and ply:GetViewOffsetDucked() or ply:GetViewOffset())
+	local src = ply:GetPos()
+		+ (ply:Crouching() and ply:GetViewOffsetDucked() or ply:GetViewOffset())
```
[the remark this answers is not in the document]
Histalek: > 100 characters

gamemodes/terrortown/entities/weapons/weapon_ttt_binoculars.lua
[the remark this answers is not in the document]
TimGoll: Honestly, I see no benefit then. Cutting the lines short by introducing different styles of functions is no improvement imho. We should select a single style and stick to it instead of randomly changing it

gamemodes/terrortown/entities/entities/ttt_traitor_button/init.lua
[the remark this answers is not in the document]
TimGoll: I see what you're trying to tell me. I just wanted to point out that to me at least there is no benefit in this change. All I see is an additional line
and obviously the first line is a bit less indented

gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
Histalek: wordwrap and long lines are bad for coding:
- wordwrap kills consistent pointer and scrollmovement
- long lines either hide content or lead to wordwrap, or even worse it forces you to scroll horizontally (see this exact github code window)

gamemodes/terrortown/gamemode/client/cl_help.lua
Histalek: this is a single function and not a variable declaration as in most other cases you commented on

gamemodes/terrortown/gamemode/client/cl_hud_editor.lua
Histalek: i'd rather not scroll horizontally

gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
TimGoll: I've used wordwrap for at least 10 years of coding and I had never problems when scrolling or cursor placements. Agree on the second point though.

Maybe chars is the problem them. It is probably due to old machines having an 80 charactter line limit. I checked our codebase (at least 10 files or so) and the longest line we had was 117 characters. Most are way shorter. I think we could solve most of my issues by increasing the line length to 120. Is there a solid reason we chose 100? That's not even half my screen.

gamemodes/terrortown/gamemode/client/cl_hud_editor.lua
TimGoll: image
 but you don't have to? It is not like the line is 500 chars long. It is 103 chars long.

gamemodes/terrortown/gamemode/client/cl_help.lua
TimGoll: but my point is, that it is inconsitent. Here it is done like this, which I quite like. In another part of the code it is done like this which is really weird imho:
image

In other places it is still like this:
image

I get this is automatically done. But I don't think it should be done this way. There are at least three different styles for hooks in this code base now

gamemodes/terrortown/gamemode/client/cl_hud_editor.lua
Histalek: that's 3 chars too long for github

gamemodes/terrortown/gamemode/client/cl_help.lua
Histalek: > There are at least three different styles for hooks in this code base now

which you will never need to think about because it is autoformatted in a deterministic way :)

gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
Histalek: github choose this as the cutoff point for horizontal scrolling. if you want to look at how our code looks with 120 chars limit take a look at the second commit in this pr.

gamemodes/terrortown/gamemode/client/cl_help.lua
TimGoll: Just because it is done autoamtically doesn't mean that it is done good. There should only be the first and the last case, Also does the beautifier understand the syntax? What happens when I add more/longer arguments to case 2? Judging by this:
image

I guess it would be:
```lua
local role, team =
	hook.Run(
		"hookname",
		activator,
		self,
		activator,
		caller,
		data
	)
```

Do In understand this correctly?

gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
TimGoll: image

At least 111 chars looks fine on my FullHD screen

gamemodes/terrortown/gamemode/client/cl_help.lua
Histalek: serious question: do you want to think about code style?

i certainly don't. i'd rather think about the code itself

TimGoll: I thought this whole PR is about code style? But yes, I aldready use style 1 and 3 (and never 2) when manually indenting code because I think style 2 and 4 look ugly.

gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
Histalek: above_100_chars_github_scroll

TimGoll: Also this discussion is not about 100 chars or 120 chars. If you prefer 100, that's ok with me. I'm mostly annoyed that this autoformatter adds so many different styles of indentation instead of sticking to a single one. I get that this is done automatic. But This makes the code not good looking to me if I scroll through a file and see 2 or 3 different styles of convar initialization directly after one another

TimGoll: > above_100_chars_github_scroll

I might be blind, but I don't get what you're trying to show to me here

gamemodes/terrortown/gamemode/client/cl_help.lua
Histalek: > I thought this whole PR is about code style?

yes code style without arbitrary choices that anyone can replicate by running an automation tool. This lessens mental load as well as eliminates unneeded changes and prevents our reviews from having random styling nits in them.

> But yes, I aldready use style 1 and 3 (and never 2) when manually indenting code because I think style 2 and 4 look ugly.

"manually indenting" oO why would you do that to yourself?
and i have to iterate: this is not about looks

gamemodes/terrortown/entities/weapons/weapon_ttt_binoculars.lua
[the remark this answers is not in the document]
Histalek: it is not changed randomly. it is perfectly deterministic.

gamemodes/terrortown/gamemode/client/cl_help.lua
TimGoll: > "manually indenting" oO why would you do that to yourself?

I mean I decide whether I use
```lua
bla(a, vb, c)
```
or
```lua
bla(
	a, 
	b, 
	c
)
```

> and i have to iterate: this is not about looks

I might be alone with this - but to me it is also about looks. Code should look clean

gamemodes/terrortown/entities/weapons/weapon_ttt_binoculars.lua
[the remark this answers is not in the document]
TimGoll: We can argue all day about that. All I'm trying to say is that we should select one style and stick to it. I mean it might be deterministic to have three cases

- try to fit the whole line
- try to split it at the `=`
- try to split it at the `,`

but the second case looks like the only thing that annoys me about wordwrap. It looks so aribitrary to wrap that line there.

gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
Histalek: > > above_100_chars_github_scroll
> 
> I might be blind, but I don't get what you're trying to show to me here

There is a scrollbar and that line is 103 chars long (i think)

gamemodes/terrortown/gamemode/client/cl_help.lua
Histalek: > Code should look clean

at the expense of having to constantly think about how to format stuff?

TimGoll: I'm not really thinking about it, it is like walking. I'm not thinking about how I have to move my legs. But aren't we able to configure the tool in a way to set how we want our code to look like?

gamemodes/terrortown/gamemode/client/cl_scoring.lua
EntranceJew: fwiw: in jetbrains IDEs, these formatting options are referred to as "chopping", as in, they only ever get split into multiline things if they exceed a breakpoint line length, and IMO most of the time this is the only thing that really matters: a line is too long, it should be split across multiple lines in a clear and consistent fashion.

TimGoll: I get it, I also think this multiline version is much more readable. But I don't get why it isn't how I suggested it. This is my main gripe with it

EntranceJew: i thought i was replying to the thread and not a specific annotation, my bad, i meant as overall; some of the other things in the chain get split up seem excessive to do _always_, like splitting if (...) then across 3 lines but not splitting the condition itself after 80 characters across each logical operator or parenthetical group

gamemodes/terrortown/gamemode/client/cl_help.lua
Histalek: > I'm not really thinking about it, it is like walking. I'm not thinking about how I have to move my legs.

I'm sorry but this is 1. not remotely comparable, just think about how some people have an inner monologue and the others don't. and 2. this is a super privileged way of thinking

> But aren't we able to configure the tool in a way to set how we want our code to look like?

Only somewhat as it is an opinionated formatter that is easily available for VSCode, Sublime Text and Neovim or any other editor as it is a standalone program.

If you want to program an alternative, by all means do it.

TimGoll:
image

I'm not sure if this is possible with the tool, but in my opinion it should only be collapsen when there is only a `return` or `continue`
Having such function calls in a one line if is a bit hard to read imho

TimGoll:
I talked with a few programmer friends and they all seem to share your sentiment of "if it is automated then it is better than having to manually set good looking code". So I guess I'm more or less conviced. But can we please manually fix some of these abnormalities like:

image
image
image

or will the tool unfix them, even if we keep them below 100 chars?

Histalek:
> image
> 
> I'm not sure if this is possible with the tool, but in my opinion it should only be collapsen when there is only a `return` or `continue` Having such function calls in a one line if is a bit hard to read imho

as far as i can tell there is no extra "support" for fencing/guard-clauses :/

Histalek:
> I talked with a few programmer friends and they all seem to share your sentiment of "if it is automated then it is better than having to manually set good looking code". So I guess I'm more or less conviced. But can we please manually fix some of these abnormalities like:
> 
> or will the tool unfix them, even if we keep them below 100 chars?


it's either do or don't. Stylua will revert any manual changes like that.

TimGoll:
> > I talked with a few programmer friends and they all seem to share your sentiment of "if it is automated then it is better than having to manually set good looking code". So I guess I'm more or less conviced. But can we please manually fix some of these abnormalities like:
> > or will the tool unfix them, even if we keep them below 100 chars?
> 
> it's either do or don't. Stylua will revert any manual changes like that.

even if the new length is below 100 chars?

Histalek:
Note that i have updated the OP with open TODOs.

`decide on whether to use collapse_simple_statement = "ConditionalOnly" ( 0d6adc3 shows the changes done by this)` is probably the most important one.

I'm against using this at the expense of simple guard-clauses occupying 3 lines.

TimGoll:
> Note that i have updated the OP with open TODOs.
> 
> `decide on whether to use collapse_simple_statement = "ConditionalOnly" ( 0d6adc3 shows the changes done by this)` is probably the most important one.
> 
> I'm against using this at the expense of simple guard-clauses occupying 3 lines.

Agree with you. These one liners with code inside the if are awful

Histalek:
> > > I talked with a few programmer friends and they all seem to share your sentiment of "if it is automated then it is better than having to manually set good looking code". So I guess I'm more or less conviced. But can we please manually fix some of these abnormalities like:
> > > or will the tool unfix them, even if we keep them below 100 chars?
> > 
> > 
> > it's either do or don't. Stylua will revert any manual changes like that.
> 
> even if the new length is below 100 chars?

honestly i don't know.

TimGoll:
> > > > I talked with a few programmer friends and they all seem to share your sentiment of "if it is automated then it is better than having to manually set good looking code". So I guess I'm more or less conviced. But can we please manually fix some of these abnormalities like:
> > > > or will the tool unfix them, even if we keep them below 100 chars?
> > > 
> > > 
> > > it's either do or don't. Stylua will revert any manual changes like that.
> > 
> > 
> > even if the new length is below 100 chars?
> 
> honestly i don't know.

would it work if I edit one of them, push and you test it?

Histalek:
> would it work if I edit one of them, push and you test it?

You can checkout this branch do your changes and run stylua yourself :D if you use vscode the extension can be found here: https://marketplace.visualstudio.com/items?itemName=JohnnyMorganz.stylua

.git-blame-ignore-revs
Histalek, line 1: This allows `git blame` to ignore this specific commit.

GitHub supports this out of the Box [1].

Using this on the command line needs either this file explicitly specified:
```
$ git blame --ignore-revs-file .git-blame-ignore-revs
```

or set the filename as the ignoreRevsFile (can be done globally with `--global`):
```
$ git config blame.ignoreRevsFile .git-blame-ignore-revs
```

[1] https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

TimGoll (approved):

PR #1256, opened 2023-12-24
ammo capitalization

EntranceJew:
adds greater icon support to community weapons

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Damage Scaling now has a help description
 - Fixed the database module setting a global variable called `callback` which breaks addons such as PointShop2
+- Coerced ammo types to lowercase for better matching in HUD.
 
 ## [v0.12.2b](https://github.com/TTT-2/TTT2/tree/v0.12.2b) (2023-12-20)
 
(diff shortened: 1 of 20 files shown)


TimGoll (approved):

TimGoll:
will this fix this issue with some third party weapons using the fallback ammo icon?

image

EntranceJew:
yes; some will still be using ammo that doesn't have a type or icon but it should fix those that uses existing ammo types, and for those that it doesn't, the table is public so it can be amended by addons

TimGoll:
Can you make sure to update this in octagonal as well, as soon as this is merged/released?

EntranceJew:
right now AmmoType isn't referenced anywhere in octagonal so it shouldn't be an issue, unless you mean you're working on something to bring them to parity

TimGoll:
> right now AmmoType isn't referenced anywhere in octagonal so it shouldn't be an issue, unless you mean you're working on something to bring them to parity

I'm putting your icon rework from TTT2 base into octagonal right now

EntranceJew:
then yeah, just tag me in the commit message itself and i'll ensure it's there

Histalek (approved):

TimGoll:
@EntranceJew not that it is merged here, can you update it in octagonal as well? Thanks!

PR #1253, opened 2023-12-23
Magnetostick: Per role ragdoll pinning permissions (fixes #1100)

EntranceJew:
also a general rewrite of magneto stick because some parts were really, really strange

PR #1257, opened 2023-12-24
Updated Turkish Localization

NovaDiablox:
- Added https://github.com/TTT-2/TTT2/pull/1241 text
- Added https://github.com/TTT-2/TTT2/pull/1253 text
- Added https://github.com/TTT-2/TTT2/pull/1142 text
 - Feel free to deploy it.


TimGoll:
Nice that you are so eager! But just so that you are aware. PRs are always changing, so you might have to redo some things

NovaDiablox:
> Nice that you are so eager! But just so that you are aware. PRs are always changing, so you might have to redo some things

Oh I agree. I check regularly so nothing will be missed hopefully.

PR #1258, opened 2023-12-24
Keyhelp: Fixed nil-table indexing

TimGoll:
The current code fails when no element is in any of the key help tables. This went unnoticed when testing, because all tables have enties by default.

This is still bad because the UI code fails if a user decides to remove the disguiser from the sourcecode for example.


EntranceJew:
Is there a way I can manually invoke this to test it?

TimGoll:
@EntranceJew delete the disguiser file. Or remove the hook in there that registers the key helper

saibotk (approved):

Histalek (approved):

PR #1260, opened 2023-12-25
scoreboard: show confirmation sources like minihud

EntranceJew:
also indicates if body info is available, instead of blindly guessing

https://github.com/TTT-2/TTT2/assets/5711436/af6abb17-f63b-4208-b30f-5351347f7180


lua/terrortown/lang/en.lua
```diff
@@ -2098,3 +2098,8 @@ L.builtin_marker = "Built-in."
 L.equipmenteditor_desc_builtin = "This equipment is built-in, it comes with TTT2!"
 L.help_roles_builtin = "This role is built-in, it comes with TTT2!"
 L.header_equipment_info = "Equipment information"
+
+-- 2023-12-24
+L.cl_sb_row_sresult_direct_conf = "Direct confirmation"
+L.cl_sb_row_sresult_in_conf = "Indirect confirmation"
```
TimGoll, line 2104: I know what you are atalking about, but this might be "programmer translation". Is this really understandable? Or should it be something like "confirmed through kill list"?

EntranceJew: I really don't know how to refer to these things, even in english it's a complicated thing that is a full phrase, and maybe that is what it should be instead. But players understand "confirmed" aspect.

TimGoll: maybe "indirect confirmation" is the right call here, dunno. I just wanted to comment on this.

Histalek: i think if "indirect confirmation" is the text used to describe this we should just show no icon at all.

if i'm already wondering how a player ended up in the "confirmed dead" part of the scoreboard with no role and no other info attached, then "indirect confirmation" doesn't tell me anything more than i already know.

EntranceJew: image
image
I do agree, but "indirect confirmation" is still the general thing that links the minihud gray X to its role-less, confirmed position on the scoreboard. My thoughts are those be more closely related.

On the other hand, reserving the icons to EXCLUSIVELY mean that some level of bodysearch information is available will drastically cut down the time players spend folding and unfolding scoreboard data, and all the tooltips could be amended to say "Click to expand bodysearch info" or something. (if it were an imgbutton, right now it can't pass-through doclick)

The purpose was to bring the two headcount trackers closer together, but maybe the other changes are stronger without it, I don't know.

TimGoll: Histalek is probably not aware of indirect confirmation. This happens when a killer is confirmed. Their kill list ist then used to confirm the other players as dead, without a role or any other information

Histalek: >  I do agree, but "indirect confirmation" is still the general thing that links the minihud gray X to its role-less, confirmed position on the scoreboard. My thoughts are those be more closely related.

In that case we should use the same icons for those, right? I think that would better be done in a scoreboard rework.

> On the other hand, reserving the icons to EXCLUSIVELY mean that some level of bodysearch information is available will drastically cut down the time players spend folding and unfolding scoreboard data, and all the tooltips could be amended to say "Click to expand bodysearch info" or something. (if it were an imgbutton, right now it can't pass-through doclick)

That's exactly what i would go for here.

> The purpose was to bring the two headcount trackers closer together, but maybe the other changes are stronger without it, I don't know.

I think bringing mini-scoreboard and scoreboard mroe closely together should really be done when we rework the scoreboard. 

> Histalek is probably not aware of indirect confirmation.

I'm aware of the kill-list, but having someone in the "confirmed dead" section of the scoreboard with no further bodysearch information already conveys that they were confirmed through a kill list. An additional icon is just superfluous in that case.

EntranceJew: we don't have icons for the minihud that match the spyglass or direct confirmed states i introduce here, and the minihud icons aren't really suited for the existing scoreboard 'cause they're single color -- and you wouldn't be able to see them in most situations against the black background

TimGoll: I'm with Entrance here. The scoreboard needs a redo anyways, so for me it is fine to use these icons in there. I see no point in making the others work just to redo it at a later point again

Histalek: i might have expressed myself poorly.

I think the original goal of "correlating miniscoreboard and scoreboard" is good, but currently not really achievable as the two scoreboards currently differ quite a lot (hence the pushback to the scoreboard rework).

But i still think the current PR would solve another issue with slight modifications. As Entrance already said:
> On the other hand, reserving the icons to EXCLUSIVELY mean that some level of bodysearch information is available will drastically cut down the time players spend folding and unfolding scoreboard data, and all the tooltips could be amended to say "Click to expand bodysearch info" or something. (if it were an imgbutton, right now it can't pass-through doclick)

This would probably boil down to removing the 'indirect confirmation'/'killlist' icon and only using the other two. Additionally changing the icon descriptions to something like "(Dete) Body search results available" should also be done.
(Maybe find something shorter here, i didn't think a lot about the specific description)


TLDR:
- Correlation of Mini-scoreboard and Scoreboard should be thought about in a scoreboard rework
- Use the changes now to convey "Hey there are body search results available for this confirmed dead person"

EntranceJew: i don't have a great deal of time to continue working on this one in particular, if precise suggestions to the code are made i can change them and test it but i can't really determine what is desired to be changed from this at the moment

EntranceJew:
fwiw this is in status of review required, not changes requested, i don't know what needs changing exactly

TimGoll:
> fwiw this is in status of review required, not changes requested, i don't know what needs changing exactly

Honestly, I don't care either way. The scoreboard is a mess anyway and if you prefer to add this, so be it. Doesn't matter to me as it has to be reworked at some point anyway.
@Histalek seems to hold a different opinion on this, so I guess this is his call

gamemodes/terrortown/gamemode/client/vgui/cl_sb_row.lua
```diff
@@ -433,13 +432,23 @@ function PANEL:UpdatePlayerData()
...
-	if ply.bodySearchResult and (LocalPlayer():GetSubRoleData().isPolicingRole or not ply.bodySearchResult.show) then
-		self.sresult:SetImageColor(Color(200, 200, 255))
+	local show_sresult = true
+	-- we have body search results
+	if ply.bodySearchResult and ply.bodySearchResult.show_sb then
+		self.sresult:SetImage("icon16/information.png")
+		self.sresult:SetTooltip(GetTranslation("cl_sb_row_sresult_direct_conf"))
+	elseif ply:OnceFound() and not ply:RoleKnown() then
+		self.sresult:SetImage("icon16/comments.png")
+		self.sresult:SetTooltip(GetTranslation("cl_sb_row_sresult_in_conf"))
```
Histalek, line 442: Remove this indirect confirmation case and i'm fine with the PR.

Like i said in the thread on the language file i think this case doesn't add any more info for the player and only adds visual clutter.
Without this case the other two cases signal the Player "Click me! There is more information here!" which would be a great addition and would save players from clicking through every single scoreboard item to see if there is any search result.

lua/terrortown/lang/en.lua
```diff
@@ -2121,6 +2121,10 @@ Note: Some weapons might not use this value which causes this modifier to be ine
 -- 2023-12-24
 L.binoc_help_reload = "Clear target."
 
+L.cl_sb_row_sresult_direct_conf = "Direct confirmation"
+L.cl_sb_row_sresult_in_conf = "Indirect confirmation"
```
Histalek, line 2125: this would need to be removed as well

Histalek (changes requested):

Histalek (approved):

ZenBre4ker (approved):

TimGoll (approved):

PR #1261, opened 2023-12-25
Global callback fix

TimGoll:
Fixed a local variable being global that breaks PointShop 2

image


TimGoll:
This fix is pretty critical because right now PS2 is incompatible with TTT2

EntranceJew:
need a linter rule to make sure that non-locals aren't being declared outside a depth level of 0 in a file

ZenBre4ker (approved):
Oof, nice catch

saibotk (approved):

PR #1263, opened 2023-12-26
Keyhelp: Check if voicechat is enabled

### LINKED ISSUE #1246: Don't show voice keys in UI if voice is disabled
## Describe the solution you'd like
The voice UI here shouldn't be shown if `sv_voiceenable` is set to zero
Unbenannt

Also you shouldn't be able to cycle through the voice modes.

Since I don't think it's a bug I open it as a feature request. It would definitly be an enhancement.

TimGoll:
Fixes #1246 by hiding the keyhelp when voicechat is disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Damage Scaling now has a help description
 - Fixed the database module setting a global variable called `callback` which breaks addons such as PointShop2
+- Fixed voicechat keybinds being shown even if voice is disabled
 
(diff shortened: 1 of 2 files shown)


saibotk (approved):

EntranceJew (approved):
tested, works good (requires new round)

Histalek (approved):

TimGoll:
> tested, works good (requires new round)

Ah, probably because the global bool is only updated on round change. Maybe we should update it on convar change

Histalek:
> > tested, works good (requires new round)
> 
> Ah, probably because the global bool is only updated on round change. Maybe we should update it on convar change

if it is easily done, sure. Otherwise i don't think we should go the extra mile for a cosmetic fix which only occurs very rarely. ( i don't expect people to switch the convar on and off all that often if ever)

PR #1265, opened 2023-12-27
bugfix/viewmodel idle recenter

EntranceJew:
makes it so that your weapon doesn't get in the way as much, or go wonk

in order to really be fully functional, i think the sipistol should be brought down from upstream so that all weapons can be tagged accordingly and the extra stuff in this commit can be dropped

alternatively we just force it across all CS:S viewmodels because the effect of one broken weapon (silent pistol) can knock on other weapons that ordinarily don't have problems centering (the TMP)

### DIFF
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -110,6 +110,10 @@ SWEP.HotReloadableKeys = {}
 -- SWEPs do not need it for anything.
 --	SWEP.WeaponID = nil
 
+-- Set this to true ONLY if a weapon uses CS:S viewmodels that fail to recenter after firing.
+-- Also requires SWEP.IdleAnim to be set to the appropriate animation, usually ACT_VM_IDLE or ACT_VM_IDLE_SILENCED.
+SWEP.idleResetFix = false
+
 --   YE OLDE SWEP STUFF
 
 if CLIENT then
@@ -158,6 +162,7 @@ SWEP.DeploySpeed = 1.4
 
 SWEP.PrimaryAnim = ACT_VM_PRIMARYATTACK
 SWEP.ReloadAnim = ACT_VM_RELOAD
+SWEP.IdleAnim = ACT_VM_IDLE
 
 SWEP.fingerprints = {}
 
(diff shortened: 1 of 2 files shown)


gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -1045,12 +1050,62 @@ function SWEP:Initialize()
...
+end
+local function IsCounterStrikeWeapon(str)
+	return string.StartsWith(string.lower(str):gsub("\\","/"), "models/weapons/cstrike/")
+end
+
+local silenced_weapons = {
+	weapon_ttt_sipistol = true,
+	weapon_ttt_tmp_s = true,
+	weapon_ttt_m4a1_s = true,
+}
```
TimGoll, line 1091: I'm not sure if I like that in the weapon base. Shouldn't that be part of the weapon itself? As `SWEP.silenced = true` or something similar

EntranceJew: this is temporary to check that I have it solved; the correct way to handle this would be to use `SWEP.IdleAnim = ACT_VM_IDLE_SILENCED` since that's what the silenced variants of the base weapon animations use

TimGoll: ah, so that is why it is a draft then, gotcha. Thanks for working on this!

EntranceJew:
i brought down the silent pistol since it was also affected

gamemodes/terrortown/entities/weapons/weapon_ttt_sipistol.lua
```diff
@@ -0,0 +1,68 @@
...
+SWEP.Primary.DefaultClip = 20
+SWEP.Primary.ClipMax = 60
+SWEP.Primary.Ammo = "Pistol"
+SWEP.Primary.Sound = Sound( "Weapon_USP.SilencedShot" )
+SWEP.Primary.SoundLevel = 50
+
+SWEP.Kind = WEAPON_EQUIP
+SWEP.CanBuy = {ROLE_TRAITOR} -- only traitors can buy
+SWEP.WeaponID = AMMO_SIPISTOL
+
```
saibotk, line 38:
```suggestion
SWEP.builtin = true

```

gamemodes/terrortown/entities/weapons/weapon_ttt_sipistol.lua
```diff
@@ -0,0 +1,68 @@
...
+end
+
+---
+-- We were bought as special equipment, and we have an extra to give
+--@ignore
+function SWEP:WasBought(buyer)
+	if IsValid(buyer) then -- probably already self:GetOwner()
+		buyer:GiveAmmo( 20, "Pistol" )
+	end
+end
```
saibotk, line 68:
```suggestion
end

```

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -1071,12 +1076,48 @@ function SWEP:Initialize()
...
+	[ACT_VM_IDLE_DEPLOYED_7] = true,
+	[ACT_VM_IDLE_DEPLOYED_6] = true,
+	[ACT_VM_IDLE_DEPLOYED_5] = true,
+	[ACT_VM_IDLE_DEPLOYED_4] = true,
+	[ACT_VM_IDLE_DEPLOYED_3] = true,
+	[ACT_VM_IDLE_DEPLOYED_2] = true,
+	[ACT_VM_IDLE_DEPLOYED_1] = true,
+	[ACT_VM_IDLE_M203] = true,
+}
+local function IsIdleActivity(vm)
```
saibotk, line 1106:
```suggestion
}

local function IsIdleActivity(vm)
```

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -1071,12 +1076,48 @@ function SWEP:Initialize()
...
+end
+
 ---
 -- Called when the swep thinks.
 -- @warning If you override Think in your SWEP, you should call BaseClass.Think(self) so as not to break ironsights
 -- @see https://wiki.facepunch.com/gmod/WEAPON:Think
 -- @realm shared
 function SWEP:Think()
+	local vm = self:GetOwner():GetViewModel()
+	if self.idleResetFix and self.ViewModel and vm:GetCycle() >= 1 and not IsIdleActivity(vm) then
```
saibotk, line 1117:
```suggestion
	local vm = self:GetOwner():GetViewModel()
	
	if self.idleResetFix and self.ViewModel and vm:GetCycle() >= 1 and not IsIdleActivity(vm) then
```

saibotk (changes requested):
Small nitpicks but nice otherwise!

Does this fix the weird issues where the weapon model is stuck sideways?

TimGoll (approved):
LGTM, can be merged if @saibotk's suggestions are done. Feel free to add your name to the changelog though

saibotk (approved):

EntranceJew:
> Does this fix the weird issues where the weapon model is stuck sideways?

yes, the issue is because every Minh Le viewmodels don't correctly have their animations set to idle after firing in the animator state tree, so some weapons may not observably have the problem but would still have a different follow-up animation

Issue #501, opened 2020-04-05
Traitor traps that cannot be interacted but only visible


ThePlatinumGhost:
If thats possible, you could add a cvar similar to "ttt_role_traitor_button" which allows the role to see traitor traps but cannot activate them, like "ttt_role_traitor_button_onlysee"
the name of the cvar of your choice of course

Alf21:
That's nothing that would make sense imo. It would make sense to have a hook that could be called to activate the visible effect, e.g. for an innocent role seeing the position of T-traps.

ThePlatinumGhost:
> That's nothing that would make sense imo. It would make sense to have a hook that could be called to activate the visible effect, e.g. for an innocent role seeing the position of T-traps.

yes, in fact maybe I explained myself wrong, I was talking about a similar thing but to give it only to some classes of your choice
for example i would give to clairvoyant 

knowing the locations of the traps / t-rooms can be an advantage

PR #1267, opened 2023-12-27
Traitor Buttons: Add cooldown circle & role convar to see buttons

EntranceJew:
Adds a pie wedge around ticking TButtons, uses a new library, but doesn't yet add the roleperm to see (but not interact with) TButtons.


TimGoll:
the new library should be in `ttt2/libraries/circle.lua`

I'm also wondering if this should be called `draw.circle.XYZ`

EntranceJew:
The API of the library (provided by [SneakySquid/Circles](https://github.com/SneakySquid/Circles#methods) involves managing objects with lifecycles, which is a little bit different from the immediate mode draw functions everything else uses.

EntranceJew:
making note that I left an issue for the outlined mode in the repo:
https://github.com/SneakySquid/Circles/issues/6

hoping that we can fix it

EntranceJew:
> You could probably do it the same way we did it in TTTC/heroes, which also better fits our coding style imho without adding another library:
>
> You could probably just move the code from TTTC to TTT2 and use it for this. Maybe you could improve the code a bit, if you see some things that could be improved: https://github.com/TTT-2/TTTC/blob/15b34a8d75067550a46811af687ab4979ef34847/lua/terrortown/autorun/client/cl_classes_hud.lua#L132-L233

maybe if these extensions were already upstream, but then they'd conflict w/ TTTC. if you wanted to PR those in to the draw extensions and simultaneously move them from TTTC i'd try it, although they look quite nice now:
image

TimGoll:
> > You could probably do it the same way we did it in TTTC/heroes, which also better fits our coding style imho without adding another library:
> > You could probably just move the code from TTTC to TTT2 and use it for this. Maybe you could improve the code a bit, if you see some things that could be improved: https://github.com/TTT-2/TTTC/blob/15b34a8d75067550a46811af687ab4979ef34847/lua/terrortown/autorun/client/cl_classes_hud.lua#L132-L233
> 
> maybe if these extensions were already upstream, but then they'd conflict w/ TTTC. if you wanted to PR those in to the draw extensions and simultaneously move them from TTTC i'd try it, although they look quite nice now: image

will do, give me an hour

TimGoll:
@EntranceJew Done in #1291

TimGoll:
@EntranceJew It's now upstream, ready for you to use

TimGoll:
I think we have to work a bit on this PR to resolve our stance on the multicore beta support. So I'll convert this back to a draft until this is done. One of us should figure out how to use our "TTTC arc code" to support the multicore mode, that would be best imho.

Even if I like the featureset of the circle library, I'd prefer not to use it in TTT2 as it deviates from the style of our rendering code.

EntranceJew:
> I think we have to work a bit on this PR to resolve our stance on the multicore beta support. So I'll convert this back to a draft until this is done. 

i don't think it should be a hardline problem because the DNA scanner is also affected by rendering issues, granted i feel as though the worst part of the draw.Arc library is how it may potentially be exceeding the vertex count or failing to stitch itself together and the base branch was just ignoring bad/concave geometry instead of stretching the triangle to a new part of the screen that's being drawn on the graphics card

> One of us should figure out how to use our "TTTC arc code" to support the multicore mode, that would be best imho.

the circle library is already stable on mcore, i don't know which part exactly can be attributed to its functionality 

1. [it generates convex polygons](https://wiki.facepunch.com/gmod/surface.DrawPoly)
2. it [doesn't round the coordinates to integers](https://github.com/SneakySquid/Circles/blob/master/circles.lua#L109-L112), as [the docs suggest](https://wiki.facepunch.com/gmod/surface.DrawPoly)
3. i assume it's generating clockwise, [and correctly stitching the final triangle to the first one](https://github.com/SneakySquid/Circles/blob/master/circles.lua#L126-L136)
4. it uses a [using a stencil to cut out the inner circle](https://github.com/SneakySquid/Circles/blob/master/circles.lua#L230-L248) with far fewer verts than would be required otherwise

> Even if I like the featureset of the circle library, I'd prefer not to use it in TTT2 as it deviates from the style of our rendering code.

i have no opinion about style vs performance when it comes to rendering, as it stands neither of them cache geometry in a way that's ideal for drawing the arcs in a performant way:

- the `draw.Arc` library caches on an ID and immediately misses the cache the next time the arc distance changes, and it could be missing as early as the very next frame because the arc progress isn't snapped to a pre-defined number of progress steps such as 60 (every 6 degrees)
- the circle library is extremely performant when only one is on screen, because the implementation would keep reusing the same exact circle and it would move around the screen via its `:Translate()` method which saves recomputing each vertex position geometrically
- the circle library does a lot more work with my original implementation b/c i only use one copy of the object resulting in excessive math and bad cache management

to staple down the finite details:
- the circles could be individually bound to a button, keeping the number of vertex tables bound to the number of visible buttons
- the circles could be snapped to a fixed amount of progress intervals, e.g. 60 to increase cache successes between frames
- the background circle which highlights the full progress of the circle could be just one vertex table instance shared with 60/60 progress intervals
- all circles, like t-button icons, draw at a fixed size so there is no necessity to scale them

in both situations they perform best inside a static HUD coordinate space, but since they move around, it might even be better if they rendered to a texture similar to the way the SVG => Material library does it that way it truly would just be manually drawing only from a small pool of 60 materials to screen in various colors

gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
```diff
@@ -210,6 +225,10 @@ function TBHUD:Draw(client)
 		scrpos = focus_but.ent:GetPos():ToScreen()
 		scrPosXMid, scrPosYMid = scrpos.x - mid, scrpos.y - mid
 
+		if (val.access or val.visible) and ent:GetDelay() > 0 and percent > 0 then
```
a7f3, line 228: When hovering make the alpha maxed out for completed/black part
```suggestion
		if (val.access or val.visible) and ent:GetDelay() > 0 and percent > 0 then
			draw.Arc(ent:EntIndex(), scrpos.x, scrpos.y, circle_radius, outline_width, 0, 360, 0, ColorAlpha(COLOR_BLACK, 200))
```

TimGoll:
@EntranceJew I was ill the last few days. I will take a look at the arc code. But judging by a short glance the arc code and the circle code are really similar.

I might answer in more detail to your text if I have any questions. Thanks!

TimGoll:
ok, so I couldn't reproduce the multicore rendering issue with the arcs, maybe it was fixed with one of my small changes to the TTTC code. Or I'm missing something. But I think this is fine and looks good. Great work!

About the second thing you added in this PR regarding the convar to let the role see T-Buttons. I'm not that big of a fan. I think this should probably be more a role-var that can be used by role devs? Or if it is a convar, then it should be done differently, because right now this is really confusing:
image

To me this looks like a traitor can use these buttons, but is unable to see them.

Imho this should be moved to a secondary PR and maybe even discussed beforehand. I'm not strictly against adding this as a convar, but I don't like the current implementation.

gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
```diff
@@ -163,25 +167,36 @@ function TBHUD:Draw(client)
...
 		d = d:Dot(d) / (usableRange * usableRange)
 
 		-- draw if this button is within range, with alpha based on distance
 		if d >= 1 then continue end
 
+		local alpha = 200 * (1 - d)
+		local outlineColorAlpha = ColorAlpha(outlineColor, alpha)
+
+		if ent:GetDelay() > 0 and percent > 0 then
+			draw.Arc(ent:EntIndex(), scrpos.x, scrpos.y, circle_radius, outline_width, 0, 360, 0, ColorAlpha(COLOR_BLACK, alpha))
```
TimGoll, line 189: FYI, that won't work because these IDs might already be used. I'd do it like this:

```lua
ent.arcCacheID = draw.Arc(ent.arcCacheID, scrpos.x, scrpos.y, circle_radius, outline_width, 0, 360, 0, ColorAlpha(COLOR_BLACK, alpha))
```

EntranceJew:
> About the second thing you added in this PR regarding the convar to let the role see T-Buttons. I'm not that big of a fan. I think this should probably be more a role-var that can be used by role devs? Or if it is a convar, then it should be done differently, because right now this is really confusing:
#501 was tagged as accepted, but no discussion has happened about it for 4 years, so i figured that it was just "something nobody got around to"

TimGoll:
> > About the second thing you added in this PR regarding the convar to let the role see T-Buttons. I'm not that big of a fan. I think this should probably be more a role-var that can be used by role devs? Or if it is a convar, then it should be done differently, because right now this is really confusing:
> 
> #501 was tagged as accepted, but no discussion has happened about it for 4 years, so i figured that it was just "something nobody got around to"

we had a stale bot for a while and everything that did not receive "accepted" got closed automatically. Hence the tag

saibotk:
I'd say we move the discussion on improving and switching the arc drawing code to a new issue.
Because i certainly like the method from circles more than our approach and also the caching can be improved wastly.

But lets just get this one in regardless of that.

gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
saibotk: Worst case the arc is recalculated isnt it? So i would not really worry about it

gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
```diff
@@ -163,25 +167,36 @@ function TBHUD:Draw(client)
 		pos = ent:GetPos()
 		scrpos = pos:ToScreen()
 
-		if util.IsOffScreen(scrpos) or not ent:IsUsable() then continue end
+		local percent = math.Clamp((ent:GetNextUseTime() - CurTime()) / ent:GetDelay(), 0, 1)
+		local end_angle = 360 * percent
+
+		-- if util.IsOffScreen(scrpos) or not ent:IsUsable() then continue end
```
saibotk, line 173: What about this one?

gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
```diff
@@ -53,9 +53,10 @@ function TBHUD:CacheEnts()
 				["overrideTeam"] = overrideTeam,
 				["roleIntend"] = roleIntend,
 				["teamIntend"] = teamIntend,
-				["admin"] = admin,
+				["admin"] =  admin,
```
saibotk, line 56:
```suggestion
				["admin"] = admin,
```

lua/ttt2/libraries/targetid.lua
```diff
@@ -192,17 +192,20 @@ function targetid.HUDDrawTargetIDTButtons(tData)
 	tData:SetTitle(ent:GetDescription() == "?" and "Traitor Button" or TryT(ent:GetDescription()))
 
 	-- set the subtitle and icon depending on the currently used mode
-	if TBHUD.focus_but.admin and not TBHUD.focus_but.access then
+	if (TBHUD.focus_but.admin or TBHUD.focus_but.visible) and (not TBHUD.focus_but.access) then
```
saibotk, line 195:
```suggestion
	if (TBHUD.focus_but.admin or TBHUD.focus_but.visible) and not TBHUD.focus_but.access then
```

lua/terrortown/entities/roles/ttt_role_base/shared.lua
```diff
@@ -273,3 +273,13 @@ function ROLE:CanUseTraitorButton()
...
+
+---
+-- Returns whether a @{ROLE} can see traitor buttons.
+-- @return boolean
+-- @realm shared
+function ROLE:CanSeeTraitorButton()
+	local cv = GetConVar("ttt_" .. self.name .. "_traitor_button_visible")
+
+	return cv and cv:GetBool() or false
+end
```
saibotk, line 285: It might be better to check if the player can use the buttons here and return true if they can.

This would be more clear and would make using this in the render code easier, since we can eliminate the access or canSee stuff with just this one function call.

We might also need to add a note to the documentation, saying that this is ignored once a role is able to use traitor buttons.

saibotk (changes requested):

gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
EntranceJew: added back the offscreen check but not the usable check, since that really isn't a concern anymore

lua/terrortown/entities/roles/ttt_role_base/shared.lua
EntranceJew: ok, set it to call CanUse*() first

gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
EntranceJew: i changed it anyway, doesn't really hurt for now

saibotk (changes requested):
Please now adjust the parts of the code where we previously used stuff like `visible or access` and reduce them to only check if its visible, because we basically now do that with the latest change.

EntranceJew:
> Please now adjust the parts of the code where we previously used stuff like `visible or access` and reduce them to only check if its visible, because we basically now do that with the latest change.

i believe i've got this now? kind of hard to see with similar named variables of different purposes nearby

lua/ttt2/libraries/targetid.lua
```diff
@@ -84,7 +84,7 @@ function targetid.FindEntityAlongView(pos, dir, filter)
 
 	-- if the user is looking at a traitor button, it should always be handled with priority
 	if TBHUD.focus_but and IsValid(TBHUD.focus_but.ent)
-		and (TBHUD.focus_but.access or TBHUD.focus_but.admin) and TBHUD.focus_stick >= CurTime()
+		and (TBHUD.focus_but.visible or TBHUD.focus_but.access or TBHUD.focus_but.admin) and TBHUD.focus_stick >= CurTime()
```
saibotk, line 87:
```suggestion
		and (TBHUD.focus_but.visible or TBHUD.focus_but.admin) and TBHUD.focus_stick >= CurTime()
```

saibotk (changes requested):

saibotk (approved):

TimGoll (changes requested):
If we want to add that convar, we should change it. Right now I'd expect a traitor to have both of them enabled. Also what happens if both are enabled? Does this clash?

I'd favor a dropdown menu with "no interaction", "able to see", "able to interact" to make it more clear. That also means that it only needs a single convar.

Maybe someone else has an idea here, but I dislike the current implementation because it is confusing

EntranceJew:
> If we want to add that convar, we should change it. Right now I'd expect a traitor to have both of them enabled. Also what happens if both are enabled? Does this clash?
> 
> I'd favor a dropdown menu with "no interaction", "able to see", "able to interact" to make it more clear. That also means that it only needs a single convar.
> 
> Maybe someone else has an idea here, but I dislike the current implementation because it is confusing

this would be easier to execute once the migrations module is done and we can safely migrate convars to change their value semantics and names

TimGoll:
> > If we want to add that convar, we should change it. Right now I'd expect a traitor to have both of them enabled. Also what happens if both are enabled? Does this clash?
> > I'd favor a dropdown menu with "no interaction", "able to see", "able to interact" to make it more clear. That also means that it only needs a single convar.
> > Maybe someone else has an idea here, but I dislike the current implementation because it is confusing
> 
> this would be easier to execute once the migrations module is done and we can safely migrate convars to change their value semantics and names

so this means we should wait to merge this until migrations are done?

EntranceJew:
> so this means we should wait to merge this until migrations are done?

i don't think it needs to be this way, but overloading the existing `ttt_$x_traitor_button` convars to have the following value table:
```
0 = no use
1 = can use
2 = can only see
```
would be convoluted, but doing an even more drastic change without migrations would be headache inducing since it could potentially break every role; also the overhead of just migrating things to a dropdown where the discrete values still don't automatically come out of parameter files with some [unpleasantries](https://github.com/TTT-2/TTT2/blob/f0e5e9ce72bb7a99382d431346beb7df00314e33/gamemodes/terrortown/gamemode/client/cl_voice.lua#L428-L438)

TimGoll:
> > so this means we should wait to merge this until migrations are done?
> 
> i don't think it needs to be this way, but overloading the existing `ttt_$x_traitor_button` convars to have the following value table:
> 
> ```
> 0 = no use
> 1 = can use
> 2 = can only see
> ```
> 
> would be convoluted, but doing an even more drastic change without migrations would be headache inducing since it could potentially break every role; also the overhead of just migrating things to a dropdown where the discrete values still don't automatically come out of parameter files with some [unpleasantries](https://github.com/TTT-2/TTT2/blob/f0e5e9ce72bb7a99382d431346beb7df00314e33/gamemodes/terrortown/gamemode/client/cl_voice.lua#L428-L438)

why would it be convoluted? because 1 and 2 are switched?

EntranceJew:
> why would it be convoluted? because 1 and 2 are switched?

yes, because if someone were relying on the GetBool of the convar (they shouldn't be) an arbitrary check on it would then result in 2 would be truthy and let someone use it when they should only see it

it could lend itself to some confusion changing the "type" of a convar from a bool to an int with extended meaning

'cause later on down the line maybe instead of being able to see the state of a tbutton, a role might have even less privilege and only be aware of the position and effect of T buttons without seeing their state, and somehow that would become 3

gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
EntranceJew: should still be resolved

WardenPotato:
In pull request #1333 @EntranceJew asked me to look at the issue regarding DrawArc spassing out. It sounded a lot like facepunch/garrysmod-issues#2731 and facepunch/garrysmod-issues#5637 ive read these before so i digged a bit deeper.

You said that circles didnt suffer this issue, going over the code of circles and this repo the most notable difference i found was [stencil rendering](https://github.com/SneakySquid/Circles/blob/master/circles.lua#L230) this becomes important later.

Ive also looked at the actual code of DrawPoly between different source engine branches (hl2 2007, tf2, csgo) and while i did find minor differences in the order of mesh drawing i dont think this is a huge deal (who know maybe thats just it).

Tests have been performed on:
A Arch Linux machine running x64 gmod with a AMD 7900XTX
A windows 10 VM running x64 gmod with a GTX 1070 passed through

I have only been able to reproduce this issue on the GTX1070 vm with `gmod_mcore_test` set to `1`

For testing i decided to work in reverse by creating a absolutely minimal reproduction in a fresh sandbox installation, which can be found here (its messy): https://gist.github.com/WardenPotato/bf948dc6de813e996dca08787a3cb7bf

So with the findings of this only happening with a specific platform, GPU and with multithreading (atleast in my testing) ive decided to try the fix that was highlighted in facepunch/garrysmod-issues#2731 by Jova which can be found [here](https://gist.github.com/Jova1106/e7cad1dc162cf511d2279f139e61ea2e), and this fix did indeed work and confirms the earlier claims that circles did not suffer from this issue, which might just have been by pure chance because it has some fancy drawing types with stencils.

The short fix for this issue is to just apply the stencil clearing fix when rendering anything with DrawPoly:
```lua
function render.FullReset()
	render.SetStencilWriteMask(0xFF)
	render.SetStencilTestMask(0xFF)
	render.SetStencilReferenceValue(0)
	render.SetStencilCompareFunction(STENCIL_ALWAYS)
	render.SetStencilPassOperation(STENCIL_KEEP)
	render.SetStencilFailOperation(STENCIL_KEEP)
	render.SetStencilZFailOperation(STENCIL_KEEP)
	render.ClearStencil()
end

function surface.DrawPolyTable(tbl)
	for i = 1, #tbl do
		render.FullReset()
		surface.DrawPoly(tbl[i])
	end
end
```

Ive also had some contact with @EntranceJew on discord about this and he confirmed this fix worked for him which is great.

So the gist of this is: Circles resets stencil rendering which is what fixes these issues and its likely this wasnt even on purpose in Circles, so to fix it here we just need to reset it.

EntranceJew:
As mentioned, I can no longer reproduce the issue. General future advisory is to just reset the stencil anywhere drawpoly is used to draw potentially concave shapes, or just in general. Huge thanks to @WardenPotato for knocking it out of the park.

TimGoll:
> As mentioned, I can no longer reproduce the issue. General future advisory is to just reset the stencil anywhere drawpoly is used to draw potentially concave shapes, or just in general. Huge thanks to @WardenPotato for knocking it out of the park.
> 
> @saibotk @TimGoll should be good to go now

I will update the draw Arc function so you can use this

EntranceJew:
> > As mentioned, I can no longer reproduce the issue. General future advisory is to just reset the stencil anywhere drawpoly is used to draw potentially concave shapes, or just in general. Huge thanks to @WardenPotato for knocking it out of the park.
> > @saibotk @TimGoll should be good to go now
> 
> I will update the draw Arc function so you can use this

the latest commit adds it tho

TimGoll:
> > > As mentioned, I can no longer reproduce the issue. General future advisory is to just reset the stencil anywhere drawpoly is used to draw potentially concave shapes, or just in general. Huge thanks to @WardenPotato for knocking it out of the park.
> > > @saibotk @TimGoll should be good to go now
> > 
> > I will update the draw Arc function so you can use this
> 
> the latest commit adds it tho

Oh, nevermind then. I'll look at it tomorrow

TimGoll (commented):
LGTM. But I sill won't approve this because the convar situation is really confusing. Once this is resolved or moved to another PR, this is fine from my side

PR #1268, opened 2023-12-27
HUD-help Elements: Use global scale factor

TimGoll:
Added scaling to keyhelp and weapon HUD help elements. I also slightly tweaked how shadows are calculated in the draw library to prevent weird glitches at some scale values

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Changed
 
 - Updated the Turkish localization file (by @NovaDiablox)
+- Keyhelp and weapon HUD Help now use the global scale factor
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


TimGoll:
Can someone on a 1440p or 4k screen verify that this is good? on my FullHD anything bigger than 1.0 causes elements to overlap but this issue shouldn't exist on higher resolutions screens

xopez:
> Can someone on a 1440p or 4k screen verify that this is good? on my FullHD anything bigger than 1.0 causes elements to overlap but this issue shouldn't exist on higher resolutions screens

I can test that tomorrow evening

TimGoll:
> > Can someone on a 1440p or 4k screen verify that this is good? on my FullHD anything bigger than 1.0 causes elements to overlap but this issue shouldn't exist on higher resolutions screens
> 
> I can test that tomorrow evening

if you've got 1440p and the scalefactor is 1.4 then it should look good. At 4k the same holds true for a factor of 2. If you set it to anything besides these values, you probably have the same "issues" I got

xopez:
Seems to work. Tested on 3440x1440px
Scalefactor 3.0
image

Scalefactor 2.0
image

Scalefactor 1.8 (default size)
image

TimGoll:
The default size for your screen should be 1.3 or 1.4. We currently calculate it based on width, but that's a bug. Thanks for the feedback!

xopez (approved):

saibotk (approved):

Histalek (approved):

PR #1269, opened 2023-12-28
Corpses: Fix looking at disconnected player's corpses

EntranceJew:
you still can't check their roles from the `was_role` property because someone has to make that info available to the client

demo of the difference, the 2nd corpse would cause a crash:
https://github.com/TTT-2/TTT2/assets/5711436/23bb4363-886b-4b4c-8510-606f17add8ed

### DIFF
--- a/lua/ttt2/libraries/targetid.lua
+++ b/lua/ttt2/libraries/targetid.lua
@@ -473,9 +473,9 @@ function targetid.HUDDrawTargetIDRagdolls(tData)
 
 	local corpse_found = CORPSE.GetFound(ent, false) or not DetectiveMode()
 	local corpse_ply = corpse_found and CORPSE.GetPlayer(ent) or false
-	local role_found = (corpse_found and ent.bodySearchResult and ent.bodySearchResult.subrole) or (corpse_ply and corpse_ply:GetSubRole())
 	local binoculars_useable = IsValid(c_wep) and c_wep:GetClass() == "weapon_ttt_binoculars" or false
-	local roleData = (corpse_ply and corpse_ply:GetSubRoleData()) or roles.GetByIndex(role_found and ent.bodySearchResult.subrole or ROLE_INNOCENT)
+	local role_found = (corpse_found and ent.bodySearchResult and ent.bodySearchResult.subrole) or (corpse_ply and IsValid(corpse_ply) and corpse_ply:GetSubRole())
+	local roleData = (corpse_ply and IsValid(corpse_ply) and corpse_ply:GetSubRoleData()) or roles.GetByIndex(role_found and ent.bodySearchResult.subrole or ROLE_INNOCENT)
 	local roleDataClient = client:GetSubRoleData()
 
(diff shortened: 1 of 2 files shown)


lua/ttt2/libraries/targetid.lua
```diff
@@ -473,9 +473,9 @@ function targetid.HUDDrawTargetIDRagdolls(tData)
 
 	local corpse_found = CORPSE.GetFound(ent, false) or not DetectiveMode()
 	local corpse_ply = corpse_found and CORPSE.GetPlayer(ent) or false
-	local role_found = (corpse_found and ent.bodySearchResult and ent.bodySearchResult.subrole) or (corpse_ply and corpse_ply:GetSubRole())
 	local binoculars_useable = IsValid(c_wep) and c_wep:GetClass() == "weapon_ttt_binoculars" or false
-	local roleData = (corpse_ply and corpse_ply:GetSubRoleData()) or roles.GetByIndex(role_found and ent.bodySearchResult.subrole or ROLE_INNOCENT)
+	local role_found = (corpse_found and ent.bodySearchResult and ent.bodySearchResult.subrole) or (corpse_ply and IsValid(corpse_ply) and corpse_ply:GetSubRole())
+	local roleData = (corpse_ply and IsValid(corpse_ply) and corpse_ply:GetSubRoleData()) or roles.GetByIndex(role_found and ent.bodySearchResult.subrole or ROLE_INNOCENT)
```
TimGoll, line 478: I'm wondering if ROLE_INNOCENT is the right choice here. Maybe ROLE_NONE would be better? I'm asking because we used innocent as unknown role in the early days before we had the none role

EntranceJew: afaik i've never actually triggered this part of the code but i'll test it out 
image
the completely blank role icon is a little jarring if this condition DOES ever get tripped

TimGoll: understandable, thanks

TimGoll (approved):

lua/ttt2/libraries/targetid.lua
EntranceJew: i think if the icon for `ROLE_NONE` were similar to the `NO TEAM` icon in this context it'd be a little more suitable
image

saibotk, line 478:
```suggestion
	local role_found = (corpse_found and ent.bodySearchResult and ent.bodySearchResult.subrole) or (IsValid(corpse_ply) and corpse_ply:GetSubRole())
	local roleData = (IsValid(corpse_ply) and corpse_ply:GetSubRoleData()) or roles.GetByIndex(role_found and ent.bodySearchResult.subrole or ROLE_INNOCENT)
```

IsValid already checks for nil afaik

saibotk (changes requested):
Is the other issue you mentioned in the description about the missing role sync already tracked as an issue? If not we should do so :)

saibotk (approved):

Issue #1271, opened 2023-12-28
Popup text has no space in-between


mexikoedi:
Leave one of the following:

- Workshop version from Steam

## Describe the bug (mandatory)

I have noticed that the space in the popups for example if you use EPOP:AddMessage has changed.
Now the popup text has no space in-between and the letters go in each other, see image below.
This wasn't the case before and I haven't changed anything.

## To reproduce

Steps to reproduce the behaviour:

1. Load into world
2. Use any item or role which uses popups
3. Trigger popup
4. See the described issue 

## Expected behaviour

This should be like it was before. Space in-between the popup text so that the letters do not go in each other.

## Context (please provide as much as you can)

- Collection link of your Server: https://steamcommunity.com/sharedfiles/filedetails/?id=1822673195
- Screenshot: ![Screenshot of popup issue](https://cdn.discordapp.com/attachments/547119387392475136/1189710380197490729/Popup_space.png?ex=659f2722&is=658cb222&hm=9a3e27e2013d7694a53bbef5dc00cce6c75524619ae473cab7033a7b90354c7c&)

PR #1276, opened 2024-01-05
Fix: Too narrow line spacings

NickCloudAT:
This fixes to narrow line spacings occuring in eg. the EPOPUP.

Fixes https://github.com/TTT-2/TTT2/issues/1271


TimGoll (approved):
As discussed on discord this is a nice fix, thank you. The `lns` variable retuned 2 for a three line text, hence this change is suitable

saibotk (changes requested):
This is only a bandaid fix and we need to dig into why this is happening instead of just adding a +1 without a proper analysis.

There is also no reason for me to believe that this is even correct, since the original code looks correct already.

Also it would be nice to not just throw in a "as discussed on discord" which is not really anything someone else can easily check up on when reviewing.

TimGoll:
> Also it would be nice to not just throw in a "as discussed on discord" which is not really anything someone else can easily check up on when reviewing.

image

saibotk:
This is not a valid argument, just because of an observation of the result and a fix to the result being applied does not mean it is correct.

Why should this start at 0? It counts the table entries in the `lines` variable using `#lines` and should never start to count at 0.

TimGoll:
> This is not a valid argument, just because of an observation of the result and a fix to the result being applied does not mean it is correct.
> 
> Why should this start at 0? It counts the table entries in the `lines` variable using `#lines` and should never start to count at 0.

You're right, this is a valid point. Maybe the table is fucked up in some way because table.Count returns 3, # returns 2. Nick is investigating this right now

TimGoll:
image

for some reason # returns 3 for this table, table.Count returns 4. Weird.

saibotk:
Yes because there is the non-numeric key "TTT2" in there

TimGoll:
> Yes because there is the non-numeric key "TTT2" in there

no, this is not part of the table, the table start at `0`, the line before is a different print

saibotk:
Maybe me or you made the mistake there to start inserting elements at 0 instead of 1.

ZenBre4ker (approved):
This fix seems okay.
But the actual mistake is, that Lua is 1-indexed. I believe I read that by someone on discord, so if Im just repeating it, Im sorry. So # will ignore 0 entries, which I think are used as strings or maybe they are just ignored by #. -1,-2 also work in lua code.

So the actual mistake with the old code was in line 583 and 597 trying to insert the first string at line 0.

saibotk (approved):

PR #1277, opened 2024-01-05
Indent Subsettings

TimGoll:
Subsettings are now indented to make it easier to spot which settings are related. The master/slave and enable/disable system was also added to helpboxes.

image
image
image

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dlabel_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dlabel_ttt2.lua
@@ -93,6 +93,26 @@ function PANEL:Paint(w, h)
 end
 
 ---
+-- @param Panel slave
+-- @realm client
+function PANEL:SetMaster(master)
+	if not IsValid(master) then return end
+
+	self.master = master
+end
+
+---
+-- @param Panel slave
+-- @realm client
+function PANEL:GetMargin()
+	if not IsValid(self.master) then
+		return 0
+	end
+
+	return 10 + self.master:GetMargin()
+end
+
+---
 -- @param string strFont
 -- @realm client
 function PANEL:SetFont(strFont)
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dlabel_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dlabel_ttt2.lua
@@ -93,7 +93,7 @@ function PANEL:Paint(w, h)
 end
 
 ---
--- @param Panel slave
+-- @param Panel master
 -- @realm client
 function PANEL:SetMaster(master)
 	if not IsValid(master) then return end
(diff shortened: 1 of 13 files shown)


gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dlabel_ttt2.lua
```diff
@@ -92,6 +92,26 @@ function PANEL:Paint(w, h)
 	return true
 end
 
+---
+-- @param Panel slave
```
saibotk, line 96:
```suggestion
-- @param Panel master
```

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dlabel_ttt2.lua
```diff
@@ -92,6 +92,26 @@ function PANEL:Paint(w, h)
...
+function PANEL:SetMaster(master)
+	if not IsValid(master) then return end
+
+	self.master = master
+end
+
+---
+-- @return number
+-- @realm client
+function PANEL:GetMargin()
```
saibotk, line 107: Usually margin means top, left, right, bottom.

So I'd suggest renaming this to clarify that this is an indentation margin

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dpanel_ttt2.lua
```diff
@@ -236,4 +236,24 @@ function PANEL:GetTooltipFont()
 	return self.tooltip.font
 end
 
+---
+-- @param Panel slave
```
saibotk, line 240:
```suggestion
-- @param Panel master
```

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dpanel_ttt2.lua
```diff
@@ -236,4 +236,24 @@ function PANEL:GetTooltipFont()
...
+function PANEL:SetMaster(master)
+	if not IsValid(master) then return end
+
+	self.master = master
+end
+
+---
+-- @return number
+-- @realm client
+function PANEL:GetMargin()
```
saibotk, line 251: Same here

saibotk (changes requested):
I like these changes! 

Just some minor renaming stuff

saibotk (approved):

ZenBre4ker (approved):
LGTM.

As a note, I personally would try to get away from calling things master/slave. So here it could be parent or controller. So for example when the parent is deactivated, the element itself is deactivated.

TimGoll:
> LGTM.
> 
> As a note, I personally would try to get away from calling things master/slave. So here it could be parent or controller. So for example when the parent is deactivated, the element itself is deactivated.

This change would break other addons though, so I'm not a big fan of that. But in general I see your point here

PR #1281, opened 2024-01-08
docs: Updated readme

Histalek:
Restructured and reworded some paragraphs.
Removed unneeded whitespace.
Added alt-text for the banners.
Removed HTML newlines.
Linked bugreport and featurerequest templates instead of generic issue page.

The different paragraphs can probably be rewritten even more, but i mostly just wanted to cleanup the markdown file.


ZenBre4ker (approved):

README.md
```diff
@@ -1,125 +1,183 @@
...
-<br />
-<br />
+The setup is fairly easy. Just subscribe to TTT2 and the addons you want to use.
+
+It’s recommended to use ULX to set it up ingame, there’s an [ULX addon for TTT2](https://steamcommunity.com/sharedfiles/filedetails/?id=1362430347).
+Commands for the admin shop editor etc. are found in [this wiki article](https://docs.ttt2.neoxult.de/).
 
 <p align="center">
-	<img src="https://i.imgur.com/pbZHURE.png">
+    <img src="https://i.imgur.com/pbZHURE.png" alt="'Custom Role Support' Banner">
```
TimGoll, line 61: is this change from tab to spaces intentional?

saibotk:
Can we just scrap all the feature sections?

This is the Readme on GH and i think its much better to only point towards the relevant resources and keeping it short :)

README.md
Histalek: not really. i wanted to scrap all html tags from the markdown and played around with different things, but there's no way to center images without html.
Seems like this happened while changing that back (i have no formatter for markdown installed, hence nothing that could respect the .editorconfig)

Histalek:
> Can we just scrap all the feature sections?
> 
> This is the Readme on GH and i think its much better to only point towards the relevant resources and keeping it short :)

So everything after `## Setup` (excluding) up to `## The History of TTT2`?

Having those on a seperate 'Sales Pitch' webpage or something would make sense. We don't currently have a page like that, right?

saibotk:
I'd consider the Workshop page the "sales pitch". I think this is more than enough.

saibotk (approved):

TimGoll (approved):

PR #1282, opened 2024-01-09
Hats: Separate Logic from Player and Hat Model

EntranceJew:
i'd make the LANG.Msg send a paramterized message but there's no way to indicate that a token should be translated locally, and TryT doesn't exist on the server, so rest in peace


TimGoll:
What exactly is the goal of this PR?

EntranceJew:
some places set the hat at arbitrary hardcoded heights, others set it at dynamic heights.
it also allows for multiple different hats that inherit from this base hat, because the existing logic is designed to explicitly spawn on a player's head and never in the world as a found item.

TimGoll:
> some places set the hat at arbitrary hardcoded heights, others set it at dynamic heights.
> it also allows for multiple different hats that inherit from this base hat, because the existing logic is designed to explicitly spawn on a player's head and never in the world as a found item.

Ahh, that sounds like a good change then. Something like this should be in the description and changelog

TimGoll:
Could you please update the changelos so it reflect what you did here?

EntranceJew:
> Could you please update the changelos so it reflect what you did here?

i think i did that just now

TimGoll (approved):
Looks good to me

gamemodes/terrortown/entities/entities/ttt_hat_deerstalker.lua
```diff
@@ -21,25 +21,27 @@ end
...
-	self:SetMoveType(MOVETYPE_NONE)
-	self:SetSolid(SOLID_NONE)
-	self:SetCollisionGroup(COLLISION_GROUP_DEBRIS)
-
 	if SERVER then
-		self.Wearer = self:GetParent()
+		if IsValid(self:GetParent()) then
+			self:EquipTo(self:GetParent())
+		else
+			self:Drop()
```
ZenBre4ker, line 35: Did this happen in your tests?
Seems weird in the Initialize logic.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Decreased shooting accuracy while sprinting or in air (by @TimGoll)
 - A player whose weapons are stripped and cached will keep `weapon_ttt_unarmed` which means they keep their crosshair (by @TimGoll)
 - Grenades have icons
+- consolidated hat logic
```
ZenBre4ker, line 33:
```suggestion
- Extended hat logic
```
or
```suggestion
- Consolidated hat logic
```

ZenBre4ker (approved):
I like the changes. Just that one question and a typing error/more common word for non-native speakers.

gamemodes/terrortown/entities/entities/ttt_hat_deerstalker.lua
```diff
@@ -65,26 +88,17 @@ if SERVER then
...
-				self:SetAngles(ang)
-			else
-				local pos = ply:GetPos()
-				pos.z = pos.z + 68
-
-				self:SetPos(pos)
-			end
+			local pos, ang = ply:GetBonePosition(bone)
+			self:SetPos(pos)
+			self:SetAngles(ang)
```
saibotk, line 101: This is wrong, bone is not defined here.

saibotk (changes requested):

gamemodes/terrortown/entities/entities/ttt_hat_deerstalker.lua
EntranceJew: In the original code, it assumes it has a parent. In this new code, it allows it to be spawned without being assigned to a parent. It's an intentional change to permit it or other hat-based items to be placed as entities or spawned freely.

gamemodes/terrortown/entities/entities/ttt_hat_deerstalker.lua
EntranceJew: good catch, i clearly meant to factor out as much of the bone lookup as possible and missed this

saibotk (approved):
Thanks!

gamemodes/terrortown/entities/entities/ttt_hat_deerstalker.lua
saibotk: @ZenBre4ker please mark this as resolved if your question is answered and merge :)

PR #1283, opened 2024-01-09
Bindings: Enable in Roundendscreen

ZenBre4ker:
This is my idea of enabling bindings while vguihandlerframes are opened.
The roundendscreen can now be also closed with the binding.

Like this we can at least manually decide, when bindings should be blocked.
I believe a more sophisticated version can be done, that handles concrete binds for special cases, because this version still enables **all** Binds. So disadvantage is, that all ingame bindings still work inside the roundendscreen.

This should help for quickbuy menus planned in #988


CHANGELOG.md
```diff
@@ -7,8 +7,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Added
 
 ### Changed
+- dframe_ttt2 panels can now manually enable bindings while they are open
```
TimGoll, line 10:
```suggestion

- dframe_ttt2 panels can now manually enable bindings while they are open
```

CHANGELOG.md
```diff
@@ -7,8 +7,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
...
 
 ### Fixed
+- The roundendscreen can now be closed with the correct Binding
```
TimGoll, line 13:
```suggestion

- The roundendscreen can now be closed with the correct Binding
```

gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -412,7 +424,6 @@ function CLSCORE:Toggle()
 end
 
 bind.Register("toggle_clscore", function()
-	CLSCORE:ClearPanel()
```
ZenBre4ker, line 415: I couldnt find out why this was in here. But it had to be removed because of two things.
First, it was hiding the panel and therefore the toggle just showed the panel again.
Second, moving it to showpanel repositioned the panel everytime. I think its better if it always has the last position you gave it.

TimGoll:
Just for reference, did you check how the bodysearch panel does it?

https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/client/cl_search.lua#L185

ZenBre4ker:
> Just for reference, did you check how the bodysearch panel does it?
> 
> https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/client/cl_search.lua#L185

I didnt see it, but that's interesting. Because one feature I would like to implement is to make it safer to investigate by being able to keep walking while investigating. Like that you could get people to actually read more. Otherwise checking out corpses is always a risk.

ZenBre4ker:
This PR is ready. Just wanted to implement it and show it with one example.

gamemodes/terrortown/gamemode/client/cl_scoring.lua
```diff
@@ -138,6 +138,10 @@ function CLSCORE:CreatePanel()
 		self:HidePanel()
 	end)
 
+	frame.IsBlockingBindings = function(slf)
+		return self:IsBlockingBindings()
```
TimGoll, line 142: not sure about this one, why is there the function parameter `slf` if you access `self` in the line below?

TimGoll (commented):
there is one thing that looks strange to me

gamemodes/terrortown/gamemode/client/cl_scoring.lua
ZenBre4ker: Well the slf is frame specific, while the self is for CLSCORE.
I generally am used to write all arguments even if you dont use them.

TimGoll: ahh, imho is is confusing, but this is fine for me as well

TimGoll (approved):

Histalek (approved):

PR #1285, opened 2024-01-09
Shop: Backend rework part 1/X

ZenBre4ker:
To start my shop backend rework, I decided to make a new SHOP class on the client.
There I will move all relevant functions and data for the shop to work.
So that UI can just use them and doesnt need to decipher old code.

The biggest change is the way equipment changes are synced.
Instead of sending the whole table of current equipment and letting the client find out, what has to be resetted/added or removed, the server now sends an enum to declare it and only send the one change.

As this is working and completed, I thought its a good start for a first review, so the changes dont get too big.
I hope we can merge those small PRs quickly and I can keep on working ^^
Also I get your opinion on changes faster.


gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -166,6 +166,54 @@ local function TargetPlayer()
...
+	local ply = LocalPlayer()
+	if not IsValid(ply) then return end
+
+	ply.equipment_credits = net.ReadUInt(8)
+end
+net.Receive("TTT_Credits", UpdateCredits)
+
+local function UpdateEquipment()
+	local ply = LocalPlayer()
+	if not IsValid(ply) then return end
```
TimGoll, line 179: we tend to name the local player `client`. Also I'm wondering if there is any scenario where `LocalPlayer()` returns a nil value. Especially if you make sure that the netmessage is sent from `TTT2PlayerReady` hook or so

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -166,6 +166,54 @@ local function TargetPlayer()
...
+		local itemName = net.ReadString()
+		local item = items.GetStored(itemName)
+
+		if mode == EQUIPITEMS_ADD then
+			equipItems[#equipItems + 1] = itemName
+			
+			if item and isfunction(item.Equip) then
+				item:Equip(ply)
+			end
+		elseif mode == EQUIPITEMS_REMOVE then
```
TimGoll, line 206: how can you have three modes if the `net.ReadUInt` only has 1 bit?

gamemodes/terrortown/gamemode/client/cl_shop.lua
```diff
@@ -0,0 +1,25 @@
+---
+-- A class to handle all shop requirements
+-- @author ZenBre4ker
+-- @class Shop
+
+SHOP = SHOP or {}
```
TimGoll, line 6: I'm not sure if this should really be `SHOP`, shouldn't it be `shop`? I thought we only use uppercase if there are instances of this, e.g. `TARGET_DATA`

I personally would put it in `libraries` as well

gamemodes/terrortown/gamemode/client/cl_shop.lua
```diff
@@ -0,0 +1,25 @@
...
+	net.SendToServer()
+end
+
+---
+-- Get available credits of the local player
+-- @realm client
+function SHOP.GetAvailableCredits()
+	local ply = LocalPlayer()
+
+	return IsValid(ply) and ply.equipment_credits or -1
```
TimGoll, line 24: same here, is it possible for the player to be invalid?

gamemodes/terrortown/gamemode/client/cl_shop.lua
TimGoll, line 1: In general I think these functions should be shared or the library should be called differently. I think it is a bit confusing to have a library called `shop` that you can use to buy equipment that doesn't exist on the server. This is right now mostly a network wrapper.

TimGoll (changes requested):

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
ZenBre4ker: Actually I dont know. In this case I only copied the statement. I can remove it, if you can confirm I dont need it ^^

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
ZenBre4ker: oops, thanks for noticing. Apparently it worked, because I didnt test the Remove mode.

gamemodes/terrortown/gamemode/client/cl_shop.lua
ZenBre4ker: you are right it should be `shop`.
But I dont think it should be a library as it will handle mostly our internal stuff and IMHO libraries should handle general reusable stuff.

gamemodes/terrortown/gamemode/client/cl_shop.lua
ZenBre4ker: Currently Im only moving the stuff out of cl_equip to make room for the UI to go there only.
And we already have a server only sv_shop.lua, which handles the serverside.
Sharing would make only sense if it will be useful for both sides.

The shop seems to be doing a lot on the client side. But I will see in the future parts, if we can merge them. Right now I am mainly trying to refactor and cleanup. So it's easier to build the UI from it.

gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -390,16 +390,16 @@ local function CreateEquipmentList(t)
 	end
 
 	if #itms == 0 and not t.notalive then
-		ply:ChatPrint("[TTT2][SHOP] You need to run 'shopeditor' as admin in the developer console to create a shop for this role. Link it with another shop or click on the icons to add weapons and items to the shop.")
+		client:ChatPrint("[TTT2][SHOP] You need to run 'shopeditor' as admin in the developer console to create a shop for this role. Link it with another shop or click on the icons to add weapons and items to the shop.")
```
TimGoll, line 393: just FYI, the command doesn't exist anymore.

gamemodes/terrortown/gamemode/client/cl_shop.lua
TimGoll: This is probably out of scope for this PR, but I'd like it if `shop.BuyEquipment(eqID)` would be basically this function:

https://github.com/TTT-2/TTT2/blob/b6df5f1fcd75f822b211c00b9e754d6c86e85636/gamemodes/terrortown/gamemode/server/sv_shop.lua#L86

If it is called on the server, this code is executed. If it is called on the client, then a netmessage is sent that executes this on the server.

This is also how I did it for markerVision: 

https://github.com/TTT-2/TTT2/blob/3659e24ed038700e929da000114f336ed4ac98f7/lua/ttt2/libraries/marker_vision.lua#L34

You can call the function on the server or the client. If you call it on the server, it is executed there and then networked to the client to be executed there as well.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Changed
 
+- Refactored client shop logic into separate shop-class (in Progress)
```
```suggestion
- Refactored client shop logic into separate shop-class
```

i get what you want to say with this but i think it's just fine without the 'in progress' label

Histalek (approved):

gamemodes/terrortown/gamemode/client/cl_shop.lua
Histalek: why would anybody need to code this serverside?

TimGoll: Maybe an addon that buys a random weapon for you

Right now such an addon has to do all the checks themselves instead of just using the already existing (local) function

ZenBre4ker: I wont include here, yet. But I also like the idea and will keep it in mind ^^

TimGoll: Alriht, this is fine with me!

gamemodes/terrortown/gamemode/client/cl_equip.lua
TimGoll: could you please rewrite this sentence to reflect the new UI? Then it is fine from my POV

TimGoll (approved):
RE: Is valid checks

I'm pretty sure you don't need them, but I have no way of verifying that. So I guess you can keep them

gamemodes/terrortown/gamemode/client/cl_equip.lua
ZenBre4ker: This is not reworked yet, so I would do that in another PR.
I just changed all plys to clients in one run.

gamemodes/terrortown/gamemode/client/cl_shop.lua
EntranceJew: seconding this that not being able to read the shop symmetrically on the client/server made the pot of greed an utter pain in my ass to write, particularly the client-sided console command that makes a purchase randomly: https://github.com/EntranceJew/gmod-addons/tree/main/ttt2_pot_of_greedier

PR #1287, opened 2024-01-11
Added Delay to Weapon Select

TimGoll:
TTT2 has the ability to remove and cache a player's inventory to be later given back to the player. It is also designed to autoselect the old weapon. However this was broken all the time.

I solved this issue here with a 100ms timer. I don't like this hook, but I spend at least two hours trying to solve the underlying issue here but I was unable to. My best guess is that it has to do something with prediction as weapon selection aparantly is predicted. Adding a 100ms delay between giving and selecting weapons solved the issue for me.

Things I tried that also failed:

- running `lastinv` on the client
- using `input.SelectWeapon` on the client as this function is predicted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Fixed
 
 - The roundendscreen can now be closed with the correct Binding
+- Fixed cached weapons not being selected after giving them back to the owner
 
 ## [v0.12.3b](https://github.com/TTT-2/TTT2/tree/v0.12.3b) (2024-01-07)
 
--- a/gamemodes/terrortown/gamemode/server/sv_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player_ext.lua
@@ -1667,7 +1667,14 @@ function plymeta:RestoreCachedWeapons()
 	end
 
 	if self.cachedWeaponSelected then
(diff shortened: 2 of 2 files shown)


ZenBre4ker (approved):
Kinda makes sense with the prediction. I had a similar problem with my demonic sheep.

Just an idea:
.1s is probably enough for most people with a low ping. Maybe an additional check if weapons are restored could help before selecting?
So you could start a simple timer aka next frame and restart it if conditions are not met and abort it in case overall wait time reached .5s

TimGoll:
> Kinda makes sense with the prediction. I had a similar problem with my demonic sheep.
> 
> Just an idea: .1s is probably enough for most people with a low ping. Maybe an additional check if weapons are restored could help before selecting? So you could start a simple timer aka next frame and restart it if conditions are not met and abort it in case overall wait time reached .5s

I'm pretty sure the weapon is already _there_ on the server in the player inventory, it is just not ready to be selected. I'm not sure how I could check that.
Maybe I run `selectWeapon` and check if it is actually selected?

Your worry with the ping is something I share. However my thought process was "it was completely broken before, so maybe it is less broken now"

saibotk (approved):
Looks fine for me, did you try just delaying this by a single tick?

TimGoll:
> Looks fine for me, did you try just delaying this by a single tick?

yep. that was my first attempt. Didn't work though

PR #1288, opened 2024-01-11
Improved Weapon Caching

TimGoll:
While working on TTTC I noticed some issues with weapon caching and fixed them in this PR. I also added the same caching for items.

Things I did in this PR:

- fixed weapon caching
- set missing `builtin` flag
- fixed an item syncing issue with table remove
- added `weapon_ttt_unarmed` to base
- made sure that by default a player keeps `weapon_ttt_unarmed` when removing all their weapons to keep the crosshair etc
- added flag for weapons to hide pickup notification and added this flag to `weapon_ttt_unarmed`

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player_ext.lua
@@ -228,7 +228,10 @@ plymeta.RemoveEquipmentWeapon = plymeta.StripWeapon
 -- @realm server
 function plymeta:SendEquipment(mode, itemName)
 	if not mode then
-		ErrorNoHalt("[TTT2] Define an EQUIPITEMS_mode for plymeta:SendEquipment(mode, itemName) to work.")
+		ErrorNoHalt("[TTT2] Define an EQUIPITEMS_mode for plymeta:SendEquipment(mode, itemName) to work.\n")
+
+		debug.Trace()
+
 		return
 	end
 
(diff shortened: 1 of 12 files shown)


saibotk (approved):
I really like this!
Very nice!

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -228,7 +228,10 @@ plymeta.RemoveEquipmentWeapon = plymeta.StripWeapon
 -- @realm server
 function plymeta:SendEquipment(mode, itemName)
 	if not mode then
-		ErrorNoHalt("[TTT2] Define an EQUIPITEMS_mode for plymeta:SendEquipment(mode, itemName) to work.")
+		ErrorNoHalt("[TTT2] Define an EQUIPITEMS_mode for plymeta:SendEquipment(mode, itemName) to work.\n")
+
+		debug.Trace()
```
Histalek, line 233: This probably shouldn't be here :D

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1625,27 +1628,36 @@ end
...
-	self.cachedWeaponInventory = cachedWeaponInventory
-	self.cachedWeaponSelected = WEPS.GetClass(self:GetActiveWeapon())
 
+	-- we have to use this hack here instead of StripWeapon because StripWeapon calls
+	-- OnDrop which is not intended for the weapon caching
 	self:StripWeapons()
+
+	if not forceAll then
+		self:Give("weapon_ttt_unarmed")
+	end
```
Histalek, line 1660: this feels somewhat weird, but i guess it's the only way to do this :/

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: This is intentional, because I got the error in the console and had no idea where it was called from. This way it is easier to debug imho

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: I had loads of issues otherwise. I agree with you, but I was unable to find a different solution

TimGoll:
should be done @Histalek

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
Histalek: why not use [ErrorNoHaltWithStack](https://wiki.facepunch.com/gmod/Global.ErrorNoHaltWithStack) in that case?

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
Histalek: the biggest issue i have with this is if something deliberately removes the unarmed weapon in another context. Because then this might lead to confusing interactions.

Can we guard against this? or do we just assume that nobody will remove the unarmed weapon?

TimGoll: if someone removed the unarmed weapon I think they have to handle it themselves

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: Because I wasn't aware that this exists, thanks. I changed it

Histalek (approved):

PR #1289, opened 2024-01-11
Fixed Last Seen Entity

TimGoll:
Added a check for a public policing search


saibotk (approved):

TimGoll:
I can't verify this as this only works if a player has targetID and bots don't have targetID. But I guess it should work

Histalek (approved):

PR #1290, opened 2024-01-11
Shop: Backend rework part 2/X

ZenBre4ker:
Here comes the second smaller part, as always there shall be more rework parts, every single one should be working by itself and is not meant to be complete. So try to concentrate on the new statusCode handlings and not single messages, as they are just taken from the old shop implementation and should be replaced in a separate language rework.

- Combined client, server and shared code of Buyable check
- Share Equipment Buy, Credits Transfer and Shop Reroll
- Use plymeta:GetCredits() instead of new shop.GetCredits(ply)
- Move out error messages to displaying sources
  - this shall be handled by the UI instead of the shop-class
  - Its doubled and not supported by translations, this shall be done in another rework to not clutter this one here

To-Dos:

- [x] Test it
- [x] Order and correctly number statusCodes


gamemodes/terrortown/gamemode/shared/sh_shop.lua
```diff
@@ -0,0 +1,320 @@
...
+}
+
+---
+-- Check if an equipment is currently buyable for a player
+-- @param Player ply The player to buy the equipment for
+-- @param string equipmentId The name of the equipment to buy
+-- @return bool True, if equipment can be bought
+-- @return number The shop.statusCode, that lead to the decision
+-- @realm shared
+function shop.IsEquipmentBuyableFor(ply, equipmentId)
```
Histalek, line 34:
```suggestion
function shop.CanBuyEquipment(ply, equipmentId)
```
just a suggestions, but how about keeping this in line with `RerollShop` and `CanRerollShop`?

Histalek (changes requested):
I like where this is going :)

Histalek (approved):

gamemodes/terrortown/gamemode/server/sv_shop.lua
```diff
@@ -66,150 +57,49 @@ function GM:TTT2OrderedEquipment(ply, cls, isItem, credits, ignoreCost)
...
+	elseif statusCode == shop.statusCode.INVALIDID then
+		ErrorNoHalt("[TTT2][ERROR] No ID was requested by:", ply)
+	elseif statusCode == shop.statusCode.NOTBUYABLE then
+		LANG.Msg(ply, "This equipment cannot be bought.", nil, MSG_MSTACK_ROLE)
+	elseif statusCode == shop.statusCode.NOTENOUGHPLAYERS then
+		LANG.Msg(ply, "Minimum amount of active players needed.", nil, MSG_MSTACK_ROLE)
+	elseif statusCode == shop.statusCode.LIMITEDBOUGHT then
+		LANG.Msg(ply, "This equipment is limited and is already bought.", nil, MSG_MSTACK_ROLE)
+	elseif statusCode == shop.statusCode.NOTBUYABLEFORROLE then
+		LANG.Msg(ply, "Your role can't buy this equipment.", nil, MSG_MSTACK_ROLE)
```
TimGoll, line 80: I think we should localize these, at least these that are in `LANG.Msg` that already supports localization. I think they also should all be moved to the `MSTACK`, there is no reason to have two of them in chat and `LANG.Msg` also supports the parameter translation needed for `equipmentID`

lua/ttt2/extensions/util.lua
```diff
@@ -150,6 +150,17 @@ function util.GetNextAlivePlayer(ply)
...
+---
+-- Returns the list of active @{Player}s that are not forced spectators
+-- @return table List of active @{Player}s
+-- @realm shared
+function util.GetActivePlayers()
+	return util.GetFilteredPlayers(
+		function(ply)
+			return IsValid(ply) and not ply:GetForceSpec()
+		end)
+end
```
TimGoll, line 162: imho this should be:
```suggestion
function util.GetActivePlayers()
	return util.GetFilteredPlayers(function(ply)
		return IsValid(ply) and not ply:GetForceSpec()
	end)
end
```

or:
```suggestion
function util.GetActivePlayers()
	return util.GetFilteredPlayers(
		function(ply)
			return IsValid(ply) and not ply:GetForceSpec()
		end
	)
end
```

gamemodes/terrortown/gamemode/server/sv_shop.lua
ZenBre4ker: I thought that someone would say these. IMHO we shouldnt do this in this PR. I try to keep functionality exactly the same. Maybe fix an error if I see one. But generally I try just sort everything and we can still modernize it later. Makes thes PRs faster to review.

gamemodes/terrortown/gamemode/shared/sh_shop.lua
```diff
@@ -0,0 +1,320 @@
...
+		timer.Simple(0.5, function()
+			if not IsValid(ply) then return end
+
+			net.Start("TTT_BoughtItem")
+			net.WriteString(equipmentId)
+			net.Send(ply)
+		end)
+
+		if GetGlobalBool("ttt2_random_shop_reroll_per_buy") then
+			shop.RerollShop(ply)
```
ZenBre4ker, line 212: Made a mistake here, this shouldnt cost any credits and should be forced either way. Will fix it.

TimGoll (approved):

Histalek (approved):

TimGoll:
@ZenBre4ker Took a look at your recent changes and it still looks fine to me. Feel free to merge

PR #1294, opened 2024-01-15
Shop: Favorites rework

ZenBre4ker:
Favorites were saved with an old manual sql query, therefore there was no Primary Key, which I believe is necessary for orm features.
It also saved the current steamid and subrole.

I find it confusing, that you favorite something and you change roles and it's "suddenly" not a favorite anymore, which always seemed like a bug, as it didnt occur to me, that it might be role-dependent.
Also saving it for the steamid seems kinda unnecessary unless you share the same computer and gmod installation, but not steam accounts.

IMHO, I think we should just fuse it all together, use the newer orm features, make a new sql-entry and only save raw favorites, without any further distinction. 

Through migration old favorites are taken from the ttt_bem_fav sql table. I merge favorites of all saved players and roles into one.


gamemodes/terrortown/gamemode/client/cl_shop.lua
```diff
@@ -4,3 +4,58 @@
...
+-- @param bool isFavorite If the equipmentId is a favorite
+-- @realm client
+function shop.SetFavoriteState(equipmentId, isFavorite)
+	local favOrm = GetFavoritesORM()
+	local favoriteItem = favOrm:Find(equipmentId)
+
+	if isFavorite and not favoriteItem then
+		favoriteItem = favOrm:New({
+				name = equipmentId,
+			})
```
TimGoll, line 61: This should probably look like this, shouldn't it?

```suggestion
		favoriteItem = favOrm:New({
			name = equipmentId,
		})
```

TimGoll (approved):
I agree with you. There are so many changes here that keeping the favorites is probably a pain.

About the favorites per role: In theory it is nice to have different favorites depending on the play style of the role. And I'm sure that there will be a few people on discord defending this.
BUT I always had the same feeling you did, that it feels like a bug and is pretty annoying as most roles have the same shop anyways.

Maybe we should split the favorites between dete and traitor? Not sure though if it is worth it.

TimGoll:
FYI, you probably broke the spectator "edit shop to select favorites" UI with this PR. Just so that you are aware that it is a thing that has to be fixed

ZenBre4ker:
> FYI, you probably broke the spectator "edit shop to select favorites" UI with this PR. Just so that you are aware that it is a thing that has to be fixed

Not sure what you mean, but it shouldnt change anything besides losing your last favorites.
There is no instance, that uses the old table beside those 4 local functions and inside the UI I only replaced them. Behavior should be the same 🤔

TimGoll:
> > FYI, you probably broke the spectator "edit shop to select favorites" UI with this PR. Just so that you are aware that it is a thing that has to be fixed
> 
> Not sure what you mean, but it shouldnt change anything besides losing your last favorites. There is no instance, that uses the old table beside those 4 local functions and inside the UI I only replaced them. Behavior should be the same 🤔

In this UI you can select the role you want to edit your favorites for. Not sure how your code will handle this, as you can no longer select the role for the favorite

ZenBre4ker:
> > > FYI, you probably broke the spectator "edit shop to select favorites" UI with this PR. Just so that you are aware that it is a thing that has to be fixed
> > 
> > 
> > Not sure what you mean, but it shouldnt change anything besides losing your last favorites. There is no instance, that uses the old table beside those 4 local functions and inside the UI I only replaced them. Behavior should be the same 🤔
> 
> In this UI you can select the role you want to edit your favorites for. Not sure how your code will handle this, as you can no longer select the role for the favorite

Ah, yeah that will update the favorites for all and is therefore still working, just maybe not as intended before. I just removed the distinction.

gamemodes/terrortown/gamemode/client/cl_shop.lua
```diff
@@ -4,3 +4,63 @@
...
+local function GetFavoritesORM()
+	local favorites = shop.favorites
+
+	if istable(favorites.orm) then
+		return favorites.orm
+	end
+
+	-- Create Sql and orm table if not already done
+	sql.CreateSqlTable(favorites.databaseName, favorites.savingKeys)
+	favorites.orm = orm.Make(favorites.databaseName)
```
saibotk, line 27: I'd say just do this when initializing the orm variable.

saibotk (changes requested):
Otherwise nice!

gamemodes/terrortown/gamemode/client/cl_shop.lua
saibotk: I mean just move all of this to the variable declaration, i see no point in having this initialized on demand.
Its much cleaner if it is done when the code is loaded. It's not like any of this is a heavy operation and even then it is pretty unusual to do something like this in a getter function. Imagine using a Web API and the first request of a day always takes 3.5s instead of the usual 50ms that would be weird right? Not that this is the case here, since it is very lightweight, but its just odd to initialize something within the getter method.

saibotk (approved):

ZenBre4ker:
I think I wait for #1300.

ZenBre4ker:
Maybe take a look at the first ever migration file.

gamemodes/terrortown/gamemode/client/cl_shop.lua
```diff
@@ -5,6 +5,50 @@
 
 shop = shop or {}
 
+shop.favorites = shop.favorites or {}
+shop.favorites.databaseName = "ttt2_shop_favorites"
+shop.favorites.savingKeys = {}
+
+sql.CreateSqlTable(shop.favorites.databaseName, shop.favorites.savingKeys)
```
ZenBre4ker, line 12: This needs to be removed.

gamemodes/terrortown/gamemode/client/cl_shop.lua
```diff
@@ -5,6 +5,50 @@
...
+
+shop.favorites.orm = orm.Make(shop.favorites.databaseName)
```
ZenBre4ker, line 14: Not sure how to handle this now. It actually needs an initialization after migration. In all other cases that should work.
Maybe only migration I reset the orm.

lua/terrortown/migrations/2024_02_16_23_10_favorites_migration.lua
```diff
@@ -0,0 +1,50 @@
...
+
+---
+-- Checks old favorites table and merges them into the new table
+-- @realm client
+return function()
+    local sqlCommand = "CREATE TABLE " .. newDatabaseName .. " (name TEXT PRIMARY KEY)"
+
+    if not sql.TableExists(newDatabaseName) and sql.Query(sqlCommand) == false then
+        error("[TTT2] Couldn't create sql table in migrations. Query failed.")
+    end
```
Histalek, line 19: i would split this up, as both failures a somewhat different.

if the table already exists (not created by us) we can tell the user this and allow them to delete the offending database table. Or even harsher we could claim this table (or rather all tables prefixed with `ttt2_` and drop or rename the table and go forward.
This would allow for a nicer errorhandling that doesn't result in a non-working gamemode.

The sql.Query could fail for various Reasons, if we take away the `table already exists` error there is no reasonable action we could take to make this work. Maybe we can get the sql error message and display it, but that's as far as it can go.

lua/terrortown/migrations/2024_02_16_23_10_favorites_migration.lua
```diff
@@ -0,0 +1,50 @@
...
+    if not sql.TableExists(oldDatabaseName) then
+        return
+    end
+
+    local retrieveOldSqlDataCommand = "SELECT * FROM ttt_bem_fav"
+    local oldFavoriteEquipments = sql.Query(retrieveOldSqlDataCommand)
+
+    if oldFavoriteEquipments == false then
+        return
+    end
```
Histalek, line 30: shouldn't we at least log a dev message that old equipment favourites could not be accessed?

additionally we would need to return on `nil` as otherwise the `pairs` call in the for loop below will fail

gamemodes/terrortown/gamemode/client/cl_shop.lua
Histalek: i'm not exactly sure what you mean here.
As long as the migration has run before this gets called everything should be alright, or am i missing something?

gamemodes/terrortown/gamemode/shared/sh_shop.lua
```diff
@@ -30,6 +30,15 @@ shop.buyTable = shop.buyTable or {}
 shop.globalBuyTable = shop.globalBuyTable or {}
 shop.teamBuyTable = shop.teamBuyTable or {}
 
+---
+-- Initializes all necessary variables for the shop
+-- @realm shared
+function shop.Initialize()
+    if CLIENT then
+        shop.favorites.orm = orm.Make(shop.favorites.databaseName)
```
saibotk, line 38: @Histalek does the orm really need a late init? Sounds like something we should address in the future and should rather not be needed

lua/terrortown/migrations/2024_02_16_23_10_favorites_migration.lua
```diff
@@ -0,0 +1,100 @@
...
+-- @realm client
+return function()
+    Dev(1, "Starting to migrate shop favorites.")
+
+    -- Drop existing table
+    local result = nil
+    if sql.TableExists(newDatabaseName) then
+        Dev(1, "SQL table", newDatabaseName, "already exists. It will be dropped!")
+        result = sql.Query("DROP TABLE " .. sql.SQLStr(newDatabaseName))
+    end
```
saibotk, line 21: I think you can safely assume this to not exist

saibotk (changes requested):

TimGoll:
I see that a lot of stuff changed since my last review. Once @saibotk comments are resolved, I can take a second look at it again as well

gamemodes/terrortown/gamemode/shared/sh_shop.lua
ZenBre4ker: Currently it does and gets all the necessary information out of the sql table. If it does not exist, it can't do that. So it is only needed for the first time migrations are done. 
https://github.com/TTT-2/TTT2/blob/master/lua/ttt2/libraries/orm.lua#L25-L32

lua/terrortown/migrations/2024_02_16_23_10_favorites_migration.lua
ZenBre4ker: This is safer though and we make sure the table has the correct format.

gamemodes/terrortown/gamemode/shared/sh_shop.lua
Histalek: > @Histalek does the orm really need a late init? Sounds like something we should address in the future and should rather not be needed

The orm does need to be initialized after any migrations, yes. `orm.Make` creates its data based on the current schema of the given database table, if you change the schema afterwards you are in for a bad time.

The explizit `Initialize` here is only necessary to guarantee that migrations have run.

lua/terrortown/migrations/2024_02_16_23_10_favorites_migration.lua
Histalek: > I think you can safely assume this to not exist

That would be a fair assumption to make, but given that we can handle this failure case we should do so.

gamemodes/terrortown/gamemode/shared/sh_shop.lua
saibotk: I understand but dont like it gotta see if we can redo that later

saibotk (approved):

ZenBre4ker:
Waiting now for Release 0.13 and Tim's second review.

lua/terrortown/migrations/2024_02_16_23_10_favorites_migration.lua
```diff
@@ -0,0 +1,100 @@
+--- @ignore
+-- Migration file for shop favorites
+if SERVER then
+    return function() end
+end
+
+local oldDatabaseName = "ttt_bem_fav"
+local newDatabaseName = "ttt2_shop_favorites"
+---
```
TimGoll, line 9:
```suggestion
local newDatabaseName = "ttt2_shop_favorites"

---
```

TimGoll (approved):
Looks good to me. I also have no problem with the removal of the favorite distinction between different roles. However I'm a bit confused, didn't you show screenshots on discord that showed that you kept the distinction?

ZenBre4ker:
Those screenshots showed, that role specific weapon favorites stay favorites in their shops and weapons that werent favorited for a role, but by another, are now included.

ZenBre4ker:
I guess it's finally time to merge the first migration :D Let the bug reports rain down on me 😈

PR #1297, opened 2024-01-15
grenade rework 1/365

EntranceJew:


TimGoll (approved):

saibotk:
Nobody noticed that we are missing the force download calls here

EntranceJew:
a lot of the base vgui is missing from the force download, should probably be a CI task to validate it

TimGoll:
oof, you're right. But I think quite a few materials are missing from that list anyway, we have to fix it and add a checker

PR #1300, opened 2024-01-16
Migrations: Base for further implementing migrations

ZenBre4ker:
Im looking at you #673 this might help 😏 

So I thought to just start it.
There is probably a lot of room for improvement.
I tested it and it generally does what it says.
The sample migration could be deleted, but I guess it's nice to have an example.
The command structure could be made as an own library and also allow standardized commands.

Problems:
- Downgrade will most probably not work without the migration files.
  - Maybe they could be handed out as a separate addon
  - Functions inside migrations also would need to be independent of library changes
  - One could think of an own small language parser for storing functions as a string in a persistent database to solve it
- Migration will only work if the database saving the version is created before the first migration
  - Therefore it is useful to release this as early as possible
- General changes should be standardized commands, that don't need to be rewritten every time
  - Like Renaming a database
  - Renaming a convar
  - Changing default values etc.

P.S.: Let the discussion begin 🔥


saibotk:
Before i do a more thorough review, i would propose to replace the explicit version fields with a system found in Laravels Migrations and other systems too: Using the migration file name as a key. Those usually also follow the standard of being prefixed with the current date and time to avoid conflicts. 
Also you would only need to have a simple migrations table with a column for the migration names and then only run those that are not yet found in the database.

I think this simplifies a lot, what do you think?

ZenBre4ker:
I like that idea with the names.

One problem I see is, how do you handle order of files? For example what if we later change something back again and someone that really late updates gets two migrations in the wrong order, wouldnt we need to additionally keep track of that?

I liked the idea, that every migration file is versioned parallel to the gamemode itself.

ZenBre4ker:
Edit: Just reread the comment with filename date and time. That would of course do the ordering.

I just dont like what that means in case we make a mistake in spelling. If we could automate those namings it would be fine.
But we probably have to name the files ourselves and for correct orders, they always need to be correct. Change the name later and it will be applied again.

saibotk:
Well no real difference to correctly having to name something else in code, right?
I think this approach is easier than additional code for handling the order.
The only thing that comes to my mind would be that we would need to ensure that the order is properly enforced, because of some differences in the results from some gmod file apis on different OS's.

@Histalek what do you think?

ZenBre4ker:
I guess it's just harder for me personally to handle files instead of tables in lua/gmod.

Otherwise the problems are the same and files sounds better, so heres a proposal.
My current imaginary draft would be:
- Sort files by name
- Find the startpoint of the latest new migrations, which filenames are not stored
- Execute and save them

Handle Edge cases:
- No database-file
  - Create it and save it as not executed
  - Dont execute them
- Not in database, but also dated before any saved one
  - Ignore and write to database as not executed

Database form:
- Filenames as Primary Key
Keys:
- Execution date
- Execution was success?

saibotk:
Much simpler:

1. Find all files in a directory, sort them
2. For each file: Check if in migrations table, otherwise execute
3. After each successful execution insert the filename to the migrations table

Nothing more needed :)

ZenBre4ker:
I just think that this behaviour can lead to errors. Thats why I outlined two edge-cases, I believe should be considered.

If we use a false name and mess up the order, they should be ignored.

If a new server is set up, migration is not needed.

saibotk:
I really think this is a pretty battle tested way of doing this and you might get problems with your approach since sometimes migrations might come in that are older than some others for example if merging a PR that was created earlier but gets merged in much later.

Well new servers might be a thing we need to discuss, because something that we missed is how we want to handle initial states (for example database structures). 
In all other systems the table schema is also only created via migrations. This is in my opinion something we should also transition to, since it makes the whole thing much easier to deal with instead of defining the schema in various code places and also migrating for older versions. 
Instead, all migrations are also executed for a new server and we dont have to differentiate between new and old servers.

EntranceJew:
> Much simpler:
> 
> 1. Find all files in a directory, sort them
> 2. For each file: Check if in migrations table, otherwise execute
> 3. After each successful execution insert the filename to the migrations table
> 
> Nothing more needed :)

this is how a lot of migration tools do it, generate a unix timestamp with a slug describing what it does. we don't have to worry about users going back in time, just like the date comments in the lang files.

ZenBre4ker:
1) Okay, but then I suggest we block them directly. Old Prs need to be updated if merged later. Actually could be included in the CI to either only check that or just do the naming for us.

2) I agree, with that way of creating databases and convars?!, migrations make more sense and can be always executed.
2.1) When doing it like this, we should add an easy way to execute them as commands during tests.

If we agree on a plan, I can live with those solutions and would integrate them.

saibotk:
No there would be no need to block them, it is perfectly fine for them to be "older", it does not matter.
And if you depend on some migrations that came in later, you can still just adjust the file name to a newer date.
This is perfectly fine. I think what you are trying to do is a bit too overprotective here. 
I don't really see the pain points in doing this.

Histalek:
There is no blocking of 'old' migrations needed. As soon as we start using migrations for a table (especially its initialization) those migrations control the table schema fully. No matter how new/old a server is, every single server will run every migration for each table and therefore end up with the same schema as the whole path to get there is the same.

As for the filename order i think if we start the file with a Unix-timestamp and keep that one distinct from other files we shouldn't run into issues. At least i'm not aware of any OS that fucks up file ordering for numbers.

Additionally i don't think we will have problems with wrongly ordering migrations because:
1. migrations aren't something that is used very often as either the migration itself or the downgrade is very likely destructive.
2. even if we screw up the ordering those migrations would need to compete in some way for this to matter e.g. modify the same table schema or read data from a table which another migration just changed.

The only thing to keep in mind would be to backup the DB while testing and discourage Users from using branches/PRs that introduce migrations. Besides that we should be fine IMO

EntranceJew:
Also I think this should introduce *something* which migrates convar values as a utility, like a simple:
```lua
function migrate.GetConVar(oldName)
  --[[ something! ]]
  return oldValue
end
```
which returns the old value of the convar, parsed from the vdf or the cfg, or by temporarily creating and removing the convar, because if a convar is renamed then it won't exist for _G.GetConVar b/c _G.CreateConVar has never run, e.g. it will just be treated like a brand new convar.
it should just be a utility for getting the old value, as the way to handle an old convar may be to port it to a new type, into the database, or a simple rename -- no assumptions.

ZenBre4ker:
I can live with that.
Didnt think about cfg files @EntranceJew , so thanks for the proposal. I think we definitely try to include Convars the same as databases and support renaming etc.

Last try to be overprotective 😂 But I accept it anyway:
How about automatically commenting on PRs with added migration files, to shortly state the position in the new filelist?
So we can at least decrease human errors in reviews.

It's not important, but I actually always wanted an opportunity to build a CI tool, so I see if I find motivation and time to try it, else we drop it.

saibotk:
Pheww i think if you want to build something we can include you in our ci revamps and maybe find something more useful 😅

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,184 @@
...
+
+	local migrationSuccess = true
+	local errorMessage = ""
+	local databaseORM = migrations.GetORM()
+
+	for fileName, commands in SortedPairs(migrations.commands) do
+		local fileInfo = databaseORM:Find(fileName)
+
+		-- As ORM currently gives out strings check it
+		if fileInfo and (fileInfo.wasSuccess == true or fileInfo.wasSuccess == "true") then continue end
```
ZenBre4ker, line 103: I guess it's also time to move savingKeys Conversion to ORM 🥲

TimGoll (approved):
LGTM, but I'm a bit OOTL, so take this with a grain of salt

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,184 @@
...
+				wasSuccess = migrationSuccess,
+				wasUpgrade = true
+			}):Save()
+		end
+
+		if not migrationSuccess then break end
+	end
+
+	if not migrationSuccess then
+		ErrorNoHalt("[TTT2] Migration failed. Error:\n" .. tostring(errorMessage) .. "\n")
```
Histalek, line 133:
```suggestion
		ErrorNoHaltWithStack("[TTT2] Migration failed. Error:\n" .. tostring(errorMessage) .. "\n")
```
This should probably never happen, but if it ever happens having the stacktrace might be valueable.

Then again, can we even ignore a failed migration? Given that the system is not in a known good state we shouldn't continue, right?

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,184 @@
...
+	else
+		MsgN("[TTT2] Migration was successful.")
```
Histalek, line 135:
```suggestion
		MsgN("[TTT2] Successfully migrated: ", fileName)
```
Some logging systems allow for out of order writes. So just make it clear what was done.

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,184 @@
...
+-- @param function downgrade The function that is used for reverting operations of this command
+-- @return @{command} A command containing necessary functionality for forward and backward operations
+-- @realm shared
+function migrations.CreateCommand(states, upgrade, downgrade)
+	if not istable(states) or not isfunction(upgrade) or not isfunction(downgrade) then
+		ErrorNoHalt("[TTT2] Couldn't create migrations command. Missing states-table, upgrade- or downgrade-function.\n")
+		states = {}
+		upgrade = function(cmd) return end
+		downgrade = function(cmd) return end
+	end
```
Histalek, line 61: should we really create an empty migration here? This should only ever happen in dev imo

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,184 @@
...
+		fileloader.LoadFolder(migrations.folderPath .. "server/", false, SERVER_FILE, function(filePath, folderPath)
+			MsgN("Added TTT2 server migration file: ", filePath)
+			fileIncludedCallback(filePath, folderPath)
+		end)
+	end
+
+	fileloader.LoadFolder(migrations.folderPath .. "shared/", false, SHARED_FILE, function(filePath, folderPath)
+		MsgN("Added TTT2 shared migration file: ", filePath)
+		fileIncludedCallback(filePath, folderPath)
+	end)
```
Histalek, line 181: Just wondering. What would be a usecase for shared migrations?

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,184 @@
...
+		for j = 1, #commands do
+			migrationSuccess, errorMessage = pcall(commands[j].Upgrade, commands[j])
+
+			if not migrationSuccess then break end
+		end
+
+		local date = os.date("%Y/%m/%d - %H:%M:%S", os.time())
+
+		if fileInfo then
+			fileInfo.date = date
```
Histalek, line 116: date isn't really describing what this information is. something like `executed_at` / `executedAt` might be a better description for this (especially in the database)

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,184 @@
...
+
+---
+-- Adds a migration command, that is executed on migration
+-- @Note Only used in `terrortown/migrations` and corresponding `client`/`server`/`shared` folders
+-- @param @{command} The command to execute on migration
+-- @return bool True, if adding was successful
+-- @realm shared
+function migrations.Add(command)
+	if not istable(command) or not command.isCommand then
+		ErrorNoHalt("[TTT2] Couldn't add migration command. Missing version or command.\n")
```
Histalek, line 77: how does the missing version come into play here? or is this a leftover?

Histalek (commented):
Not a full review yet, just some thoughts.
I'll probably need to run some migrations myself to get a better feel for this.

But it already looks quite good :)

lua/ttt2/libraries/migrations.lua
ZenBre4ker: Same ConvarNames or database entries. Not sure if thats important though. Just wanted to be inclusive for the general name schemes.

lua/ttt2/libraries/migrations.lua
ZenBre4ker: Leftover.

lua/ttt2/libraries/migrations.lua
ZenBre4ker: Maybe a command triggering an error inside the migration is better?

We somehow need to help us developing migrations and make sure they are complete.

lua/terrortown/migrations/shared/2024_01_22_10_26_sampleMigration.lua
```diff
@@ -0,0 +1,19 @@
...
+	print(cmd.endMessage)
+end
+
+migrations.Add(
+	migrations.CreateCommand(
+		states,
+		Upgrade,
+		Downgrade
+	)
+)
```
saibotk, line 19: Id rather imagine an auto load of the folder and something we do already eg with hudelements and just have some global table or just return the up and down methods from the file.

Also i dont get what this states thingy is for.

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,184 @@
...
+			fileInfo.date = date
+			fileInfo.wasSuccess = migrationSuccess
+			fileInfo.wasUpgrade = true
+			fileInfo:Save()
+		else
+			databaseORM:New({
+				name = fileName,
+				date = date,
+				wasSuccess = migrationSuccess,
+				wasUpgrade = true
```
saibotk, line 125: This field is not really needed and true anyways

lua/ttt2/libraries/migrations.lua
saibotk: What about just making them shared and have the developer take the decision as usual via ifServer then stuff? Would be much simpler

saibotk (changes requested):

lua/terrortown/migrations/shared/2024_01_22_10_26_sampleMigration.lua
ZenBre4ker: Yeah sounds good, just didnt want to spend any time for it until we are sure about the overall workings.

States is derived from the [command pattern](https://en.wikipedia.org/wiki/Command_pattern) and shall make clear that everything that command needs is inside the class. This would contain the before and after states, that can be saved for example. This makes a command absolutely standalone in programming languages were you have to manually address memory for example or the variable contents change.

Generally used for forward and revert mechanics in photoshop or similar.

lua/ttt2/libraries/migrations.lua
ZenBre4ker: Not yet, but I want to keep the option open for downgrades.

lua/ttt2/libraries/migrations.lua
ZenBre4ker: Would be fine by me.

lua/ttt2/libraries/migrations.lua
saibotk: I really don't see what you try to do here and if it ever will come in again you can just assume them as true for all older migrations.

lua/terrortown/migrations/shared/2024_01_22_10_26_sampleMigration.lua
saibotk: This is overly complicated to me and i don't see any issue that is mitigated by using this. 
Also the migration functions don't need any state. I'd prefer if we keep it simple here.

Really i'd envision this to replicate the Laravel system, as it is quite simple and easy.

lua/ttt2/libraries/migrations.lua
EntranceJew: fwiw this is how laravel handles migration stages: [keys: id, name, batch](https://github.com/illuminate/database/blob/015ecbc43bfa49422371f370d4f35ae53f6ee3da/Migrations/DatabaseMigrationRepository.php#L165-L171) -- if an error is encountered during migration it either does not log it or it does log it and keeps some sort of record to describe that an error happened; idk if it automatically runs the `down()` step

ZenBre4ker:
Okay, how about this? 😄

lua/ttt2/libraries/migrations.lua
saibotk: Yes but batch is only for convenience to rollback all migrations done in a single execution.

Also it does not execute down.
It will just stop doing migrations entirely when one of them errors because the invariant is that all previous migrations have run.

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,122 @@
...
+
+---
+-- Loads migration-files and starts inheriting from base files
+-- @realm shared
+function migrations.Load()
+	MsgN("[TTT2] Loading base Migrations for inheritance ...")
+	local baseClasses = classbuilder.BuildFromFolder(migrations.baseFolderPath, SHARED_FILE, "MIGRATION", OnInitialization, true, ShouldInherit, {}, PostInherit)
+
+	MsgN("[TTT2] Loading Migrations ...")
+	migrations.migrations = classbuilder.BuildFromFolder(migrations.folderPath, SHARED_FILE, "MIGRATION", OnInitialization, true, ShouldInherit, baseClasses, PostInherit)
```
saibotk, line 119: The class builder is also rather overkill. I would not like to have the inheritance included for migrations. Just make the migrations return the up and down function and load the files.

saibotk (changes requested):
We are getting pretty close very nice!

But some open points:

We need to provide a command or method to downgrade and it would need to have a parameter for providing the number of migrations to rollback.
Honestly though the more i think about this the more i think it is entirely useless and we should discuss on wether we would rather want to just forward migrate instead. Because when would a user ever downgrade? Only devs might want to do so. And imagine you downgrade to an earlier release, then you cant even downgrade because the migrations are only available in the newest version. 
Additionally, our auto migration on game load can be a problem. Users who want to downgrade will have their downgrade being reversed when restarting. 

We really need to discuss the invariants and expectations of this system.

Especially we also need to think on how to provide users with some accessible methods to still continue playing. Because in Systems like Laravel obviously the administrator is responsible to make sure everything went fine but we cant expect that from Normal Workshop users.
Maybe some approach like resetting the DB entirely and rerunning all migrations? What about non DB migrations like convars for example?

Thats a tough question lets discuss! @histalek gotta hear your opinion here too

lua/ttt2/libraries/migrations.lua
ZenBre4ker: I actually like it way more. With the earlier command setup I wanted to give a way for default behaviours.
With this inheritance instead we can make sure to program a few base migrations and reuse them over and over again. Makes it way safer, then copying code over and over again, IMHO.

ZenBre4ker:
I actually left the down migration out on purpose, with the previous versioning I had a proper way, but with the new system I have no good idea to solve it. Therefore I leave that to future TTT2 Dev team to solve :3

lua/ttt2/libraries/migrations.lua
saibotk: Migrations are really nearly never connected to each other. You will never need this as far as i can tell from my experience

EntranceJew: i don't think it hurts. if i wanted to implement a generic convar migration or some other situation it would save some juggling.

ZenBre4ker: Yes, thats what I thought too.
We wanted:
- Create databases here
- Rename some databases
- Maybe change savingKeys
- Change convarnames
- Change convardefaults

I cant really think of much more we ever did, having those few cases standardized was my goal with this PR. We can implement them along the way and can be sure they work.

saibotk: I strongly disagree. Implementing stuff just because it does not hurt is not our ambition here, we would rather have something simple and clear to solve the problem.
I think you both are only thinking theoretically about this. You wont need any of this through inheritance. This concept is really nothing that brings benefits here. We do not need any hierarchical structure to provide reusable tools to do these common changes but rather utilities. 
We should focus on making this as accessible and easy as possible. It is way easier to just write a migration file that returns a single function ("upgrade" or "migrate") or maybe two if we want to keep both directions.

Composition based on utilities is much more flexible for such a scenario, where object oriented design is rather inflexible and does not allow to be combined as much. Especially since we do not need any of the structural benefits that come with such a design. So we only get the downsides.

ZenBre4ker: That I can agree with. We definitely need clear utilities. I thought that classes can give us exactly that. But in our case migrations are limited to maybe 2? areas. We dont need to handle files or any other Apis.

For me it's important that stuff that gets used more than once is never rewritten.
So I strongly disagree to have any viable code in the migration files that get dirctly applied itself, they have to be inherited or given by utilities. And IMO should therefore also directly allow for downgrade, even if we never properly implement it.

And even if we dont want to inherit, using the classbuilder is a plus IMO.
The IsValid check can be surely done by the migration.apply function, but using the Migration globals seems useful.

Histalek:
> We need to provide a command or method to downgrade and it would need to have a parameter for providing the number of migrations to rollback. Honestly though the more i think about this the more i think it is entirely useless and we should discuss on wether we would rather want to just forward migrate instead. 

I think simply migrating forward is the way to go. This would make the whole system way less complex. Given that we struggle to come up with a usecase for downgrades and given that we could always map a downgrade to another migration this seems like the best way forward.

> We really need to discuss the invariants and expectations of this system.

I would keep it simple (at least for the database case):
- Migrations should control the Database schema (e.g. DDL Statements should only be run in migrations)
- All migrations are expected to be up before we interact with the database tables controlled by migrations
(I haven't thought much about the ConVar usecase yet)

> Especially we also need to think on how to provide users with some accessible methods to still continue playing. Because in Systems like Laravel obviously the administrator is responsible to make sure everything went fine but we cant expect that from Normal Workshop users. Maybe some approach like resetting the DB entirely and rerunning all migrations? What about non DB migrations like convars for example?

Resetting the db seems a bit harsh, but i can't think of another way.
We could probably just rename all (migration controlled) tables so this would at least be recoverable manually.

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,122 @@
...
+migrations.databaseName = "ttt2_migrations"
+migrations.savingKeys = {
+	executedAt = {
+		typ = "string",
+		default = ""
+	},
+	statusMessage = {
+		typ = "string",
+		default = "Success"
+	},
```
Histalek, line 22: This is somewhat redundant, `executedAt` should only be set after the migration was fully executed and therefore implies success

ZenBre4ker: executedAt is just the time, while statusMessage could contain an error message in case of failure. Could make it easier to see when migrations failed and why for bug reports.

ZenBre4ker: But maybe an actual log is better.

lua/ttt2/libraries/migrations.lua
saibotk: I can't seem to get across my points to you, but i still want to say that this is just making things unnecessarily complex and migrations do not share code. When looking at bigger Laravel projects for example i know that they would not benefit from inheritance.
Also the entire base migration has no use. It does not provide any functionality that makes it reasonable to introduce this complexity. Having to define a base class for migrations is just not needed. We just get more dead code for some potentially highly theoretical benefit, that we do not even know a use case of yet. This is rather over-engineering the problem to be honest.

What are your opinions @Histalek or @TimGoll ?

gamemodes/terrortown/gamemode/shared/sh_main.lua
```diff
@@ -182,6 +182,9 @@ hook.Add("PreRegisterSWEP", "TTT2RegisterSWEP", TTT2RegisterSWEP)
 -- @hook
 -- @realm shared
 function GM:TTT2Initialize()
+	-- Migrate all changes of TTT2
+	migrations.Apply()
```
ZenBre4ker, line 186: Maybe move this to an earlier place. Gotta see where it makes the most sense, but it should be early enough to catch new Convars and database calls.

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,73 @@
...
+
+	local fileExecutionCounter = 0
+
+	-- Add Hook to OnLuaError to catch if migrations fail
+	local isSuccess = true
+	local hookName = "OnLuaError"
+	local hookIdentifier = "TTT2MigrationErrorCatcher"
+	hook.Add(hookName, hookIdentifier, function(error, realm, stack, name, id)
+		isSuccess = false
+	end)
```
ZenBre4ker, line 37: This hook seems to be pretty new. It didnt work on my half a year old server. But after an update it did. Not sure if we should use it or warn to update the server?

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,73 @@
...
+
+		MsgN("[TTT2] Migrating: ", fileName)
+
+		if SERVER then
+			AddCSLuaFile(fullPath)
+		end
+		include(fullPath)
+
+		if not isSuccess then
+			Error("\n[TTT2] Migration failed.\n\n")
```
ZenBre4ker, line 55: This errors the migration but not the gamemode. You can still play after that. Maybe we should crash our gamemode or should we just ignore it?
It will be erroneous on every server and map restart as the filename doesnt get entered into the database.

ZenBre4ker (commented):
A few more discussions.

lua/ttt2/libraries/migrations.lua
TimGoll: Sure about that? Doesn't seem to be new judging by the wiki:
image

lua/ttt2/libraries/migrations.lua
```diff
@@ -9,114 +9,65 @@ end
...
 
-	for fileName, migration in SortedPairs(migrations.migrations) do
-		if not IsValid(migration) then continue end
+	-- Add Hook to OnLuaError to catch if migrations fail
+	local isSuccess = true
+	local hookName = "OnLuaError"
+	local hookIdentifier = "TTT2MigrationErrorCatcher"
+	hook.Add(hookName, hookIdentifier, function(error, realm, stack, name, id)
+		isSuccess = false
+	end)
```
saibotk, line 37: Maybe use ProtectedCall here instead?
And use `ProtectedCall(include(...))` because we still want it to print errors etc but this might be much cleaner than relying on the hook

saibotk (changes requested):

lua/ttt2/libraries/migrations.lua
ZenBre4ker: I can test that, but pcall already didnt work, because the include call itself is without error.

lua/ttt2/libraries/migrations.lua
ZenBre4ker: Weird. It didnt work for me before the update.

lua/ttt2/libraries/migrations.lua
ZenBre4ker: Does also not work.

lua/ttt2/libraries/migrations.lua
Histalek: We can't just run the gamemode after a failed migration.
All code that interacts with whatever the migration was supposed to do expects the migration to be run successfully. If we just run the gamemode after a migration failure we will fail on e.g. table insertions instead or even risk some sort of corruption.

lua/ttt2/libraries/migrations.lua
saibotk: Mhh if include is really not catchable, then we might just consider changing the migrations to just contain an anonymous function which is returned at the end of the file.
Then we could pcall the function and it would be cleaner than using the hook.

TimGoll:
Wanted to throw in another example of how [IvenTree KiCad](https://github.com/afkiwers/inventree_kicad/tree/main/inventree_kicad/migrations) does migrations

lua/ttt2/libraries/migrations.lua
ZenBre4ker: Done.

ZenBre4ker:
Okay, I hope that is early enough and now properly crashes the gamemode if it fails.

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,62 @@
...
+	local fileExecutionCounter = 0
+	for i = 1, #files do
+		local fileName = files[i]
+		local fullPath = migrations.folderPath .. files[i]
+
+		local fileInfo = migrations.orm:Find(fileName)
+
+		if fileInfo then continue end
+
+		MsgN("[TTT2] Migrating: ", fileName)
```
saibotk, line 38: @Histalek What's our policy now on those messages?

I think we need to write that down somewhere.

saibotk (changes requested):
Otherwise looks good, we would then need to figure out on how to help people getting into a playable state again.
Maybe add an info that they can delete the database? 
Or do we assume that our migrations should always succeed?
We just need to decide on how to handle these cases.

lua/ttt2/libraries/migrations.lua
TimGoll: This should be `Dev(1, xx)`

lua/ttt2/libraries/migrations.lua
```diff
@@ -0,0 +1,62 @@
...
+
+		Dev(1, "Migrating: ", fileName)
+
+		if SERVER then
+			AddCSLuaFile(fullPath)
+		end
+		local isSuccess, errorMessage = pcall(include(fullPath))
+
+		if not isSuccess then
+			error("[TTT2] Migration failed.\n" .. errorMessage, 1)
```
TimGoll, line 46:
```suggestion
			ErrorNoHalt("[TTT2] Migration failed.\n" .. errorMessage, 1)
```

ZenBre4ker: This is on purpose to crash the gamemode. `Error` doesnt work, as it is` ErrorNoHalt`. So i chose `error`

TimGoll: Ahh, sorry, then nevermind :D

ZenBre4ker:
Normally I'd say our migrations always succeed and they only error if we made a mistake implementing it.
In many cases I guess we just try to save old configs, if that doesnt work, we should catch errors, so the migration itself doesnt fail.
I'd say it should only ever fail if settings are done needed for the gamemode, like creating an sql table. And if that doesnt work, I believe theres no easy fix.

saibotk (approved):
Codewise fine, but we might want to wait until after release so this can get some more testing etc. 

Also before we merge, i'd like @Histalek's view on the failure case here.

ZenBre4ker:
This module here currently does nothing after I remove the sampleMigration.
So we could merge it and then wait for release to test the first actual implementation with the favorites reworks?

Histalek (approved):
Nice!

We probably need a bit more documentation/comments on this, but that can wait until we've tried using migrations in practice i think

PR #1305, opened 2024-01-16
Make crosshair dot centered

a7f3:
## before
image
## after
image


gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -360,8 +360,9 @@ if CLIENT then
 
 		-- draw crosshair dot
 		if cvCrosshairMode:GetInt() == CROSSHAIR_MODE_DOT_AND_LINES or cvCrosshairMode:GetInt() == CROSSHAIR_MODE_DOT_ONLY then
-			local xDot = mathFloor(xCenter - thicknessLine * 0.5)
-			local yDot = mathFloor(yCenter - thicknessLine * 0.5)
+			local xDot = xCenter - mathFloor(thicknessLine * 0.5)
+			local yDot = yCenter - mathFloor(thicknessLine * 0.5)
```
ZenBre4ker, line 364: Doesnt that mean a thicknessLine of 1 is invisible?

a7f3: this is just for the dot's position, if the thicknessLine is 1, it means there's no offset

EntranceJew:
i need to double check this when i have a moment later but iirc i think at certain scale combinations and outline weights it is still possible to offset the line weights to a similar effect, i think this PR could be an opportunity to fix that also

EntranceJew:
image

image

it's a little hard to see but with the right settings i think the lines are misaligned also, but not nearly as bad as they were before without integer multipliers

ZenBre4ker:
> it's a little hard to see but with the right settings i think the lines are misaligned also, but not nearly as bad as they were before without integer multipliers

I think you are right, in the code you can see that in x and y direction the `thicknessOutline` is subtracted for all 4 rectangles. This does not make sense for one direction per rectangle.
For example the left and right rectangle, dont need the `thicknessOutline` subtracted for the y direction (height), as it should have no effect on the center.
Same goes for upper and lower rect in x direction (width).

ZenBre4ker (approved):
This change is good and is also theoretically more precise, as we also only floor for the offsetLine, which is subtracted from the center calculation for the lines.

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -360,8 +360,8 @@ if CLIENT then
...
```
ZenBre4ker, line 364: Just use the offsetLine variable that has these values already calculated.

ZenBre4ker (changes requested):
Actually just use the variable.

TimGoll (approved):
It is fine once @ZenBre4ker's comment is addressed. Thank you!

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
EntranceJew:
```suggestion
			local xDot = xCenter - offsetLine
			local yDot = yCenter - offsetLine
```
should just be able to click this to make that change afaik

ZenBre4ker (approved):

PR #1306, opened 2024-01-19
deleted redundant model files + ignored

EntranceJew:
[Redox said so, and I trust them](https://github.com/CFC-Servers/gm_addon_optimization_tricks/wiki/Optimizations-without-downsides)


TimGoll (approved):

.gitignore
```diff
@@ -58,3 +58,8 @@ addon.json
 .luarc.json
 
 ttt2.jpg
+
+# redundant source files
+*.dx80.vtx
+*.sw.vtx
+*.xbox.vtx
```
Histalek, line 65:
```suggestion

# redundant source files
*.dx80.vtx
*.sw.vtx
*.xbox.vtx
```

Please don't do this.

These files aren't expected to exist locally so they should not be ignored by git.

Histalek (changes requested):
Looks good for the most part.

.gitignore
EntranceJew: the files are generated by most model compilers, if someone else were to recompile or compile new files for this addon then the redundant files could potentially be reintroduced

Histalek: That is for reviewers to catch. 

With this it is way easier for the local directory content to diverge from the upstream repository content. Possibly leading to changes in behaviour.

PR #1162, opened 2023-12-11
UI: Add missing equipment/weapon tooltips

a7f3:
Issue: #1151

done | name | current | expected
---- | ---- | ------- | ---
[[x](https://github.com/TTT-2/TTT2/assets/124642739/0314f69d-4a73-457e-a7ea-6ecc57d88c66)] | **weapon_zm_carry/Magneto-stick** | none | Primary push, secondary pick up/pull 
[[x](https://github.com/TTT-2/TTT2/assets/124642739/b4672e60-3fd4-4369-9441-49e6766ad88a)] | **weapon_ttt_knife** | none | Primary stab, secondary throw
[x] | **weapon_ttt_push/newton launcher** | none | Primary normal force shot, Secondary charged shot
[x] | **weapon_ttt_phammer/poltergeist** | none | Primary shoot, delayed shot at a long distance

## Weapons whos lua files need to be copied from [base ttt](https://github.com/Facepunch/garrysmod/tree/master/garrysmod/gamemodes/terrortown/entities/weapons): 
done | name | current | expected
---- | ---- | ------- | ---
[[x](https://github.com/TTT-2/TTT2/assets/124642739/e7c8e9f4-2c63-405c-9e49-9fcc3dac49d5)] | **C4** | Currently has none | primary throw, secondary stick
[x] | **defuser** | Uses old HudHelp instead of TTT2HudHelp | primary defuse c4
[x] | **Flare gun** | None | Primary burn body
[x] | **Radio** | Using old HudHelp | Primary throw, Secondary stick to surface
[x] | **Health station** | Using old HudHelp | primary & secondary throw 

## Added previously
- **weapon_ttt_binoculars**
- **weapon_ttt_wtester/DNA scanner**
- **weapon_ttt_cse/Visualizer**
- **weapon_ttt_teleport**

## changed wording 
- **weapon_ttt_decoy**

PR #1308, opened 2024-01-21
HUD: Safeguard against max_length not being defined in HUDHelp struct

EntranceJew:


TimGoll:
Does this fix the issue where sometimes a HUD help box is shown even though it has no content?

EntranceJew:
Not to my knowledge, but it might help shake it out. This is specifically for when HUDDraw might get blanked on a swep or might not be defined at all prior to #1162 getting in, there are more cases of this than are easy to deal with.

TimGoll (approved):
I'm not entirely sure how this issue can arise sometimes, but I gues a simple `or 0` to make sure that an acutal number is passed to the function is fine

saibotk (approved):

PR #1311, opened 2024-01-22
SWEP: Add SWEP.InvisibleViewModel

EntranceJew:
- the Deploy() code is useless
- the DrawViewModel check happens each DrawHUD so it's pointless to set it once in Deploy
- the only reason it appeared to work is because the FOV of those items were set to something incredibly small

### DIFF
--- a/gamemodes/terrortown/entities/weapons/weapon_ttt_unarmed.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_ttt_unarmed.lua
@@ -36,6 +36,7 @@ SWEP.InLoadoutFor = {ROLE_INNOCENT, ROLE_TRAITOR, ROLE_DETECTIVE}
 SWEP.AllowDelete = false
 SWEP.AllowDrop = false
 SWEP.NoSights = true
+SWEP.InvisibleViewModel = true
 
 SWEP.silentPickup = true
 
@@ -69,18 +70,6 @@ end
 -- @ignore
 function SWEP:Reload()
 
-end
-
----
--- @ignore
-function SWEP:Deploy()
-	if SERVER and IsValid(self:GetOwner()) then
-		self:GetOwner():DrawViewModel(false)
-	end
-
-	self:DrawShadow(false)
-
-	return true
 end
 
 ---
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -68,6 +68,10 @@ if CLIENT then
 	-- if the files are different, it only looks at the name. I recommend you
 	-- create your own directory so that this does not happen,
 	-- eg. /materials/vgui/ttt/mycoolserver/mygun.vmt
+
+	-- The ViewModel should not draw, in any situation. Prevents the need for hacks which
+	-- overload drawing functions.
+	SWEP.InvisibleViewModel = false
 end
 
 --   MISC TTT-SPECIFIC BEHAVIOUR CONFIGURATION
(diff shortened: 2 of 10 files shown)


TimGoll (approved):

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -68,6 +68,10 @@ if CLIENT then
 	-- if the files are different, it only looks at the name. I recommend you
 	-- create your own directory so that this does not happen,
 	-- eg. /materials/vgui/ttt/mycoolserver/mygun.vmt
+
+	-- The ViewModel should not draw, in any situation. Prevents the need for hacks which
+	-- overload drawing functions.
+	SWEP.InvisibleViewModel = false
```
saibotk, line 74: Which hacks are you talking about? Can you give an example?

gamemodes/terrortown/entities/weapons/weapon_ttt_unarmed.lua
```diff
@@ -71,18 +72,6 @@ function SWEP:Reload()
...
 end
 
----
--- @ignore
-function SWEP:Deploy()
-	if SERVER and IsValid(self:GetOwner()) then
-		self:GetOwner():DrawViewModel(false)
-	end
-
-	self:DrawShadow(false)
```
saibotk, line 81: This one had DrawShadow false too, can you please test, that all entities that did that really do not need to have this? 
Just make sure that none of them now still render a shadow / test ingame.

saibotk (changes requested):

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
EntranceJew: hacks as in overriding DrawViewModel in the deploy (only works once), setting the FOV < 10 (the viewmodel still exists, just out of frame)

saibotk: I still need a real example like is this even done anywhere? I would assume to see a deletion of such a hacky way in the code when you say you removed the need for it. But i cant find anything

EntranceJew: [yes, but i don't have any suggestions for what FOV an invisible viewmodel should be, so I didn't change it](https://github.com/search?q=repo%3ATTT-2%2FTTT2%20%22ViewModelFOV%20%3D%2010%22&type=code), the rest of the [DrawViewModel changes are in the changelist](https://github.com/TTT-2/TTT2/pull/1311/files)

TimGoll: > I still need a real example like is this even done anywhere?

The unarmed weapon does this, with my cam rework you are sometimes able to see the crowbar model

saibotk (approved):

TimGoll:
@EntranceJew I just spent 5 hours debugging my game until I noticed that this PR breaks the view models of all weapons. If you change maps after a while no weapon has a view model anymore, only a client restart helps, not even a map change.

Any idea why? Because I don't get it

EntranceJew:
you could change it to:
```
client:DrawViewModel(not self.InvisibleViewModel)
```

TimGoll:
> you could change it to:
> 
> ```
> client:DrawViewModel(not self.InvisibleViewModel)
> ```

well it is already that on the master, but that isn't working

EntranceJew:
i only had issues like this on first rounds and i attributed them to other weapon first round bugs like the slots being incorrect; [the docs](https://wiki.facepunch.com/gmod/Player:DrawViewModel) suggest that it is fine to have in a hook like that.

the alternative is to take it out of that function and move it to the base:
```lua
SWEP:PreDrawViewModel(vm, wep, ply)
  if wep.InvisibleViewModel then
    ply:DrawViewModel(false)
  end
end

SWEP:PostDrawViewModel(vm, wep, ply)
  if wep.InvisibleViewModel then
    ply:DrawViewModel(true)
  end
end
```

moving things back to the deploy/holster as we've already seen is unreliable

TimGoll:
> i only had issues like this on first rounds

In my debug session I tested it with only TTT2 installed, no other addons. So it has to be buggy, maybe the docs are wrong here. Or this is a GMod issue.

I'll give this a try

TimGoll:
Seems to fix the issue, I'll open a PR

PR #1312, opened 2024-01-22
made beacon model and icon unique from decoy

EntranceJew:
the decoy was the beacon all along

- brought down `icon_beacon`
- renamed icon_beacon to icon_decoy
- created new `icon_beacon` files
- changed the beacon model to match


TimGoll (approved):

saibotk (changes requested):
Add them both to force download and we are good to go

saibotk (approved):

PR #1315, opened 2024-01-23
Shop: Backend rework 3/X

ZenBre4ker:
This time I removed global BuyTables and moved them to proper getter and setter.
Should enable future UI to more easily check those separately.


gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -995,17 +999,12 @@ local function ReceiveBought()
 	local num = net.ReadUInt(8)
 
 	for i = 1, num do
-		local s = net.ReadString()
-		if s ~= "" then
-			client.bought[#client.bought + 1] = s
+		local equipmentId = net.ReadString()
+		if equipmentId ~= "" then
+			client.bought[#client.bought + 1] = equipmentId
```
TimGoll, line 1004:
```suggestion
		local equipmentClass = net.ReadString()
		if equipmentClass ~= "" then
			client.bought[#client.bought + 1] = equipmentClass
```

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -299,31 +299,16 @@ end
...
-
-		if SERVER then
-			net.Start("TTT2ReceiveTBEq")
-			net.WriteString(cls)
-			net.Send(GetTeamFilter(team))
-		end
-	end
+	shop.SetEquipmentBought(self, equipmentId)
+	shop.SetEquipmentGlobalBought(equipmentId)
+	shop.SetEquipmentTeamBought(self, equipmentId)
```
TimGoll, line 311: I can't add a suggestion, but I think here would be `equipmentClass` more fitting as well

gamemodes/terrortown/gamemode/server/sv_shop.lua
```diff
@@ -104,6 +104,54 @@ local function NetOrderEquipment(len, ply)
 end
 net.Receive("TTT2OrderEquipment", NetOrderEquipment)
 
+function shop.BroadcastEquipmentGlobalBought(equipmentId)
```
TimGoll, line 107: same here

gamemodes/terrortown/gamemode/shared/sh_shop.lua
```diff
@@ -67,19 +181,19 @@ function shop.CanBuyEquipment(ply, equipmentId)
...
-		or equipment.limited
-		and ply:HasBought(equipment.ClassName) then
-		return false, shop.statusCode.LIMITEDBOUGHT
+		and shop.IsGlobalBought(equipmentId) then
+		return false, shop.statusCode.GLOBALLIMITEDBOUGHT
+	end
+
+	if equipment.teamLimited
+		and shop.IsTeamBoughtFor(ply,equipmentId) then
+		return false, shop.statusCode.TEAMLIMITEDBOUGHT
```
TimGoll, line 196: I think this is pretty unreadable, in our new style it should be

```suggestion
	if
		equipment.teamLimited
		and shop.IsTeamBoughtFor(ply,equipmentId)
	then
		return false, shop.statusCode.TEAMLIMITEDBOUGHT
```

or if you prefer the "current" style:
```lua
	if equipment.teamLimited
		and shop.IsTeamBoughtFor(ply,equipmentId)
	then
		return false, shop.statusCode.TEAMLIMITEDBOUGHT
```

TimGoll (commented):
Looks mostly good to me, my main concern is the variable name `equipmentID`, imho it should reflect that it is a class, like `equipmentClass`

gamemodes/terrortown/gamemode/shared/sh_shop.lua
ZenBre4ker: Did we already agree on the new style?
As far as I know we try to let a linter do all the work, so current style should be fine until we fully switch to it.

gamemodes/terrortown/gamemode/client/cl_equip.lua
TimGoll: Hmm, not sure this is an improvement, but I don't care that much. But then you should add this to the docs here as well: 
image

Something like ` @param string equipmentId The class name of the equipment used as identifier`

Otherwise it is not really clear what a "equipmentId" imho

gamemodes/terrortown/gamemode/shared/sh_shop.lua
TimGoll: Not sure, but your style here is pretty weird in my opinion. Maybe it is just me but that the `and` and `return` are on the same level without any sort of devider gave me the impression that the return is missing

gamemodes/terrortown/gamemode/client/cl_equip.lua
ZenBre4ker: Its like last time, I guess. It's either obvious or should not be used.
But currently all equipments equal in `classname = id = name`.

So I wouldn rather not add a description everywhere for something that should be known.

But I just took a look at the rest of our codebase and we only use `equipmentId` in the newest shop rework parts 1-3.

We use `.id` to check nearly everything shop- relevant. And to give items we use `cls` for `classname`.

So either we agree on changing everything to `id` or revert back to `classname`.
Im just not sure why in the first place we used `.id` and not `.name`.

TimGoll: I think name is pretty fitting as well. Even though it is a bit conusing to me why we call something "name" that is derived from a class. I guess this is just what happens when 100 people work on the same code over years.

I don't have a good solution here. The thing I don't like about id is that in vanilla TTT equipment items have an numeric id and this is the first thing I thought of. If I'd have to decide I's say `classname` is the most descriptive name for this.

TimGoll: See this line in original TTT: https://github.com/Facepunch/garrysmod/blob/87aa68cb8da3d70e6e7265bc61ca222c55e3f8e5/garrysmod/gamemodes/terrortown/gamemode/equip_items_shd.lua#L80

ZenBre4ker: I believe that id was introduced by @Alf21 to make getting the classname shorter. But as we tend to give descriptive names, maybe we should instead try to get rid of id instead.

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
saibotk, line 311: Tbh I don't like the design of this api.
This rather should be done by a single call. 

Also i don't think its nice to have 3 tables for something that just derives from a single one. Because team and global limitations can be checked by checking the bought table.

I know this is how we did it before and not a concept of yours but nonetheless i propose that we try to get rid of this and make it easier

saibotk (changes requested):

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
ZenBre4ker: I'll see what I can do ^^

ZenBre4ker (commented):
Generally would be nice to have an opinion from both of you @Histalek @saibotk  about naming convention concerning equipmentId.
Generally why have Id, name and classname if they are the same.
https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/shared/sh_equip_items.lua#L59-L69

gamemodes/terrortown/gamemode/client/cl_equip.lua
saibotk: I like ID the most tbh because it fits both old and new code also we treat it as the unique identifier and it is not just a class which is rather far from being a word to identify a single element.

TimGoll: But there is an ID in the old ttt Items which isn't this one

saibotk: Need to check the codebase then

saibotk: I still think ID is more fitting because in the items module it is also used as ID and class can be confused with the weapon classnames.

ZenBre4ker: Well, class is definitely wrong, which for me should only be handled for weapon and item tables.

But classname, id and name are all the same in ohr codebase. We habe that one function, that sets id and name to the classname.

Also garrysmod uses classname with weapons.GetStored and the give function.

So in general that does not make a big difference. But I will get rid of `cls` in favor of one of the three, we just gotta try to be consistent about it.

TimGoll: I vote for classname, id sounds wrong for me

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
ZenBre4ker: Nah, lets do it in another PR.

gamemodes/terrortown/gamemode/shared/sh_shop.lua
```diff
@@ -223,7 +339,7 @@ function shop.BuyEquipment(ply, equipmentId)
 		---
 		-- @note Add our own hook with more consistent class parameter
 		-- @realm server
-		hook.Run("TTT2OrderedEquipment", ply, equipmentId, isItem, credits, ignoreCost)
+		hook.Run("TTT2OrderedEquipment", ply, equipmentName, isItem, credits, ignoreCost)
```
TimGoll, line 342: You probably have to update the docs for `GM:TTT2OrderedEquipment` as well

TimGoll (approved):

saibotk (approved):

PR #1316, opened 2024-01-24
Icons: Fix missing icons

EntranceJew:
fixes the issue in which there are icons sometimes but actually there are not icons because there are is not the anti icons of having them, paradoxically not really icons


EntranceJew:
hashtag no changed log

TimGoll (approved):

saibotk (approved):

PR #1318, opened 2024-01-24
Moved Fire and Radius Damage to Game Effects Library

EntranceJew:
because it looks good instead of bad

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Removed some crosshair related convars and replaced them with other ones, see the crosshair settings menu for details
 - Removed DX8/SW models that aren't used
+- Removed `ttt_fire_fallback`, there's no situation where the fire shouldn't draw anymore.
 
 ### Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Throwables (grenades) now have a `:GetPullTime()` accessor
 - Throwables (grenades) show UI for the amount of time remaining before detonation (fuse time) (by @EntranceJew)
 - UI for grenade throw arcs from [colemclaren's TTT fork](https://github.com/colemclaren/ttt/blob/master/addons/moat_addons/lua/weapons/weapon_tttbasegrenade.lua#L293-L353) (integrated by @EntranceJew)
+- `gameEffects` library for global effects that are useful, such as starting fires (by @EntranceJew)
 
 ### Changed
 
@@ -60,6 +61,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Converted `ttt_ragdoll_pinning` and `ttt_ragdoll_pinning_innocents` into per-role permissions.
 - Magneto stick now allows right-clicking to instantly drop something, while left-clicking still releases/throws it.
 - Magneto stick now shows tooltips respective to its current state.
+- `ttt_flame` is visible while it is moving  (by @EntranceJew)
+- `ttt_flame`'s hurtbox is more accurate to its visuals (by @EntranceJew)
 
 ### Fixed
 
--- a/gamemodes/terrortown/entities/entities/ttt_flame.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_flame.lua
@@ -13,29 +13,44 @@ ENT.Model = Model("models/weapons/w_eq_flashbang_thrown.mdl")
 -- @realm shared
 AccessorFunc(ENT, "dmgparent", "DamageParent")
 
----
--- @accessor boolean
--- @realm shared
-AccessorFunc(ENT, "die_explode", "ExplodeOnDeath")
-
----
--- @accessor number
--- @realm shared
-AccessorFunc(ENT, "dietime", "DieTime")
-
 ENT.firechild = nil
-ENT.fireparams = {
-	size = 120,
-	growth = 1
-}
+
+ENT.fire_ignite_delay = 0
+
+ENT.real_scale = nil
+ENT.fire_scale_base = 0.9
+ENT.fire_scale_variance = 0.2
 
 ENT.next_hurt = 0
 ENT.hurt_interval = 1
+
+ENT.low_detail_fps = 12
+ENT.low_detail_size = 10
+
+ENT.hurt_radius = nil
+ENT.hurt_base = 5
+ENT.hurt_variance = 1
+ENT.hurt_radius_scale_factor = 0.4
+
+ENT.trail = nil
+ENT.trail_color = Color(255, 56, 56, 255)
+ENT.trail_width = 32
+ENT.trail_end_width = 0
+ENT.trail_lifetime = 0.25
+ENT.trail_res = 16
+ENT.trail_texture = "trails/physbeam"
+
+ENT.debug_viz = false
 
 ---
 -- @realm shared
 function ENT:SetupDataTables()
 	self:NetworkVar("Bool", 0, "Burning")
+	self:NetworkVar("Bool", 1, "Immobile")
+	self:NetworkVar("Bool", 2, "ExplodeOnDeath")
+	self:NetworkVar("Float", 0, "FlameSize")
+	self:NetworkVar("Float", 1, "LifeSpan")
+	self:NetworkVar("Float", 2, "DieTime")
 end
 
 ---
@@ -43,12 +58,7 @@ end
 function ENT:Initialize()
 	self:SetModel(self.Model)
 	self:DrawShadow(false)
-	self:SetNoDraw(true)
-
-	if CLIENT and GetConVar("ttt_fire_fallback"):GetBool() then
-		self.Draw = self.BackupDraw
-		self:SetNoDraw(false)
-	end
+	self:SetNoDraw(false)
 
 	self:PhysicsInit(SOLID_VPHYSICS)
 	self:SetMoveType(MOVETYPE_VPHYSICS)
@@ -56,11 +66,21 @@ function ENT:Initialize()
 	self:SetCollisionGroup(COLLISION_GROUP_DEBRIS)
 	self:SetHealth(99999)
 
+	if SERVER then
+		self.trail = util.SpriteTrail( self, 0, self.trail_color, true, self.trail_width, self.trail_end_width, self.trail_lifetime, self.trail_lifetime, self.trail_texture )
+		self:DeleteOnRemove(self.trail)
+	end
+
+	self.real_scale = self:GetFlameSize() * (self.fire_scale_base + math.Rand(-self.fire_scale_variance, self.fire_scale_variance))
+	self.hurt_radius = self.real_scale * self.hurt_radius_scale_factor
+
 	self.next_hurt = CurTime() + self.hurt_interval + math.Rand(0, 3)
 
 	self:SetBurning(false)
 
-	if self:GetDieTime() == 0 then self:SetDieTime( CurTime() + 20 ) end
+	if self:GetDieTime() == 0 then
+		self:SetDieTime( CurTime() + self:GetLifeSpan() )
+	end
 end
 
 ---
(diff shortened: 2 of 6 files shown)


lua/ttt2/libraries/game_effects.lua
```diff
@@ -0,0 +1,114 @@
...
+-- @param Entity inflictor
+-- @todo C4 uses a completely different method for this, so do all grenades, maybe fix that?
+-- @todo This seems to do similar checking as util.BlastDamageInfo might, consider comparing them
+-- @realm shared
+function gameEffects.RadiusDamage(dmginfo, pos, radius, inflictor)
+	for k, vic in ipairs(ents.FindInSphere(pos, radius)) do
+		if IsValid(vic) and inflictor:Visible(vic) and vic:IsPlayer() and vic:Alive() and vic:IsTerror() then
+			vic:TakeDamageInfo(dmginfo)
+		end
+	end
```
TimGoll, line 113:
```suggestion
	local vics = ents.FindInSphere(pos, radius)
	for i = 1, #entsFound do
		local vic = vics[i]

		if IsValid(vic) and inflictor:Visible(vic) and vic:IsPlayer() and vic:Alive() and vic:IsTerror() then
			vic:TakeDamageInfo(dmginfo)
		end
	end
```

lua/ttt2/libraries/game_effects.lua
```diff
@@ -0,0 +1,114 @@
...
+-- @param nil|Entity parent The thing to attach the fire to.
+-- @return nil|Entity The fire it created, or nil if it was merged / couldn't be created.
+-- @realm server
+function gameEffects.SpawnFire(pos, scale, ignite_delay, life_span, owner, parent)
+	if CreateVFire then
+		return CreateVFire(nil, pos, vector_up, life_span, owner)
+	end
+
+	local fire = ents.Create("env_fire")
+	if not IsValid(fire) then return end
```
TimGoll, line 79:
```suggestion
	local fire = ents.Create("env_fire")

	if not IsValid(fire) then return end
```

lua/ttt2/libraries/game_effects.lua
```diff
@@ -0,0 +1,114 @@
...
+		local phys = flame:GetPhysicsObject()
+		if IsValid(phys) then
+			-- the balance between mass and force is subtle, be careful adjusting
+			phys:SetMass(2)
+			phys:ApplyForceCenter(ang:Forward() * spread_force)
+			phys:AddAngleVelocity(Vector(ang.p, ang.r, ang.y))
+		end
+		flames[#flames + 1] = flame
+	end
+	return flames
```
TimGoll, line 60:
```suggestion
	local flames = {}
	for i = 1, num do
		local ang = Angle(-math.Rand(0, 180), math.Rand(0, 360), math.Rand(0, 360))
		local vstart = pos + tr.HitNormal * 64
		local ttl = lifetime + math.Rand(-lifetime_variance, lifetime_variance)

		if CreateVFireBall then
			flames[#flames + 1] = CreateVFireBall(ttl, size, vstart, ang:Forward() * spread_force, dmgowner)
			continue
		end

		local flame = ents.Create("ttt_flame")
		flame:SetPos(vstart)
		flame:SetFlameSize(size)
		flame:SetLifeSpan(ttl)
		flame:SetImmobile(immobile)

		if IsValid(dmgowner) and dmgowner:IsPlayer() then
			flame:SetDamageParent(dmgowner)
			flame:SetOwner(dmgowner)
		end

		flame:SetDieTime(CurTime() + ttl)
		flame:SetExplodeOnDeath(explode)
		flame:Spawn()
		flame:PhysWake()

		local phys = flame:GetPhysicsObject()

		if IsValid(phys) then
			-- the balance between mass and force is subtle, be careful adjusting
			phys:SetMass(2)
			phys:ApplyForceCenter(ang:Forward() * spread_force)
			phys:AddAngleVelocity(Vector(ang.p, ang.r, ang.y))
		end

		flames[#flames + 1] = flame
	end

	return flames
```

TimGoll (changes requested):
almost good imho, just a few small style changes. Also don't use `ipairs` please

TimGoll (approved):

gamemodes/terrortown/entities/entities/ttt_flame.lua
```diff
@@ -189,73 +122,81 @@ if SERVER then
...
+		physobj:EnableMotion(false)
+	end
+end
 
-		render.SetMaterial(fakefire)
-		render.DrawBeam(vstart, vend, 80, 0, 0, COLOR_RED)
+if CLIENT then
+	local flamesprites = {}
+	for i = 1, 5, 1 do
+		flamesprites[#flamesprites + 1] = Material("particles/flamelet" .. i)
```
saibotk, line 186: Write this out instead, so the material names are more easily found when looking / searching for their usage.

gamemodes/terrortown/entities/entities/ttt_flame.lua
```diff
@@ -189,73 +122,81 @@ if SERVER then
...
 	function ENT:Draw()
+		if self:GetBurning() and self.debug_viz then
+			render.DrawWireframeSphere(self:GetPos(), self.hurt_radius, 16, 9, COLOR_RED)
+		elseif not self:GetBurning() then
+			local frame = math.floor(((SysTime() * self.low_detail_fps) + self:EntIndex()) % #flamesprites) + 1
+			cam.Start3D()
+			render.SetMaterial(flamesprites[ frame ])
+			render.DrawSprite(self:GetPos(), self.low_detail_size, self.low_detail_size, color_white)
+			cam.End3D()
+		end
```
saibotk, line 200: What is this exactly? 
This was not in here before and i don't know why this entity is drawing anything at all in this state?

lua/ttt2/libraries/game_effects.lua
```diff
@@ -0,0 +1,124 @@
...
+	local flames = {}
+	for i = 1, num do
+		local ang = Angle(-math.Rand(0, 180), math.Rand(0, 360), math.Rand(0, 360))
+		local vstart = pos + tr.HitNormal * 64
+		local ttl = lifetime + math.Rand(-lifetime_variance, lifetime_variance)
+
+		if CreateVFireBall then
+			flames[#flames + 1] = CreateVFireBall(ttl, size, vstart, ang:Forward() * spread_force, dmgowner)
+			continue
+		end
```
saibotk, line 36: I'd remove this, because otherwise our extra handling in `SpawnFire` is never used and all the logic in ttt_flame is also not controlling this anymore.

lua/ttt2/libraries/game_effects.lua
```diff
@@ -0,0 +1,124 @@
...
+	fire:SetParent(parent)
+	fire:SetOwner(owner)
+	fire:SetPos(pos)
+	--no glow + delete when out + start on + last forever
+	fire:SetKeyValue("spawnflags", tostring(128 + 32 + 4 + 2 + 1))
+	fire:SetKeyValue("firesize", tostring(scale)) -- hardly controls size, hitbox is goofy, impossible to work with
+	fire:SetKeyValue("fireattack", tostring(ignite_delay))
+	fire:SetKeyValue("health", tostring(life_span))
+	fire:SetKeyValue("ignitionpoint", "64")
+	fire:SetKeyValue("damagescale", "0")
```
saibotk, line 97: Previously this was negative with a comment that only negative values lead to preventing damage, is this not true anymore?

saibotk (changes requested):

gamemodes/terrortown/entities/entities/ttt_flame.lua
```diff
@@ -13,54 +13,74 @@ ENT.Model = Model("models/weapons/w_eq_flashbang_thrown.mdl")
...
 	self:PhysicsInit(SOLID_VPHYSICS)
 	self:SetMoveType(MOVETYPE_VPHYSICS)
 	self:SetSolid(SOLID_VPHYSICS)
 	self:SetCollisionGroup(COLLISION_GROUP_DEBRIS)
 	self:SetHealth(99999)
 
+	if SERVER then
+		self.trail = util.SpriteTrail( self, 0, self.trail_color, true, self.trail_width, self.trail_end_width, self.trail_lifetime, self.trail_lifetime, self.trail_texture )
+		self:DeleteOnRemove(self.trail)
+	end
```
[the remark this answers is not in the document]
EntranceJew: it was added because these flame entities are invisible until they find a resting position, fling far away from their point of origin, and those issues compounded the flames not matching the size of their hurtbox

they look like a short red trail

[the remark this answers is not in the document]
saibotk: I am kind of against it from my imagination. I'd need a video but otherwise i tend to suggest removing this, dunno what @TimGoll or @Histalek thinks about this.

gamemodes/terrortown/entities/entities/ttt_flame.lua
```diff
@@ -13,54 +13,74 @@ ENT.Model = Model("models/weapons/w_eq_flashbang_thrown.mdl")
...
--- @accessor boolean
--- @realm shared
-AccessorFunc(ENT, "die_explode", "ExplodeOnDeath")
+ENT.firechild = nil
 
----
--- @accessor number
--- @realm shared
-AccessorFunc(ENT, "dietime", "DieTime")
+ENT.fire_ignite_delay = 0
```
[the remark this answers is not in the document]
EntranceJew: This was changed again because the flame already builds up in size as the particle system begins emitting -- but the hurtbox does not scale over time, so you will be hurt by something with no corresponding visual.

The engine code doesn't really make much use of this, but I pass it through in the off chance that someone overwrites the "env_fire" entity with a better implementation that uses the engine parameters as documented.

gamemodes/terrortown/entities/entities/ttt_flame.lua
EntranceJew: The entity is not burning because it has not gone from invisible to visible yet, so to give it a material to track it, it is drawing a billboarded sprite at its center of mass. GetBurning is poorly named but it was already like that.

lua/ttt2/libraries/game_effects.lua
EntranceJew: Similar to the `fireattack` parameter, the negative damagescale value still caused the "fake damage" screen pulse -- and having it at 0 doesn't prevent it either. Having it at a flat 0 fits the intent since we're managing the hurtbox ourselves.

gamemodes/terrortown/entities/entities/ttt_flame.lua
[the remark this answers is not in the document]
EntranceJew: You could pull the branch and see it in motion, isn't that why we make these small? So people can look at them? I'm sure I posted lots of footage of this over the last 6 months.

https://github.com/TTT-2/TTT2/assets/5711436/d52748f3-b7b9-469a-b61f-5f93a0f91cb2

lua/ttt2/libraries/game_effects.lua
saibotk: Then please add a small comment, why this is set to 0.

gamemodes/terrortown/entities/entities/ttt_flame.lua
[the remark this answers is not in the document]
saibotk: Since this is just a new variable, lets just remove it. This would be a very niche use case and then someone could still overwrite the function and do so.

gamemodes/terrortown/entities/entities/ttt_flame.lua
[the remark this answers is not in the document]
saibotk: I could but i am reviewing many PRs today and adding such videos to the PR descriptions helps a lot to speed up discussion about such changes.

But yes this seems fine to me thanks!

lua/ttt2/libraries/game_effects.lua
EntranceJew: It still creates a managed instance of an entity. If we didn't create a fireball then it wouldn't spawn fire that can be merged with the vfire ball for anything that calls it directly.

saibotk: I don't understand what you are trying to explain sorry

gamemodes/terrortown/entities/entities/ttt_flame.lua
```diff
@@ -189,73 +122,81 @@ if SERVER then
...
-
-			self:SetBurning(true)
 		end
 	end
 end
 
-if CLIENT then
-	---
-	-- @realm client
-	CreateConVar("ttt_fire_fallback", "0", FCVAR_ARCHIVE)
```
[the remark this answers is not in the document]
EntranceJew: ok

lua/ttt2/libraries/game_effects.lua
```diff
@@ -0,0 +1,124 @@
...
+---
+-- greatly simplified version of SDK's game_shard/gamerules.cpp:RadiusDamage
+-- does no block checking, radius should be very small
+-- @note only hits players!
+-- @param DamageInfo dmginfo
+-- @param Vector pos
+-- @param number radius
+-- @param Entity inflictor
+-- @todo C4 uses a completely different method for this, so do all grenades, maybe fix that?
+-- @todo This seems to do similar checking as util.BlastDamageInfo might, consider comparing them
```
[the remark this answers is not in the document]
EntranceJew: ok

gamemodes/terrortown/entities/entities/ttt_flame.lua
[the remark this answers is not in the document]
EntranceJew: The variable isn't new exactly, it's just externalized from its original position.

lua/ttt2/libraries/game_effects.lua
EntranceJew: ok

gamemodes/terrortown/entities/entities/ttt_flame.lua
EntranceJew: ok

lua/ttt2/libraries/game_effects.lua
EntranceJew: vfire when it exists is supposed to deal with everything pertaining to fire, all the appropriate parameters are remapped accordingly

gamemodes/terrortown/entities/entities/ttt_flame.lua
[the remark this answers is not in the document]
saibotk: Yes but its nothing that someone would already rely on and we do not really need to expose this. This is a really hypothetical usecase so i suggest to just remove it

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added `SWEP:ClearHUDHelp()` to allow blanking the help text, for dynamically updating help text on equipment (by @EntranceJew)
 - Throwables (grenades) now have a `:GetPullTime()` accessor
 - Throwables (grenades) show UI for the amount of time remaining before detonation (fuse time) (by @EntranceJew)
+- `gameEffects` library for global effects that are useful, such as starting fires (by @EntranceJew)
```
saibotk, line 34: Maybe mention that vFire integration is now builtin?

lua/ttt2/libraries/game_effects.lua
saibotk: Then please add a note to the function doc block that if vFire is enabled, parameters such as immobile, explode, lifetime_variance, lifetime, num are ignored.
But then to be honest this function is not really doing what someone would expect, so my suggestion is to remove this if part here entirely and let this function take care of controlling how multiple fires are spawned while you can check for CreateVFireBall directly inside the firegrenade and call that instead of this function.

I think this would be the best solution.

saibotk (changes requested):

lua/ttt2/libraries/game_effects.lua
EntranceJew: the functions aren't meant to just replace the grenade, they're meant to replace functions that other addons which create fire may have been using, like the C4 for example

saibotk: It can't, because no other addon uses this new gameeffects function. So i don't get your point here. It just does not make sense to have an entire SpawnFires function that just does not spawn fires but a single fireball.

EntranceJew: yeah, anything that did use them, should now use these

counter to its name, a fireball also explodes into several fires after impact which is why it's functionally similar

saibotk: Yeah no i still disagree here. Maybe @Histalek or @TimGoll can chime in

CHANGELOG.md
EntranceJew: not needed

saibotk: I think it is and don't resolve my comments

TimGoll: I second this, just add `  - this library has also built-in vFire support meaning that TTT2 now fully supports vFire`

EntranceJew: image
?

EntranceJew: if there are objections to how vFire itself functions then there's no reason to support it

saibotk: You just removed it wow. That's not remotely what we told you to do, but idc

saibotk (approved):

PR #1322, opened 2024-01-25
HUD: Scaling for player names when spectating

EntranceJew:
image


TimGoll (approved):
Thank you! That was something that was on my todo list for years but I never came around actually doing it

saibotk (approved):

PR #1314, opened 2024-01-22
UI: 3D Weapon Preview

TimGoll:
Adds a 3D weapon preview that could be used in the shop rework.

Things done:

- [ ] Trigger attack animation [[discord]](https://discord.com/channels/565105920414318602/567617926991970306/1199019906050572329) (probably too much, so it is scrapped for now)
- [x] Support weapons such as boomerang and melon mine [[discord]](https://discord.com/channels/565105920414318602/567617926991970306/1199087704470716426)
- [x] reduce player opacity on hover to highlight weapon

image

Demovideo:
https://github.com/TTT-2/TTT2/assets/13639408/98e4bf81-6ca7-420d-ac57-fbaeee3c4bbc

PR #1325, opened 2024-01-25
Added Custom World and View Model Rendering

TimGoll:
The [SWEP Construction Kit](https://steamcommunity.com/sharedfiles/filedetails/?id=109724869) is a popular addon that allows the creation of custom view and world models. Addons such as the boomerang and the melon mine use it.
I liked the core functionality so much, that I refactored the code and added it to TTT2. The toolkit has more features than this implementation, but the core functionality of adding world and viewmodels is there. The data tables are also compatible with those created by the toolkit.

image
Here's a reworked spring mine with custom world and view models

As can be seen in [this commit](https://github.com/TTT-2/ttt2-wep_spring_mine/commit/14d0ad00854de0816348c36586530e65911c03a5) this PR cleans up addons a lot.

This opens the possibilities to:
- add custom world and view models to some TTT2 weapons like radiio, beacon, ... (done here: https://github.com/TTT-2/TTT2/pull/1327)
- integrate it into the rendering of #1314 (in a third PR, once these two are merged)

## How to add them to your addon

```lua
SWEP.HoldType = "slam"

SWEP.ViewModelFOV = 70
SWEP.ViewModelFlip = false

SWEP.UseHands = true
SWEP.ShowDefaultViewModel = false
SWEP.ShowDefaultWorldModel = false

SWEP.ViewModel = "models/weapons/cstrike/c_c4.mdl"
SWEP.WorldModel = "models/weapons/w_c4.mdl"

function SWEP:Initialize()
	self:AddCustomViewModel("vmodel", {
		type = "Model",
		model = "models/props_phx/smallwheel.mdl",
		bone = "ValveBiped.Bip01_R_Finger2",
		rel = "",
		pos = Vector(1.557, 4.9, 0),
		angle = Angle(120, 0, 20),
		size = Vector(0.55, 0.55, 0.55),
		color = Color(50, 50, 50, 255),
		surpresslightning = false,
		material = "models/debug/debugwhite",
		skin = 0,
		bodygroup = {}
	})

	self:AddCustomWorldModel("wmodel", {
		type = "Model",
		model = "models/props_phx/smallwheel.mdl",
		bone = "ValveBiped.Bip01_R_Hand",
		rel = "",
		pos = Vector(5, 6, 0),
		angle = Angle(120, 20, 0),
		size = Vector(0.625, 0.625, 0.625),
		color = Color(50, 50, 50, 255),
		surpresslightning = false,
		material = "models/debug/debugwhite",
		skin = 0,
		bodygroup = {}
	})

	self.BaseClass.Initialize(self)
end
```

**Note:** The names `vmodel` and `wmodel` can be set to anything, I chose them only like this to make their intention clear. There can also be multiple models, sprites or quads added. They are then all rendered at once!


EntranceJew:
be sure to check [the graph](https://github.com/Clavus/SWEP_Construction_Kit/network) for the most recent versions people are working on, it looks like [MagicSwap](https://github.com/MagicSwap/SWEP_Construction_Kit) has the best maintained fork, and you may want to revisit the portions of the code you utilized from the base SWEPCK for any changes or improvements they may have made.

TimGoll:
> be sure to check [the graph](https://github.com/Clavus/SWEP_Construction_Kit/network) for the most recent versions people are working on, it looks like [MagicSwap](https://github.com/MagicSwap/SWEP_Construction_Kit) has the best maintained fork, and you may want to revisit the portions of the code you utilized from the base SWEPCK for any changes or improvements they may have made.

I took a look at [this file](https://github.com/MagicSwap/SWEP_Construction_Kit/blob/f57f02b17859eed86f5afce1a33eb228aead940a/lua/weapons/swep_construction_kit/base_code.lua#L49) and it seems like it is unchanged

TimGoll:
image

I tested it with the melon mine and it works great as well

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
...
+-- @param Entity wep The weapon for which the model should be created
+-- @param ModelData modelData The model data for the model
+-- @realm client
+function modelbuilder.CreateModel(wep, modelData)
+	-- handle a model being added to the view or world model
+	if modelData.type == "Model" and modelData.model and modelData.model ~= ""
+		and (not IsValid(modelData.modelEnt) or modelData.createdModel ~= modelData.model)
+		and string.find(modelData.model, ".mdl") and file.Exists (modelData.model, "GAME")
+	then
+		modelData.modelEnt = ClientsideModel(modelData.model, RENDER_GROUP_VIEW_MODEL_OPAQUE) --todo check correct render group
```
saibotk, line 46: Check the todo and remove afterwards

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
...
+			modelData.modelEnt:SetNoDraw(true)
+			modelData.createdModel = modelData.model
+		else
+			modelData.modelEnt = nil
+		end
+
+	-- handle a sprite being added to the view or world model
+	elseif modelData.type == "Sprite" and modelData.sprite and modelData.sprite ~= ""
+		and (not modelData.spriteMaterial or modelData.createdSprite ~= modelData.sprite)
+		and file.Exists ("materials/" .. modelData.sprite .. ".vmt", "GAME")
```
saibotk, line 61:
```suggestion
		and file.Exists("materials/" .. modelData.sprite .. ".vmt", "GAME")
```

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
...
+	end
+
+	for identifier, dataTable in pairs(allBones) do
+		local bone = viewModel:LookupBone(identifier)
+
+		if not bone then continue end
+
+		local s = Vector(dataTable.scale.x, dataTable.scale.y, dataTable.scale.z)
+		local p = Vector(dataTable.pos.x, dataTable.pos.y, dataTable.pos.z)
+		local ms = Vector(1, 1, 1)
```
saibotk, line 137: no abbreviations here

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
...
+			model:DrawModel()
+
+			render.SetBlend(1)
+			render.SetColorModulation(1, 1, 1)
+
+			if modelData.surpresslightning then
+				render.SuppressEngineLighting(false)
+			end
+		elseif modelData.type == "Sprite" and sprite then
+			local drawpos = pos + ang:Forward() * modelData.pos.x + ang:Right() * modelData.pos.y + ang:Up() * modelData.pos.z
```
saibotk, line 302: move this outside, this is used in every single case

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
...
+		end
+	end
+
+	return pos, ang
+end
+
+---
+-- Renders the world or view model.
+-- @param Entity wep The weapon whose view model should be rendered
+-- @param table renerOrder The render order of the view model elements
```
saibotk, line 244:
```suggestion
-- @param table renderOrder The render order of the view model elements
```

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
+---
+-- Used to build custom world and view models.
```
saibotk, line 2: Also rendering.

But please also describe that this needs the weapon base to contain specific entries for this to work.

Maybe we need to also add a comment to all `wep` params that this only works with tttbase weapons.

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
...
+---@field pos Vector The position offset
+---@field angle Angle The rotation offset
+
+if SERVER then
+	AddCSLuaFile()
+
+	return
+end
+
+modelbuilder = {}
```
saibotk, line 31: I think we should rename this, because it is much more connected to the weapon base and also mainly takes care of rendering instead of building the models.

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
...
+		else
+			allBones[bonename] = {
+				scale = Vector(1, 1, 1),
+				pos = Vector(0, 0, 0),
+				angle = Angle(0, 0, 0)
+			}
+		end
+	end
+
+	for identifier, dataTable in pairs(allBones) do
```
saibotk, line 130: this loop could be combined with the loop above i think

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
...
+		-- you can get in an infinite loop. Let's just hope nobody's that stupid.
+		pos, ang = modelbuilder.GetBoneOrientation(wep, baseDataTable, tbl, boneEntity)
+
+		if not pos then return end
+
+		pos = pos + ang:Forward() * tbl.pos.x + ang:Right() * tbl.pos.y + ang:Up() * tbl.pos.z
+
+		ang:RotateAroundAxis(ang:Up(), tbl.angle.y)
+		ang:RotateAroundAxis(ang:Right(), tbl.angle.p)
+		ang:RotateAroundAxis(ang:Forward(), tbl.angle.r)
```
saibotk, line 217: This seems pretty inconsistent to me, why are these rotations etc only applied to the result if it is coming from the "rel" lookup.

This whole method really seems very specifically crafted for another function instead of some public available function. If so, add comments and make this local

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
...
+		elseif dataTable.type == "Sprite" or dataTable.type == "Quad" then
+			table.insert(newRenderOrder, identifier)
+		end
+	end
+
+	return newRenderOrder
+end
+
+---
+-- Gets the orientation of a given bone.
```
saibotk, line 190: Describe that it ignores the bone override etc. when a "rel" value is found.
Tbh. i still don't know how someone would use this or why

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
...
+	for i = 0, boneCount do
+		viewModel:ManipulateBoneScale(i, Vector(1, 1, 1))
+		viewModel:ManipulateBoneAngles(i, Angle(0, 0, 0))
+		viewModel:ManipulateBonePosition(i, Vector(0, 0, 0))
+	end
+end
+
+---
+-- Updates the bone positions of the view model of a weapon. Is used to hide a
+-- default view model of a weapon.
```
saibotk, line 103: How is this used to hide the default view model?

And please explain what "update" means in this context. Like update the bone positions to what?

saibotk (changes requested):

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -659,6 +668,99 @@ if CLIENT then
...
+	end
+
+	---
+	-- Adds a custom world model.
+	-- @note Multiple world models can be added, they are all rendered at once
+	-- @param string identifier The name of the added world model
+	-- @param ModelData modelData The model data table
+	-- @realm client
+	function SWEP:AddCustomWorldModel(identifier, modelData)
+		self.WElements = self.WElements or {}
```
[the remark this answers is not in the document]
EntranceJew: people that are used to using swepck are used to the velement/welement naming convention, this would make migrating output more difficult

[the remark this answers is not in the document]
TimGoll: Agreed, I kept all names from the tool so that it is identical. Also that way our third person renderer for the UI is compatible with both my implementation and the basic implementation without me adding something

However I'm wondering if having the same variable names can lead to clashes of the two renderers, hm. The boomerang is working fine for me, but maybe some addons do some weird shit that could break here? I'm not 100% sure.

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -659,6 +668,99 @@ if CLIENT then
...
+		then return end
+
+		if self.ShowDefaultWorldModel then
+			self:DrawModel()
+		end
+
+		if not self.WElements then return end
+
+		-- we build a render order because sprites need to be drawn after models
+		self.WRenderOrder = modelbuilder.BuildRenderOrder(self.WElements, self.WRenderOrder)
```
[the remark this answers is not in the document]
TimGoll: I have to cash it somewhere to improve performance. And I thought it would be cleaner to have this outside the rendering code. But I don't care too much. Imho it is nicer that way, because generating a render order is not really a part of rendering it

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
+---
+-- Used to build custom world and view models. 
+-- @author Mineotopia
+-- @module modelbuilder
```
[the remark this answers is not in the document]
TimGoll: Not much code was reused here, this is mostly done from scratch after learning how to do it. But you're right, linking this would probably be a good idea

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -659,6 +668,99 @@ if CLIENT then
...
+	-- Adds a custom view model.
+	-- @note Multiple view models can be added, they are all rendered at once
+	-- @param string identifier The name of the added view model
+	-- @param ModelData modelData The model data table
+	-- @realm client
+	function SWEP:AddCustomViewModel(identifier, modelData)
+		self.VElements = self.VElements or {}
+		self.VElements[identifier] = modelData
+
+		modelbuilder.CreateModel(self, self.VElements[identifier])
```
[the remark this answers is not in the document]
TimGoll: Hmm, how exactly? This funtion takes the data in `self.VElements[identifier]` and stores the modified output in the same table

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,316 @@
...
+modelbuilder = {}
+
+local propertiesMaterial = {"nocull", "additive", "vertexalpha", "vertexcolor", "ignorez"}
+
+---
+-- Creates a clientside model that is used as a view or world model with the provided model data.
+-- @param Entity wep The weapon for which the model should be created
+-- @param ModelData modelData The model data for the model
+-- @realm client
+function modelbuilder.CreateModel(wep, modelData)
```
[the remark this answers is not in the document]
TimGoll: This is how it was done in the original code, I just assumed I can do this without actually copying the table

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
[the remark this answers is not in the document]
saibotk: Then please describe the workflow on how someone adds such models etc since im not familiar with the outputs or anything of that tool.

lua/ttt2/libraries/modelbuilder.lua
TimGoll: I did, just forgot to remove it. The thing is, it works and I took this line from the old addon. HOWEVER this render group doesn't exist in the docs. So I'm not sure here

lua/ttt2/libraries/modelbuilder.lua
TimGoll: Agreed. Regarding the last point: Don't we already assume that all weapons in TTT are based on the ttt weapon base?

lua/ttt2/libraries/modelbuilder.lua
TimGoll: modelrenderer then?

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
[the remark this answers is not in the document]
saibotk: Well its much cleaner because it is something the renderer has to take into account and no one else has to know about this. Also you can still cache it somewhere on the entity.

lua/ttt2/libraries/modelbuilder.lua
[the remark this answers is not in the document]
saibotk: I think this would make it much clearer from the API perspektive.
Modifying the modelData was the least i expected from this function.

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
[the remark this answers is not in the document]
saibotk: Closing this, see the comment on the function instead

lua/ttt2/libraries/modelbuilder.lua
TimGoll: Honestly, I don't know. This part with the `rel` is not something I used so far. But it is something that was done by original addon and I wanted to support it, so I kept it.

lua/ttt2/libraries/modelbuilder.lua
saibotk: Mhh fair, but still mention it, because it helps with understanding that this can only be used on weapon entities and is tightly coupled to some fields and functions therein

lua/ttt2/libraries/modelbuilder.lua
TimGoll: Maybe I should remove this then. This is taken from the old addon and how they tried to hide the weapon in the view model. But I found a better way that I'm using now. Therefore this is no longer needed. However I thought I keep it since I already put the work in it and it might be useful for something else as well, I'm not sure

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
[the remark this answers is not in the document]
TimGoll: I've also never used that tool, but isn't this already a good description:

https://github.com/TTT-2/TTT2/blob/0f6b4b325a3ebe82705efa6d14a6d88c08247459/lua/ttt2/libraries/modelbuilder.lua#L6-L23

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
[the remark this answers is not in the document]
TimGoll: If you favor it, I can move it. I don't care that much

lua/ttt2/libraries/modelbuilder.lua
saibotk: weaponmodelrenderer ? idk lets ask Chief Naming Officer @Histalek

lua/ttt2/libraries/modelbuilder.lua
saibotk: Weird stuff, but we surely can refactor this. Either remove these inconsistencies and try to apply those after the function call if possible or we make this a local function instead of some module function then i wouldn't care.

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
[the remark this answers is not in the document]
saibotk: Yes but this does not care about the variable naming we use to just store these elements.
I don't think this is an issue.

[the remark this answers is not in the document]
TimGoll: > Then please describe the workflow on how someone adds such models etc since im not familiar with the outputs or anything of that tool.

If they are named @TimGoll they manually try and error with those variables until it looks nice. If they use the linked tool, the tool gives them this exact table. Therefore I kept all variable names inside the table identical, so that it is as low friction as possible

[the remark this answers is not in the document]
EntranceJew: also a lot of existing TTT1 equipment is made using said tool, so it will make porting older stuff significantly easier

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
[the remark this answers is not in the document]
TimGoll: Hm, as this function both renders the world and the viewmodel, I have to pass the correct render order as an argument. So I'm not sure how I should implement your suggestion

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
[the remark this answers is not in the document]
TimGoll: I've thought about this for a while and I think it would be best to rename the variables `WElements` and `VElements`. This way we would be safe from clashes with external addons that implement the renderer from the tool.

The variable names inside the table should stay unchanged though. Otherwise porting over addons will be tricky for devs just like @EntranceJew said

lua/ttt2/libraries/modelbuilder.lua
TimGoll: I think this can also be used to make more advanced stuff, like custom finger positions etc for addons such as the harpoon. So I will keep it, but I'm not knowledgeable enough yet to give this a good documentation, sorry

lua/ttt2/libraries/modelbuilder.lua
TimGoll: I did this, I think it works, but I could not test it

lua/ttt2/libraries/modelbuilder.lua
TimGoll: made it local

TimGoll:
all done @saibotk

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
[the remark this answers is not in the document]
saibotk: Just move the function call inside the function. There are no dependencies on any other parameters. Or what are you talking about?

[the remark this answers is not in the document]
TimGoll: How would I cache it though? Should I create the cache outside the render function and pass the table reference? Because I need one view model cache and one world model cache

lua/ttt2/libraries/modelbuilder.lua
saibotk: If it works, maybe add a comment, that it is known that this is unknown. Just to prevent someone from changing it just because it does not exist in the docs

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,317 @@
...
+		else
+			dataTable = {
+				scale = Vector(1, 1, 1),
+				pos = Vector(0, 0, 0),
+				angle = Angle(0, 0, 0)
+			}
+		end
+
+		local scale = Vector(dataTable.scale.x, dataTable.scale.y, dataTable.scale.z)
+		local position = Vector(dataTable.pos.x, dataTable.pos.y, dataTable.pos.z)
```
saibotk, line 191:
```suggestion
		local scale = dataTable.scale
		local position = dataTable.pos
```

lua/ttt2/libraries/modelbuilder.lua
saibotk: Just describe what the code below tells you it does. Like literally that this applies custom bone modifications on the weapon. And if none are there it will just reset the bone positions.

lua/ttt2/libraries/modelbuilder.lua
saibotk: I think you missed this

lua/ttt2/libraries/modelbuilder.lua
```diff
@@ -0,0 +1,317 @@
...
+		local modelScale = Vector(1, 1, 1)
+
+		local currentBone = viewModel:GetBoneParent(i)
+
+		while (currentBone >= 0) do
+			modelScale = modelScale * allBones[viewModel:GetBoneName(currentBone)].scale
```
saibotk, line 197: This now needs to make a lookup in the customViewModelBoneMods instead. 

And allBones needs to be removed

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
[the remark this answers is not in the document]
saibotk: Well you can still just access the wep entity as before just store it in there or make a lookup table.

lua/ttt2/libraries/modelbuilder.lua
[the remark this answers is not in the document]
saibotk: Missing a return doc here

gamemodes/terrortown/gamemode/client/cl_main.lua
```diff
@@ -421,6 +421,9 @@ local function RoundStateChange(o, n)
 
 		-- clear decals in cache from previous round
 		util.ClearDecals()
+
+		-- resets bone positions that fixes broken fingers on bad addons
+		modelbuilder.ResetBonePositions(LocalPlayer():GetViewModel())
```
saibotk, line 426: Seeing this, i really dislike to have this on the modelbuilder module. Especially since we now use this function in a context different than the weapon base.

saibotk (changes requested):

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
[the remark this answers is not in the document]
TimGoll: yes, but both have to be called differently. How do I do this? pass two difference references? pass a string identifier?

lua/ttt2/libraries/modelbuilder.lua
Histalek: do we really need the 'model' part? shouldn't `weaponrenderer` be just fine?

TimGoll: that's fine as well, yes

lua/ttt2/libraries/modelbuilder.lua
TimGoll: image

not, it is updated

lua/ttt2/libraries/modelbuilder.lua
TimGoll: done. I checked the wiki and it seems like the vectors are not modified and therefore a copy of that vector is not needed. But I'm not 100% sure here. If there are issues down the line, this is most likely the cause

lua/ttt2/libraries/modelbuilder.lua
TimGoll: oh wow, good catch. Totally missed this

gamemodes/terrortown/gamemode/client/cl_main.lua
TimGoll: hm, I still think it fits nicely in that library - especially after we rename it

TimGoll:
@saibotk I think everything is done from my side. We still have to decide on a new name though

TimGoll:
ok shit, I noticed an issue with view model rendering of weapon such as the boomerang. I have to look into this

TimGoll:
Fixed the issue. It was a bit tricky to solve as the root of the problem was that these custom weapoins use the same `SWEP:ViewModelDrawn` hook without calling the base class function. The issue here is that I also use this draw hook to apply the "do not render the view model" debug material. This is applied to the **view model** not the current weapon meaning that switching from our holstered (which has the view model hidden) to a weapon that overwrites `SWEP:ViewModelDrawn` results in the view model not being drawn even for this new weapon.

My solution now uses `SWEP:Holster` and `SWEP:OnRemove` to circumvent this issue. This way this can be controlled by weapons that actually use our new functions. Meaning that any weapon that uses the new systems introduced in this PR has to make sure that the `SWEP:Holster` and `SWEP:OnRemove` functions of the base class are called. Either by not overwriting them or by making sure that they are called via a base class call.

In this last commit I did exactly this. To be extra safe I added it to every weapon that overwrites these functions, even if they don't use a custom view model.

lua/ttt2/libraries/weaponrenderer.lua
```diff
@@ -0,0 +1,321 @@
...
+		end
+
+		if viewModel:GetManipulateBonePosition(i) ~= position then
+			viewModel:ManipulateBonePosition(i, position)
+		end
+	end
+end
+
+---
+-- Builds a render order or returns the cache if already built. The render order is necessary
```
saibotk, line 226:
```suggestion
-- Returns an ordered elements table. The render order is necessary
```

saibotk (changes requested):

saibotk (approved):

lua/ttt2/libraries/weaponrenderer.lua
```diff
@@ -0,0 +1,321 @@
...
+-- Used to build custom world and view models. It also renders those custom models.
+-- This module can only be used by entities that are based on `weapon_ttt_base` as this
+-- relies on some functions defined in this class.
+-- This code is based on the SWEP construction kit.
+-- @ref https://github.com/MagicSwap/SWEP_Construction_Kit
+-- @author Mineotopia
+-- @module weaponrenderer
+
+---@class ModelData
+---@field type string The type of this model, it can be "Model", "SPrite" or "Quad"
```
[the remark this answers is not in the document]
TimGoll: I'm not sure if there is any documentation for that. It is taken from the SWEP construction kit

Histalek (approved):

PR #1331, opened 2024-01-27
do not leak SendFullStateUpdate events happen

EntranceJew:
if someone has developer 1 on then they can see that an event which triggered a full state rebroadcast, aka, someone's role changed


TimGoll (approved):

TimGoll:
Add a changelog entry please

saibotk (approved):
I don't think this needs a changelog

PR #1332, opened 2024-01-27
Logging: Unify usage of log levels

EntranceJew:
if we expect people to fix them then we should help them find them


gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
```diff
@@ -651,16 +651,16 @@ function addonChecker.Check()
 		local detectedAddon = addonChecker.curatedList[tostring(addon.wsid)]
 		if not detectedAddon then continue end
 
-		ErrorNoHalt(((detectedAddon.type == ADDON_OUTDATED) and "Outdated add-on detected: " or "Incompatible add-on detected: ") .. addon.title .. "\n")
+		ErrorNoHaltWithStack(((detectedAddon.type == ADDON_OUTDATED) and "Outdated add-on detected: " or "Incompatible add-on detected: ") .. addon.title .. "\n")
 
 		if detectedAddon.reason then
-			ErrorNoHalt("Reason: " .. detectedAddon.reason .. "\n")
+			ErrorNoHaltWithStack("Reason: " .. detectedAddon.reason .. "\n")
```
TimGoll, line 657: Im this file it makes no sense

gamemodes/terrortown/gamemode/server/sv_karma.lua
```diff
@@ -635,7 +635,7 @@ function KARMA.Remember(ply)
 	end
 
 	if ply:SteamID64() == nil then
-		print("[TTT2] ERROR: Player has no steamID64")
+		ErrorNoHaltWithStack("[TTT2] ERROR: Player has no steamID64:" .. ply)
```
TimGoll, line 638: Does this work without a tostring

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -356,13 +356,13 @@ function GM:Initialize()
 	WaitForPlayers()
 
 	if cvars.Number("sv_alltalk", 0) > 0 then
-		ErrorNoHalt("TTT2 WARNING: sv_alltalk is enabled. Dead players will be able to talk to living players. TTT2 will now attempt to set sv_alltalk 0.\n")
+		ErrorNoHaltWithStack("TTT2 WARNING: sv_alltalk is enabled. Dead players will be able to talk to living players. TTT2 will now attempt to set sv_alltalk 0.\n")
```
TimGoll, line 359: I think a stack isn't useful here

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -356,13 +356,13 @@ function GM:Initialize()
...
 
 		RunConsoleCommand("sv_alltalk", "0")
 	end
 
 	if not IsMounted("cstrike") then
-		ErrorNoHalt("TTT2 WARNING: CS:S does not appear to be mounted by GMod. Things may break in strange ways. Server admin? Check the TTT readme for help.\n")
+		ErrorNoHaltWithStack("TTT2 WARNING: CS:S does not appear to be mounted by GMod. Things may break in strange ways. Server admin? Check the TTT readme for help.\n")
```
TimGoll, line 365: Same here

gamemodes/terrortown/gamemode/shared/sh_equip_items.lua
```diff
@@ -726,7 +726,7 @@ function GenerateNewEquipmentID()
 
 			if v.oldId ~= val or not v.id then continue end
 
-			print("[TTT2][WARNING] TTT2 doesn't support old items completely since they are limited to an amount of 16. If the item '" .. v.id .. "' with id '" .. val .. "' doesn't work as intended, modify the old item and use the new items system instead.")
+			ErrorNoHaltWithStack("[TTT2][WARNING] TTT2 doesn't support old items completely since they are limited to an amount of 16. If the item '" .. v.id .. "' with id '" .. val .. "' doesn't work as intended, modify the old item and use the new items system instead.")
```
TimGoll, line 729: Stack isn't helpful here I think

lua/ttt2/libraries/items.lua
```diff
@@ -172,7 +172,7 @@ function items.Get(name, retTbl)
 		local base = items.Get(retval.Base)
 
 		if not base then
-			Msg("ERROR: Trying to derive item " .. tostring(name) .. " from non existant item " .. tostring(retval.Base) .. "!\n")
+			ErrorNoHaltWithStack("ERROR: Trying to derive item " .. tostring(name) .. " from non existant item " .. tostring(retval.Base) .. "!\n")
```
TimGoll, line 175: As this is also only internally called, I'm not sure a stack is any helpful here

lua/ttt2/libraries/plyspawn.lua
```diff
@@ -203,7 +203,7 @@ function plyspawn.GetRandomSafePlayerSpawnPoint(ply)
 
 		if not riggedSpawnPoint then continue end
 
-		ErrorNoHalt("TTT2 WARNING: Map has too few spawn points, using a riggedSpawnPoints spawn for " .. tostring(ply) .. "\n")
+		ErrorNoHaltWithStack("TTT2 WARNING: Map has too few spawn points, using a riggedSpawnPoints spawn for " .. tostring(ply) .. "\n")
```
TimGoll, line 206: This is on the map maker and happens quite a lot. Please remove that here, there is nothing the dev can change about that

TimGoll (changes requested):
This is already quite nice, but there are some places where this shouldn't be changed. I might have missed some, maybe you should check them yourself again and think if this adds any benefit.

Also I'm not sure if we should add the trace to the language stuff as these old addons will likely never be updated and my console is already flooded with depreciation warnings.

EntranceJew:
FWIW my point of reference for severity is as follows:
- **ErrorNoHaltWithStack** things that should be fixed, or would otherwise get an addon put in the addonchecker.
- **Dev(1, ...)** anything that qualifies as an indication of program flow / features being debugged
- **Dev(2, ...)** extraneous debugging noise
- **Dev(N, ...)** even noisier noise
- **print** should only happen as a direct consequence of debug commands 

banished:
- **MsgN**does not convey a priority
- **ErrorNoHalt** functionally the same as MsgN, but with an error connotation
- **Error** identical to ErrorNoHalt, easy to mistake for `error`, therefore, same treatment
- **error** if we're going to error, we may as well promote a stack via **ErrorNoHaltWithStack** -- unless we need to redirect blame via `errorlevel`

TimGoll:
I absolutely agree with you. My raised points still hold true because we don't want to spam the console when internal code raises these flags as the developers have no chance of fixing it

lua/ttt2/libraries/items.lua
EntranceJew: If it happens at all, it should be fixed, right? Some indication of how someone got there is important. I definitely utilize this inside PotOfGreedier.

TimGoll:
But the ones I marked are unfixable, like this one: 
image

this could only be fixed by the map dev, but this isn't going to happen

EntranceJew:
> this could only be fixed by the map dev, but this isn't going to happen. Also a trace is not useful at all in this case

i hear you, i just can only respond to things to keep track of what i changed and didn't change, might contest individual situations :U

gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
Histalek, line 1: Please revert the changes in this file. We want people to always see the Output of the addon checker without any extra steps involved.

Histalek (changes requested):
Anything besides the addonchecker changes looks good :)

gamemodes/terrortown/entities/entities/ttt_credit_adjust.lua
```diff
@@ -20,7 +20,7 @@ function ENT:KeyValue(key, value)
 		self.Credits = tonumber(value) or 0
 
 		if not tonumber(value) then
-			ErrorNoHalt(tostring(self) .. " has bad 'credits' setting.\n")
+			ErrorNoHaltWithStack(tostring(self) .. " has bad 'credits' setting.\n")
```
TimGoll, line 23: Could you explain here the helpfulness of this? Because this is also called from a map entity, meaning that the stacktrace will probably not point to any error that actually exists.

I'm not entirely sure though, that's just my impression. Feel free to say I'm wrong here

gamemodes/terrortown/entities/entities/ttt_game_text.lua
```diff
@@ -39,7 +39,7 @@ function ENT:KeyValue(key, value)
 		self.Receiver = tonumber(value)
 
 		if not self.Receiver or self.Receiver < 0 or self.Receiver > 5 then
-			ErrorNoHalt("ERROR: ttt_game_text has invalid inputReceiver value\n")
+			ErrorNoHaltWithStack("ERROR: ttt_game_text has invalid inputReceiver value\n")
```
TimGoll, line 42: same here for these two. I'm not saying this is wrong, I just wonder if it adds any meaningful information here?

gamemodes/terrortown/entities/entities/ttt_logic_role.lua
```diff
@@ -29,7 +29,7 @@ function ENT:KeyValue(key, value)
 		self.checkingRole = tonumber(value)
 
 		if not self.checkingRole then
-			ErrorNoHalt("ttt_logic_role: bad value for Role key, not a number\n")
+			ErrorNoHaltWithStack("ttt_logic_role: bad value for Role key, not a number\n")
```
TimGoll, line 32: and here as well, this is also a map entity that is never triggered inside lua. This is an error the mapmaker has to take care of, hence I think the stack is useless

gamemodes/terrortown/entities/entities/ttt_map_settings.lua
```diff
@@ -66,11 +66,11 @@ function ENT:AcceptInput(name, activator, caller, data)
 		local mdlname = tostring(data)
 
 		if not mdlname then
-			ErrorNoHalt("ttt_map_settings: Invalid parameter to SetPlayerModels input!\n")
+			ErrorNoHaltWithStack("ttt_map_settings: Invalid parameter to SetPlayerModels input!\n")
```
TimGoll, line 69: here as well, this is also a map ent and a few lines down below as well

gamemodes/terrortown/entities/entities/ttt_weapon_check.lua
```diff
@@ -102,7 +102,7 @@ function ENT:TestWeapons(weptype)
 	local check = self:GetWeaponChecker(weptype)
 
 	if check == nil then
-		ErrorNoHalt("ttt_weapon_check: invalid parameter\n")
+		ErrorNoHaltWithStack("ttt_weapon_check: invalid parameter\n")
```
TimGoll, line 105: here as well, these two checks are map entities

gamemodes/terrortown/gamemode/client/cl_lang.lua
```diff
@@ -69,7 +69,7 @@ function LANG.AddToLanguage(langName, stringName, stringText)
 	langName = LANG.GetNameFromAlias(langName)
 
 	if not LANG.IsLanguage(langName) then
-		ErrorNoHalt(Format("Failed to add '%s' to language '%s': language does not exist.\n", tostring(stringName), tostring(langName)))
+		ErrorNoHaltWithStack(Format("Failed to add '%s' to language '%s': language does not exist.\n", tostring(stringName), tostring(langName)))
```
TimGoll, line 72: as I said in my last review, my console is already spammed with lang deprecation warning. If we add t more lines to each warning I think the console becomes unusable, don't you think?

gamemodes/terrortown/gamemode/client/cl_radio.lua
```diff
@@ -273,21 +273,21 @@ end
...
+		ErrorNoHaltWithStack("ttt_radio failed, invalid player")
 
 		return
 	end
 
 	---
 	-- @realm client
 	if hook.Run("TTT2ClientRadioCommand", cmd) then
-		print("ttt_radio, execution prevented by a hook")
+		ErrorNoHaltWithStack("ttt_radio, execution prevented by a hook")
```
TimGoll, line 284: no, this is bad. If a hook prevents the radio, it shouldn't throw an error. But tbh, this print should probably be removed or moved to a MStack message

gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
TimGoll: agreed, the addonchecker should always be visible to everyone

TimGoll (commented):
This is now mostly good, thanks! There are a few questions and one major issue here

gamemodes/terrortown/gamemode/client/cl_lang.lua
TimGoll: well this comment is tied to the wrong line, sorry about that. This specific change here is good.

But in the same file is a deprecation warning. I'm talking about the one in line 223, the rest seems fine

gamemodes/terrortown/gamemode/client/cl_radio.lua
Histalek: If this should be moved to a MStack message this would also need a lang string.

i'd say reverting this for now matches the scope of this PR.

TimGoll: or put it as a DEV(1 for now

gamemodes/terrortown/gamemode/client/cl_lang.lua
EntranceJew: the problem is that this error does not indicate the oirigin of the problem. even if i wanted to fix this, i would need to unpack every addon i have in order to locate the problematic language file, and it's so far removed from the source file directory that it cannot be associated _without_ a stack trace

gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
EntranceJew: > This looks mostly sane to me, i would change 2 things though.
> 
> * `print()` should not be in our codebase

wat do

Histalek: > > This looks mostly sane to me, i would change 2 things though.
> > 
> > * `print()` should not be in our codebase
> 
> wat do

**should** :D 
besides that, this is a special case as it is more akin to program output and not a logged error/problem. `print` therefore is fine.

Histalek (approved):

gamemodes/terrortown/gamemode/client/cl_lang.lua
TimGoll: I absolutely understand that. But 99% of the people only want to play TTT2 and will be pretty confused when there are errors on every mapchange that they can't fix

TimGoll: I reverted this, @Histalek and I talked about a better solution in our debug library that can be done in a coming PR so that we can merge this one!

TimGoll (approved):

PR #1333, opened 2024-01-28
Fix issues with marks library

WardenPotato:
This PR addresses the marks library having self zfailing and color issues.

This originated from helping @TimGoll on the [gmod discord](discord.gg/gmod) with this issue, more context can be found [there](https://discord.com/channels/565105920414318602/567617926991970306/1200805871303786506)

For completeness sake i will also mirror his original help request here:
```
It is me again, this time with a more in depth question about stencils and rendering. It is about this part in our TTT2 codebase: https://github.com/TTT-2/TTT2/blob/master/lua/ttt2/libraries/marks.lua#L76-L108

It is used to render "marks" (or wallhacks) of an entity through a wall (see picture 1)

But there is a problem. We color the entity with ent:SetColor on spawn to the player's role color. The entity itself is white. This works fine (see image 2) if no mark/wallhack is used.

But if I combine the wallhack with ent:SetColor() the color is removed in places where the entity intersects with itself or another entity that is colored.

Do note, that this issue only exists with entities that use ent:SetColor. If the material itself is already colored, then this works fine without issues.

So my question is, does anyone here know how to resolve this? I guess it shuld be possible somehow, but I'm unable to figure it out. Thanks in advance!
```

I will also attach the original three screenshots from the help request:
broken1
broken2
broken3

And one from my fix:
the test props, 4 of each are from top to bottom:
Original
Colored red
debugwhite material + red color
the mosaic texture (this was important to test funky material stuff)

Heres another screenshot highlighting self zfailing on the barrel (this was done on purpose to highlight the issue, my fix does not suffer from this), its subtle but you can see it on the ridges of the barrel:
brokenbarrel

Excuse the complete lack of a TTT2 dev env in my screenshot as i normally have little to do with TTT (this was tested by just copying marks.lua into my existing dev env)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed last seen player being wrongly visible for every search instead of only public policing role search (by @TimGoll)
 - Fixed the crosshair being offcenter on some UI scales (by @TimGoll)
 - Fixed to wrong line calculations for wrapped text (by @NickCloudAT)
+- Fixed marks library having self zfailing and color issues (by @WardenPotato)
 
 ### Removed
 
--- a/lua/ttt2/libraries/marks.lua
+++ b/lua/ttt2/libraries/marks.lua
@@ -9,8 +9,6 @@
 local render = render
 local table = table
 local IsValid = IsValid
-local cam = cam
-local surface = surface
 local hook = hook
 local pairs = pairs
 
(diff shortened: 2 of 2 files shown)


TimGoll:
Awesome, thank you so much! I will test this in an hour or so

EntranceJew:
code here looks clean and reasonable.

if you have any extra time it'd be really interesting to get your insight on whatever is happening [on this issue](https://github.com/TTT-2/TTT2/pull/1267), there are some [serious seizure inducing problems[(https://cdn.discordapp.com/attachments/547119387392475136/1196219794882576594/Garrys_Mod_x64_33360_2024-01-14_16-30-04.mp4) on the arcs, but only sometimes, and only on x64, and it doesn't occur with circles but does for our implementation.

TimGoll:
OMG, it is actually working. I did not believe it until I tried it myself!

image
image

TimGoll (approved):

saibotk (approved):
Thank you very much for going through the extra work and even opening up a PR to fix this issue!
We really really appreciate your help here, so a huge thank you!

I am not that proficient with the stencil operations but the code itself looks good and documented in a way that is understandable to someone who knows about stencils from the wiki i'd say. So this is fine for me!

PR #1335, opened 2024-01-29
Voice: safe guard against invalid DB volume values

### LINKED ISSUE #1334: Voice Chat Bug
TTT2 - v0.12.3b

One player is having an unreproducable issue with voice chat. 
Console shows following error whenever that player attempts to talk.

[ERROR] gamemodes/terrortown/gamemode/client/cl_voice.lua:499: bad argument #1 to 'SetVoiceVolumeScale' (number expected, got nil)
  1. SetVoiceVolumeScale - [C]:-1
   2. UpdatePlayerVoiceVolume - gamemodes/terrortown/gamemode/client/cl_voice.lua:499
    3. unknown - gamemodes/terrortown/gamemode/client/cl_voice.lua:176

All other players do not have this issue, and I have been unable to replicate it. 
Error continues when server is run with no other addons enabled. 
Player can use voice chat normally when base TTT is run instead.

EntranceJew:
maybe fixes #1334 if the database stuff doesn't

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_voice.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_voice.lua
@@ -496,6 +496,12 @@ function VOICE.UpdatePlayerVoiceVolume(ply)
 		out_vol = func( vol )
 	end
 
+	-- if the database fails for strange people then blast their ears out and save it
+	if out_vol ~= 0 and not out_vol then
+		out_vol = 1
+		VOICE.SetPreferredPlayerVoiceVolume(ply, out_vol)
+	end
+
 	ply:SetVoiceVolumeScale( out_vol )
 
 	return out_vol, mute


ZenBre4ker:
This shouldnt  be necessary. As far as I can see, when the  sql.query returns nil or anything else, its always converted to a number in `VOICE.GetPreferredPlayerVoiceVolume`. Therefore I believe, that the scalingfunction could be the problem. Maybe just add an
`out_vol = func( vol ) or vol`

ZenBre4ker:
I tested all three scaling functions with weird values and it seemed fine, linear is also default, which also just puts `out_vol = vol`, therefore the error is rather in an overwrite of a function.The sql table getter outputs well defined numbers in any case.

Im not fan of this fix, as it might not neutralize the problem itself. In case the db is fine, we have a database write everytime someone starts talking.

EntranceJew:
The code that I made is pretty airtight, and all the checks have been working flawlessly for users on a server with an existing database and for users client-sided in a new database that were specifically set up to join and play.

I can't find a reason why this would happen. All the checks that are in place are designed to minimize the amount of values flowing into C functions and to the database, but still have the potential to update dynamically to prevent a player from being dramatically louder or quieter between circumstances.

The only probable causes in my mind are:
- numeric comparison precision accuracy problems?
- database type coercion?
- a client or server has a read-only database and thus has the schema but not the new records

ZenBre4ker:
In case this should be worked again. IMO we should just implement a default value with `or` in case of `nil`. I'm against setting the value, because it could be something else is broken and this code just wirtes to the db everytime `nil` comes up. It's not safe to say, that by setting the preferred value, that the actual error will be fixed in the next call.

EntranceJew:
> This shouldnt be necessary. As far as I can see, when the sql.query returns nil or anything else, its always converted to a number in `VOICE.GetPreferredPlayerVoiceVolume`. Therefore I believe, that the scalingfunction could be the problem. Maybe just add an `out_vol = func( vol ) or vol or 1`

I don't think so. Similar to the shop issues we discussed, the only scaling function that could have this flaw is linear, but that is because the data is arriving to it as a string and not a number. The existing conditions assume `nil` or a falsey value which is not the expected number. Why would something recorded as a numeric type be permitted to come back as a string?

saibotk:
Thanks for your contribution, but the conclusion here is that we'd rather fix root causes for this than setting the volume to be the maximum volume which can be very weird for players.

PR #1336, opened 2024-01-29
MarkerVision: Added Static Role/Team Support and API Rework

TimGoll:
While reworking TTT2 Heroes a bit I noticed that we need markers that are not only bound to a player, but to a role or team. Therefore I extended the code so that the owner can be a player or a role/team. While doing this I noticed a few other issues with the markerVision library, most notably that is was impossible to have two marks (for different players) on the same identity. This pullequest resolves this.

E.g. the heroes crystal now works like this:

```lua
local mvObject = self:AddMarkerVision("crystal_supervillain")
mvObject:SetOwner(ROLE_SUPERVILLAIN)
mvObject:SetVisibleFor(VISIBLE_FOR_ROLE)
mvObject:SyncToClients()
```

This also already tracks role changes. So when a player changes their role, they lose access to the wallhack. But they can gain it back, if they change their role back.

## Note

If you'd replace `ROLE_SUPERVILLAIN` with a player, the player would keep the mark ownership and take them to their new role.

```lua
local mvObject = self:AddMarkerVision("crystal_supervillain")
mvObject:SetOwner(ply)
mvObject:SetVisibleFor(VISIBLE_FOR_ROLE)
mvObject:SyncToClients()
```

This would still keep it to the role (the role the player had when registering the mark). But if they'd update their role to become an innocent, every innocent would then be able to see it.

Take a look at the entity diffs to see what the API changes mean in reality:
https://github.com/TTT-2/TTT2/pull/1336/files#diff-25ef72df9ab41eb109b09f41c0d8612a92b99b77e1501762366ca1b09be33e27


TimGoll:
I added skip changelog here because marker vision is still new and I hope this can be released pretty quickly. It also won't break any existing uses - if there are any

lua/ttt2/libraries/none.lua
```diff
@@ -86,7 +86,7 @@ end
 -- @return boolean
 -- @realm shared
 function IsPlayer(ent)
-	return ent and IsValid(ent) and ent:IsPlayer()
+	return ent and isentity(ent) and IsValid(ent) and ent:IsPlayer()
```
Histalek, line 89: I suppose this solves some kind of edge-case? Could you explain what that edge-case is?

Also this would be a change in behaviour and should be noted in the changelog.

lua/ttt2/libraries/marker_vision.lua
```diff
@@ -24,7 +25,8 @@ markerVision.registry = {}
 -- Registers an entity that should be rendered on screen with a wallhack and an optional
 -- UI element that works similar to targetID.
 -- @param Entity ent The entity that should be rendered
--- @param Player owner The owner of the wallhack that takes their ownershipo with them on team change
+-- @param number|string|Player owner The owner of the wallhack that takes their ownership with them on team change; can
+-- also be a team (string) or role (number) if it shouldn't be bound to a player
 -- @param number visibleFor Visibility setting: `VISIBLE_FOR_PLAYER`, `VISIBLE_FOR_TEAM`, `VISIBLE_FOR_ALL`
```
Histalek, line 30:
```suggestion
-- @param number visibleFor Visibility setting: `VISIBLE_FOR_PLAYER`, `VISIBLE_FOR_ROLE`, `VISIBLE_FOR_TEAM`,  `VISIBLE_FOR_ALL`
```

lua/ttt2/libraries/marker_vision.lua
```diff
@@ -33,10 +35,13 @@ markerVision.registry = {}
...
-		local plysTeam = GetTeamFilter(owner:GetTeam(), true, true)
+		-- handle the player being in a team that is not known, like the unknown
+		if IsPlayer(owner) then
+			local plysTeam = GetTeamFilter(owner:GetTeam(), false, true)
 
-		if #plysTeam == 0 then -- happens for TEAM_NONE for example
-			visibleFor = VISIBLE_FOR_PLAYER
+			if #plysTeam == 0 then -- happens for TEAM_NONE for example
+				visibleFor = VISIBLE_FOR_PLAYER
+			end
```
Histalek, line 44: whoops, seems like i missed this in the last review...

If I understand correctly this means that in this case the MarkerEntity would not carry over to a new role or team if the player owner would change roles/teams, right? This is probably irrelevant for TEAM_NONE but if other teams would suffer from this i'd like to fix this somehow.

Suppose the dev of the melon mine intends for the team of the player to be informed about the location of the mine. They would set `VISIBLE_FOR_TEAM`, but maybe the player is currently in a team that suffers from this condition and then gets its `visibleFor` value overridden. The player then changes team, but the team does not get informed about the markerEntity as it is now only `VISIBLE_FOR_PLAYER`.
Did i understand this correctly?

Histalek (changes requested):

lua/ttt2/libraries/none.lua
TimGoll: IsPlayer checks if a entity is a player. However if you give a string or a number to that function, it errors. So I added this check to simplify the code

Histalek: tbh that qualifies as a bug fix then :D (and should be in the changelog)

lua/ttt2/libraries/marker_vision.lua
TimGoll: I think this only can happen for players in team none, because this is the only team that set the `.alone` flag. And I don't know to which team I should attach it here as the player has "team none"

TimGoll: If I'd attach it to team none a unknow could see the marks of an amnesiac

Histalek: Couldn't we stuff the handling for this case into the if-clause on line 57 then? that way we don't mangle the intention of the consumer.

TimGoll: so you mean it will still get changed to "visible only to you" internally, gut it is registered as whatever it was before so that it can carry over on a team change?

Histalek: yes exactly

TimGoll: Sounds doable, let me try

TimGoll:
@Histalek @saibotk The scope of this PR changed and this is a full rework of the marker vision system

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -146,6 +146,16 @@ function plymeta:SetRole(subrole, team, forceHooks, suppressEvent)
 		---
 		-- @realm shared
 		hook.Run("TTT2UpdateSubrole", self, oldSubrole, subrole)
+
+		if SERVER then
+			-- trigger the update in the next tick to make sure the role is updated
+			-- on the client first before any markerVision updates are sent
```
saibotk, line 152: Why is this important?
This just describes what the code already describes.

lua/ttt2/libraries/marker_vision.lua
```diff
@@ -21,160 +22,203 @@ markerVision = {}
...
 end
 
 ---
--- Returns the visibleFor flag for a provided entity, nil if not set
--- @param Entity ent The tracked entity
--- @return number|nil The visibility flag
+-- Returns the marker vision element if it exists.
+-- @param Entity ent The entity which should receive the marker vision element
+-- @param string identifier The unique identifier of this marker vision element
+-- @return MARKER_VISION_OBJECT The marker vision object
```
saibotk, line 46: Missing the 2nd return value.
And missing the possible nil case

Also is the second param even needed?

lua/ttt2/libraries/marker_vision.lua
```diff
@@ -21,160 +22,203 @@ markerVision = {}
...
-	if not IsValid(ent) or not markerVision.registry[ent] then return end
+function markerVision.Get(ent, identifier)
+	for i = 1, #markerVision.registry do
+		local mvObject = markerVision.registry[i]
 
-	return markerVision.registry[ent].visibleFor
+		if mvObject:IsSearchedObject(ent, identifier) then
+			return mvObject, i
+		end
+	end
```
saibotk, line 55: Personally, i would have used a 2-dimensional lookup table from ent -> identifier -> mvObject instead. 

But this is rather an optional suggestion.

lua/ttt2/libraries/marker_vision.lua
```diff
@@ -21,160 +22,203 @@ markerVision = {}
...
+
+			if mvObjectData.visibleFor ~= VISIBLE_FOR_TEAM then continue end
+
+			-- case 1: the owner is a player and they are the one that changed their team
+			-- this means that the entity is taken to the new team
+			if IsPlayer(mvObjectData.owner) and mvObjectData.owner == ply then
+				mvObject:UpdateRelations(mvObjectData.owner, VISIBLE_FOR_TEAM)
+				mvObject:SyncToClients()
+
+				--markerVision.UpdateEntityOwnerTeam(entry, oldTeam, newTeam)
```
saibotk, line 141: Remove

saibotk (changes requested):
Okay mainly some clarifications needed.

Also I think we should talk about the syncing logic as my main concern, as outlined by the comments. If not please tell me and i will try to elaborate more.

Thank you, over all some nice additions!

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
```diff
@@ -0,0 +1,244 @@
...
+	function MARKER_VISION_OBJECT:SyncToClients(receiverListOverwrite, passThroughData)
+		if not self.data.owner then return end
+
+		-- if data was previously set, it should be removed first on the client
+		if self.lastReceiverList then
+			net.Start("ttt2_marker_vision_entity_removed")
+			net.WriteEntity(self.data.ent)
+			net.WriteString(self.data.identifier)
+			net.Send(self.lastReceiverList)
+		end
```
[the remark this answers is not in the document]
TimGoll: It is the easiest way to handle two things:
- a player that lost access to a markerVision element 
- a marker vision element that changed color as one has to remove and readd the mark client-side to achieve a color change

lua/ttt2/libraries/marker_vision.lua
TimGoll: My main concern was performance when rendering. Therefore I chose this structure

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
```diff
@@ -0,0 +1,244 @@
...
+			end
+		elseif self.data.visibleFor == VISIBLE_FOR_TEAM then
+			if IsPlayer(self.data.owner) then
+				self.receiverList = GetTeamFilter(self.data.owner:GetTeam(), false, true)
+			else
+				-- handle static team
+				self.receiverList = GetTeamFilter(self.data.owner, false, true)
+			end
+		else
+			self.receiverList = nil
```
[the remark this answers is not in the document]
TimGoll: I see what you mean but it is expected to call this function and the sync function together. I thought it looked cleaner to split it up

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
```diff
@@ -0,0 +1,244 @@
...
+	self.data.passThroughData = passThroughData
+end
+
+---
+-- Gets the pass through data to the element.
+-- @return any The pass through data
+-- @realm shared
+function MARKER_VISION_OBJECT:GetPassThroughData()
+	return self.data.passThroughData
+end
```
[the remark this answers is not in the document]
TimGoll: Yes it is unused but I thought it would be nice. I can remove it though

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
```diff
@@ -0,0 +1,244 @@
+---
+-- @author Mineotopia
+-- @class MARKER_VISION_OBJECT
```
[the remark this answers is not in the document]
TimGoll: I see what you are saying. The data object is named the same as the targetID one therefore the name.

I don't have a good name for the synced one (the object) but it simplifies it a lot because I can have functions on it and don't need any manual table manipulations.

I'm not a fan of merging these two as one is purely for rendering and the other is for syncing and managing.

The data object also is not per entity but one that gets changed its content per loop iteration

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
```diff
@@ -0,0 +1,244 @@
...
+function MARKER_VISION_OBJECT:GetVisibleFor()
+	return self.data.visibleFor
+end
+
+---
+-- Gets the visible for flag as a language string to the element.
+-- @return string The visible for flag as a language string
+-- @realm shared
+function MARKER_VISION_OBJECT:GetVisibleForString()
+	return "marker_vision_visible_for_" .. tostring(self:GetVisibleFor())
```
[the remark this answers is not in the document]
TimGoll: Because it is used in energy entity and I had to use the same line everywhere. This simplified it and made it more robust for future possible changes

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
```diff
@@ -0,0 +1,244 @@
...
+
+---
+-- Updates the relations of this marker vision element.
+-- @note This does not sync it to the client, call @{MARKER_VISION_OBJECT:SyncToClients} to sync to client.
+-- @param number|string|Player owner The owner of the wallhack that takes their ownership with them on
+-- team change; can also be a team (string) or role (number) if it shouldn't be bound to a player
+-- @param number visibleFor Visibility setting: `VISIBLE_FOR_PLAYER`, `VISIBLE_FOR_ROLE`, `VISIBLE_FOR_TEAM`,
+-- `VISIBLE_FOR_ALL`
+-- @realm shared
+function MARKER_VISION_OBJECT:UpdateRelations(owner, visibleFor)
```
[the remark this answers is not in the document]
TimGoll: But these two alre always set together. Not sure if splitting them up is actually helpful?

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
```diff
@@ -0,0 +1,244 @@
...
+			streamData.color = streamData.color or TEAMS[LocalPlayer():GetTeam()].color
+		end
+
+		marks.Add({streamData.ent}, streamData.color)
+	end)
+
+	net.Receive("ttt2_marker_vision_entity_removed", function()
+		markerVision.Remove(net.ReadEntity(), net.ReadString())
+	end)
+end
```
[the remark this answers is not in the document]
TimGoll: I'm not sure how I would do the syncing in the library if it is triggered by the object. It kinda feels like a loop then. The library includes the object and the object includes the library

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
TimGoll: Liie the first comment says: it makes sure that it is executed after the role updates is finished

lua/ttt2/libraries/marker_vision.lua
TimGoll: Yew it is needed as only a combination of ent and identifier is unique

lua/ttt2/libraries/marker_vision.lua
saibotk: Table lookups are faster though

TimGoll: I could cache the table index on the object. That way we would have the best performance imho

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
saibotk: Yes but why?

lua/ttt2/libraries/marker_vision.lua
saibotk: I meant second return value

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
TimGoll: Because if the order is wrong, the client still has the old role information. I therefore either have to delay the transmission or send all the role data that might be needed along with this update

lua/ttt2/libraries/marker_vision.lua
TimGoll: This is also needed. Otherwise I'd need a direct table access eith the index which I try to avoid

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
saibotk: But what makes it important that the clients role info is fully correct here? What calculations are done that are not reactive? Thats what the comment should describe

lua/ttt2/libraries/marker_vision.lua
TimGoll: This is executed on every frame. It is always advised against using pairs loops in rendering. Which I would have to do if I'm using your suggestion

If I'd use two sequential tables together, I don't see any benefit except having a more complex structure for no real reason

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
TimGoll: Right now mainly the color. But it is up to the dev that puts their code in the hook.

TimGoll: But the comment literally says that this is done to update the role on the client first. And this is the reason: update the role on the client first.

Adding a comment like "to make sure that the role is updated on the client first first that one can access the correct role" seems redundant

lua/ttt2/libraries/marker_vision.lua
saibotk: I think you dont get what im saying here. I suggest using a table indexed by entity and then id. You are currently looping over it

TimGoll: Yes, I did that previously but this would mean two pairs loops combined which I really don't want to have inside a rendering hook

saibotk: No it does not its a Lookup Table that means -> registry[entitiy][identifier] and you got it

TimGoll: I know. But I have to iterate over it each frame. Or should I have two tables? One for rendering, one for lookup?

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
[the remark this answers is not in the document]
TimGoll: I want to add here that a more in depth update system would probably work as well, but this would be way more complex code wise. So I chose this simple approach as it does the job without much complexity

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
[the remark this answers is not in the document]
saibotk: Please do so

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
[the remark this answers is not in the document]
saibotk: I only said you should clarify this in a comment, its fine that there are 2 they are just very confusing

lua/ttt2/libraries/marker_vision.lua
saibotk: It is still undocumented. Thats all what this is about.

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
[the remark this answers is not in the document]
saibotk: You are misunderstanding me again. I am asking those questions not because i necessarily want you to change this but rather as something that i asked myself when reading this and thus i noted that you might want to elaborate on the comment. 
These cases that you just listed are a perfect example for a comment. Something i was not aware of when reading the code. This makes it so much easier to refactor this later, because i now know why this was done here.

lua/ttt2/libraries/marker_vision.lua
TimGoll: oh, sorry, I misunderstood this

lua/ttt2/libraries/marker_vision.lua
saibotk: First of: 2 loops are not really an issue for such code. This would really be a micro optimization i believe, but as i said this is an optional suggestion so will close this.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
saibotk: You don't get what im saying here. Answering the why is important here. Explaining which cases otherwise fail is important. It is still not obvious why we would care about the order of updates. Also your example comment is badly worded and does not answer the question of why doing this.

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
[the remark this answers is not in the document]
saibotk: Just move the functions there, i don't get what you are trying to say here.
The library then has a function with the DTO as a parameter and the DTO does not have any specific functionality on its own.

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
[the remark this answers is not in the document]
saibotk: It would certainly be much cleaner and make the naming easier.
Sometimes or rather most of the time the owner is the only thing that dynamically changes.

gamemodes/terrortown/gamemode/shared/sh_marker_vision_object.lua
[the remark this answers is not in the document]
saibotk: I think you missed my point here again. This is more about keeping these lists in sync:
Now you need to update every marker vision object when a player leaves or connects etc. and the playerlist and thus the potential receivers change. This can easily be avoided by not precomputing these and moving this to the latest possible computation, where it is actually needed.
It just makes it easier to not accidentally miss updating the receivers.

saibotk (changes requested):

TimGoll:
@saibotk I think all should be done

gamemodes/terrortown/gamemode/shared/sh_marker_vision_element.lua
```diff
@@ -0,0 +1,198 @@
...
+-- @realm shared
+function MARKER_VISION_ELEMENT:GetColor()
+	if self.data.visibleFor == VISIBLE_FOR_ALL then
+		return self.data.color or TEAMS[TEAM_INNOCENT].color
+	elseif self.data.visibleFor == VISIBLE_FOR_PLAYER then
+		return self.data.color or TEAMS[TEAM_NONE].color
+	elseif self.data.visibleFor == VISIBLE_FOR_ROLE then
+		return self.data.color or LocalPlayer():GetRoleColor()
+	elseif self.data.visibleFor == VISIBLE_FOR_TEAM then
+		return self.data.color or TEAMS[LocalPlayer():GetTeam()].color
```
saibotk, line 33: I think the `self.data.color` case can be moved to some early return here :)

saibotk (changes requested):

lua/ttt2/libraries/marker_vision.lua
```diff
@@ -21,176 +22,198 @@ markerVision = {}
...
+				-- nothing needs to be updated here, the new receiver list is generated
+				-- in SyncToClients
+				mvObject:SyncToClients()
 
-			markerVision.UpdateEntityOwnerTeam(ent, oldTeam, newTeam)
+				continue
+			end
+
+			-- case 2: it is bound to the role and the player's new/old role matches
+			if mvObjectData.owner == oldRole or mvObjectData.owner == newRole then
```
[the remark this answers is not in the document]
TimGoll: the owner can be an entity, a role or a team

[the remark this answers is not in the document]
saibotk: Not according to the docs on the owner function, please correct those and add a comment there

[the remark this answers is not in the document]
TimGoll: ah, that was in UpdateRelations what I removed. Will fix this

TimGoll:
done again @saibotk

TimGoll:
Just tested again to verify that everything is still working

saibotk (approved):

Histalek (approved):

PR #1338, opened 2024-01-31
Over Head Icon improvements

TimGoll:
This PR brings multiple small and minor improvements to the over head icons:

- they track the head bone positions: This means they move with the player animation, move down when crouched and also change position when the model scale is changed (i.e. when the minifier is used)
- it now uses a proper Cam3D2D environment: This fixed some minor issues with the icon drop shadow
- rendering order is based on distance: Previously the icons were drawn in the player order which could cause icons to be rendered in front of icons that they shouldn't be; it is now sorted by distance
- when observing a player in their eyes, their icon is now hidden, it is still shown when in third person
- changed the angle the icon is drawn: the rotation is still the same, but the pitch is locked on 90 meaning that the arrow is always pointing at the head, even when looking from the top

image
Only a trained eye can spot a visual difference here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -244,6 +244,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Moved reset buttons onto the left (by @a7f3)
 - Added ammo icons to the weapon switch HUD and player status HUD elements (by @EntranceJew)
 - Changed the disguiser icon to be more fitting (by @TimGoll)
+- Changed the way the role overhead icon is rendered (by @TimGoll)
+  - It now tracks the players head position
+  - Rendering order is based on distance, no more weird visual glitches
+  - Hidden when oberserving a player in their eyes
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -247,7 +247,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Changed the way the role overhead icon is rendered (by @TimGoll)
   - It now tracks the players head position
   - Rendering order is based on distance, no more weird visual glitches
-  - Hidden when oberserving a player in their eyes
+  - Hidden when observing a player in first person view
 
 ### Fixed
 
--- a/gamemodes/terrortown/gamemode/client/cl_targetid.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_targetid.lua
@@ -6,7 +6,6 @@ local util = util
 local render = render
 local surface = surface
 local draw = draw
-local GetPlayers = player.GetAll
 local math = math
 local IsValid = IsValid
 local hook = hook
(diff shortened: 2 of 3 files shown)


gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -53,54 +52,37 @@ local materialBaseOverlay = Material("vgui/ttt/dynamic/sprite_base_overlay")
...
+-- @param Material iconRole The role icon that will be drawn
+-- @param Color colorRole The role color for the background
 -- @realm client
-function DrawOverheadRoleIcon(ply, ricon, rcolor)
-	local client = LocalPlayer()
-	if ply == client then
-		return
-	end
+function DrawOverheadRoleIcon(client, ply, iconRole, colorRole)
+	local pos = ply:GetBoneMatrix(ply:LookupBone("ValveBiped.Bip01_Head1")):GetTranslation()
```
saibotk, line 61: What if a model does not have this? Can we just add a fallback position then?

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -53,54 +52,37 @@ local materialBaseOverlay = Material("vgui/ttt/dynamic/sprite_base_overlay")
...
-	-- draw icon
-	render.SetMaterial(ricon)
-	render.DrawQuadEasy(Vector(pos.x, pos.y, pos.z + 0.5), dir, 8, 8, util.GetDefaultColor(rcolor), 180)
+local function DistanceSorter(a, b)
+	local client = LocalPlayer()
 
-	-- stop linear filter
-	render.PopFilterMag()
-	render.PopFilterMin()
+	return client:GetPos():Distance(a.ply:GetPos()) > client:GetPos():Distance(b.ply:GetPos())
```
saibotk, line 85: Cache client:GetPos instead of client here

saibotk (changes requested):
Just minor nits

gamemodes/terrortown/gamemode/client/cl_targetid.lua
TimGoll: good idea, I will add the old code as a fallback here

CHANGELOG.md
```diff
@@ -244,6 +244,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Moved reset buttons onto the left (by @a7f3)
 - Added ammo icons to the weapon switch HUD and player status HUD elements (by @EntranceJew)
 - Changed the disguiser icon to be more fitting (by @TimGoll)
+- Changed the way the role overhead icon is rendered (by @TimGoll)
+  - It now tracks the players head position
+  - Rendering order is based on distance, no more weird visual glitches
+  - Hidden when oberserving a player in their eyes
```
Histalek, line 250:
```suggestion
  - Hidden when observing a player in first person view
```

Histalek (approved):

TimGoll:
Done @saibotk! While adding the fallback, I moved the code to the player class so that we now have a function that tells the exact height of the player. It is not perfect because I assumed a head size, but it is better then it was before:

```lua
---
-- Returns the player height vector based on the curren thead bone
-- position. X and Y move along the head bone, Z contains the actual
-- height.
-- @note Uses the player's bounding box as a fallback if the head
-- bone is not defined
-- @note Respects the model scale for the height calculation
-- @return vector The player height
-- @realm shared
function plymeta:GetHeightVector()
```

I also noticed an issue in the draw check code that I fixed: 
image

saibotk (changes requested):

TimGoll:
In the last commit I added support for the head bone scale used in Alf's minigames:

image

saibotk (approved):

Issue #1207, opened 2023-12-17
Console error on Clients


mexikoedi:
## Your version of TTT2 (mandatory)

- Workshop version from Steam

## Describe the bug (mandatory)

Some Clients get an error in the console which appears to happen random? 
I cannot pinpoint it but it only happens on some Clients.
For example I didn't get this error.
This error was not present in earlier TTT2 versions. (prior 0.12.0)

## To reproduce

Steps to reproduce the behaviour:

1. Join a game
2. Play around
3. An error should appear in the console after some time/some action

## Expected behaviour

This error shouldn't appear.

## Context (please provide as much as you can)

- Collection link: https://steamcommunity.com/sharedfiles/filedetails/?id=1822673195
- Client error: 
[TTT2 (Base) - v0.12.1b] gamemodes/terrortown/entities/weapons/weapon_tttbase.lua:447: attempt to index local 'line' (a nil value)
  1. DrawHelp - gamemodes/terrortown/entities/weapons/weapon_tttbase.lua:447
   2. unknown - gamemodes/terrortown/entities/weapons/weapon_tttbase.lua:300

TimGoll:
I don't really see how this can break. The error is happening in this line:
![image](https://github.com/TTT-2/TTT2/assets/13639408/aafaea52-1317-405e-b03d-53d2a79b9545)

However I can't see any scenario where line is a nil value:
![image](https://github.com/TTT-2/TTT2/assets/13639408/a244a811-9873-41b4-a06e-707d981ce2a8)

mexikoedi:
I also don't know why. I hadn't this issue, only some other players.
This appeared randomly but I think most of the time right after round start. 
And it's a new issue, this only started to appear recently after 0.12.0.
But I couldn't pinpoint it down. They also redownloaded TTT2 from the Steam Workshop but it didn't help them.

sbzlzh:
> I also don't know why. I hadn't this issue, only some other players. This appeared randomly but I think most of the time right after round start. And it's a new issue, this only started to appear recently after 0.12.0. But I couldn't pinpoint it down. They also redownloaded TTT2 from the Steam Workshop but it didn't help them.

I think it might be a plugin conflict.

sbzlzh:
I don't seem to have that problem either.

mexikoedi:
More information for this issue:
This error seems to appear if you switch to a specific weapon.
For example from holstered to a specific weapon.
But only while the round is active and even then not always.
[...]

PR #1340, opened 2024-02-02
Fix Weapon Help Width

### LINKED ISSUE #1339: HUD Help Size is Bugged
image

The size is bugged in HUD help, not sure why, but it happens for multiple weapons (at least crowbar and Kiss). Maybe it has something to do with the localization

TimGoll:
fixes #1339


mexikoedi:
Does this also fix this issue: https://github.com/TTT-2/TTT2/issues/1207#issue-2045418359

AddHUDHelp function seemed to not work correctly and was the cause for the error.
I have updated my addons (Kiss, ...) to AddTTT2HUDHelp and it seemed to work correctly.

If yes then the issue can be also closed.

TimGoll:
no, this only fixes the rendering

saibotk (approved):

Histalek (approved):

PR #1348, opened 2024-02-03
ci(lang-parser): Use app token instead of GITHUB_TOKEN

Histalek:
This is needed as creating a pr/pushing a commit using the GITHUB_TOKEN will not trigger other workflow runs. Which is done to prevent circular workflow runs.

Using a token generated from a github app does not have this limitation.

### DIFF
--- a/.github/workflows/lang-parser.yml
+++ b/.github/workflows/lang-parser.yml
@@ -12,6 +12,9 @@ on:
 
 jobs:
   update-language-files:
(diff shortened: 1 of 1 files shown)


TimGoll (approved):
sounds good!

Histalek:
Oh well, the implementation is not that sturdy if it runs on any other branch than master.

Meaning this change was smuggled in via #1346 

Whoops. closing this as it is obsolete now

PR #1351, opened 2024-02-03
MSTACK: Fix missing sound for notification cue

EntranceJew:
[that noise was missing](https://github.com/Facepunch/garrysmod/pull/1983) and i'm tired of seeing developer prints about noises that don't exist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed font initialization to not trip engine font fallback behavior (by @EntranceJew)
 - Fixed the decoy producing a wrong colored icon for other teams (by @NickCloudAT)
 - Fixed door health displaying as a humongous string of decimals
+- Fixed the MSTACK noise not playing when it should (by @figardo in base TTT)
 
 ### Removed
 
--- a/gamemodes/terrortown/gamemode/client/cl_msgstack.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_msgstack.lua
@@ -12,12 +12,12 @@ local table = table
 local net = net
 
 sound.Add({
-	name = "Hud.Hint",
-	channel = CHAN_STATIC,
-	volume = 1.0,
-	level = SNDLVL_NONE,
-	pitch = 100,
-	sound = "ui/hint.wav"
+    name = "Hud.Hint",
+    channel = CHAN_STATIC,
+    volume = 1.0,
+    level = SNDLVL_NONE,
+    pitch = 100,
+    sound = "ui/hint.wav",
 })
 
 local traitor_msg_bg = Color(255, 0, 0, 255)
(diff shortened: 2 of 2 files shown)


TimGoll:
huh, interesting. Thanks for bringing this over. I'm wondering if we should also add a convar to disable this, but I guess we probably shouldn't as long as nobody is complaining.

I thought for a moment if we should also rename this sound like they did in TTT, but I don't think that it is likely to clash with anything really

TimGoll (approved):
but you should probably mention that it is taken from baseTTT, similar to how we did it in the past: https://github.com/TTT-2/TTT2/pull/1156/files

EntranceJew:
it is actually incredibly annoying, but this should motivate people to use less MSTACK prompts, or for us to innovate a recyclable MSTACK id to refresh it like a Toast Notification

TimGoll:
> it is actually incredibly annoying

My fears as well, but this sentence made me chuckle

saibotk:
Actually if it is that annoying, we should really consider either removing this sound cue entirely or we might also add the config convar. Even though i dislike convars, i feel this is nothing many people will enjoy.

EntranceJew:
sounds complicated

TimGoll:
agreed on the sound being annoying. Maybe a different sound would be nice. Maybe the sound used in https://wiki.facepunch.com/gmod/chat.PlaySound

EntranceJew:
> agreed on the sound being annoying. Just tested it. Maybe a different sound would be nice. Maybe the sound used in https://wiki.facepunch.com/gmod/chat.PlaySound

wouldn't it be confusing for the player to potentially hear `HudChat.Message` while both chatting and getting MSTACK popups?

TimGoll:
> > agreed on the sound being annoying. Just tested it. Maybe a different sound would be nice. Maybe the sound used in https://wiki.facepunch.com/gmod/chat.PlaySound
> 
> wouldn't it be confusing for the player to potentially hear `HudChat.Message` while both chatting and getting MSTACK popups?

Probably right, yes. Then let's not do it then.

I've thought about it for a while and do we really want it then? Wouldn't it just be better to not add this annoying feature and remove the bug by removing the sound play that causes an error in the console?

EntranceJew:
if this is blocking something it's gonna need some direction

TimGoll:
nah, it's not blocking. I jsut think we're not sure how to proceed with this. While we all agree it is good to keep feature parity to TTT, it is also kinda annoying

EntranceJew:
attaching here so it doesn't get lost, i did an alternate "heavier" version of the "talk" sfx that is referenced earlier to further distinguish them from each-other
[talk2.zip](https://github.com/TTT-2/TTT2/files/14232219/talk2.zip)

saibotk:
I'd still favor it, if we could just remove this altogether.

ZenBre4ker:
If it's about a vote, I also would vote against the sound and remove the bug.

Histalek:
given that this hasn't existed/worked in the last 12 years or so i would also vote to remove this completely

PR #1353, opened 2024-02-04
Radar in Meters

TimGoll:
The builtin radar now shows meters instead of hammer units

### DIFF
--- a/lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
+++ b/lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
@@ -170,7 +170,7 @@ function RADAR:Draw(client)
 		end
 	end
 
-	-- vorpse calls
+	-- corpse calls
 	if not table.IsEmpty(self.called_corpses) then
 		surface.SetTexture(det_beacon)
 		surface.SetTextColor(255, 255, 255, 240)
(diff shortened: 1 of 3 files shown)


lua/terrortown/entities/items/item_ttt_radar/cl_init.lua
```diff
@@ -168,7 +170,7 @@ function RADAR:Draw(client)
 		end
 	end
 
-	-- Corpse calls
+	-- vorpse calls
```
saibotk, line 173:
```suggestion
	-- corpse calls
```

lua/terrortown/lang/en.lua
```diff
@@ -2174,3 +2174,6 @@ L.header_weapon_settings = "Weapon Settings"
 
 -- 2024-01-24
 L.grenade_fuse = "FUSE"
+
+-- 2024-02-04
+L.distance_meter = "{distance}m"
```
saibotk, line 2179: I think "m" is a pretty universal standard globally and does not need to be translated

saibotk (changes requested):

lua/terrortown/lang/en.lua
TimGoll: I'm not sure it is m in chinese etc

saibotk: It is the official SI unit symbol there is nothing to translate

saibotk (approved):

EntranceJew:
is this consistent with the DNA scanner's display?

TimGoll:
> is this consistent with the DNA scanner's display?

good point, no. Will look at this in another PR

Histalek (approved):

PR #1355, opened 2024-02-04
DNA Scanner Now Uses Meters

TimGoll:
Similar to #1353 now the DNA scnanner uses meters:

image


saibotk (approved):

Histalek (approved):

Spanospy:
A side effect of this change is that the distance readout is much less precise now. Personally I think it makes it more interesting, but perhaps there could be an option to bring back the old readout or use decimal places?

Histalek:
Is this kind of precision really needed though?

Even if it is we would probably need to remove the unit (m) to make space for at least 1 decimal, which would then roughly take up the same amount of space.
But even then this would "only" give a fifth of the precision we had with hammer units. Without any use cases i can't really reason if this would be enough.

PR #1364, opened 2024-02-04
TargetID: Disable seeing own spectator nametag

EntranceJew:
image
vs
image

does not affect propspec nametag because that helps find the prop's origin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
   - It now tracks the players head position
   - Rendering order is based on distance, no more weird visual glitches
   - Hidden when observing a player in first person view
+- Your own spectator nametag will not display when looking directly up in post-round (by @EntranceJew)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


saibotk (approved):

TimGoll:
This is good, thanks. I think this issue also exists if you're dead, spectate a player and that player looks in the sky. This at least existed for overhead icons. There I solved it this way:

https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/client/cl_targetid.lua#L176-L177

Maybe you could confirm it / fix it for that as well?

EntranceJew:
image
yeah it works with that snippet included

TimGoll:
Not a cgabge request, more a question: do you think it is a good idea to show the name when in third person observer mode?

If you think that it isn't, then remove the part with the IN_EYE, otherwise give me your go and I'll approve this

Thank you!

EntranceJew:
> Not a change request, more a question: do you think it is a good idea to show the name when in third person observer mode?
> 
> If you think that it isn't, then remove the part with the IN_EYE, otherwise give me your go and I'll approve this
> 
> Thank you!

yeah, i forgot third person orbit existed, i definitely prefer it not being visible if advanced spectator already has the nameplate

TimGoll:
Just FYI, the nameplate is part of base TTT2, only the additional info comes from AS

TimGoll (approved):

PR #1371, opened 2024-02-05
fix font goofs

EntranceJew:
based on [rubat's input here](https://discord.com/channels/565105920414318602/565108080300261398/1203788317846671390) which divulges the inner mechanisms of source's engine, and the [following discussion](https://discord.com/channels/442107660955942932/547119387392475136/1203809434338468000) -- this specifies to utilize "Tahoma" instead of various forms of Trebuchet which cannot be modified, thus fallback on every client version.

the fix for this, as used here, maintains the game's appearance across all platforms as-is, just prevents the errors from occurring


TimGoll:
@EntranceJew If you don't complain I will take this over and make this ready to merge today in the evening (maybe 6 hours from now)

TimGoll:
I don't know what happened when merging ther master here, but somehow your changes about the updated icon location were added again?

PR #1372, opened 2024-02-05
relocate content to the content folder

EntranceJew:
things in a gamemode specific subfolder aren't loaded during other gamemodes preventing less resources from being used


saibotk (approved):
Nvm didn't see that it had files in both the models folder and the weapons folder for the viewmodel.

TimGoll (approved):

PR #1374, opened 2024-02-05
Entity Base: Added Pickup Handling

TimGoll:
I added pickup handling to the recently added entity base.

To add this to an entity, the following two parameters have to be set:
```lua
ENT.CanUseKey = true -- to enable the ENT:UseOverwrite hook
ENT.pickupWeaponClass = "weapon_ttt_radio" -- to define which weapon should be added to the inventory
```

There are two hooks that are called during the pickup process:
```lua
---
-- Run if a valid player tries to pick up this entity to check if this pickup is accepted.
-- @param Player activator The player that used their use key
-- @return[default=true] boolean Return true to allow pickup
-- @hook
-- @realm server
function ENT:PlayerCanPickupWeapon(activator)
```

```lua
---
-- Called when this entity is picked up and about to be removed.
-- @param Player activator The player that used their use key
-- @param Weapon wep The weapon that is added to their inventory
-- @hook
-- @realm server
function ENT:OnPickup(activator, wep)
```

It can handle the following scenarios:
- weapons that are "stackable" like the beacon (it adds to the weapon's clip)
- weapons that are not stackable
- weapons that are not dropable and therefore might prevent the pickup of the placed entity

Also it adds sounds and notifications that improve the player feedback, while also making sure that all entities behave the same

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
   - Added `AddCustomWorldModel` to add custom world models
   - Added an automatic fix for badly coded addons that break the view model fingers
 - Added `ttt_base_placeable` entity that is used to handle any placeable / destroyable entity (by @TimGoll)
-  - moved `ttt_c4`, `ttt_health_station`, `ttt_beacon`, `ttt_decoy`, `ttt0_radio` and `ttt_cse_proj` to that base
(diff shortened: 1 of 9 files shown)


saibotk (approved):

EntranceJew:
i pulled the branch and tested all the affected weapons, i'd say: good work, only oversight is this
image

TimGoll:
@EntranceJew Are you talking about the translation being missing?

TimGoll:
Huh, I already did this in another PR, I guess merging the branches broke this

EntranceJew (approved):

Histalek (approved):

PR #1378, opened 2024-02-06
Fonts: Fix wrong setup

TimGoll:
Based on @EntranceJew's work I created this new PR that fixes the wrong font initialization. Due to issues with his branch that I was unable to resolve, I created this new branch and cherry picked his commit.

While fixing the Font in the DNA scanner I noticed that it was initialized in its init function meaning that every spawn of a DNA scanner ent would create a new font. Therefore I moved that part of the code.


EntranceJew:
`antialias = true` doesn't actually need to be set, i think it's enabled by default and it was a mistake I overlooked, but we should provide `extended = true`  and test that it functions on Cyrillic letters as most system fonts or future replacement fonts we might ctrl+f replace should attempt to fallback gracefully

TimGoll:
> `antialias = true` doesn't actually need to be set, i think it's enabled by default and it was a mistake I overlooked, but we should provide `extended = true`  and test that it functions on Cyrillic letters as most system fonts or future replacement fonts we might ctrl+f replace should attempt to fallback gracefully
> 
> i'm not sure if `symbol = true` is worth providing, we'd have to check the code rubat linked to really understand what charsets it encompasses or how it behaves

That's true. I'm not available today, but I keep that in mind

CHANGELOG.md
```diff
@@ -100,6 +100,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed draw.Arc when `gmod_mcore_test` is set to 1 (by @WardenPotato)
 - Fixed weapon help box width for wide bindings with short descriptions (by @TimGoll)
 - Fixed `GM:TTTBodySearchPopulate` using the wrong data variable (by @TimGoll)
+- Fixed the way fonts are initialized in TTT2 meaning that it is no longer broken and also works on linux (by @EntranceJew)
```
saibotk, line 103: This is misleading, remove the linux part. Because Tahoma is not a font you will find on Linux (needs to be extracted from windows isos or has to be downloaded from some other 3rd party projects).

Maybe add a note to linux users that they should install Tahoma for correct fonts

saibotk (changes requested):
This is a good first fix.
But for the future we may want to look into a vendored font, so that there are no differences when someone does not have the font installed.

EntranceJew:
> This is a good first fix. But for the future we may want to look into a vendored font, so that there are no differences when someone does not have the font installed.

Google Fonts' `Open Sans` is the go-to Tahoma replacement, and gmod already ships Roboto, so after testing a lot of other options, between that and Google Fonts' `Ubuntu`, it'd be my primary choice.

The only reason I haven't done it yet is that font weights and sizes will need to be tweaked per-specimen to match their display height/width/weight. Specifically, with Google Font specimens we cannot request a font weight of `1200` when they only hit regular denominations of `300/500/700`/etc and heavier versions might carry a different internal name. I also don't know if gmod correctly utilizes variable TTF fonts as the most popular fonts are served as that, with a bunch of individual family files for all the variants.

CHANGELOG.md
TimGoll: could you please modify it so that it represents your situation on linux because what I wrote is what I assumed this did on linux. And I don't have a linux installation on hand to test it. Thanks!

TimGoll: Done I think, if you 2ant something else, please just change it yourself

saibotk (approved):

ZenBre4ker (approved):

EntranceJew (approved):
looks good, anything else i can do with a font if we agree on a replacement. but resizing all the things to replace is going to be a chore if we change our minds on a suitable replacement.

TimGoll:
> looks good, anything else i can do with a font if we agree on a replacement. but resizing all the things to replace is going to be a chore if we change our minds on a suitable replacement.

Once v0.13 is released, this is a good project to continue with. I'd suggest first deciding on a font on discord, then opening a low effort PR that only changes the font without tweaked sizes. Than we can tweak the sizes together before we merge it.

gamemodes/terrortown/entities/weapons/weapon_ttt_wtester.lua
```diff
@@ -497,11 +470,18 @@ if SERVER then
     -- @hook
     -- @realm server
     function GAMEMODE:TTTFoundDNA(finder, toucher, ent) end
-else -- CLIENT
+end
```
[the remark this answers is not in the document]
TimGoll: we didn't do that anywhere else before. But I like that, I think we should start doing this, because these big blocks are sometimes hard to understand. I'll merge this and we should make this a style rule for ourself.

PR #1381, opened 2024-02-07
Fix: Radar decoy fix

NickCloudAT:
This fixes and issue with the radar showing a grey radar symbol instead of the green one for other teams.

I'm not sure, but was this just an oversight or am I just not seeing something else that may cause this issue?
Setting the team to TEAM_INNOCENT too fixes the issue.


saibotk (approved):
Seems alright

TimGoll (approved):

PR #1382, opened 2024-02-08
Scoreboard: Fixed Icons sometimes missing

TimGoll:
This PR fixes the scoreboard being stuck open after an icon is missing. This PR does this in two levels

- adding a safety check to make sure the scoreboard never stays stuck
- fixing the root cause that causes the breakage in the first place

## The problem

Some entities register custom projectiles (like the boomerang) that are then registered as the inflictor. This breaks our current UI as I expected that the weapon is always a real TTT weapon. The change in this PR searches for an entity with the given name class, if no weapon was found.

image
image

I'm pretty sure that it fixes the two linked issues.


lua/ttt2/libraries/bodysearch.lua
```diff
@@ -930,11 +945,23 @@ if CLIENT then
...
+                end
+
+                search[type] = {
+                    img = materialFile,
+                    text = transText,
+                    p = index, -- sorting number
+                }
+
+                index = index + 1
+            end
```
saibotk, line 964: 2 things here:

1. You are now not showing anything when no icon is available, maybe we should just display some kind of fallback? or at least display this one without an icon?
2. This special png handling should not be needed. Because DImage, which is used in SetIcon in SimpleIcon does already turn the string into a Material, which should work automatically with .png. See https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/vgui/dimage.lua#L91

saibotk (changes requested):

lua/ttt2/libraries/bodysearch.lua
TimGoll: 1. But what should I show if the weapon has no icon and also no inflictor? The missing icon material?
2. But it did not work when I did pass it without an .png, so who knows

TimGoll: Done the first thing, not sure what I could do about the second

saibotk: Why is the searchData even providing the material instead of the filename? Seems like we may be able to change that?

TimGoll: because we cache the materials and I don't want to recreate the materials based on their file

lua/ttt2/libraries/bodysearch.lua
```diff
@@ -819,7 +821,23 @@ if CLIENT then
 
     local function TypeToMaterial(type, data)
         if type == "wep" then
-            return util.WeaponForClass(data.wep).iconMaterial
+            -- in most cases the inflictor is a weapon and the weapon has a cached
+            -- material that can be used
+            local wep = util.WeaponForClass(data.wep).iconMaterial
+
+            if wep then
+                return wep.iconMaterial
```
ZenBre4ker, line 829:
```suggestion
            local wep = util.WeaponForClass(data.wep)

            if wep then
                return wep.iconMaterial
```
umm.. Shouldnt that be like this or at least check weapon.iconMaterial?

ZenBre4ker (approved):

lua/ttt2/libraries/bodysearch.lua
TimGoll: you are abolutely right, REALLY good catch. I only tested it on the boomerang of course, oof

lua/ttt2/libraries/bodysearch.lua
saibotk: I'd rather have this fixed correctly instead of only for png images, so I'd suggest adjusting the scoreboard code.
But i would also check this out on my own and see if i can find a solution, if you don't want to.

saibotk: I see, can you then just add cases for jpeg and jpg, i think these are currently the only 2 other types that are allowed in the addon whitelist.

TimGoll: As a disclaimer: I don't know if any addon is actually using this information, but it could be

TimGoll: I considered jpeg but couldn't think of a single case I've seen jpegs used for icons. But I can add this

TimGoll: Also note, this whole function here is removed as soon as we have a new scoreboard. This only converts the new bodysearch data to the legacy table

EntranceJew:
image
image
image
minor overcorrection

TimGoll:
The latent commit based on a suggestion by @ZenBre4ker should solve this. Are you sure you are on that commit?

EntranceJew:
image
what's on the branch is on the branch

TimGoll:
I have to test this tomorrow again, but looking at the code I don't see how this could happen

ZenBre4ker:
But its not?

EntranceJew:
for some reason the pull i performed an hour ago to be made aware of the branch did not include a commit performed over 3 hours ago

image
image
image
image
it isn't getting stuck but the boomerang isn't displaying anymore either

TimGoll:
Oh well, I should test stuff after merging suggestions. No idea why it is broken, but I will take a look. Thanks for testing!

TimGoll:
okay, I fixed and tested it. It works now in both cases.

The problem here was that `util.WeaponForClass(data.wep)` already does what I want and finds the correct projectile entity, therefore my check failed. The solution therefore is to check if a cached material exists.

Maybe it would be an idea to add material caching to entities as well in the future, I'm not so sure though. Nevertheless, this is a discussion for the future, this fallback for non-cached icons is useful, even if we decide to add caching to entities as this is a good way to prevent issues for edge cases

lua/ttt2/libraries/bodysearch.lua
saibotk: You've missed ".jpeg"

EntranceJew (approved):

EntranceJew:
tested, couldn't find a situation that broke the scoreboard or any equipment that didn't have an attribution, LGTM

saibotk (approved):

PR #1393, opened 2024-02-08
Playersettings: Fixed unwanted shared rederence

TimGoll:
This PR fixes an issue with the synced player settings. Initializing the synced player settings as `plymeta.playerSettings = plymeta.playerSettings or {}` results in one table reference being created that is shared between all instances of the `Player` class. This pullrequest changes this so that each player has their own instance of said class.


saibotk (approved):
Welp, we need to keep such pointer like stuff more in mind for future reviews :D

TimGoll:
Yes! But I wasn't aware that it was working like that. I assumed the creation of a Player instance copied all its contents. Aparantly not

PR #1395, opened 2024-02-09
C4 UI: Fixed C4 UI accessing wrong BaseClass

### LINKED ISSUE #1387: C4 Disarm UI is not functional

- Latest GitHub version

## Video

image
image

C4 UI locks up and can't be interacted with

## To reproduce

1. Attempt to disarm C4

## Expected behaviour

- Disarming C4 works

Pretty sure this got broken during an entity bringdown

## Additional context

the `arm c4` dialog works just fine

EntranceJew:
image
it works again [this evil line broke things](https://github.com/TTT-2/TTT2/commit/a52f830afb6b2912d2568c344c923ce0bcfb660d#diff-6f8517483f2d960a76d2a50181ebdda8ae10df6174cdf1713e75c5d5f9636102L189)

### DIFF
--- a/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua
@@ -236,7 +236,7 @@ end
 ---
 -- @realm client
 function WIREPANEL:OnCursorExited()
-    self.PaintOver = BaseClass.PaintOver
+    self.PaintOver = self.BaseClass.PaintOver
 end
 
 ---
@@ -248,7 +248,7 @@ function WIREPANEL:DoClick()
 
     self.IsCut = true
 
-    self.PaintOver = BaseClass.PaintOver
+    self.PaintOver = self.BaseClass.PaintOver
 
     self.m_Image:SetMaterial(c4_wirecut_mat)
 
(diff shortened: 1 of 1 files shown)


EntranceJew:
and if nobody likes it then you'll have to move the derma panels somewhere else

TimGoll:
Ok this is on me, I did not expect there to be code that isn't `ENT` releated. Should have checked more thoroughly. Just a note: Changing the `BASECLASS` call at the top of the file should resolve this as well. I'll remove this call as this is not needed in this file and then this is good to go imho

TimGoll (commented):
removed DEFINE_BASECLASS

LOL: I tried to commit via the GitHub web editor. Aparantly I didn't commit anything :D

TimGoll (approved):

EntranceJew:
well, we can keep the DEFINE

TimGoll:
> well, we can keep the DEFINE

we can, but it is not used in that file. Hence I removed it

EntranceJew:
> > well, we can keep the DEFINE
> 
> we can, but it is not used in that file. Hence I removed it

oh, yeah, it should have been available in shared anyway, you're right

TimGoll:
> > > well, we can keep the DEFINE
> > 
> > 
> > we can, but it is not used in that file. Hence I removed it
> 
> oh, yeah, it should have been available in shared anyway, you're right

But that was a good fix by you. Thank you. I'd probably have spend like an hour probing around wondering why it isn't working!

Histalek (approved):

PR #1404, opened 2024-02-10
Dynamic FOV: Rewrite

TimGoll:
Due to problems with my previous implementation on low-tickrate servers, I had to redo the whole system. It now uses the `CalcView` hook as well. It should work the same (and better) than before.

Things that I could not test: How does this behave when spectating a player that scopes etc. I think it is fine, but maybe it isn't. I need some feedback on this, best from @EntranceJew.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
@@ -17,6 +17,14 @@ local cvEnableBobbing = CreateConVar("ttt2_enable_bobbing", "1", FCVAR_ARCHIVE)
 -- stylua: ignore
 local cvEnableBobbingStrafe = CreateConVar("ttt2_enable_bobbing_strafe", "1", FCVAR_ARCHIVE)
 
+-- @realm client
+-- stylua: ignore
+local cvEnableDynamicFOV = CreateConVar("ttt2_enable_dynamic_fov", "1", {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+
+cvars.AddChangeCallback("ttt2_enable_dynamic_fov", function(_, _, valueNew)
+    LocalPlayer():SetSettingOnServer("enable_dynamic_fov", tobool(valueNew))
+end)
+
 ---
 -- Applies a animation gesture
 -- @param ACT act The @{ACT} or sequence that should be played
@@ -398,6 +406,8 @@ local position = 0
 local frameCount = 10
 
 local lastStrafeValue = 0
+local lastFOVValue = 85
+local multiplier = 40
 
 -- heavily inspired from V92's "Head Bobbing": https://steamcommunity.com/sharedfiles/filedetails/?id=572928034
 hook.Add("CalcView", "TTT2ViewBobbingHook", function(ply, origin, angles, fov)
@@ -412,18 +422,51 @@ hook.Add("CalcView", "TTT2ViewBobbingHook", function(ply, origin, angles, fov)
         return
     end
 
-    if (not ply:IsOnGround() and ply:WaterLevel() == 0) or ply:InVehicle() then
-        airtime = math.Clamp(airtime + 1, 0, 300)
-
-        return
-    end
-
     local view = {
         ply = ply,
         origin = origin,
         angles = angles,
         fov = fov,
+        drawviewer = false,
     }
+
+    local dynFOV = fov
+    local mul = ply:GetSpeedMultiplier() * SPRINT:HandleSpeedMultiplierCalculation(ply)
+    local desiredFOV = dynFOV * mul ^ (1 / 6)
+
+    if ply.triggeredFOV then
+        ply.triggeredFOV = false
+
+        multiplier = (desiredFOV - lastFOVValue) / ply.timeFOV
+    end
+
+    if not ply.fixedFOV then
+        if desiredFOV > lastFOVValue then
+            dynFOV = math.min(desiredFOV, lastFOVValue + FrameTime() * multiplier)
+        elseif desiredFOV < lastFOVValue then
+            dynFOV = math.max(desiredFOV, lastFOVValue - FrameTime() * multiplier)
+        else
+            dynFOV = lastFOVValue
+
+            multiplier = 40
+        end
+
+        lastFOVValue = dynFOV
+    else
+        -- cache the current FOV value while in fixed FOV mode to use this value
+        -- to transition out of this mode
+        lastFOVValue = fov
+    end
+
+    if cvEnableDynamicFOV:GetBool() then
+        view.fov = dynFOV
+    end
+
+    if (not ply:IsOnGround() and ply:WaterLevel() == 0) or ply:InVehicle() then
+        airtime = math.Clamp(airtime + 1, 0, 300)
+
+        return view
+    end
 
     local eyeAngles = ply:EyeAngles()
     local strafeValue = 0
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -286,6 +286,7 @@ util.AddNetworkString("TTT2OrderEquipment")
 util.AddNetworkString("TTT2RoleGlobalVoice")
 util.AddNetworkString("TTT2MuteTeam")
 util.AddNetworkString("TTT2UpdateHoldAimConvar")
+util.AddNetworkString("TTT2UpdateFOV")
 
 -- provide menu files by loading them from here:
 fileloader.LoadFolder("terrortown/menus/score/", false, CLIENT_FILE)
@@ -352,7 +353,6 @@ function GM:Initialize()
 
     -- register synced player variables
     player.RegisterSettingOnServer("enable_dynamic_fov", "bool")
-    player.RegisterSettingOnServer("fov_desired", "float")
 
     -- Force friendly fire to be enabled. If it is off, we do not get lag compensation.
     RunConsoleCommand("mp_friendlyfire", "1")
@@ -1442,7 +1442,6 @@ function GM:OnReloaded()
 
     -- register synced player variables
     player.RegisterSettingOnServer("enable_dynamic_fov", "bool")
-    player.RegisterSettingOnServer("fov_desired", "float")
 
     ---
     -- @realm shared
--- a/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
@@ -1261,28 +1261,75 @@ end
 -- @param number fov The angle of perception (FOV); set to 0 to return to default user FOV
 -- @param[default=0] number time The time it takes to transition to the FOV expressed in a floating point
 -- @param[default=self] Entity requester The requester or "owner" of the zoom event; only this entity will be able to change the player's FOV until it is set back to 0
--- @param[default=false] boolean isSprinting Whether this is called to set sprinting FOV
 -- @realm shared
-function plymeta:SetFOV(fov, time, requester, isSprinting)
-    if isSprinting then
-        self.sprintingFOV = fov
+function plymeta:SetFOV(fov, time, requester)
+    -- if dynamic FOV is disabled, the default function should be used
+    if not self:GetPlayerSetting("enable_dynamic_fov") then
+        self:SetOldFOV(fov, time, requester)
 
-        if self.externalFOV then
+        return
+    end
+
+    -- normal case, SetFOV is used to set a fixed FOV value, we should track
+    -- this so we know that we should no apply custom FOV in GM:CalcView
+    if fov and fov ~= 0 then
+        self.fixedFOV = true
+        self.triggeredFOV = false
+        self.timeFOV = time
+
+        if SERVER then
+            net.Start("TTT2UpdateFOV")
+            net.WriteEntity(self)
+            net.WriteBool(true) -- FOV is fixed
+            net.WriteFloat(time)
+            net.Broadcast()
+        end
+
+    -- when we zoom out, this has to trigger our custom zoom out that handles different
+    -- desired FOV values instead of 0, we therefore set it without transition to handle
+    -- the transistion ourself
+    else
+        -- the SetFOV function is called multiple times, this check makes sure
+        -- we only trigger on the first call after a zoom out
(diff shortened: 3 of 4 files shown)


saibotk:
Can you explain what parts you exactly had to redo and what is now handled better?

TimGoll:
> Can you explain what parts you exactly had to redo and what is now handled better?

The whole dynamic FOV stuff is redone. It is now fully clientside (with some additional syncing to support weapons that call SetFOV only on the server)

TimGoll:
I fixed the issue Entrance was heaving with it (some weapons only call SetFOV on the server instead of calling it shared which is done by most weapons). I reworked the networking a bit to handle that as well.

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -352,7 +353,6 @@ function GM:Initialize()
 
     -- register synced player variables
     player.RegisterSettingOnServer("enable_dynamic_fov", "bool")
-    player.RegisterSettingOnServer("fov_desired", "float")
```
saibotk, line 355: I think you migth also need to remove the F1 menu slider for this one?

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -1261,28 +1261,80 @@ end
...
 
-    self:SetOldFOV(fov, time, requester, isSprinting)
+    self:SetOldFOV(fov, time, requester)
+end
+
+if CLIENT then
+    -- we have to update this via netmessage as well, because some addons only call
+    -- the SetFOV on the server, so we have to manually sync it
+    net.Receive("TTT2UpdateFOV", function()
+        local ply = net.ReadEntity()
```
saibotk, line 1321: I think we can remove this, as each client only needs to set this for themselves, right?

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -398,6 +406,8 @@ local position = 0
 local frameCount = 10
 
 local lastStrafeValue = 0
+local lastFOVValue = 85
+local multiplier = 40
 
 -- heavily inspired from V92's "Head Bobbing": https://steamcommunity.com/sharedfiles/filedetails/?id=572928034
 hook.Add("CalcView", "TTT2ViewBobbingHook", function(ply, origin, angles, fov)
```
saibotk, line 413: this is now more than just view bobbing

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -1261,28 +1261,80 @@ end
...
+            self.triggeredFOV = true
+            self.timeFOV = time
+
+            if SERVER then
+                net.Start("TTT2UpdateFOV")
+                net.WriteEntity(self)
+                net.WriteBool(false) -- FOV isn't fixed anymore
+                net.WriteFloat(time)
+                net.WriteFloat(self:GetFOV())
+                net.Broadcast()
```
saibotk, line 1306: This doesn't seem like it needs a broadcast

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -398,6 +406,8 @@ local position = 0
 local frameCount = 10
 
 local lastStrafeValue = 0
+local lastFOVValue = 85
+local multiplier = 40
```
saibotk, line 410: Rename this, so it is clear that it belongs to FOV stuff

saibotk (changes requested):
First round of feedback

For further feedback I would need you to explain the use of the fixed and plyTriggered variables.

What is their purpose? When are they true / false and why? What problem do they solve?

Also, currently I think there exist edge cases in your if else logic which we might want to check for. E.g. paths when fixedFOV is not true and fov is 0 (in the SetFOV method) and maybe elsewhere too. But I did not look into those that deeply, because I need to understand the variables first.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
TimGoll: I intended to support it as a spec as well, but I had no means to test it and it appears to be doing nothing right now. So I can remove it for now and readd it in a PR when I properly test it for specs

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: true

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
TimGoll: same as above, this was to support specs

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: ok

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
TimGoll: We have to test it today in our group session, but removing this might cause issues when spectating. So I'd like to keep it as a broadcast

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
saibotk: Rename the hook

TimGoll:
> For further feedback I would need you to explain the use of the fixed and plyTriggered variables

Let me try: In a perfect world every weapon would call `SetFOV` shared. Not just because of my changes, but also because it is intended do be done. E.g. calling it only on the server is pretty jarring on a 16 tick server and really smooth when it is shared on the same server.
As I hook into this system as well, I somehow have to "notify" the hook when a player started transitioning out of the zoom in, because it uses a different FOV change multiplier then the normal sprinting does. It is a bit like a semaphore, but without the multithreading race conditions.

> What is their purpose? When are they true / false and why? What problem do they solve?

I guess you're talking about these?

```lua
ply.fixedFOV = net.ReadBool()
ply.timeFOV = net.ReadFloat()
ply.lastFOV = net.ReadFloat()
```

1. Defines whether the FOV is fixed or not - meaning: If `setFOV` is used, the FOV is fixed and my dynamic FOV should not be applied on top of it
2. This is the time for the transition
3. This is the last FOV value, as a starting point for the transition

> E.g. paths when fixedFOV is not true and fov is 0

I don't see the edge case here. If it is set to 0, it is never fixed by design of `SetFOV`: `the angle of perception (FOV). Set to 0 to return to default user FOV. ( Which is ranging from 75 to 90, depending on user settings )` [[link]](https://wiki.facepunch.com/gmod/Player:SetFOV)

TimGoll:
I will try to see if I can rework this to work similar as the ironsight with network vars

TimGoll:
Completely overhauled the way dynamic FOV works with network vars. It is now really smooth as before, but without the trigger and weapon sights hacks that I used previously.

Tested on 66 and 16 ticks, with shared and server side SetFOV. All cases are smooth. So I think this works as intended. Would apreciated @EntranceJew checking it out as well.

EntranceJew (approved):
tested it up and down and the latest version doesn't have any snags at 16tick, meaning it'll be super smooth for people on generally bad connections as well

ZenBre4ker (approved):

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -412,19 +423,71 @@ hook.Add("CalcView", "TTT2ViewBobbingHook", function(ply, origin, angles, fov)
...
+        -- if the transition progress has reached 100%, we should enable the sprint
+        -- FOV smoothing algorithm
+        if progressTransition >= 1.0 then
+            if desiredFOV > ply.lastFOVValue then
+                desiredFOV = math.min(desiredFOV, ply.lastFOVValue + FrameTime() * 40)
+            elseif desiredFOV < ply.lastFOVValue then
+                desiredFOV = math.max(desiredFOV, ply.lastFOVValue - FrameTime() * 40)
+            end
+
+            ply.lastFOVValue = desiredFOV
```
saibotk, line 462: Why does this not also use the last fov below from the network var? 

Regardless it is very confusing that we define another variable with the same name / purpose a few lines down.

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -412,19 +423,71 @@ hook.Add("CalcView", "TTT2ViewBobbingHook", function(ply, origin, angles, fov)
...
-    end
-
     local view = {
         ply = ply,
         origin = origin,
         angles = angles,
         fov = fov,
     }
 
+    ply.lastFOVValue = ply.lastFOVValue or fov
```
saibotk, line 433: Isn't this always the current FOV? Because we did not yet apply the next fov change?

saibotk (changes requested):

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -17,6 +17,14 @@ local cvEnableBobbing = CreateConVar("ttt2_enable_bobbing", "1", FCVAR_ARCHIVE)
 -- stylua: ignore
 local cvEnableBobbingStrafe = CreateConVar("ttt2_enable_bobbing_strafe", "1", FCVAR_ARCHIVE)
 
+-- @realm client
+-- stylua: ignore
+local cvEnableDynamicFOV = CreateConVar("ttt2_enable_dynamic_fov", "1", {FCVAR_NOTIFY, FCVAR_ARCHIVE})
+
+cvars.AddChangeCallback("ttt2_enable_dynamic_fov", function(_, _, valueNew)
+    LocalPlayer():SetSettingOnServer("enable_dynamic_fov", tobool(valueNew))
```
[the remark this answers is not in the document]
TimGoll: This is done in the code already:

https://github.com/TTT-2/TTT2/blob/88849ec06924779df67abda37c0b6f5eb03e14f2/gamemodes/terrortown/gamemode/client/cl_main.lua#L322

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -412,19 +423,71 @@ hook.Add("CalcView", "TTT2ViewBobbingHook", function(ply, origin, angles, fov)
...
+    -- fixed mode: when SetZoom is set to something different than 0, this value should be used
+    -- without any custom modification
+    if ply:GetFOVFixedValue() then
+        dynFOV = fov
+
+    -- dynamic mode: transition between different FOV values
+    else
+        local time = math.max(
+            0,
+            (CurTime() - ply:GetFOVTime()) * game.GetTimeScale() * cvHostTimescale:GetFloat()
```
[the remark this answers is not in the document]
TimGoll: This is taken from the ironsights code. GMod has two variables. A convar that can be set by the server owner and an internal API that can be used by addons. The code should respect both.

As this is used for our custom ironsight zoom-out I thought it would be appropriate to support this here as well.

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: one of them is for the aimation on a frame by frame basis (`ply.lastFOVValue`) and the other is on a more marco level, that is used to handle the overall transition

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: Only the first time the hook is called fov is used as the value, after that the lastFOVValue is used.

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -412,19 +423,71 @@ hook.Add("CalcView", "TTT2ViewBobbingHook", function(ply, origin, angles, fov)
...
+
+        local progressTransition = math.min(1.0, time / ply:GetFOVTransitionTime())
+
+        -- if the transition progress has reached 100%, we should enable the sprint
+        -- FOV smoothing algorithm
+        if progressTransition >= 1.0 then
+            if desiredFOV > ply.lastFOVValue then
+                desiredFOV = math.min(desiredFOV, ply.lastFOVValue + FrameTime() * 40)
+            elseif desiredFOV < ply.lastFOVValue then
+                desiredFOV = math.max(desiredFOV, ply.lastFOVValue - FrameTime() * 40)
```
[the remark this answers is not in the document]
saibotk: Sure 👍 its good to know that this was chosen by testing and does not refer to something special

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
saibotk: No i am talking about the general value of this is baically always the previous fov value right? Why bother creating a variable when this is already available via the current fov on the player

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
saibotk: But do they differ in any way? Cant they be merged?

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: you mean I should use `ply:GetFOV()` here? This might work, I'm not sure if this is influenced by GMods own transitioning. Also I'm not setting the previous FOV value in all cases, there are some cases where it is intentionally not changed. So I think the getter function probably won't work here

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: no, they can't be merged. Let me explain:

the current transition progress is calculated by difference in the network vars for lastFOV and nextFOV times the time that passed. The result of that is applied to the current FOV. The result is also set to the players lastFOV value cache.
However this lastFOV value is not set on every frame. Sometimes it is intentionally kept on the last value

saibotk: Hmm then please try to rename the variables accordingly and maybe add a small comment explaining this. Otherwise its very confusing

Issue #1403, opened 2024-02-09
unarmed errors because it's ejected


EntranceJew:
it has an .Icon but it never gets converted to .iconMaterial

because shopeditor tries to eject it

ZenBre4ker:
To know how to fix it, we would need further info, how to reproduce it and what exactly errors.

PR #1408, opened 2024-02-11
fix buy menu error #skip-changelog

EntranceJew:
it's gotta go someway somehow

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_shopeditor.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_shopeditor.lua
@@ -26,7 +26,6 @@ function ShopEditor.GetEquipmentForRoleAll()
 
     local eject = {
         weapon_fists = true,
-        weapon_ttt_unarmed = true,
         bobs_blacklisted = true,
     }
 


TimGoll:
@saibotk What is your opnion on that? I don't care that much. It is the easiest fix to that problem.

saibotk:
Closed in favor of #1421

TimGoll:
> Closed in favor of #1421

I don't think #1421 is fixing the issue that exists here. @ZenBre4ker should comment on that, but I'm pretty sure these are different issues

ZenBre4ker:
Well, it seems like this was the idea to make unarmed being editable, but I also don't know how to reproduce #1403 to confirm, that it fixes anything. This just allows it being shown in the shopeditor, right?

TimGoll:
Yes. But the problem here is, that the material is only cached if it isn't ejected. If the material isn't cached, it throws an error at a later point. I didn't try to reproduce the issue though, but this is the issue that exists here.

ZenBre4ker:
Still seems wrong to not eject it in the equipmentForRoleAll.
If caching is somehow important, it should be done somewhere else entirely.
Weird enough, that those two calls are entangled:
ShopEditor.GetEquipmentForRoleAll() and hook.run("TTT2RegisterWeaponID")

That is something I was on the way to untangle, but that was probably inside one of those PRs that a took a long to get reviewed.

TimGoll:
yes, it is bad that these are entangled! This is why the fix in this PR is more a bandaid fix then a real fix. I just wanted to point out that your PR does not fix the issue at hand.

saibotk:
Oh then nvm i was wrong but will keep this closed as we need a fix that does not just allow this to be editable

EntranceJew:
THE PROBLEM IS THAT EVERY TIME YOU OPEN THE BUY MENU IT ERRORS

EntranceJew:
_ERRORING EVERY TIME YOU OPEN THE MENU SUCKS_

TimGoll:
Please keep the discussion civil and on topic. We are aware of the issue, but we want a fix, not a bandaid. Therefor this PR is not a good solution to us. This is nothing against you, this is only about this project as we told you on discord before you opened this PR.

EntranceJew:
"this should be done better" with no indication of what better means is the opposite of useful, it destroys labor

Issue #1406, opened 2024-02-11
Spawning ttt_random_weapon after round start does not create a weapon


pontMakesThings:
**TTT2 version:**
workshop v0.12.3b

**Description:**
Spawning ttt_random_weapon after the round has started does not spawn a weapon. 

**To reproduce:**
`ent_create ttt_random_weapon`, or spawn ttt_random_weapon via point_template. No weapon will spawn, and TTT2 Base will encounter a lua error.

**Expected behaviour:**
A random weapon spawns.

**Context:**
Spawning in random weapons and ammo after the initial round start is handy for maps with round variants and mid-round mechanics. I'm currently playtesting an in-development map that makes heavy use of this. ttt_random_weapon functions correctly in the original TTT. I have verified that random weapons are failing to spawn entirely, not just being misplaced. ttt_random_ammo works correctly.

**Relevant console readout:**
```
] ent_create ttt_random_weapon

[TTT2 (Base) - v0.12.3b] gamemodes/terrortown/entities/entities/ttt_random_weapon.lua:18: attempt to call field 'GetSpawnableSWEPs' (a nil value)
  1. unknown - gamemodes/terrortown/entities/entities/ttt_random_weapon.lua:18
```

**Map:**
While this is simple to create a test map for (or use any map where you know the origin), here's the map I'm working on where the bug is clearly observed between loading it in TTT and TTT2. A majority of weapons in the map are templated ttt_random_weapon. The origin is on the lower floor. Mind the round variant where it might fixed-spawn HUGEs everywhere.
https://steamcommunity.com/sharedfiles/filedetails/?id=3147823480

EntranceJew:
have you checked out how these maps do it?
https://steamcommunity.com/sharedfiles/filedetails/?id=1326819008
https://steamcommunity.com/sharedfiles/filedetails/?id=1358920571

if you can help it, you should be using `entspawn.SpawnRandomWeapon` or `entspawn.SpawnRandomAmmo` within TTT2 -- but the source of ttt_random_weapon doesn't currently reference GetSpawnableSWEPs at all so i'm confused as to which addon you may have that is overwriting it
![image](https://github.com/TTT-2/TTT2/assets/5711436/ec614a4c-9791-4eef-9762-964a1cc4cf47)

can you run `which gamemodes/terrortown/entities/entities/ttt_random_weapon.lua` as a client to see where it originates from?

pontMakesThings:
[...]

PR #1409, opened 2024-02-12
Addonchecker: Added another addon

TimGoll:
As described in #1406 the added addon breaks our random spawn system and is therefore incompatible.


EntranceJew:
unfortunate, because as i mention on #1150, it's one of those things that is pretty interesting

ZenBre4ker (approved):

saibotk (approved):

PR #1410, opened 2024-02-12
EquipmentMenu: Added check before function call

TimGoll:
Last night I discussed some things with @mexikoedi. One thing we noticed that there are quite a few badly done weapons for TTT that forget to set the weapon base correctly. While this is bad practice and shouldn't be done, most of these weapons still work just fine. Adding this check here makes sure that this populate function is only called if it actually uses the correct base.

### DIFF
--- a/lua/terrortown/menus/gamemode/equipment/base_equipment.lua
+++ b/lua/terrortown/menus/gamemode/equipment/base_equipment.lua
@@ -186,7 +186,9 @@ function CLGAMEMODESUBMENU:Populate(parent)
     end
 
     -- now add custom equipment settings
-    equipment:AddToSettingsMenu(parent)
+    if isfunction(equipment.AddToSettingsMenu) then
+        equipment:AddToSettingsMenu(parent)
+    end
 
     -- stylua: ignore
     hook.Run("TTT2OnEquipmentAddToSettingsMenu", equipment, parent)
--- a/lua/terrortown/menus/gamemode/equipment/base_equipment.lua
+++ b/lua/terrortown/menus/gamemode/equipment/base_equipment.lua
@@ -193,7 +193,7 @@ function CLGAMEMODESUBMENU:Populate(parent)
             1,
             "Weapon '"
                 .. equipment:GetClass()
-                .. "' does't use the weapon_tttbase and can therefore not populate the settings panel."
+                .. "' doesn't use the weapon_tttbase and cannot be added to the settings panel."
         )
     end
 
--- a/lua/terrortown/menus/gamemode/equipment/base_equipment.lua
+++ b/lua/terrortown/menus/gamemode/equipment/base_equipment.lua
@@ -188,6 +188,13 @@ function CLGAMEMODESUBMENU:Populate(parent)
     -- now add custom equipment settings
     if isfunction(equipment.AddToSettingsMenu) then
         equipment:AddToSettingsMenu(parent)
+    else
+        Dev(
+            1,
+            "Weapon '"
+                .. equipment:GetClass()
+                .. "' does't use the weapon_tttbase and can therefore not populate the settings panel."
+        )
     end
 
     -- stylua: ignore
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,7 +89,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Fixed
 
-- Fixed database now properly saving boolean `false` values
+- Fixed database now properly saving boolean `false` values (by @ZenBre4ker)
 - Fixed cached weapons not being selected after giving them back to the owner (by @TimGoll)
 - The roundendscreen can now be closed with the correct Binding (by @ZenBre4ker)
 - Fixed last seen player being wrongly visible for every search instead of only public policing role search (by @TimGoll)
@@ -103,7 +103,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed font initialization to not trip engine font fallback behavior (by @EntranceJew)
 - Fixed the decoy producing a wrong colored icon for other teams (by @NickCloudAT)
 - Fixed the scoreboard being stuck open sometimes if the inflictor was no weapon (by @TimGoll)
-- Fixed door health displaying as a humongous string of decimals
+- Fixed door health displaying as a humongous string of decimals (by @EntranceJew)
+- Fixed weapons that use the wrong weapon base from throwing errors in the F1 menu (by @TimGoll)
 
 ### Removed
 


ZenBre4ker (approved):

EntranceJew:
how can something set their base as something other than weapon_tttbase and still be set accordingly? don't we explicitly filter out things that aren't based on weapon_tttbase? shouldn't we check or correct this even higher up the chain rather than mitigate consequences so they're even harder to detect down the chain?

TimGoll:
> how can something set their base as something other than weapon_tttbase and still be set accordingly? don't we explicitly filter out things that aren't based on weapon_tttbase? shouldn't we check or correct this even higher up the chain rather than mitigate consequences so they're even harder to detect down the chain?

This addon for example uses their own base: https://steamcommunity.com/sharedfiles/filedetails/?id=2705642928

There are quite a few of them and realistically no one will update them. We could also throw them out of the valid weapon list, but I think that will only result in complains by player.

My decision here to do it that way is founded on the fact that this weapon is working perfectly fine, it is just causing an error in the F1 menu:

```
[ttt2-test-0.13] addons/ttt2-test-0.13/lua/terrortown/menus/gamemode/equipment/base_equipment.lua:189: attempt to call method 'AddToSettingsMenu' (a nil value)
  1. Populate - addons/ttt2-test-0.13/lua/terrortown/menus/gamemode/equipment/base_equipment.lua:189
   2. BuildContentArea - addons/ttt2-test-0.13/gamemodes/terrortown/gamemode/client/cl_help.lua:319
    3. GenerateSubmenuList - addons/ttt2-test-0.13/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenulist_ttt2.lua:149
     4. OnValueChange - addons/ttt2-test-0.13/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenulist_ttt2.lua:72
      5. OnValueChange - addons/ttt2-test-0.13/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsearchbar_ttt2.lua:42
       6. unknown - lua/vgui/dtextentry.lua:194
```

TimGoll:
However, we should probably iterate over all weapons and check their base. If it isn't the correct one, then we should throw an error in the addon checker. I think that will get best of both worlds

EntranceJew:
> This addon for example uses their own base: https://steamcommunity.com/sharedfiles/filedetails/?id=2705642928
> 
But this one doesn't even reference weapon_tttbase because it immediately overwrites it, my understanding was these lines were true and we were meant to eject things that did not even utilize the weapon_tttbase:
- [skipWeapon](https://github.com/TTT-2/TTT2/blob/289895ee4dc7428cde7bbbe79d1eaefaaed140d4/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua#L199-L216)
- [comment](https://github.com/TTT-2/TTT2/blob/289895ee4dc7428cde7bbbe79d1eaefaaed140d4/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua#L99-L102)

> Also, I want to add here, that this error is pretty unhelpful to the average developer because it doesn't point in any way to their weapon. You have to understand the depths of the TTT2 code base in order to know the reason behind this error.
i agree with this, i just thought we were more selective with what does and doesn't get rejected

TimGoll:
> > This addon for example uses their own base: https://steamcommunity.com/sharedfiles/filedetails/?id=2705642928
> 
> But this one doesn't even reference weapon_tttbase because it immediately overwrites it, my understanding was these lines were true and we were meant to eject things that did not even utilize the weapon_tttbase:
> 
> * [skipWeapon](https://github.com/TTT-2/TTT2/blob/289895ee4dc7428cde7bbbe79d1eaefaaed140d4/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua#L199-L216)
> * [comment](https://github.com/TTT-2/TTT2/blob/289895ee4dc7428cde7bbbe79d1eaefaaed140d4/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua#L99-L102)
> 
> > Also, I want to add here, that this error is pretty unhelpful to the average developer because it doesn't point in any way to their weapon. You have to understand the depths of the TTT2 code base in order to know the reason behind this error.
> 
> i agree with this, i just thought we were more selective with what does and doesn't get rejected

huh, then this part of the code is not working. I did not know our policy was to reject all weapons that are not based on the weapon_tttbase. The we should probably change that

HOWEVER this weapon in particular isn't buyable, it is spawned in the world. Maybe there is the issue

mexikoedi:
> > > This addon for example uses their own base: https://steamcommunity.com/sharedfiles/filedetails/?id=2705642928
> > 
> > 
> > But this one doesn't even reference weapon_tttbase because it immediately overwrites it, my understanding was these lines were true and we were meant to eject things that did not even utilize the weapon_tttbase:
> > 
> > * [skipWeapon](https://github.com/TTT-2/TTT2/blob/289895ee4dc7428cde7bbbe79d1eaefaaed140d4/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua#L199-L216)
> > * [comment](https://github.com/TTT-2/TTT2/blob/289895ee4dc7428cde7bbbe79d1eaefaaed140d4/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua#L99-L102)
> > 
> > > Also, I want to add here, that this error is pretty unhelpful to the average developer because it doesn't point in any way to their weapon. You have to understand the depths of the TTT2 code base in order to know the reason behind this error.
> > 
> > 
> > i agree with this, i just thought we were more selective with what does and doesn't get rejected
> 
> huh, then this part of the code is not working. I did not know our policy was to reject all weapons that are not based on the weapon_tttbase. The we should probably change that
> 
> HOWEVER this weapon in particular isn't buyable, it is spawned in the world. Maybe there is the issue

It spawns in the world and is also buyable by default. 
But yes the error is only a bit annoying for example if you want to edit something in the "Equipment menu" and use the search bar or click on the addon then the error pops up but otherwise this "weapon" seems to be working fine.

EntranceJew:
Should definitely still consider removing things that share no lineage with weapon_tttbase, it'll be even more difficult to keep things sorted -- if a user is playing peer to peer with friends all their sandbox addons will leech into TTT and trip errors, because very few people set their weapons/ents to go into `/gamemodes/terrortown/`, it can get really messy. 

In ***this circumstance***, most of the MHS base is just SCK code that could be ported to the new code @TimGoll just introduced to the weapon_tttbase and isn't providing any critical functionality.

TimGoll:
> Should definitely still consider removing things that share no lineage with weapon_tttbase, it'll be even more difficult to keep things sorted -- if a user is playing peer to peer with friends all their sandbox addons will leech into TTT and trip errors, because very few people set their weapons/ents to go into `/gamemodes/terrortown/`, it can get really messy.
> 
> In _**this circumstance**_, most of the MHS base is just SCK code that could be ported to the new code @TimGoll just introduced to the weapon_tttbase and isn't providing any critical functionality.

I completely agree with you here. But we also have to be realistic, no one is going to update these addons. Maybe mexikoedi will, but not the original dev

EntranceJew:
> I completely agree with you here. But we also have to be realistic, no one is going to update these addons. Maybe mexikoedi will, but not the original dev

As a separate effort from the addonchecker, if there are old addons that people would like to see brought forward that have no equivalent then we should make a form or something to just be made aware of addons which aren't up to par with TTT2 features even if they're not "breaking" things. Things which could be using the MSTACK / STATUS libs or the new enhanced bases, or anything that is using SCK code that could be gutted and given tooltips / more consistent controls.

EntranceJew (approved):

TimGoll:
> > I completely agree with you here. But we also have to be realistic, no one is going to update these addons. Maybe mexikoedi will, but not the original dev
> 
> As a separate effort from the addonchecker, if there are old addons that people would like to see brought forward that have no equivalent then we should make a form or something to just be made aware of addons which aren't up to par with TTT2 features even if they're not "breaking" things. Things which could be using the MSTACK / STATUS libs or the new enhanced bases, or anything that is using SCK code that could be gutted and given tooltips / more consistent controls.

I'm 100% with you that the addonchecker should warn if the weapon uses the wrong base. Besides that I'm open to removing them completely like you suggested or doing it as mexikoedi suggested. I think @ZenBre4ker @saibotk and @Histalek should voice their opinion on this as well as this is a pretty important decision imho.

lua/terrortown/menus/gamemode/equipment/base_equipment.lua
```diff
@@ -186,7 +186,9 @@ function CLGAMEMODESUBMENU:Populate(parent)
     end
 
     -- now add custom equipment settings
-    equipment:AddToSettingsMenu(parent)
+    if isfunction(equipment.AddToSettingsMenu) then
+        equipment:AddToSettingsMenu(parent)
+    end
```
Histalek, line 191: maybe we could have an else-block with a message/warning? That way this problem is at least discoverable.

Histalek (commented):
Generally i'm fine with this sanity check, but we should make it discoverable so that there is at least the option that someone will fix it.

lua/terrortown/menus/gamemode/equipment/base_equipment.lua
TimGoll: As a `Dev(1, ....)`. Could do that! But a report in the addon checker that this addon is using the wrong base, would solve the issue as well. We could still do both though

Histalek: yes but the addonchecker is manually curated and we would need something to realize that an equipment is setting the wrong base so that we can put it onto the addonchecker list.
A Dev(1) message would allow us to do so.

saibotk (approved):

lua/terrortown/menus/gamemode/equipment/base_equipment.lua
TimGoll: yes, but my plan was to add an automatic check to the addonchecker, that checks the base of all weapons loaded. So you get the error on startup, not just when opening the UI. And the error would actually be more usable I think

Histalek: tbh i'd rather have a Dev(1) message now.

I'm not sure about the addonchecker addition, but that is a discussion for another time. And we can always have both or replace this once it's added to the addonchecker

TimGoll:
@Histalek please confirm that my dev message is to your liking

lua/terrortown/menus/gamemode/equipment/base_equipment.lua
```diff
@@ -186,7 +186,16 @@ function CLGAMEMODESUBMENU:Populate(parent)
...
     -- now add custom equipment settings
-    equipment:AddToSettingsMenu(parent)
+    if isfunction(equipment.AddToSettingsMenu) then
+        equipment:AddToSettingsMenu(parent)
+    else
+        Dev(
+            1,
+            "Weapon '"
+                .. equipment:GetClass()
+                .. "' does't use the weapon_tttbase and can therefore not populate the settings panel."
```
Histalek, line 196:
```suggestion
                .. "' doesn't use the weapon_tttbase and cannot be added to the settings panel."
```

TimGoll: ah shit, I merged it but your suggestion was wrong. It is added to the panel, only custom settings can't populate. I should probably revert it?

Histalek (approved):

PR #1413, opened 2024-02-14
Entity Base: Added throw space check

TimGoll:
Previously you were always able to throw devices (beacon, radio etc). This could cause issues when really close to a wall as they would glitch through it. I added the same check to it, that the weapon dropping also uses.

While working on that I decreased the min distance to a wall when dropping a weapon as I noticed that a smaller distance works as well and improves the weapon pickup in tight corners. My testing also showed no glitches with this change.

This should fix the issue @mexikoedi was having.

### DIFF
--- a/gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
@@ -306,12 +306,19 @@ if SERVER then
     -- @return boolean Returns true on success
     -- @realm server
     function ENT:ThrowEntity(ply, rotationalOffset)
+        local posThrow = ply:GetShootPos() - Vector(0, 0, 15)
+        local vecAim = ply:GetAimVector()
+
+        if not ply:HasDropSpace(nil, posThrow, vecAim) then
+            LANG.Msg(ply, "throw_no_room", nil, MSG_MSTACK_WARN)
+
+            return false
+        end
+
         ply:SetAnimation(PLAYER_ATTACK1)
 
         rotationalOffset = rotationalOffset or Angle(0, 0, 0)
 
-        local posThrow = ply:GetShootPos() - Vector(0, 0, 15)
-        local vecAim = ply:GetAimVector()
         local velocity = ply:GetVelocity()
         local velocityThrow = velocity + vecAim * 250
 
--- a/gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
@@ -309,7 +309,7 @@ if SERVER then
         local posThrow = ply:GetShootPos() - Vector(0, 0, 15)
         local vecAim = ply:GetAimVector()
 
-        if not ply:HasDropSpace(nil, posThrow, vecAim) then
+        if not ply:HasDropSpace(posThrow, vecAim) then
             LANG.Msg(ply, "throw_no_room", nil, MSG_MSTACK_WARN)
 
             return false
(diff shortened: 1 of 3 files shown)


gamemodes/terrortown/gamemode/server/sv_player_ext.lua
```diff
@@ -1316,6 +1316,20 @@ function plymeta:Give(weaponClassName, bNoAmmo)
     return wep
 end
 
+---
+-- Checks if the player has space in front of them to drop a weapon.
+-- @param[opt] Weapon wep The weapon that should be dropped; add it as a parameter
```
ZenBre4ker, line 1321: Optional parameters should come last

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
```diff
@@ -306,12 +306,19 @@ if SERVER then
     -- @return boolean Returns true on success
     -- @realm server
     function ENT:ThrowEntity(ply, rotationalOffset)
+        local posThrow = ply:GetShootPos() - Vector(0, 0, 15)
+        local vecAim = ply:GetAimVector()
+
+        if not ply:HasDropSpace(nil, posThrow, vecAim) then
```
ZenBre4ker, line 312: For whatever reasons this function needs a weapon. The entity to throw should also be included here or not?

ZenBre4ker (changes requested):

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
TimGoll: no, as the weapon drop and the entity throw are really different imho

gamemodes/terrortown/gamemode/server/sv_player_ext.lua
TimGoll: done

lua/terrortown/lang/en.lua
```diff
@@ -2188,3 +2188,6 @@ L.magneto_stick_help_carry_rag_pin = "Pin ragdoll"
 L.magneto_stick_help_carry_rag_drop = "Drop ragdoll"
 L.magneto_stick_help_carry_prop_release = "Release prop"
 L.magneto_stick_help_carry_prop_drop = "Drop prop"
+
+-- 2024-02-14
+L.throw_no_room = "You have to place here to throw this device"
```
saibotk, line 2193: This sentence is wrong 😅

saibotk (changes requested):

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
ZenBre4ker: But doesnt that check just include the entity in the trace as a filter, so it doesnt count as blocking entity?
Should be the same for weapon and placeable.

TimGoll: I don't have the reference for the weapon in this case. I could loop in through, but that is not needed. Probably it isn't needed in the filter anyway, as it wasn't part of the filter before my change, but I added it where I had it because there might be some weird edge cases with really big weapons

ZenBre4ker: isnt it just self?

TimGoll: no, self is the entity that is about to be spawned, not the weapon the player is holding

ZenBre4ker (approved):

lua/terrortown/lang/en.lua
TimGoll: done

saibotk (approved):

PR #1414, opened 2024-02-14
Weapon Base: Auto select last weapon on remove

TimGoll:
While fixing bugs in weapons I was pretty annoyed about two things:

- many weapons forget to call `lastinv` on remove, meaning that you don't have a weapon selected, not even unarmed (e.g. when placing a boom body the boom body placer weapon is removed)
- fixing this means copying the same code to each weapon

Therefore I added it to the weapon base with a SWEP flag to disable it if desired. But I don't see a reason why you wouldn't want it, my testing showed no problems.

### DIFF
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -128,10 +128,6 @@ SWEP.silentPickup = false
 -- Can be useful if you have multiple instances, that rely on global variables stored via weapons.GetStored()
 SWEP.HotReloadableKeys = {}
 
--- Auto select the last weapon, when this weapon is removed from the player's iventory
--- Only works if OnRemove in the base class is called
-SWEP.SelectLastWeaponOnRemove = true
-
 -- If this weapon should be given to players upon spawning, set a table of the
 -- roles this should happen for here
 --	SWEP.InLoadoutFor = {ROLE_TRAITOR, ROLE_DETECTIVE, ROLE_INNOCENT}
@@ -750,10 +746,6 @@ if CLIENT then
     ---
     -- @realm client
     function SWEP:OnRemove()
-        if not self.SelectLastWeaponOnRemove then
-            return
-        end
-
         local owner = self:GetOwner()
 
         if IsValid(owner) and owner == LocalPlayer() and owner:IsTerror() then
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -128,6 +128,10 @@ SWEP.silentPickup = false
 -- Can be useful if you have multiple instances, that rely on global variables stored via weapons.GetStored()
 SWEP.HotReloadableKeys = {}
 
+-- Auto select the last weapon, when this weapon is removed from the player's iventory
+-- Only works if OnRemove in the base class is called
+SWEP.SelectLastWeaponOnRemove = true
+
 -- If this weapon should be given to players upon spawning, set a table of the
 -- roles this should happen for here
 --	SWEP.InLoadoutFor = {ROLE_TRAITOR, ROLE_DETECTIVE, ROLE_INNOCENT}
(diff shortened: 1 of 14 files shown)


Histalek (approved):
nice

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -128,6 +128,10 @@ SWEP.silentPickup = false
 -- Can be useful if you have multiple instances, that rely on global variables stored via weapons.GetStored()
 SWEP.HotReloadableKeys = {}
 
+-- Auto select the last weapon, when this weapon is removed from the player's iventory
+-- Only works if OnRemove in the base class is called
+SWEP.SelectLastWeaponOnRemove = true
```
ZenBre4ker, line 133: I dont think this needs to be adjustable. If somebody doesnt want it, just override OnRemove.

TimGoll:
> Why did you add BaseClass.OnRemove only to a few weapons? Are they all calling that already?

Why should I add it to the other weapons? I'm confused?

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
TimGoll: I'm a fan of explicit variables to explain what happens here, but if I'm alone with that, I can remove that

Histalek:
> Why did you add BaseClass.OnRemove only to a few weapons? Are they all calling that already?

It's added to weapons that implement their own `OnRemove` function which shadows the base implementation. So it has to be called explicitly or it won't be called at all.

TimGoll:
> > Why did you add BaseClass.OnRemove only to a few weapons? Are they all calling that already?
> 
> It's added to weapons that implement their own `OnRemove` function which shadows the base implementation. So it has to be called explicitly or it won't be called at all.

Ah, now I got Sven's question. I don't have to add `BaseClass.OnRemove(self)` to the other weapons because I don't overwrite it and therefore the base class function is called automatically.

ZenBre4ker:
Yeah, on the phone I wasnt sure, but yeah, you only used it where it's necessary, so it's fine :D
Im not so sure about calling BaseClass.OnRemove as first call. I guess in most cases it doesnt matter, but I also dont know which problems could occur with calling lastinv while still removing as weapon.

TimGoll:
> Yeah, on the phone I wasnt sure, but yeah, you only used it where it's necessary, so it's fine :D Im not so sure about calling BaseClass.OnRemove as first call. I guess in most cases it doesnt matter, but I also dont know which problems could occur with calling lastinv while still removing as weapon.

I intentionally called it first so that when programmers do some wild stuff in the function, the owner is still valid. Probably not nessecary, but it shouldn't matter either way

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
ZenBre4ker: Not sure what you mean with explicit variables. But this is not really an attribute a weapon needs imho. Normally the selectlastweapononremove shouldnt even be handled by a weapon itself, but by the player-entity.
Also using a consolecommand seems like a lazy old global call to something you dont want to get the reference for.

But this is just me seeing those few lines. I didnt investigate the overall workings closely, just what I expect it to do.

TimGoll: > Normally the selectlastweapononremove shouldnt even be handled by a weapon itself, but by the player-entity.

But it has to trigger on weapon remove, that's easily done in the weapon base. I could also put this in a global hook, that would probably work as well.

> Also using a consolecommand seems like a lazy

I think that is honestly the best way to do. Otherwise I would have to track the weapon switches so I always know the last valid weapon. There is no function to get the last valid weapon.
Example: If I switch from A to B to C and B is removed before C, then I should auto switch back to A. So the tracking would be a bit more complex. GMod already does that, but only exposes this as a clientside command.

Another thing is that for some reason you can't trigger the selection of a weapon in the same tick you give it to a player. Even a tick later is not late enough. My testing showed that you need at least a delay of 0.2 seconds. Which is kinda weird.
This would probably not be an issue here, but something that you have to keep in mind

ZenBre4ker: Yeah i get that, I sometimes forget, that we have to deal with gmod in many of those weird cases.

Still think the same about removing that variable.

TimGoll: I can remove it, it is fine by me

TimGoll: done

ZenBre4ker (approved):

PR #1419, opened 2024-02-16
remove(sh_sql): Remove adding columns to existing database table

### LINKED ISSUE #1418: Sqlite PRAGMA will return nil in the future

TTT2-0.12.3b (workshop)

Garrysmod will not support PRAGMA and will start to return `nil` in the future.
(currently only the dev branch is affected)

Affected files i found:

https://github.com/TTT-2/TTT2/blob/88849ec06924779df67abda37c0b6f5eb03e14f2/gamemodes/terrortown/gamemode/shared/sh_sql.lua#L251

https://github.com/TTT-2/TTT2/blob/88849ec06924779df67abda37c0b6f5eb03e14f2/lua/ttt2/extensions/sql.lua#L16-L72

This will cause a error like this:

```
[TTT2 (Base) - v0.12.3b] gamemodes/terrortown/gamemode/shared/sh_sql.lua:237: attempt to get length of local 'clmns' (a nil value)
  1. CreateSqlTable - gamemodes/terrortown/gamemode/shared/sh_sql.lua:237
   2. Register - lua/ttt2/libraries/database.lua:960
    3. OnLoaded - lua/ttt2/libraries/items.lua:115
     4. unknown - gamemodes/terrortown/gamemode/server/sv_main.lua:422
```

image
(screenshot of next-update channel on the garrysmod discord)

Histalek:
Before this change it was possible to add columns to an existing database table through the `sql.CreateSqlTable()` function.

This is a troublesome and undocumented(!) functionality which should not exist going further.
Troublesome as once a column is added there is no way back. This whole functionality also sways from the `CreateSqlTable` expectation as running a `CREATE TABLE` statement on an existing sqlite table fails normally. (In our case we used this function to ensure a table exists, hence we now return `true` if the table already exists)

`sql.CreateSqlTable` is now also "soft" deprecated through our code annoations and therefore in our api-docs.
Handling database table definitions should be done through migrations from now on.

"Hard" deprecation (`ErrorNoHalt` on function call) should follow once we have migrated our codebase to use migrations to prevent spamming of deprecation messages.

Additionally this partially fixes #1418 as one of the offending pragma statements is removed.

A few notes:
- i'm not super sure on the `Dev` messages, maybe those could be improved/changed
- i've gone with the "soft" deprecation route so this **could** be merged for v0.13.0, but it is definitely not required.

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/sh_sql.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_sql.lua
@@ -233,11 +233,26 @@ end
 -- @param table keys the keys for the data @{table}
 -- @return boolean Whether the database table was created successfully
 -- @realm shared
--- @todo usage
+-- @deprecated
 function sql.CreateSqlTable(tableName, keys)
-    local result
-
-    if not sql.TableExists(tableName) then
+    Dev(
+        2,
+        "`sql.CreateSqlTable` is deprecated and should be handled by migrations.",
+        "This ran for table:",
+        tableName
+    )
+    if sql.TableExists(tableName) then
+        Dev(
+            2,
+            "The database table",
+            tableName,
+            "already exists.",
+            "Adding columns to an existing database table via `sql.CreateSqlTable` was removed."
+        )
+        -- We need to return true here to not break our current assumptions
+        -- in our current codebase (database library, roleselection, huds)
+        return true
+    else
         local str = "CREATE TABLE " .. sql.SQLStr(tableName) .. " (name TEXT PRIMARY KEY"
 
         for key, data in pairs(keys) do
@@ -246,39 +261,8 @@ function sql.CreateSqlTable(tableName, keys)
 
         str = str .. ")"
 
-        result = sql.Query(str)
-    else
-        local clmns = sql.Query("PRAGMA table_info(" .. sql.SQLStr(tableName) .. ")")
-
-        for key, data in pairs(keys) do
-            local exists = false
-
-            for i = 1, #clmns do
-                if clmns[i].name ~= key then
-                    continue
-                end
-
-                exists = true
-            end
-
-            if exists then
-                continue
-            end
-
-            local res = sql.ParseDataString(key, data)
-            if not res then
-                continue
-            end
-
-            local resArr = string.Explode(",", res)
-
-            for i = 1, #resArr do
-                sql.Query("ALTER TABLE " .. sql.SQLStr(tableName) .. " ADD " .. resArr[i])
-            end
-        end
+        return sql.Query(str) ~= false
     end
-
-    return result ~= false
 end
 
 ---
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed the scoreboard being stuck open sometimes if the inflictor was no weapon (by @TimGoll)
 - Fixed door health displaying as a humongous string of decimals
 
+### Deprecated
+
+- Deprecated `sql.CreateSqlTable`, use migrations to manage database schemas instead
+
 ### Removed
 
 - Removed some crosshair related convars and replaced them with other ones, see the crosshair settings menu for details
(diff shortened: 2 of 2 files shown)


gamemodes/terrortown/gamemode/shared/sh_sql.lua
```diff
@@ -233,11 +233,26 @@ end
...
 function sql.CreateSqlTable(tableName, keys)
-    local result
-
-    if not sql.TableExists(tableName) then
+    Dev(
+        2,
+        "`sql.CreateSqlTable` is deprecated and should be handled by migrations.",
+        "This ran for table:",
+        tableName
+    )
```
ZenBre4ker, line 243: Deprecating this functionality idk.
Could be still used in migration files.
Maybe check a flag, so that it works without dev message in migrations?

gamemodes/terrortown/gamemode/shared/sh_sql.lua
```diff
@@ -233,11 +233,26 @@ end
...
+        Dev(
+            2,
+            "The database table",
+            tableName,
+            "already exists.",
+            "Adding columns to an existing database table via `sql.CreateSqlTable` was removed."
+        )
+        -- We need to return true here to not break our current assumptions
+        -- in our current codebase (database library, roleselection, huds)
+        return true
```
ZenBre4ker, line 254: Maybe we should check keys and return false if somebody tries to change an existing database.

ZenBre4ker (changes requested):
Generally fine to deprecate it. But I dont like it, that there is no alternative besife writing sql code yourself. I'd like to keep the create sql table function and rather limit it to migrations.

gamemodes/terrortown/gamemode/shared/sh_sql.lua
Histalek: this would be somewhat of a catch-22 with the `pragma` problem :D given that we would need to check the existing keys of the database table

ZenBre4ker: Damn, I see. Didnt know that.

ZenBre4ker (approved):

gamemodes/terrortown/gamemode/shared/sh_sql.lua
ZenBre4ker: Writing own sql statements directly in migration files seems to be the best way.

TimGoll (approved):
Was this functionality used somewhere in our codebase? It is really weird and unexpected behaviour. I therefore agree with your change here. Not sure if it should be Dev 1 or 2, but I think it is fine that way!

ZenBre4ker:
> Was this functionality used somewhere in our codebase? It is really weird and unexpected behaviour. I therefore agree with your change here. Not sure if it should be Dev 1 or 2, but I think it is fine that way!

Everytime we changed the savingKeys for the shop for example, this was used.

TimGoll:
> > Was this functionality used somewhere in our codebase? It is really weird and unexpected behaviour. I therefore agree with your change here. Not sure if it should be Dev 1 or 2, but I think it is fine that way!
> 
> Everytime we changed the savingKeys for the shop for example, this was used.

I'm a bit OOTL on all the recent database/migration changes. So was this changed already? or is this PR breaking things?

ZenBre4ker:
Lets explain it like this:
For example we added in version 0.12.? the DeathOnDropOverride for the shopeditor.
There we added this "item" to the savingKeys as additional Column.
When Sql.CreateSqlTable is called, it checks if the stored Keys are equivalent to the ones that shall be applied.
If that is not the case, we just add them.
Therefore the table is changed.

With this PR this will be forbidden or just not used anymore.
If the savingKeys change, they will not update the sql table anymore and will therefore result in missing data.

TimGoll:
Ok, I get it now, thank you!

saibotk (approved):

Histalek:
Given that this would break users that have played ttt2 before v0.12.0 but haven't started v0.12.0 or above yet i'm not exactly keen on merging this without having the affected tables be fully controlled by migrations.
Holding off on merging this seems like the best idea for now.

TimGoll:
> Given that this would break users that have played ttt2 before v0.12.0 but haven't started v0.12.0 or above yet i'm not exactly keen on merging this without having the affected tables be fully controlled by migrations. Holding off on merging this seems like the best idea for now.

Migrations are done now, in case you want to finish this PR!

PR #1310, opened 2024-01-22
Equipment: All builtin items can configured via AllowDrop now

EntranceJew:
- does not affect the default drop-ability of weapons
- unarmed got an icon because it didn't have one
- unarmed got removed from the shop blacklist because there's no way to change AllowDrop without it being in there
- removed ShouldDropOnDie because we [never meet the criteria for calling it](https://wiki.facepunch.com/gmod/WEAPON:ShouldDropOnDie)

PR #1421, opened 2024-02-17
Equipment: Always remove weapon_ttt_unarmed

ZenBre4ker:
Currently it is possible for weapon_ttt_unarmed to be dropped, but as it is no real weapon it's not inside the shopeditor, therefore settings made to it via mysterious ways cant be reversed.

One example are the wrongly saved values to the database introduced in v0.11.7b with this PR #1041.
As only in v0.12.0b with PR #984 the way no default values are saved was introduced, servers before this PR now have partly unusable values saved for some weapons and changing their defaults now doesnt change that anymore.

So this partly reverts changes done in #1310  at least for unarmed.
To make sure this never has any consequences on other weapons we also block the ability to choose the unarmed slot.

### DIFF
--- a/gamemodes/terrortown/entities/weapons/weapon_ttt_unarmed.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_ttt_unarmed.lua
@@ -68,6 +68,18 @@ function SWEP:Holster()
     return true
 end
 
+---
+-- @ignore
+function SWEP:OnDrop()
+    self:Remove()
+end
+
+---
+-- @ignore
+function SWEP:ShouldDropOnDie()
+    return false
+end
+
 if CLIENT then
     ---
     -- @realm client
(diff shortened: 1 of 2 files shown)


TimGoll (approved):
This looks good to me. We can discuss changing that in the future if there is any need for that, but right now I don't see any reason to be able to drop the unarmed "weapon" as it is no real wweapon anyway.

EntranceJew:
image
will this break the infected's ability to not have the unarmed?

TimGoll:
If it works in 0.12 then it will continue to work

saibotk (approved):

TimGoll:
@ZenBre4ker is there a reason you didn't merge this yet?

ZenBre4ker:
Beside the fact, that I wasnt online yet to test it, no.
Feel free to merge it on your own risk :P

PR #1423, opened 2024-02-18
Markervision: Generalize for all placeable entities

ZenBre4ker:
This enables markervision for all placeable entities, by initializing and removing it in the baseclass.
To overwrite, change icons or add additional infos, you can use the ent specific functions.

This shall also be the start of remote triggerable entities, that make use of the markervision system, by enabling remote configuration when a markervision enabled entity is in line of sight.

### DIFF
--- a/gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
@@ -408,6 +408,11 @@ if SERVER then
         return true
     end
 
+    ---
+    -- Gets the current markerOwner
+    -- @note Override if not the originator should own the marker
+    -- @return Player|Role The Player or role be marker owner
+    -- @realm server
     function ENT:GetMarkerOwner()
         return self:GetOriginator()
     end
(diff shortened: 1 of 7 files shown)


gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
```diff
@@ -14,6 +14,10 @@ ENT.isDestructible = true
 
 ENT.pickupWeaponClass = nil
 
+-- MarkerVision related
+ENT.iconMaterial = Material("vgui/ttt/tid/tid_big_role_not_known")
```
ZenBre4ker, line 18: We need to do a missing icons icon.

gamemodes/terrortown/entities/entities/ttt_health_station.lua
```diff
@@ -12,6 +12,7 @@ DEFINE_BASECLASS("ttt_base_placeable")
 if CLIENT then
     ENT.Icon = "vgui/ttt/icon_health"
     ENT.PrintName = "hstation_name"
+    ENT.iconMaterial = Material("vgui/ttt/icon_health")
```
ZenBre4ker, line 15: better health icon

gamemodes/terrortown/entities/entities/ttt_decoy.lua
```diff
@@ -19,6 +19,12 @@ ENT.CanHavePrints = false
 ENT.CanUseKey = true
 ENT.pickupWeaponClass = "weapon_ttt_decoy"
 
+if CLIENT then
+    ENT.Icon = "vgui/ttt/icon_decoy"
+    ENT.PrintName = "decoy_name"
+    ENT.iconMaterial = Material("vgui/ttt/icon_decoy")
```
ZenBre4ker, line 25: better decoy icon

TimGoll:
One thing I'm not a fan of is the `.iconMaterial`. For two reasons:

1. The `.iconMaterial` for roles and weapons is autogenerated
2. The `.Icon` variable also has a different icon then the `.iconMaterial` which would be really confusing

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
```diff
@@ -24,6 +28,11 @@ function ENT:Initialize()
 
     if SERVER then
         self:PrecacheGibs()
+
+        local mvObject = self:AddMarkerVision("ttt_base_placeable_owner")
+        mvObject:SetOwner(self:GetOriginator())
```
TimGoll, line 33: not a fan that the owner is hard set to the originator here, the owner can also be a role or a team

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
```diff
@@ -392,4 +407,58 @@ if SERVER then
...
+    end
+
+    function ENT:CustomizeMarkerVision(mvData) end
+
+    function ENT:GetMarkerVisionIcons(mvData)
+        return { self.iconMaterial }
+    end
+
+    function ENT:GetMarkerVisionColors(mvData)
+        return { COLOR_WHITE }
```
TimGoll, line 427: Why is this a color table? Is this one color for each icon? One for each line of text?

TimGoll: Ah, I see this is for the icon, this is a bit confusing, because the text can have colors as well

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
ZenBre4ker: For what is the owner relevant anyways? I just hardcoded it as I saw no benefit to it being customizable. Normally you place it you own it.

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
ZenBre4ker: Good to know. Will adjust.

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
TimGoll: The heroes crystal, that also uses this base, has the owner set to ROLE_SUPERVILLAIN. As any supervillain can see it, but a supervillain changing their role, will lose access to the marker vision

ZenBre4ker:
Tried to do your review comments. Would start documenting later and then try myself with the remote trigger.

ZenBre4ker:
This is finished from my side, if Tim supplies the icons, it should be ready again.
It has no hurry for me personally, as I wont rely for now on the placeable entity, but on markervision only.

TimGoll:
> This is finished from my side, if Tim supplies the icons, it should be ready again. It has no hurry for me personally, as I wont rely for now on the placeable entity, but on markervision only.

I'll try to add the icons this week!

TimGoll:
I took a quick glimpse at the code and I think @saibotk has to give his opinion here on the structure. I'm not entirely sure if this complicates it too much or so, seems a bit arbitrary to me, hm

PR #1424, opened 2024-02-18
Equipment Caching: Rework

### LINKED ISSUE #1091: [Hotreload] Hotreload breaks shops
Shops with standard items (not with manually selected items) are broken after hotreload. This is because the functions has to be called after a hotreload: https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/client/cl_equip.lua#L1117-L1130

### LINKED ISSUE #1403: unarmed errors because it's ejected

TimGoll:
Fixes #1403 
Partly fixes some of #1091 (it does not fix this issue, it fixes a part of this issue, but after hotreload weapons are still missing (items are there) and everything is not buyable)

It looks like lots of changes, but I only moved things around. It is mostly the old code with renamed variables split into three functions. Notable changes:

- caching is now independant of eject and other changes, meaning that also `weapon_ttt_unarmed` has a cached icon
- caching now happens on every load / reload of the game instead of the first call this improved the first opening time of the shop for me by a lot
- it fixes the issue that the cache is not rebuilt on hoteload until the shop is reopened, improving some things mentioned in in 1091

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_equip.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_equip.lua
@@ -1195,40 +1195,6 @@ function GM:OnContextMenuOpen()
     end
 end
 
----
--- Caches the material or model
--- @param table item
--- @realm client
-function TTT2CacheEquipMaterials(item)
-    item.isValidEquipment = true
-
-    if item.material then
-        item.iconMaterial = Material(item.material)
-
-        if item.iconMaterial:IsError() then
-            -- setting fallback error material
-            item.iconMaterial = fallback_mat
-        end
-    elseif item.model then
-        -- do not use fallback mat and use model instead
-        item.itemModel = item.model
-    end
-
-    --if there is no sensible material and model, the item should probably not be editable in the equipment Editor
-    if item.material == "vgui/ttt/icon_id" and item.model == "models/weapons/w_bugbait.mdl" then
-        item.isValidEquipment = false
-    end
-end
-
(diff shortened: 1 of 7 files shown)


TimGoll:
This should fix the issue @EntranceJew was experiencing. Please confirm

saibotk (approved):
Very nice!

TimGoll:
> Looking good so far. Gonna check in detail at home.
> 
> Im only wondering if this is shopeditor only or if this actually should be done for shop or other parts as well. Because then I would move this to another namespace like shop or equipment related.

The shop UI already uses this, the shop backend has its own handling. Maybe one can unify this, but probably not in this PR

gamemodes/terrortown/gamemode/client/cl_equip.lua
```diff
@@ -1195,40 +1193,6 @@ function GM:OnContextMenuOpen()
     end
 end
 
----
--- Caches the material or model
--- @param table item
--- @realm client
-function TTT2CacheEquipMaterials(item)
```
ZenBre4ker, line 1202: This function is used in sh_main after a hotreload of a weapon you changed.

ZenBre4ker (changes requested):

ZenBre4ker (approved):
Fixed the only error I saw. Otherwise looking good and works for me.

PR #1425, opened 2024-02-18
AddonChecker: added base checking

TimGoll:
This is still a draft as I have to test how useful this actually is. But this is what I envisioned: Print a warning in the addon checker if an addon uses the wrong base. Sadly I think there is no way to detect the addon where this file came from.


Histalek:
Given that e.g. TFA weapons ( see #1432 ) do not use `weapon_tttbase` as a base, this would have the potential of spamming the addonchecker for no good reason. The TFA devs made these weapons explicitly compatible with ttt.

Maybe we could gate this weaponbase scanning behind a flag and not run it by default? Or split this out into another function which can be triggered by a different concommand?

If this is split and not run by default i would be fine with merging this. Giving serverowners a tool to discover these kinds of possibly problematic weapons definitely has value.

PR #1436, opened 2024-02-25
Entity Base: Add remote triggerable

ZenBre4ker:
This PR aims to gather the focussed markers and enables a clientside remote trigger to placeable entities.

Shall further enable the removal of radio tabs in shop.
Hopefully can be used for a radar as well, if we make it placeable for example.


gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
```diff
@@ -392,4 +392,43 @@ if SERVER then
...
+        })
+
+        -- Check if we want to pickup a weapon or check a corpse
+        if tr.Hit and IsValid(tr.Entity) and (tr.Entity.CanUseKey or tr.Entity.player_ragdoll) then
+            return
+        end
+
+        focussedEnt:RemoteUse()
+    end)
+end -- CLIENT
```
TimGoll, line 434: I know why you put it on the client, but should it really be on the client? Two things:

- `ENT:UseOveerride` is serverside
- there are quite a few role specific checks that can only be done serverside because the data is only valid on the server

My suggestion would complicate this whole system a bit, but:

- integrate this hook into the already existing system:
  https://github.com/TTT-2/TTT2/blob/2ae4f047ad44db09e214c87d5e9751a7c856a7ef/gamemodes/terrortown/gamemode/server/sv_player.lua#L523-L556
- call a function like `ENT:RemoteUseOverride` if it exists, that way it would also work for **all** entities, not just placeable entities
- populate the placeables UseOverride with a hook that can be overriden that contains a check if the player is able to open the UI, similar to this:
  https://github.com/TTT-2/TTT2/blob/bb659f30a2d3281ce6ccf1b0d5e1c8824c97363e/gamemodes/terrortown/entities/entities/ttt_base_placeable.lua#L245
- send a netmessage to the client that triggers a clientside RemoteUseOverride where the UI can be added.

But now that I think about this, this might be problematic in high ping scenarios. So see this as a starting point for a discussion. But I'm not sure that I like the current implementation that much.

lua/ttt2/libraries/marker_vision.lua
```diff
@@ -227,6 +228,24 @@ if CLIENT then
...
+            local entTbl = markerVision.focussedMarkers[i]
+            if entTbl.screenDistanceSquared < closestEntTbl.screenDistanceSquared then
+                closestEntTbl = entTbl
+            end
+        end
+
+        markerVision.focussedMarkers = { closestEntTbl }
+
+        return closestEntTbl.entity
+    end
```
TimGoll, line 247: just a note; this could probably unnecessary if we implemented a marker sorting system on a frame by frame base that makes sure they are also rendered in the correct order (which they aren't right now)

lua/ttt2/libraries/marker_vision.lua
```diff
@@ -294,6 +313,13 @@ if CLIENT then
                 continue
             end
 
+            if isOnScreenCenter then
+                markerVision.focussedMarkers[#markerVision.focussedMarkers + 1] = {
+                    entity = ent,
+                    screenDistanceSquared = screenDistSquared,
+                }
```
TimGoll, line 320: you seem to add one focussed marker each frame here, am I correct? Do you clear them as well besides when a player presses their use key? Otherwise this seems like a big problem. Also I'm a bit confused with the screenDistSquared thing. Why does it have to be squared?

TimGoll:
Overall I really like this feature, I'm not so sure on its implementation though

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
TimGoll: Or - if it is client side only like you implemented it currently - it should probably be a feature of markerVision, not the placeable entity. That way one could add UIs to all kind of things

ZenBre4ker: Well so here are the main points:
- Markervision is currently rendered and calculated clientside only.
- The ideal TTT2 GM:KeyRelease implementation for pickup and bodysearch is serverside only, but generally predicted and shared

So this was my idea to simply use that hook and just react to the usekey clientside, if a marker is focussed and do all the calls to the server from here.
It wouldnt make sense to combine this generally with the server as long as the server has no clue which markers you focus.
It would be much better to change our general use key system to a predicted one, that would also solve weapons not being properly picked up. But then we would also need to predict the marker system. We cant use eyetrace on the server, which is already quite bad with client vision.

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
```diff
@@ -392,4 +392,43 @@ if SERVER then
...
     end
-end
+end -- SERVER
+
+if CLIENT then
+    ---
+    -- This is triggered, when you focus a marker of an entity and press 'Use'-Key
+    -- Override to handle remote use
+    -- @realm client
+    function ENT:RemoteUse() end
```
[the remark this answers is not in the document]
ZenBre4ker: I dont get the Override part. This is more a OnRemoteUse for sure, but whats with that UseOverride?

lua/ttt2/libraries/marker_vision.lua
TimGoll: Exactly, or what I recently did with the overhead icons

lua/ttt2/libraries/marker_vision.lua
TimGoll: I see, I didn't see that connection while looking at the code, but now I get it

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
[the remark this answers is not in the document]
TimGoll: Because it is a hook that is intended to be overriden. In general I prefer the `OnXYZ` naming as well, but the other use hook already has the name `UseOverride`, hence I shuggested that as a name

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
ZenBre4ker: No, thats not what I meant to say. Just wanted to point out my problems. I agree, that this is not a good solution, but I wanted feedback, so thanks for that <3

ZenBre4ker (commented):
I guess I'm gonna put this to draft again. As it was mainly a quick way to get remote use to work. But I will try to work on a proper implementation in GM:KeyRelease instead and combine it with weapon pickup and bodysearch to be client-authorative and just checked by the server if it is generally possible.

ZenBre4ker:
This shall be based on #1447 instead of this unrelated hook.

ZenBre4ker:
Updated and waiting for #1447

ZenBre4ker:
Next one is directly ready 😆

gamemodes/terrortown/gamemode/client/cl_keys.lua
```diff
@@ -76,29 +77,38 @@ function GM:PlayerBindPress(ply, bindName, pressed)
...
+            useEnt = tr.Entity
+            if not tr.Hit or not IsValid(useEnt) then
+                return
+            end
+
+            if isfunction(useEnt.ClientUse) then
+                isClientOnly = useEnt:ClientUse()
+            end
+        elseif isfunction(useEnt.RemoteUse) then
+            isClientOnly = useEnt:RemoteUse(ply)
```
TimGoll, line 101: Is there a reason `ply` is only passed to one of these two? If not, couldn't these two be combined?

gamemodes/terrortown/gamemode/client/cl_keys.lua
```diff
@@ -76,29 +77,38 @@ function GM:PlayerBindPress(ply, bindName, pressed)
...
-        if not tr.Hit or not IsValid(tr.Entity) then
-            return
-        end
-
+        -- Find out if a marker is focussed otherwise check normal use
         local isClientOnly = false
-        if isfunction(tr.Entity.ClientUse) then
-            isClientOnly = tr.Entity:ClientUse()
+        local useEnt = markerVision.GetFocussedEntity()
+        local isRemote = IsValid(useEnt)
```
TimGoll, line 83: maybe something like `isRemoteMarker` so that you see that this has something to do with marker vision?

lua/ttt2/libraries/marker_vision.lua
TimGoll: I guess you won't implement a sorting for the rendering in this PR?

TimGoll (commented):
I like it much more then the previous imlementation. Some minor things. Also I'm wondering if we should use your current method for the focus detection, or if we should just add sorting so we can ditch that part of the code

gamemodes/terrortown/gamemode/client/cl_keys.lua
ZenBre4ker: Not sure if that is relevant. This internally just remotes use an entity, that is remote controlled. So the moment we get the focussed entity from markervision we dont do anything with the marker anymore, so we only need to know if this is a remote entity or a close one.

gamemodes/terrortown/gamemode/client/cl_keys.lua
ZenBre4ker: Well `RemoteUse` shall be a shared implementation, which was not possible with `Use` as it could have broken old implementations. Therefore ClientUse didnt need it, but RemoteUse does.

lua/ttt2/libraries/marker_vision.lua
ZenBre4ker: Not really, I only get the foremost marker, therefore the interface/API stands and if we wanna sort it at one point internally and just return the foremost marker inside this function without the sorting its fine as well.

gamemodes/terrortown/gamemode/client/cl_keys.lua
TimGoll: I think I forgot how to read, to me it seemed like you call the same two functions here. My bad!

gamemodes/terrortown/gamemode/client/cl_keys.lua
TimGoll: I just thought it would make it easier at a later point to read the code

gamemodes/terrortown/gamemode/client/cl_keys.lua
Histalek, line 101: This would prioritize focused marker ents over "normal" use scenarios even if the marker doesn't have RemoteUse functionality, right?

From a gameplay perspective i'm not sure if i like this or not.
It is probably good because it is always the same no matter what marker ent happens to be focused.

Histalek (approved):
LGTM

gamemodes/terrortown/gamemode/client/cl_keys.lua
ZenBre4ker: Not sure I understand that. Because the variable isRemote does not contain the marker entity, but a boolean to detect if we should trigger RemoteUse or normal ClientUse + serverside Use.

gamemodes/terrortown/gamemode/client/cl_keys.lua
ZenBre4ker: Yes, the current implementation of markers are drawn over everything and it even gets bigger on focus, so I think its clear, that nothing else should get triggered.
Even if the marker has no implementation clienside, it still might be that there is an implementation serverside.

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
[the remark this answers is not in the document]
ZenBre4ker: Comin back to this, as Im gonna remove UseOverride in my cleanup, I dont think it should be called OnXYZ.
The gmod hook is also just called Use and is called on using it.
You can always call it yourself, so it's not even wrong to not be called OnXYZ or Override, because hooks are always called and can be overriden.

gamemodes/terrortown/gamemode/client/cl_keys.lua
TimGoll: I guess you're right

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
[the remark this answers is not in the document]
Histalek: @TimGoll this conversation is still open but i think it was solved. Could you double check?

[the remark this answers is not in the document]
TimGoll: yep

PR #1437, opened 2024-02-25
Radio: Create new UI for remote useable radio

ZenBre4ker:
Before this you had to navigate to the shop tabs to use it, which is really unintuitive and currently blocks our idea to do a new shop.
grafik

Now you can use the 'Use'-Key from any distance on the focussed radio marker
grafik
and get the following new UI, which triggers the sounds in the same way as before
grafik

This will be waiting for the approval of the remote trigger feature in #1436

### DIFF
--- a/gamemodes/terrortown/entities/entities/ttt_radio.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_radio.lua
@@ -368,10 +368,14 @@ if CLIENT then
 
     ---
     -- This is triggered, when you focus a marker of an entity and press 'Use'-Key
-    -- Overriden to open the radio menu
+    -- Opens the radio menu
+    -- @param Player ply The player that used this entity. Always LocalPlayer here.
+    -- @return bool True, if this shouldnt be used serverside
     -- @realm client
-    function ENT:RemoteUse()
+    function ENT:RemoteUse(ply)
         TRADIO:Toggle(self)
+
+        return true
     end
 end
 
--- a/gamemodes/terrortown/entities/entities/ttt_radio.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_radio.lua
@@ -362,6 +362,7 @@ if CLIENT then
 
         mvData:AddDescriptionLine(ParT("marker_vision_owner", { owner = nick }))
         mvData:AddDescriptionLine(ParT("marker_vision_distance", { distance = distance }))
+        mvData:AddDescriptionLine(ParT("use_entity", { usekey = Key("+use", "USE") }))
 
         mvData:AddDescriptionLine(TryT(mvObject:GetVisibleForTranslationKey()), COLOR_SLATEGRAY)
     end)
--- a/gamemodes/terrortown/entities/entities/ttt_radio.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_radio.lua
@@ -371,7 +371,7 @@ if CLIENT then
     -- This is triggered, when you focus a marker of an entity and press 'Use'-Key
     -- Opens the radio menu
     -- @param Player ply The player that used this entity. Always LocalPlayer here.
-    -- @return bool True, if this shouldnt be used serverside
+    -- @return bool True, because this shouldn't be used serverside
     -- @realm client
     function ENT:RemoteUse(ply)
         TRADIO:Toggle(self)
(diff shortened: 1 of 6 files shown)


TimGoll:
We should add some sort of indication that the radio supports this kind of stuff, right now no one can figure that out I think

gamemodes/terrortown/gamemode/client/cl_tradio.lua
```diff
@@ -21,92 +20,80 @@ local sound_names = {
...
+    self.menuFrame:SetPadding(
+        self.sizes.padding,
+        self.sizes.padding,
+        self.sizes.padding,
+        self.sizes.padding
+    )
+
+    -- any keypress closes the frame
+    self.menuFrame:SetKeyboardInputEnabled(false)
+    self.menuFrame.OnKeyCodePressed = util.BasicKeyHandler
```
TimGoll, line 73: This means that the UI is closed as soon as the player moves, doesn't it? Not sure if this is the right approach here as it is probably annoying to stand this to interact. With the old shop implementation you were also able to move around while the UI is open

ZenBre4ker: I believe line 73 is irrelevant, as I deactivated it in 72. So you can move while running.
I still gotta do some cleanups 😄

ZenBre4ker:
This is ready as well now.
grafik

I added a general name for the translation string, because I could imagine to use this generally for all entities Use / UseOverride / ClientUse and RemoteUse. Normally UI descriptions should be kept shortm if you want more information it should be standing somewhere else. So I dont want to add "this {radio_name}" or "remotely" as this is kinda self explaining if you just try to press E.

gamemodes/terrortown/entities/entities/ttt_radio.lua
```diff
@@ -362,9 +362,22 @@ if CLIENT then
...
+        mvData:AddDescriptionLine(ParT("use_entity", { usekey = Key("+use", "USE") }))
 
         mvData:AddDescriptionLine(TryT(mvObject:GetVisibleForTranslationKey()), COLOR_SLATEGRAY)
     end)
+
+    ---
+    -- This is triggered, when you focus a marker of an entity and press 'Use'-Key
+    -- Opens the radio menu
+    -- @param Player ply The player that used this entity. Always LocalPlayer here.
+    -- @return bool True, if this shouldnt be used serverside
```
Histalek, line 374:
```suggestion
    -- @return bool True, because this shouldn't be used serverside
```

Histalek (approved):
Nice!

TimGoll:
Could you please put the keybind in the subtitle instead of a "random" description line? Right now it gets kind of lost and it is also not in-line with the other targetID implementations:

https://github.com/TTT-2/TTT2/blob/master/lua/ttt2/libraries/targetid.lua#L621-L628

ZenBre4ker:
I get what you mean, but you could also improve it yourself when its implemented.
I'd rather merge it, get some feedback earlier, then wait for a feature thats not necessarily needed, its only an improvement. 😄 
It's also just a placeholder, as I would rather generalize it for all remote enabled entities, so it probably wont stay like this in the radar anyway.

gamemodes/terrortown/gamemode/client/cl_tradio.lua
```diff
@@ -21,92 +20,77 @@ local sound_names = {
...
+    buttonField:Dock(TOP)
+
+    for sound, translationName in pairs(sounds) do
+        local buttonReport = vgui.Create("DButtonTTT2", buttonField)
+        buttonReport:SetText(LANG.GetTranslation(translationName))
+        buttonReport:SetSize(self.sizes.widthButton, self.sizes.heightButton)
+        buttonReport.DoClick = function(btn)
+            if IsValid(self.radio) then
+                RunConsoleCommand("ttt_radio_play", self.entIndex, sound)
+            end
```
TimGoll, line 93:
```suggestion
            if not IsValid(self.radio) then return end

            RunConsoleCommand("ttt_radio_play", self.entIndex, sound)
```

TimGoll (approved):
LGTM, I'll rework the tooltip today or tomorrow then! Please merge the two open suggestions before merging the PR though, thanks.

PR #1439, opened 2024-02-26
Custom View: Fixed broken 3rd party addons

TimGoll:
There were quite a few reports that the reworked custom view is breaking 3rd party addons such as super sheep. @EntranceJew and I were unable to reproduce it. This is when it it me, it is due to the loading order. On my testing setup, TTT2 is loaded last, because it is a file system addon. For people using the workshop version this is not true. If TTT2 is loaded before the supersheep, the CalcView hook breaks the supersheep.

The super sheep registers its own `CalcView` hook:
image

However due to the way hooks work in GMod, the next hook is only called if the previous one returned `nil`. Due to an oversight in my CalcView hook implementation, there was no early return for observing non-player entities. I fixed this. This alone should have fixed the issue at hand here.

But to make it extra safe, I moved the view calculation to `GM:CalcView`. This way our hook is always called last, which is what we want.


TimGoll:
To make your review easier, there were only two things I fixed in the `CalcView` hook besides moving it to a different file:

1. `obersverTarget` typo was fixed
2. the `if` was modified, the `elseif` was added:
  image

EntranceJew (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Fixed
 
+- Fixed the new view changes affecting some weapons by moving it to the GM:CalcView hook and adding a new break condition (by @TimGoll)
```
```suggestion
- Fixed the new view changes preventing weapons from modifying the playerview (by @TimGoll)
- Fixed the new view changes affecting non-player entities (by @TimGoll)
```

is `playerview` an accurate description?
My main point is that these are 2 related but different fixes
Feel free to change my wording

Histalek (changes requested):

CHANGELOG.md
TimGoll: yeah, that's fine!

Histalek (approved):

PR #1440, opened 2024-02-26
Bodysearch: Added safeguard to support undefined title or text

TimGoll:
image

There was this bugreport. I wasn't able to track down the root of this problem, but I'm pretty sure addons try to add things without a title via the hook (which is fine) breaking the rendering code. So I made sure that not providing a title won't break the rendering.


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Fixed
 
+- Fixed bodysearch entries not having a title no longer breaking the rendering
```
```suggestion
- Fixed bodysearch entries not having a title breaking the rendering
```

Histalek (approved):

Histalek (approved):

saibotk (approved):

PR #1441, opened 2024-02-26
ResetBones: Added player validity check

TimGoll:
Fixes this reported issue:
image

This issue happens because when late-joining (or any join that isn't the first I'd guess), the round state is updated before the player is ready. Meaning that `GetViewModel` does not yet exist. This change prevents this issue


gamemodes/terrortown/gamemode/client/cl_main.lua
```diff
@@ -456,8 +456,14 @@ local function RoundStateChange(o, n)
...
-        -- resets bone positions that fixes broken fingers on bad addons
-        weaponrenderer.ResetBonePositions(LocalPlayer():GetViewModel())
+        local client = LocalPlayer()
+
+        -- Resets bone positions that fixes broken fingers on bad addons.
+        -- When late-joining a server this function is executed before the local player
+        -- is completely set up. Therefore we safeguard this with this check.
+        if IsValid(client) and isfunction(client.GetViewModel) then
+            weaponrenderer.ResetBonePositions(client:GetViewModel())
+        end
```
Histalek, line 466: This would mean that a late-joining might have broken bone positions until the roundstate changes right?

It should fix the immediate issue, but we really need to find out how to prevent these late-join issues in general.

TimGoll: > This would mean that a late-joining might have broken bone positions until the roundstate changes right?

No as these broken bones only happen when bad addons break them. Strictly speaking this line isn't needed in TTT2, it just adds a protection for addons such as the harpoon that break bones.

EntranceJew: in addition to doing broken bone position checks, an enhancement would be to reset the alphas as well, sometimes the slide on the deagle or everything but the buttons on C4 get disappeared as well -- anything bad that could possibly happen to a viewmodel part probably will happen

resetting everything between rounds is pretty inconsequential as well

TimGoll: The buttons disappearing could be due to broken bones as well. It might sound illogical, but the C4 has bones too. The "classic" SWEP Construction Toolkit weapons hide the view model by modifying its bones so it is no longer visible. When this is done, this can lead to artifacts such as the buttons not showing on the C4

Histalek (approved):

saibotk (approved):

PR #1442, opened 2024-02-26
Weapon Renderer: Fixed rendering for weapons not based on the weapon base

### LINKED ISSUE #1432: Non-stock weapons (such as TFA weapons) do not render their base viewmodels

v0.13.0b (both workshop and git)

Non-stock weapons do not render their base viewmodels by default. This appears to be due to #1342 being **opt-in** to rendering the base viewmodel rather than **opt-out**. Any weapons which are not aware of this property will not render correctly. 

This can be fairly simply fixed by setting that property to `true` on any weapon which does not have it set.

## To reproduce

1. Install some custom weaponry
2. Enable it in TTT (through whatever means)
3. Pick up that weapon

## Expected behaviour

The viewmodels render.

Mostly the same as #1411.

## Screenshots

image

TimGoll:
Fixes #1432 

If a weapon is not based on the weapon base, `ShowDefaultWorldModel` is `nil`. In this case the viwe model should be rendered as well. We therefore check for an explicit `false` instead.


TimGoll:
oh - I did not fix the view model, but the world model. Will take a look at the view model as well.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Fixed
 
+- Fixed renering for weapons not based on the TTT2 base (like TFA) (by @TimGoll)
```
```suggestion
- Fixed rendering for weapons not based on the TTT2 base (e.g. TFA) (by @TimGoll)
```

saibotk (approved):
Looks good, but are there other assumptions about functions we recently added to the ttt_base where we also might need to apply fixes like this?

TimGoll:
> Looks good, but are there other assumptions about functions we recently added to the ttt_base where we also might need to apply fixes like this?

I thought about that as well, yes. But then again, should we really add a `if isfunction(wep.xyz)` for every call? seems a bit tedious. Let us release this and see if any issues occur.

ZenBre4ker (approved):

PR #1447, opened 2024-02-28
Entity: Rework Player Use Handling

ZenBre4ker:
The idea is to combine the different handling of the use key on client and server through the same workflow.
First on the server the gmod own entity use is permanently blocked by returning false in GM:PlayerUse.
So the client gathers information about the "+use"-key, which entity or traitor button is hit and sends that as net-message to the server.
The server then checks if that entity is in range and handles the use case scenario just as always.
Specs are supported, the PlayerUse-Hook is called before the ENTITY:Use-Hook is called and afterwards we check UseOverride, player_ragdolls and the weapon switch.

In this first draft we dont make use of the new ENTITY:ClientUse-Hook (name still up for discussion), generally the weapon switch delay is not fully implemented, ttt2 binding for use and the GM:KeyRelease hook are removed.

On this PR I want to build proper clientside entity use for UI (e.g. C4 and bodysearch), enable markervision for traitor buttons and with that also implement the possibility of remote use, which shall be handled on the same way as this Use workflow.


gamemodes/terrortown/gamemode/client/cl_keys.lua
```diff
@@ -66,19 +66,40 @@ function GM:PlayerBindPress(ply, bindName, pressed)
...
-
-            return true
-        elseif TBHUD:PlayerIsFocused() then
+        if TBHUD:PlayerIsFocused() then
             if ply:KeyDown(IN_WALK) then
                 -- Try to change the access to the button for your current role or team
                 return TBHUD:ToggleFocused(input.IsButtonDown(KEY_LSHIFT))
-            else
-                -- Try to use the button that is currently focused
-                return TBHUD:UseFocused()
```
ZenBre4ker, line 79: This needs to be reworked as well, as it internally just sends the entity to the server. Not sure if I implement it with this PR or wait for the remote PR to do it together.

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -382,7 +385,7 @@ end
 -- @ref https://wiki.facepunch.com/gmod/GM:PlayerUse
 -- @local
 function GM:PlayerUse(ply, ent)
-    return ply:IsTerror()
+    return overrideDoPlayerUse
```
TimGoll, line 388: what's the reason here for having this in a variable?

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -511,88 +514,76 @@ function GM:KeyPress(ply, key)
...
-
-            return true
+    if ent.player_ragdoll then
+        if not ply:KeyDown(IN_WALK) then
+            CORPSE.ShowSearch(ply, ent)
         else
-            -- do nothing, can't +use held objects
-            return true
+            ply:Spectate(OBS_MODE_IN_EYE)
+            ply:SpectateEntity(ent)
```
TimGoll, line 533: I wasn't aware that you could observe corpses, lol. Since you already simplify this system I think you could probably remove that feature? I don't see the point of it

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -511,88 +514,76 @@ function GM:KeyPress(ply, key)
...
         end
-    elseif tr.Entity.player_ragdoll then
-        CORPSE.ShowSearch(ply, tr.Entity, ply:KeyDown(IN_WALK) or ply:KeyDownLast(IN_WALK)) -- Body Corpse Search Identify TODO
-
-        return true
+    elseif ent:IsPlayer() and ent:IsActive() then
```
TimGoll, line 535: shouldn't this be `ent:IsTerror()`?

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -511,88 +514,76 @@ function GM:KeyPress(ply, key)
...
         return
     end
 
-    -- Do not allow the spectator to gather information if they're about to revive.
-    if ply:IsReviving() then
-        LANG.Msg(ply, "spec_about_to_revive", nil, MSG_MSTACK_WARN)
-
+    -- Check if the use interaction is possible
+    local distance = ply:GetShootPos():Distance(ent:GetPos())
+    if distance > 150 then
```
TimGoll, line 552: in the game we use a use distance of 100 so far, so I think it should be 100 here as well. Also we should probably put the use distance in a global variable, maybe even a convar?

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -511,88 +514,76 @@ function GM:KeyPress(ply, key)
...
-    elseif tr.Entity:IsPlayer() and tr.Entity:IsActive() then
-        ply:Spectate(ply.spec_mode or OBS_MODE_IN_EYE)
-        ply:SpectateEntity(tr.Entity)
-    else
-        PROPSPEC.Target(ply, tr.Entity)
+
+        ent:UseOverride(ply)
+    elseif ent.player_ragdoll then
+        CORPSE.ShowSearch(ply, ent, ply:KeyDown(IN_WALK) or ply:KeyDownLast(IN_WALK)) -- Body Corpse Search Identify TODO
+    elseif ent:IsWeapon() and ply:Alive() then
```
TimGoll, line 583: just FYI, even a spectator is "alive", therefore we use `IsTerror`

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -511,88 +514,76 @@ function GM:KeyPress(ply, key)
...
         end
-    elseif tr.Entity:IsPlayer() and tr.Entity:IsActive() then
-        ply:Spectate(ply.spec_mode or OBS_MODE_IN_EYE)
-        ply:SpectateEntity(tr.Entity)
-    else
-        PROPSPEC.Target(ply, tr.Entity)
+
+        ent:UseOverride(ply)
+    elseif ent.player_ragdoll then
+        CORPSE.ShowSearch(ply, ent, ply:KeyDown(IN_WALK) or ply:KeyDownLast(IN_WALK)) -- Body Corpse Search Identify TODO
```
TimGoll, line 582: you should handle this todo

gamemodes/terrortown/gamemode/client/cl_keys.lua
TimGoll: if you see a benefit to do this in another PR, then it is fine with me

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: Ok, I get it now. But I would do it differently. I'd add a third parameter to `GM:PlayerUse(ply, ent, overrideDoPlayerUse)`

And then when you hook.Run: `Hook.Run("PlayerUse", ply, ent, true)`

I think that should handle all cases the way you want without using a local global variable.

TimGoll (commented):
I really like this already

gamemodes/terrortown/gamemode/client/cl_keys.lua
ZenBre4ker: I think I leave this PR to focus on just combining all 'use' cases and not changing any behaviour for now.

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: I think I wont remove or change any functionality in this PR.

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: I believe this has the same constraints as before.

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: The convar sounds like a good idea, but the server check here should be more than 100.

The client already checks that it's only in a distance of 100, but the server might be in a slightly different position and especially when the client aims at an outer position of an ent, but the server only compares to the center Position this can lead to the use being denied.
Not sure, if an additional trace should be done when range is over a hundred, but I thought this can make it more robust for the client and not lead to any behavior that can be abused a lot.

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: Thats not mine 😅

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: Isterror is checked some lines above.

ZenBre4ker:
This is working as expected. Some general comments though:
- I didnt try to change any behavior. As far as Alive, IsTerror, IsPlayer Checks go, they should be the same.
- Every Use code is still only executed serverside, but I moved the trace to the client and added an ENTITY:ClientUse to finally be able to use Entities clientside for UI
- I check the range of the weapon to the player clientside with 100 as usual, but serverside I give a little extra and check for 150 to make sure that a desync doesnt affect it too much. A problem that could occur is the following:
grafik
The eyetrace on the client sees a shorter distance, than the safety check distance the server does.

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: Nevermind, `IsActive` is the same as `IsTerror`m but it also checks the round state:

https://github.com/TTT-2/TTT2/blob/ded4f19fa53f737c223c62f23f05bd1fc971d933/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua#L559-L561

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: But what happens if you look at a corpse? This already has a size of 160 I think, breaking this implementation because the center has a distance of 80 from the corners. Or even more if it is diagonal. I think you should use the EyeTrace on the server as well. Maybe add a 10% margin to it or so.

EyeTrace is also cached, so it shouldn't add any performance penalty.

However, now that I think about it: This would also break if the model is in a slightly different position on the server compared to the client. So maybe the entity shouldn't be checked at all on the server?

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: I'm not sure why this is added there at all. If a player `IsTerror` he is also always alive, because IsTerror is a subset of Alive. So I guess you could remove that part completely.

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: I think our policy is always to remove ToDos and put them in issues. But i think @Histalek is the expert with that

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ### Added
 
+- Added hook ENTITY:ClientUse(), which is triggered clientside if an entity is used
+  - Return true if only clientside should be used
```
TimGoll, line 10: I'm not sure I like that. Is there a problem if ClientUse is called and after that also Use on the server? Because if you did not add any code to the Use function on the server, it wouldn't matter anyway

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -511,88 +513,89 @@ function GM:KeyPress(ply, key)
...
-    local tr = util.QuickTrace(ply:GetShootPos(), ply:GetAimVector() * 128, ply)
+    ---
+    -- Enable addons to allow handling PlayerUse
+    -- Otherwise default to old IsTerror Check
+    -- @realm server
+    -- stylua: ignore
+    if hook.Run("PlayerUse", ply, ent, ply:IsTerror()) then
+        ent:Use(ply, ply)
+    end
+    overrideDoPlayerUse = false
```
TimGoll, line 574: I think that line causes a bug. Because it now declared a global name with the same name as the parameter in the hook. Meaning that the UseHook will now probably always return false when not called by the hook.Run above.

But I think this is only a leftover line anyway?

TimGoll (commented):
I'm almost happy with it. I like the changes you did here

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: Oops, good catch. Could have been a nasty bug to find. Not sure what would happen if the global variable is set and the parameter is nil.

CHANGELOG.md
ZenBre4ker: This only stops it from being sent to the server. We just omit a net message like that.
As it is a hook, you generally dont call it yourself. You just write code clientside that is executed on use. If you do not return true, after its called clientside, its automatically called serverside. Its meant to prevent server usage in UI only cases.

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: The entity has to be checked on the server, otherwise you enable cheats like instant corpse reveal even if you are not close to it.

Instead of a trace you could maybe do a bounding box check, gotta look into that.
The eyetrace of course could be done, but it will only help in a few cases, where you are that far away, taht the check fails, but you are still close enough and looking at the server entity correctly. As this check was bad before, it would only help a little.

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: Mhh, I can remove it, but I have no idea why its there in the first place, so I am clueless what I should put in the issue.

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: IsTerror is just a team check, while IsAlive is about players living. So while it might not be possible in our gameloop to be alive and not from team terror, the checks themselves dont make sure of that. So I think it should stay for safety.

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: Maybe this:
https://wiki.facepunch.com/gmod/util.IsBoxIntersectingSphere

gamemodes/terrortown/gamemode/server/sv_player.lua
Histalek:
```
$ git blame -L 552,552 gamemodes/terrortown/gamemode/server/sv_player.lua
ff244a6bd (Alf Müller 2019-10-29 03:27:54 +0100 552)         CORPSE.ShowSearch(ply, tr.Entity, ply:KeyDown(IN_WALK) or ply:KeyDownLast(IN_WALK)) -- Body Corpse Search Identify TODO
```

remove it tbh

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: This here is perfect:
https://wiki.facepunch.com/gmod/Entity:BoundingRadius

You can just add it to the distance and it scales for all entities.
Flat/Boxed ragdoll => 45/40
C4 => 13
This will add the necessary distance to be safe.

CHANGELOG.md
TimGoll: Hmm, I think I'm ok with that then. But this definitely needs a second opinion, so I'll leave this thread open so the other reviewer cag give their opinion on that as well

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -511,88 +513,88 @@ function GM:KeyPress(ply, key)
...
-        ply:Spectate(ply.spec_mode or OBS_MODE_IN_EYE)
-        ply:SpectateEntity(tr.Entity)
-    else
-        PROPSPEC.Target(ply, tr.Entity)
+
+        ent:UseOverride(ply)
+    elseif ent.player_ragdoll then
+        CORPSE.ShowSearch(ply, ent, ply:KeyDown(IN_WALK) or ply:KeyDownLast(IN_WALK)) -- Body Corpse Search Identify TODO
+    elseif ent:IsWeapon() and ply:Alive() then
+        ply:SafePickupWeapon(ent, false, true, true, nil) -- force pickup; drop blocking weapon, autoselect is set automatically
```
Histalek, line 595:
```suggestion
        ply:SafePickupWeapon(ent, false, true, true, nil)
```

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: I wrote that code in question here and I'm pretty sure I did it that way because I didn't know anything about IsTerror. If IsTerror and Alive diverge, the whole gamemode is broken anyway

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: Perfect! This looks good to me then!

TimGoll (approved):
I really like it, thank you for reworking it! I'll keep the two remaining threads open, maybe someone else will chime in as well. But other than that this LGTM.

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: Ahh well, `Alive` doesn't make any sense here. Like I said a spectator is alive is as well and they **shouldn't** be able to pick up weapons! This is not a problem because of the previously mentioned `IsTerror` check, but that just underlines that the current `Alive` check is pretty pointless.

So while this is not really part of your PR, I'd vote in favor of removing that check to clean up the code and to avoid future confusion.

gamemodes/terrortown/gamemode/client/cl_weapon_pickup.lua
```diff
@@ -45,13 +11,3 @@ net.Receive("ttt2_switch_weapon_update_cache", function()
...
-
--- register a binding for the weapon switch, the default should be the use key
-bind.Register(
-    "ttt2_weaponswitch",
-    AttemptWeaponSwitch,
-    nil,
-    "header_bindings_ttt2",
-    "label_bind_weaponswitch",
-    input.GetKeyCode(input.LookupBinding("+use") or KEY_E)
-)
```
Histalek, line 57: targetid.lua is still using this

Histalek (changes requested):

gamemodes/terrortown/gamemode/server/sv_player.lua
ZenBre4ker: I can do this. I dont know anything about those checks.
I see we have those terror and spec checks, that differ.
But in many parts of our codebase alive is always checked as well.
Might be, that is was just not knowing better. But as I dont know better, I would leave it in unless you are all sure its safe to remove. I dont really want to test it....

gamemodes/terrortown/gamemode/client/cl_weapon_pickup.lua
TimGoll: good catch, targetID would have to be updated as well then. Also the language file needs updating!

ZenBre4ker: grafik
Is it?

ZenBre4ker: Oh yeah I can see where...

TimGoll: yes, looking at entities always used "+use", but look at a weapon which uses the pickup binding

TimGoll: > Oh yeah I can see where... grafik

well **THIS** is a bug you just accidentally fixed with your rework. These two bindings should be the same :D

ZenBre4ker: grafik
Yeah just noticed that as well

TimGoll: should be a a easy change though as all external addons already use "+use", only our builtin weapon pickup uses something different

TimGoll:
Change looks good to me

gamemodes/terrortown/gamemode/server/sv_player.lua
TimGoll: When I started with TTT2 I used the alive checks as well as I thought that was the way to go, but it isn't. Probably something we should clean up as well.

Histalek (approved):

PR #1449, opened 2024-03-03
ci: Add automated release workflow

Histalek:
This is a first draft (with an absolutely horrendous implementation) of automating our release steps after a PR with a `release/*` label has been merged to the `release` branch.

Outstanding TODOs:

- I 100% need to implement this with some proper file templating and double check all if-conditions. This is a garbled mess (but probably somehow functional ...)

- add the actual release tagging to this as well

- document the new release process

This is a highly WIP PR and mostly intended to get some more eyes on this.


.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+  contents: write
+  pull-requests: write
+
+env:
+  PANDOC_VERSION: 3.1.1.0
+
+on:
+  pull_request:
+    types:
+      - closed
```
ZenBre4ker, line 13: Is there no merged type, do we actually have to check it later in the action?

.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+    branches:
+      - release
+
+jobs:
+  release-pr:
+    # guard against running on forks
+    if: github.repository == 'TTT-2/TTT2'
```
ZenBre4ker, line 20: Is there no "this" repository? Do you need to hardcode the name?

.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+
+      - name: Check if PR was merged
+        uses: actions-ecosystem/action-release-label@v1
+        id: release-label
+        if: ${{ github.event.pull_request.merged == true }}
+
+      - name: Get latest tag if a release label exists on last merged PR
+        uses: actions-ecosystem/action-get-latest-tag@v1
+        id: get-latest-tag
+        if: ${{ steps.release-label.outputs.level != null }}
```
ZenBre4ker, line 36: Is an action skipped or does the whole workflow fail on a failed if condition?
If it does fail, I wonder why we recheck it in other actions.

.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+      - name: Generate app token
+        uses: actions/create-github-app-token@v1
+        id: generate-token
+        if: ${{ steps.release-label.outputs.level != null }}
+        with:
+          app-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+
+      - name: Add release heading (version, tag link and date) to CHANGELOG.md
+        run: 'sed -i "/^## Unreleased$/a\\\n## [v${TTT2_VERSION}](https://github.com/TTT-2/TTT2/tree/v${TTT2_VERSION}) ($(date +%Y-%m-%d))" CHANGELOG.md'
```
ZenBre4ker, line 55: Probably would be nice in a proper action with all other file changes.

.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+          TTT2_VERSION: ${{ steps.bump-semver.outputs.new_version }}
+
+      - name: Generate html from changelog for cl_changes.lua
+        uses: "docker://docker.io/pandoc/core:${PANDOC_VERSION}"
+        if: ${{ steps.release-label.outputs.level != null }}
+        with:
+          args: >-
+            --output=output.html
+            --read=markdown_strict
+            <(sed -e '1,/^## \[/d' -e '/^## \[/,$d' CHANGELOG.md)
```
ZenBre4ker, line 67: i guess we try to do it without html

ZenBre4ker (commented):
General workflow looks good. No idea about syntax or the actions used, but beside the manual file editing it looks good.

We could probably add a steam account in secrets and also auto publish on merge of a release pr?

.github/workflows/release.yml
Histalek: jup ...
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

.github/workflows/release.yml
Histalek: "this" repository would mean the forked repo on a fork, which is exactly what i want to prevent. (besides i don't think that something like that even exists)

.github/workflows/release.yml
Histalek: only this specific step is skipped.

i'm not sure if i want all of these steps in a single job like it is now but it is the way i went with for now.
Additionally i need to double check all if-conditions anyway

.github/workflows/release.yml
Histalek: the html itself is not the hard part, mixing in lua and getting proper formatting is.
i would really like to have some kind of templating like jinja2 or something, but i have yet to find a useable action for it.

.github/workflows/release.yml
ZenBre4ker: Yes, an own program. Doesnt look clean like this.

Histalek:
I need to go back to the drawing board with this one so i'll close this for now

PR #1452, opened 2024-03-04
AFK timer / incomplete player join bookkeeping

EntranceJew:
- add plymeta:IsFullySignedOn
 - overzealous AFK reigned in

now the player will not get booted the very moment they receive a role while joining or between rounds, as long as the idle_limit is greater than 5, the hardcoded check interval

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,13 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Added hook ENTITY:ClientUse(), which is triggered clientside if an entity is used
   - Return true to prevent also using this on the server for clientside only usecases
+- Added `plymeta:IsFullySignedOn()` to allow excluding players that have not gotten control yet (by @EntranceJew)
 
 ### Changed
 
 ### Fixed
+
+- Fixed the AFK timer accumulating while player not fully joined (by @EntranceJew)
 
 ## [v0.13.1b](https://github.com/TTT-2/TTT2/tree/v0.13.1b) (2024-02-27)
 
--- a/gamemodes/terrortown/gamemode/client/cl_main.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_main.lua
@@ -1000,6 +1000,7 @@ local idle = {
     mx = 0,
     my = 0,
     t = 0,
+    last_check = 0,
 }
 
 ---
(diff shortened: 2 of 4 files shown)


EntranceJew:
FWIW I think this should immediately be applied pretty much everywhere, but especially to role distribution, so that players do not feel the inclination to beat down players that have not even fully connected but have a role.

TimGoll (approved):
Huh, that's neat!

I'm wondering I implemented something similar a few years ago, called `plymeta:IsReady`. Should this be replaced by your implementation?

https://github.com/TTT-2/TTT2/blob/4c05a87956b919a159a833b42ac49716a309339f/gamemodes/terrortown/gamemode/client/cl_player_ext.lua#L240-L266

Histalek (approved):
Definitely a nice addition.

I guess we would need to see if `SIGNONSTATE_FULL` is always the state we need in different cases. (maybe `_PRESPAWN` or `_SPAWN` might be better suited for some things)
For the idle timer i'm fine with either way though.

TimGoll:
From my POV this is ready to merge, but I want @EntranceJew's feedback on my `ply:IsReady` question first so this doesn't get lost. The code in this PR is good imho

EntranceJew:
> Huh, that's neat!
> 
> I'm wondering I implemented something similar a few years ago, called `plymeta:IsReady`. Should this be replaced by your implementation?
> 
> https://github.com/TTT-2/TTT2/blob/4c05a87956b919a159a833b42ac49716a309339f/gamemodes/terrortown/gamemode/client/cl_player_ext.lua#L240-L266

one major benefit of my implementation is that it allows clients and the server to know which players are pending connection as opposed to just the client, so I would say yes for this case, even if the first SetupMove is closer to the first rendered visual frame.

TimGoll:
> > Huh, that's neat!
> > I'm wondering I implemented something similar a few years ago, called `plymeta:IsReady`. Should this be replaced by your implementation?
> > https://github.com/TTT-2/TTT2/blob/4c05a87956b919a159a833b42ac49716a309339f/gamemodes/terrortown/gamemode/client/cl_player_ext.lua#L240-L266
> 
> one major benefit of my implementation is that it allows clients and the server to know which players are pending connection as opposed to just the client, so I would say yes for this case, even if the first SetupMove is closer to the first rendered visual frame.

My implementation is also shared. But I don't care about the fact if the first frame is rendered, I only needed a way to determine if the player is ready to receive net messages. So I guess we should switch over to your system!

PR #1454, opened 2024-03-04
bring down loose content files

EntranceJew:
i keep losing track of these

this was made a lot easier to do ever since we relocated our own content folders so thanks for doing that


saibotk (approved):

TimGoll:
these are unused:
image

this and its traitor and inno counterpart as well:
image

I'd vote to remove those and then I'm fine with this PR

TimGoll (approved):

PR #1458, opened 2024-03-06
ENTITY: New Use workflow

ZenBre4ker:
This is implementing the new use workflow and gets rid of UseOverride.
Also C4 is now the first to properly use ClientUse to display a UI without routing it over the server.
As sidenote I fixed the Visualizer so that it can be picked up by the originator. This was only a problem if a non policing role should be able to buy it. Thats more in line with our general shop setup possibilities.


gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
```diff
@@ -41,6 +43,38 @@ function ENT:SetupDataTables()
...
+    function ENT:ClientUse()
+        local client = LocalPlayer()
+        if not IsValid(client) or not client:IsTerror() or not self.pickupWeaponClass then
+            return true
+        end
+
+        if not self:PlayerCanPickupWeapon(client) then
+            LANG.Msg(client, "pickup_fail", nil, MSG_MSTACK_WARN)
+
+            self:EmitSound(soundDeny)
```
TimGoll, line 71: there is one thing I'm not so happy here: Previously everyone was able to hear the pickup and deny sound. Now everyone is able to hear the pickup sound, but the deny sound is only audible to the player hittin use. If you could broadcast this easily, it would be really neat.

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
```diff
@@ -41,6 +43,38 @@ function ENT:SetupDataTables()
...
+
+if CLIENT then
+    ---
+    -- Hook that is called if a player uses their use key while focusing on the entity.
+    -- Implement this to predict early if entity can be picked up
+    -- @return bool True to prevent pickup
+    -- @realm client
+    function ENT:ClientUse()
+        local client = LocalPlayer()
+        if not IsValid(client) or not client:IsTerror() or not self.pickupWeaponClass then
```
TimGoll, line 64:
```suggestion
        local client = LocalPlayer()

        if not IsValid(client) or not client:IsTerror() or not self.pickupWeaponClass then
```

gamemodes/terrortown/entities/entities/ttt_cse_proj.lua
```diff
@@ -233,14 +229,13 @@ if CLIENT then
...
+        if ent:GetOriginator() == client then
             tData:SetSubtitle(ParT("target_pickup", { usekey = Key("+use", "USE") }))
-            tData:SetKeyBinding("+use")
         else
-            tData:SetSubtitle(TryT("vis_no_pickup"))
-            tData:AddIcon(roles.DETECTIVE.iconMaterial)
+            tData:SetSubtitle(TryT("entity_pickup_owner_only"))
         end
 
+        tData:SetKeyBinding("+use")
```
TimGoll, line 238: now everbody can see the keybinding, could you please change it back so that only people that are able to pick it up see the binding?

gamemodes/terrortown/entities/entities/ttt_health_station.lua
```diff
@@ -200,6 +199,18 @@ else
...
     }
 
+    ---
+    -- Hook that is called if a player uses their use key while focusing on the entity.
+    -- Early check if client can use the health station
+    -- @return bool True to prevent pickup
+    -- @realm client
+    function ENT:ClientUse()
+        local client = LocalPlayer()
+        if not IsValid(client) or not client:IsPlayer() or not client:IsActive() then
```
TimGoll, line 209:
```suggestion
        local client = LocalPlayer()

        if not IsValid(client) or not client:IsPlayer() or not client:IsActive() then
```

TimGoll (changes requested):

gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
ZenBre4ker: While it sounds lazy, I actually think those sounds shouldnt be broadcasted at all.

Before this rework use was always serverside and therefore those sounds were broadcasted, but I think those are clientside only things. I'd rather remove the sounds completely from being broadcasted.

But what do you think @Histalek @saibotk?

Histalek: tbh i consider this behaviour change a bugfix.

IMO the deny sound is a audible aid for the player trying to pick something up, giving them feedback that the attempt was registered but they are denied. Nothing more.

Making this be heard globally is somewhat unexpected and also makes this part of the deduction process which i think is a mistake.

TimGoll: I see your point. While working on the sound stuff yesterday I added a client side sound to body search. This made me wonder if it should be audible to everyone. Hence the question here.

So to be in line with your reasoning, the body search sound should be broadcasted as well I think

TimGoll: But for example picking up a weapon is a feedback sound as well. But this is handled by the engine and the sound is audible to everyone.

I just want to get consistency here

Histalek: i don't think there is any consistency to be gotten here. There are many ways to reason about different sounds being globally heard or not.
e.g. weapon pickup is not a feedback sound at all (see CS)

my main point is that something like this shouldn't give any information away about your role/team etc. and in this case forcefully so. (there is no counterplay: if it is your equipment you can't generate the deny sound and you're cooked)

Aside: having bodysearch sounds be audible for all doesn't sound bad.

TimGoll: I was mostly asking for us to define a consistent behavior with the limitations we are facing. Thanks

EntranceJew: FWIW i don't think I've ever heard another player's Use fail sound if it wasn't a scripted map entity button specifically designed to make a "deny" noise

TimGoll (approved):

TimGoll:
@saibotk @Histalek Do you want to have anything changed here?

Histalek:
> @saibotk @Histalek Do you want to have anything changed here?

Nothing particular, no

i'd like to playtest this a bit but that's not blocking the merge IMO

TimGoll:
> > @saibotk @Histalek Do you want to have anything changed here?
> 
> Nothing particular, no
> 
> i'd like to playtest this a bit but that's not blocking the merge IMO

I tested it for a few minutes and noticed no issues. So I'll go ahead and merge it, we have to playtest the whole update anyways

PR #1459, opened 2024-03-06
MarkerVision: Added subtitle option

TimGoll:
Added the possibility to add a subtitle to marker vision
image

Also targetID is now hidden, when a marker vision element is focused as it is quite confusing to see two mutually exclusive use interactions on screen at the same time.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_marker_vision_data.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_marker_vision_data.lua
@@ -134,6 +134,20 @@ function MARKER_VISION_DATA:SetTitle(text, color, inline_icons)
 end
 
 ---
+-- Sets the subtitle of the specific targetID element
+-- @param[default=""] string text The text that should be displayed
+-- @param[default=Color(210, 210, 210, 255)] Color color The color of the line
+-- @param[opt] table inline_icons A table of materials that should be rendered in front of the text
+-- @realm client
+function MARKER_VISION_DATA:SetSubtitle(text, color, inline_icons)
+    self.params.displayInfo.subtitle = {
+        text = text or "",
+        color = IsColor(color) and color or COLOR_LLGRAY,
+        icons = inline_icons or {},
+    }
+end
+
+---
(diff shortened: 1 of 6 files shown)


gamemodes/terrortown/gamemode/client/cl_marker_vision_data.lua
```diff
@@ -133,6 +133,20 @@ function MARKER_VISION_DATA:SetTitle(text, color, inline_icons)
     }
 end
 
+---
+-- Sets the subtitle of the specific targetID element
```
saibotk, line 137:
```suggestion
-- Sets the subtitle of the specific radar vision element
```

radar vision is what its called in all other functions

gamemodes/terrortown/gamemode/client/cl_targetid.lua
```diff
@@ -308,6 +308,11 @@ function GM:HUDDrawTargetID()
         return
     end
 
+    -- if the entity also is a focused marker vision element, then targetID should be hidden
+    if ent == markerVision.GetFocussedEntity() then
```
saibotk, line 312:
```suggestion
    if ent == markerVision.GetFocusedEntity() then
```

Seems like this slipped through, optional if you want i'd appreciate if we can fix this one too

saibotk (changes requested):

gamemodes/terrortown/gamemode/client/cl_marker_vision_data.lua
TimGoll: yeah, copy pasta, best pasta, oof

TimGoll:
done @saibotk

saibotk (approved):

Histalek (approved):
LGTM

TimGoll:
@ZenBre4ker We probably have to check if the "use" button text is hown to the wrong people. Right now it doesn't check if the player can actually use the key, I think. This is probably a bug?

ZenBre4ker:
Yeah @TimGoll you're right. But I was thinking we just might want to check if the function remoteuse is available?

PR #1460, opened 2024-03-06
Sounds: Added multiple UI sounds

TimGoll:
After testing #1437 I notices that it feels really weird to not have any kind of feedback when interacting with entities. When picking up entities, there already is a sound, but not when opening UIs. Therefore I added `sound.ConditionalPlay` that plays a sound if the sound type is enabled in the settings.

I also extended this to body search and UI buttons.

See this demo video: https://www.youtube.com/watch?v=c6ixkuFSYfw

Note: I also pulled in the changes from #1351 and added them to the same system. They are disabled by default.

Menu to disable sounds by type:
image


lua/terrortown/menus/gamemode/gameplay/voiceandvolume.lua
Histalek, line 1: i think the gameplay sounds should go into another submenu(?) in the navbar `Sounds` or maybe `Sounds & Volume`.
This makes it a lot more obvious as to where to find these settings and also saves a potential scrollbar and therefore mouse actions needed.

lua/ttt2/extensions/sound.lua
```diff
@@ -0,0 +1,65 @@
...
+    end
+end
+
+---
+-- Plays the chat "tick" sound.
+-- @note This is overwritten so that this function respects the TTT2 convar.
+-- @realm client
+function chat.PlaySound()
+    sound.ConditionalPlay("HudChat.Message", SOUND_TYPE_MESSAGE)
+end
```
Histalek, line 65: Maybe we want to put stuff like this into some kind of `gmod_overrides.lua` so this would be more obvious? At least for stuff we can't really put anywhere else?

Histalek (changes requested):

lua/terrortown/menus/gamemode/gameplay/voiceandvolume.lua
TimGoll: Makes sense

lua/ttt2/extensions/sound.lua
TimGoll: I could, yes. I could also put it in `chat.lua`

Histalek: ah well yes, that would make straightforward as well

TimGoll:
all things should be done @Histalek

Histalek (approved):

lua/terrortown/lang/en.lua
```diff
@@ -2191,3 +2191,20 @@ L.throw_no_room = "You have no space here to throw this device"
...
 L.use_entity = "Press [{usekey}] to use"
+
+-- 2024-03.06
+L.submenu_gameplay_sounds_title = "Client-Sounds"
+
+L.header_sounds_settings = "UI Sound Settings"
+
+L.help_enable_sound_interact = "Interaction sounds are those sounds that are played when opening an UI. Such a sound is played for example when interacting with the radio marker."
+L.help_enable_sound_buttons = "Button sounds are clicky sounds that are played when clicking a button."
+L.help_enable_sound_message = "Message or notification sounds are played for chat and mStack messages. They can be quite obnoxious."
```
saibotk, line 2202: Players do not know what a "mStack" is

saibotk (changes requested):

lua/terrortown/lang/en.lua
TimGoll: What should I use instead then? Notifications?

saibotk: yes

saibotk (approved):

PR #1461, opened 2024-03-07
Bugfix: Fixed issue with +use for traitors

TW1STaL1CKY:
This PR fixes traitors sometimes not being able to use entities (doors, weapons, etc), and spectators not being able to use entities at all.

**Detail:**
Traitors were denied pressing Use on all entities except traitor buttons if they ever hovered their crosshair over a traitor button, as TBHUD:PlayerIsFocused would report they were always focused on one for the rest of the round.
_This has been fixed by adding a TBHUD.focus_stick check to TBHUD:PlayerIsFocused._

Spectators were permanently denied pressing Use. This was because the "pressed" parameter in GM:PlayerBindPress always returned false for spectators while the +use check inside it wanted "pressed" to be true in order to pass.
_This has been fixed by removing the "pressed" check for +use in GM:PlayerBindPress. The "pressed" parameter seems to always return true only for non-spectators anyway, but if there's an issue let me know._

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_keys.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_keys.lua
@@ -65,7 +65,7 @@ function GM:PlayerBindPress(ply, bindName, pressed)
 
             return true
         end
-    elseif bindName == "+use" then
+    elseif bindName == "+use" and pressed then
         -- Do old traitor button check
         if TBHUD:PlayerIsFocused() then
             if ply:KeyDown(IN_WALK) then
--- a/gamemodes/terrortown/gamemode/client/cl_keys.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_keys.lua
@@ -65,7 +65,7 @@ function GM:PlayerBindPress(ply, bindName, pressed)
 
             return true
         end
-    elseif bindName == "+use" and pressed then
+    elseif bindName == "+use" then
         -- Do old traitor button check
         if TBHUD:PlayerIsFocused() then
             if ply:KeyDown(IN_WALK) then
--- a/gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
@@ -77,7 +77,6 @@ function TBHUD:PlayerIsFocused()
         and ply:IsActive()
         and self.focus_but
         and (self.focus_but.access or self.focus_but.admin)
-        and self.focus_stick
         and self.focus_stick >= CurTime()
         and IsValid(self.focus_but.ent)
 end
--- a/gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_tbuttons.lua
@@ -77,6 +77,8 @@ function TBHUD:PlayerIsFocused()
         and ply:IsActive()
         and self.focus_but
         and (self.focus_but.access or self.focus_but.admin)
+        and self.focus_stick
+        and self.focus_stick >= CurTime()
         and IsValid(self.focus_but.ent)
 end
 


ZenBre4ker (commented):
Not a fan of those fixes. Going to have a look at it tonight.
Just wondering why, because I tested spectator and had no such problems.
Same goes for the traitor button, its weird that it is doing it like that.

Could you please create issues instead for this?

TimGoll:
So you're saying that spectators were unable to press use ok corpses or so? Because if this was your problem, there is something on your addon selection causing this because it works flawless on my 200 addon test server.

Spectators should be unable to interact with doors, weapons etc though.

EntranceJew:
> So you're saying that spectators were unable to press use on corpses or so? Because if this was your problem, there is something on your addon selection causing this because it works flawless on my 200 addon test server.
> 
> Spectators should be unable to interact with doors, weapons etc though.

what about prop spectator

ZenBre4ker:
> > So you're saying that spectators were unable to press use on corpses or so? Because if this was your problem, there is something on your addon selection causing this because it works flawless on my 200 addon test server.
> > Spectators should be unable to interact with doors, weapons etc though.
> 
> what about prop spectator

Tested it rn and it works as well. So spectator works as always, you can watch others by pressing use, you can look at corpses and you can control props.

ZenBre4ker:
But you are right about the traitor button :)

ZenBre4ker (changes requested):
I'd accept the focus_stick change. Traitor buttons have weird behaviour, which Im about to change anyways, so its fine like that for now.

TW1STaL1CKY:
I did do debug.traces to see where PlayerUse was being called from, and only TTT2PlayerUseEntity came up so there was definitely no addon conflicts there.
I'll do some with PlayerBindPress to confirm what, if anything, is causing the spectator bug.

I'll do the requested changes for you so you can merge

ZenBre4ker (approved):

ZenBre4ker:
> I did do debug.traces to see where PlayerUse was being called from, and only TTT2PlayerUseEntity came up so there was definitely no addon conflicts there. I'll do some with PlayerBindPress to confirm what, if anything, is causing the spectator bug.
> 
> I'll do the requested changes for you so you can merge

Be aware that gmod itself calls it as hook. So things that intervene with the hook can be a problem.

TW1STaL1CKY:
I've found that the spectator issue was caused by the TTT2 Spectator Deathmatch addon... [It has its own PlayerBindPress hook that tries to replicate the base TTT2 one with a few tweaks](https://github.com/NickCloudAT/TTT2_Spectator_Deathmatch/blob/e29190191e4f9ee3f71d611375ba39925ee6e250/lua/cl_spectator_deathmatch.lua#L282).
Sorry to waste your time on that one! Anyways thank you for approving 😊

ZenBre4ker:
> I've found that the spectator issue was caused by the TTT2 Spectator Deathmatch addon... [It has its own PlayerBindPress hook that tries to replicate the base TTT2 one with a few tweaks](https://github.com/NickCloudAT/TTT2_Spectator_Deathmatch/blob/e29190191e4f9ee3f71d611375ba39925ee6e250/lua/cl_spectator_deathmatch.lua#L282). Sorry to waste your time on that one! Anyways thank you for approving 😊

Thank you for taking your time to look into the issue. Much appreciated <3

saibotk:
For someone not familiar with the focus_stick property. Could someone quickly explain when and why it is set?

Also i'd say we add a changelog for this fix.

TimGoll:
> For someone not familiar with the focus_stick property. Could someone quickly explain when and why it is set?
> 
> Also i'd say we add a changelog for this fix.

I looked through the codebase and I think the focus sticks for `0.1` seconds even after looking somewhere else.

(This is something I considered adding to the whole targetID, but I'm not yet sure. It would make it easier to see the name of a player at a glance. This might make it too easy, I dunno.)

EntranceJew:
> For someone not familiar with the focus_stick property. Could someone quickly explain when and why it is set?
> 
> Also i'd say we add a changelog for this fix.

it may be used for the ability to do traitor/suspicious callouts without having the targetID active

ZenBre4ker:
I did out the changelog skip there because it's just fix to my newest rework, so this bug didnt exist on the latest release.

The focus_stick stores the last time a button was looked on + 0.1 seconds. Because of this sticky behaviour focused buttons are not cleared, thats why the bug exists.
Im not a fan of this sticky behaviour and will remove it with the port to markervision. In my opinion, targetid should only ever be shown for stuff this actually focused. This can lead to weird behaviour and has to be properly handled.

saibotk (approved):

PR #1466, opened 2024-03-08
EquipmentMenu: Fix GetClass error

TimGoll:
This error was reported by @mexikoedi on Discord:

image

I fixed this by correctly using `WEPS.GetClass` here instead of `equipment:GetClass`. While doing this I noticed that items have no equivalent of `GetClass` and `IsValid` so I added that to the base class. These changes are not strictly needed for this fix, but imho they are useful for consistency.

### DIFF
--- a/lua/terrortown/entities/items/item_base/shared.lua
+++ b/lua/terrortown/entities/items/item_base/shared.lua
@@ -100,9 +100,32 @@ function ITEM:Bought(ply) end
 function ITEM:Initialize() end
 
 ---
--- useable, but do not modify this!
--- @return boolean whether this @{ITEM} is an equipment
+-- Checks whether this items is a valid equipment item.
+-- @note Useable, but do not modify or overwrite this!
+-- @return boolean Return whether this @{ITEM} is an equipment
 -- @realm shared
 function ITEM:IsEquipment()
     return WEPS.IsEquipment(self)
+end
+
+---
+-- Returns the classname of an item. This is often the name of the Lua file or folder containing the files for the item.
+-- @return string The item's class name
+-- @realm shared
+function ITEM:GetClass()
+    return self.ClassName
+end
+
+---
+-- Checks if the item is valid. Returns true if valid.
+-- @return boolean Returns true if item is valid
+-- @realm shared
+function ITEM:IsValid()
+    -- this is probably always valid if there is no custom weird stuff done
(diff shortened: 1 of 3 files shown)


Histalek (approved):
LGTM

lua/terrortown/menus/gamemode/equipment/base_equipment.lua
```diff
@@ -179,9 +179,11 @@ function CLGAMEMODESUBMENU:Populate(parent)
         database = DatabaseElement(accessName, itemName, "damageScaling"),
     })
 
+    local equipmentClass = WEPS.GetClass(equipment)
```
saibotk, line 182: Isnt this only intended for weapons?

I think this should not be used for items too

lua/terrortown/entities/items/item_base/shared.lua
```diff
@@ -100,9 +100,32 @@ function ITEM:Bought(ply) end
...
 function ITEM:IsEquipment()
     return WEPS.IsEquipment(self)
 end
+
+---
+-- Returns the classname of an item. This is often the name of the Lua file or folder containing the files for the item.
+-- @return string The item's class name
+-- @realm shared
+function ITEM:GetClass()
+    return self.ClassName
```
saibotk, line 116: This is very confusing, what is the difference between the class name here and the class name returned by the weps function?

saibotk (changes requested):

lua/terrortown/entities/items/item_base/shared.lua
TimGoll: It's identical. But entities (like weapons) have a function called `wep:GetClass()`. For some reason items didn't have that function and only a variable named `.ClassName` (which I think also exists on entites). In our code we mostly handle items and weapons as similar things, so imho it should be good to have the same basic functions.

The weps function basically checks if the equipment is a table and then returns `.ClassName` or returns `:GetClass()` it the equipment `:IsValid()`

lua/terrortown/menus/gamemode/equipment/base_equipment.lua
TimGoll: we always use this for both in our code base and it works as well. But yes, the name is misleading

saibotk (approved):

PR #1468, opened 2024-03-08
ViewModels: Fix error texture bug

TimGoll:
Fixes this issue reported by @EntranceJew and @mexikoedi:
image
image

The bug seen in the image happens because some weapons set a custom view model texture. Our view model hider resets this custom texture by calling `viewModel:SetMaterial("")`. I did not anticipate that weapons might set custom view models, so I wasn't aware that this is an issue.
image
image

We *have to* reset the view model texture because the view model does not change on weapon change. So if weapon A has a hidden view model and we apply the debug texture, the same debug texture is applied to weapon B making it invisible as well. The first solution of applying the texture reset if it should not be hidden caused the bug mentioned above.

The solution therefore is to set a trigger, if the debug texture is set before the view model is rendered and reset the material after the view model is rendered. If a player then switches to weapon B, the view model works as expected without a continuous material reset.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Fixed
 
 - Fixed the AFK timer accumulating while player not fully joined (by @EntranceJew)
-- Fixed view models of some weapons having an error texture
+- Fixed weapons which set a custom view model texture having an error texture
 - Fixed the equipment menu throwing errors when clicking on some items
 
 ### Removed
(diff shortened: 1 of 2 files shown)


saibotk (approved):

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Fixed
 
 - Fixed the AFK timer accumulating while player not fully joined (by @EntranceJew)
+- Fixed view models of some weapons having an error texture
```
Histalek, line 30:
```suggestion
- Fixed weapons which set a custom view model texture having an error texture
```
we know which weapons are affected so we should mention this

Histalek (changes requested):
a small changelog nit, otherwise LGTM!

Histalek (approved):

PR #1469, opened 2024-03-08
Crosshair: Use default crosshair renderer for binoculars

TimGoll:
The binoculars previously had custom crosshair rendering code. It now uses the same rendering code from the base that all weapons use.

old:
image_2024-03-08_12-11-08

new (with custom color set in the settings):
image


saibotk (approved):

Histalek (approved):
a negative LoC PR by Tim :astonished: 

:heart:

TimGoll:
> a negative LoC PR by Tim 😲
> 
> ❤️

Ey, I'm pretty sure this is at least the second time in 5 years! :D

PR #1471, opened 2024-03-08
Crosshair: Prevent rendering GMod crosshair

TimGoll:
This started as a fix for a bug introduced in #1464 because HUD Help was now drawn twice for grenades with HUD help. So I decided to make the crosshairq function callable, so it can be called seperately in the grenade base.

While doing so I noticed [`SWEP:DoDrawCrosshair`](https://wiki.facepunch.com/gmod/WEAPON:DoDrawCrosshair). This function is called when `SWEP.DrawCrosshair` is set to true. We have set that to false in our base, because we do not want the GMod crosshair do be visible at all. However, some weapons set this to true, probably because the SWEP author thinks setting this to false disables the TTT2 crosshair. This results in two crosshairs being drawn at once:
image

In a perfect world we would use `SWEP:DoDrawCrosshair` to draw our own crosshair and then return `true` at the end. That way the GMod crosshair is never drawn and SWEP authors can use `SWEP.DrawCrosshair` to disable the crosshair for their weapon. However I know for a fact that there are a few weapons that explicitly set this variable to `false` which would result in no crosshair being drawn at all, even though this is not the intention.

My solution here was to use this hook to draw the crosshair and pass another variable to it. If it is called by us, the crosshair should be drawn. If it is called by the engine, the hook should just return true. This way all weapons always have a crosshair and only the TTT2 crosshair.

We could also opt to use `SWEP:DoDrawCrosshair` in its intended way hoping that those authors fix their weapons, but I think that this would cause people complaining about TTT2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 - Fixed the AFK timer accumulating while player not fully joined (by @EntranceJew)
 - Fixed the equipment menu throwing errors when clicking on some items
+- TTT2 now ignores Gmods SWEP.DrawCrosshair and always draws just its own crosshair to prevent two crosshairs at once
 
 ### Removed
 
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -363,15 +363,30 @@ if CLIENT then
             self:DrawHelp()
         end
 
-        if not cvEnableCrosshair:GetBool() then
-            return
+        self:DoDrawCrosshair(mathCeil(ScrW() * 0.5), mathCeil(ScrH() * 0.5), true)
+    end
+
+    ---
+    -- Called when the crosshair is about to get drawn, and allows you to override it.
+    -- @note This is a hook that is used to draw the crosshair. We use the function to prevent
+    -- the crosshair from being drawn and then use the same hook to draw our own custom
+    -- crosshair. The third parameter has therefore be set to true if the crosshair should
+    -- actually be drawn, otherwise the function only returns true to prevent the default one.
+    -- @param number xCenter The x center position of the crosshair
+    -- @param number yCenter The y center position of the crosshair
+    -- @param boolean shouldDraw Should the crosshair be drawn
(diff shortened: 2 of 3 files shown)


TimGoll:
I thought about this a bit yesterday. Merging it like this wouldn't change anything besides fixing the two bugs mentioned in the description. We can still decide later on if we want addons to handle hiding the crosshair later on, these changes won't interfere with that decision

saibotk (changes requested):
I am not really happy about this solution to be honest.

First, we are now ignoring the Swep.DrawCrosshair parameter entirely which is really sad to do, because authors for very sophisticated weapons, imagine some targeting device where you manage a helicopter gun like in CoD for example, cannot disable the crosshair anymore. I'd rather have no crosshair on buggy weapons than forcing our crosshair onto cool specially designed weapons. This now "fixes" some but breaks others, so no real win here.

TimGoll:
> First, we are now ignoring the Swep.DrawCrosshair parameter entirely which is really sad to do

We / TTT is doing this since 2013. The Crosshair in TTT is always forced, if you wanted to hide it, you'd overwrite SWEP:DrawHUD. And this change doesn't change that, this would still hide it. This change only hides the GMod default crosshair.

This is also why I suggested we might change it. But this change would break many weapons because some authors accidentally set it to false.

saibotk:
Ah okay thats so scuffed fr that even TTT does not adhere to the base game stuff.

Also interesting:

https://wiki.facepunch.com/gmod/Structures/SWEP#AccurateCrosshair

Idk how these vars are handled currently, just fyi that these exist.

TimGoll:
> Ah okay thats so scuffed fr that even TTT does not adhere to the base game stuff.

Yes, the way TTT does its crosshair is really bad. I'm inclined to change it because it won't matter for 95% of addons. But this PR has no influence on this. It only fixes two bugs.

If we'd go the proper route later on, we only have to remove the third parameter from the `DoDrawCrosshair` and remove its call from `DrawHUD`.

saibotk (approved):
Okayish for me, but we should investigate a better way in the future that is more integrated with the native crosshair handling.

For example stuff like https://wiki.facepunch.com/gmod/Player:CrosshairDisable should be respected.

TimGoll:
I absolutely agree with you, the current implementation is really shit and the 3D crosshair sounds like fun as well!

ZenBre4ker (approved):

PR #1472, opened 2024-03-08
[RELEASE] Prepare for v0.13.2b

Histalek:
I cherry-picked all 9 commits mentioned in https://github.com/TTT-2/TTT2/milestone/11?closed=1

While conflicts in the changelog were expected and pretty straightforward to resolve, the conflicts in the language files would've been quite a chore. Therefore i have not pulled in f2e7c7ed 
Those are meaningless anyway so this should be fine.

Still this was a bit more manual work than i anticipated and i'm therefore not sure if we really want to start doing this for releases...

i have yet to try what happens if i merge master into this branch now (which would be the same as if we would do a minor release after any patch release). Maybe that would be painless, maybe that would again mean conflicts that need to be resolved...

This is a draft for now as at least #1471 is still missing.


Histalek:
Picked #1471 and #1480 as well.

Should now be good to go i think.

TimGoll (approved):

saibotk (approved):

PR #1476, opened 2024-03-09
Preventwin: Also prevent timelimit

NickCloudAT:
Debug preventwin states "prevent any win condition". I think the time limit is a win condition as well but was not prevented with this.


TimGoll (approved):
That's awesome, thank you

Histalek (approved):

PR #1479, opened 2024-03-09
bodysearch: TTT2GiveFoundCredits bugfix & new TTT2CanTakeCredits hook

Spanospy:
Noticed in 9cff5af that the hook run got re-added with the old name!

Used this opportunity to re-evaluate the hook, too. Decided to split it so that the original intention of the hook is intact whilst enabling the ability to handle the side effects of the hook.


Spanospy:
Finally got around to testing this and it works perfectly, as long as addons properly hook onto `TTT2CanTakeCredits` on both client and server.

Addons I've used to test the hooks: https://github.com/Spanospy/ttt2-wep_pickpocketer and https://github.com/Spanospy/ttt2-det_corpse_protection

TimGoll (approved):

saibotk (approved):
Thanks 🙏

PR #1487, opened 2024-03-20
playermodels: Unique Playermodels at round start

Exonen2:
At Round Start, if the option is activated, each player gets assigned a different playermodel from the pool.


TimGoll:
You know that this is already a thing in TTT2? :D

Exonen2:
> You know that this is already a thing in TTT2? :D

Huh? Maybe I am blind lol...

Doesn't the current "model_per_Round" assign the same model to everyone at round start?

The Changes im proposing is that at round start everyone gets assigned a different model.

TimGoll:
Oh sorry, then nevermind! I'll review it later today, thank you

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -1041,6 +1046,11 @@ function PrepareRound()
         and GAMEMODE.force_plymodel ~= ""
     then
         GAMEMODE.playermodel = GAMEMODE.force_plymodel
+    elseif cvSelectUniqueModelPerRound:GetBool() then
+        local plys = player.GetAll()
```
TimGoll, line 1050:
```suggestion
        local plys = select(2, player.Iterator())
```

This is a little bit faster and something that was added in the most recent gmod update

TimGoll (changes requested):
Sorry for the late review, it kind of got lost in my todo list.

I'd like to change this a bit. Your implementation now results in three modes:
- random identical playermodel on map chanage
- random identical playermodel on round change
- random unique playermodel on round change

Imho it would be better to keep the two old modes as is (new model assignment on round or map change) and just modify your convar so it affects both modes. If your convar is enabled, a random unique player model is selected no matter the selection mode.

Does that make sense to you? I think this would be better and cleaner.

gamemodes/terrortown/gamemode/server/sv_main.lua
TimGoll: well, nevermind, we reverted this for now. So don't apply this yet

TimGoll:
No worries, I'm out of town as well. The proposed changes should probably be pretty easy for you

TimGoll (approved):

TimGoll:
Maybe someone else takes a look at this as I reworked this a lot compared to the initial PR

gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -953,7 +958,19 @@ function GM:TTT2PrePrepareRound(duration)
...
+                plys[i].defaultModel = playermodels.GetRandomPlayerModel()
+            end
+        else
+            local plys = player.GetAll()
+            for i = 1, #plys do
+                plys[i].defaultModel = nil
+            end
+
+            self.playermodel = playermodels.GetRandomPlayerModel()
+        end
```
Histalek, line 973: @TimGoll 
Can we find out what addons etc. write to `.defaultModel`? Because until now we (ttt2) never wrote to it and only read it.

If we start writing to and even resetting(!) it that might be a problem. Even more so now that the default behaviour is changed here.

Don't get me wrong this could all be completely fine, but i have no idea where any value for `.defaultModel` would have come from. Certainly not from ttt2

Histalek (changes requested):
What exactly is the goal of this change?
* diversify the selectable playermodels. So that not everyone looks the same (but allow duplicates)
* give a hard guarantee that any two players won't have the same model. So that you can instantly recognize a specific player (besides identity-swapper shenanigans etc.)

Because currently the implementation does the former and the settings implies the latter.
We need to consolidate this

gamemodes/terrortown/gamemode/server/sv_main.lua
TimGoll: I had the same thoughts, however I'm unable to find any use of that variable in our whole code base. My test server also has loads of addons and I didn't encounter any issues. My suspicion is that this is a very old part of the code that is no longer used. But I could also introduce a new variable for that.

TimGoll:
> What exactly is the goal of this change?

The former. I guess the wording in the UI is then a bit unfitting. Unique models are more complex and are probably not worth it

gamemodes/terrortown/gamemode/server/sv_main.lua
Histalek: `.defaultModel` is used twice in our codebase.

[First occurrence](https://github.com/TTT-2/TTT2/blob/b03a912682a572b6b3b9e341f8662f2c62e020e7/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua#L1105) in `plymeta:SetModel(mdlName)` as a fallback if the given `mdlName` is not a valid model.

[Second occurrence](https://github.com/TTT-2/TTT2/blob/b03a912682a572b6b3b9e341f8662f2c62e020e7/gamemodes/terrortown/gamemode/server/sv_player.lua#L268) which calls the function in the first occurrence with `ply.defaultModel or GAMEMODE.playermodel`. Which is somewhat stupid in itself ...

If we can reasonably believe that this does not break playermodelselector addons i'm fine with merging this. streamlining the playermodel handling can come later.

TimGoll: I meant that it is set nowhere. But you are right. I might ask some EPMS users on discord to test this branch

TimGoll: image

seems to work

TimGoll: @Histalek I think you probably missed this comment

Histalek (approved):

PR #1488, opened 2024-03-24
fix magneto stick not targeting certain props anymore

homonovus:


TimGoll (approved):

saibotk (approved):
Thanks 🙏

EntranceJew:
We still should not let the base gamemode do things that invalidate the eyetrace cache and affect performance

TimGoll:
> We still should not let the base gamemode do things that invalidate the eyetrace cache and affect performance

The trace is still cached in the weird TTT implementation. But I kind of agree, but that shouldn't be part of this PR but more of a general rework of that part

PR #1489, opened 2024-03-29
Fixed DynamicCamera breaking if a weapon's CalcView doesn't return complete values

TW1STaL1CKY:
If a SWEP's CalcView function doesn't return the three expected values, DynamicCamera would spam errors and not function anymore.
This has been fixed by ensuring the active SWEP's CalcView returns valid values before giving them to DynamicCamera.


saibotk (approved):
Seems reasonable! 
Thank you very much for creating a PR 😍

TimGoll (approved):
Uh, nice one. Thanks for fixing it!

PR #1490, opened 2024-03-29
Disable locational voice during the preparing phase by default

ruby0b:
(didn't get a chance to test this with someone else yet so I'm setting it as a draft, should work though)


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added a globally audible sound when searching a body
 - Added the option to add a subtitle to a marker vision element
 - Added `TTT2CanTakeCredits` hook for overriding whether a player is allowed to take credits from a given corpse. (by @Spanospy)
+- Disabled locational voice during the preparing phase by default
+  - Added a ConVar `ttt_voice_locational_prep` to reenable it
```
TimGoll, line 21:
```suggestion
  - Added a ConVar `ttt_locational_voice_prep` to reenable it
```

gamemodes/terrortown/gamemode/server/sv_voice.lua
```diff
@@ -30,9 +30,15 @@ local cv_ttt_limit_spectator_voice = CreateConVar("ttt_limit_spectator_voice", "
 -- stylua: ignore
 local loc_voice = CreateConVar("ttt_locational_voice", "0", {FCVAR_NOTIFY, FCVAR_ARCHIVE})
 
+---
+-- @realm server
+-- stylua: ignore
+local loc_voice_prep = CreateConVar("ttt_locational_voice_prep", "0", {FCVAR_NOTIFY, FCVAR_ARCHIVE})
```
TimGoll, line 36: imho, I'd change this to include post as well, don't you think? During post spectators shold also be able to speak witrh everyone again imho (which is probably already the case?)

gamemodes/terrortown/gamemode/server/sv_voice.lua
```diff
@@ -43,14 +49,24 @@ cvars.AddChangeCallback(loc_voice:GetName(), function(cv, old, new)
     SetGlobalBool(loc_voice:GetName(), tobool(tonumber(new)))
 end)
 
+cvars.AddChangeCallback(loc_voice_prep:GetName(), function(cv, old, new)
+    SetGlobalBool(loc_voice_prep:GetName(), tobool(tonumber(new)))
+end)
```
TimGoll, line 54: I don't think you need the global bool stuff here (or in line 41)

gamemodes/terrortown/gamemode/server/sv_voice.lua
Histalek: yes, that is already the case and i don't think it needs to be configurable.

TimGoll: I was talking about the locational voice. It should be disabled in post as well, if it is disabled in prep. At least in my opinion. No secondary convar, this one should handle both

Histalek: It is already disabled in post and i don't think that needs changing. Especially not depending on if locational voice is disabled in prep.

TimGoll: Locational voice is already disabled in post? I wasn't aware of that

TimGoll: Then the UI should probably have a helpbox with "Locational voice is always disabled during the post round phase" to make it a bit more clear

Histalek: maybe, but that's unrelated to this PR

ruby0b: yeah, I find it hard to imagine a use case where you want it enabled during post anyway since all but a few players are probably spectators at that point, but it definitely shouldn't be linked to the preparing phase.

Histalek (approved):
Looks good to me

refactoring the locational voice conditions into a local function is a really nice change!

TimGoll:
I will approve it as soon as you tested it

ruby0b:
tested, works

TimGoll (approved):

PR #1492, opened 2024-03-30
Roundendscreen: Don't show karma changes if karma is disabled

### LINKED ISSUE #1456: No karma on round end screen when karma is disabled
When karma is disabled in settings make it so the round end screen doesn't show karma lost/gained.

Histalek:
Fixes #1456

Before this change the Roundendscreen showed karma changes even if karma was disabled:

To prevent confusion this change now prevents the Roundendscreen from showing these changes:


saibotk (approved):

TimGoll (approved):
Awesome!

EntranceJew:
This is huge, because it caused a lot of people to believe karma was still affecting role distribution despite lots of things being in the way of that (such as the traitor convar never being defined within TTT2)

Well done

PR #1498, opened 2024-03-31
Text: Added text and nickname limiting

TimGoll:
Adds the following two functions to the game:

```lua
---
-- Creates a text that is limited to the provided length. Adds a limiting chat (e.g. '...') after the
-- text if so desired.
-- @param string text The text that may be limited in length
-- @param number width The maximum width that should be used to limit the text
-- @param[default="DefaultBold"] string font The font ID
-- @param[opt] string limitChar The limiting character(s) that might be appended to the end
-- @param[default=1.0] number scale The UI scale factor
-- @return string The length limited text
-- @realm client
function draw.GetLimitedLengthText(text, width, font, limitChar, scale)
```

```lua
---
-- Creates and caches a elliptic nick for a given length.
-- @param number width The maximum width that should be used to limit the nick
-- @param[default="DefaultBold"] string font The font ID
-- @param[default="..."] string limitChar The limiting character(s) that might be appended to the end
-- @param[default=1.0] number scale The UI scale factor
-- @return string The length limited nick
-- @realm client
function plymeta:NickElliptic(width, font, limitChar, scale)
```

These can be used to limit text / nickname to a given maximum length, while also adding a limiting character.

**Example:**

```lua
print(LocalPlayer():NickElliptic(50, "DefaultBold", "", 1.0))

>>> "Tim | Mi"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added `TTT2CanTakeCredits` hook for overriding whether a player is allowed to take credits from a given corpse. (by @Spanospy)
 - Disabled locational voice during the preparing phase by default
   - Added a ConVar `ttt_locational_voice_prep` to reenable it
+- Added Text / Nickname length limiting and caching (by @TimGoll)
 
 ### Changed
 
--- a/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
@@ -115,6 +115,25 @@ function plymeta:AnimUpdateGesture()
 end
 
 ---
+-- Creates and caches a elliptic nick for a given length.
+-- @param number width The maximum width that should be used to limit the nick
+-- @param[default="DefaultBold"] string font The font ID
+-- @param[default="..."] string limitChar The limiting character(s) that might be appended to the end
+-- @param[default=1.0] number scale The UI scale factor
+-- @return string The length limited nick
+-- @realm client
+function plymeta:NickElliptic(width, font, limitChar, scale)
+    limitChar = limitChar or "..."
+
+    self.cachedEllipticNicks = self.cachedEllipticNicks or {}
+
+    self.cachedEllipticNicks[width] = self.cachedEllipticNicks[width]
+        or draw.GetLimitedLengthText(self:Nick(), width, font, limitChar, scale)
+
+    return self.cachedEllipticNicks[width]
+end
+
+---
 -- @hook
 -- @param Player ply The player to update the animation info for.
 -- @param Vector vel The player's velocity.
--- a/lua/ttt2/extensions/draw.lua
+++ b/lua/ttt2/extensions/draw.lua
@@ -678,6 +678,35 @@ function draw.GetTextSize(text, font, scale)
     return w * scale, h * scale
 end
 
+---
+-- Creates a text that is limited to the provided length. Adds a limiting chat (e.g. '...') after the
+-- text if so desired.
+-- @param string text The text that may be limited in length
+-- @param number width The maximum width that should be used to limit the text
+-- @param[default="DefaultBold"] string font The font ID
+-- @param[opt] string limitChar The limiting character(s) that might be appended to the end
+-- @param[default=1.0] number scale The UI scale factor
+-- @return string The length limited text
+-- @realm client
+function draw.GetLimitedLengthText(text, width, font, limitChar, scale)
+    scale = scale or 1.0
+
+    local widthText = draw.GetTextSize(text, font, scale)
+
+    if widthText <= width then
+        return text
+    end
+
+    if limitChar and limitChar ~= "" then
+        width = width - draw.GetTextSize(limitChar, font, scale)
+    end
+
+    -- we use this function here that splits the text in multiple lines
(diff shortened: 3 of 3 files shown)


gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -114,6 +114,25 @@ function plymeta:AnimUpdateGesture()
...
+-- @param[default="..."] string limitChar The limiting character(s) that might be appended to the end
+-- @param[default=1.0] number scale The UI scale factor
+-- @return string The length limited nick
+-- @realm client
+function plymeta:NickElliptic(width, font, limitChar, scale)
+    limitChar = limitChar or "..."
+
+    self.cachedEllipticNicks = self.cachedEllipticNicks or {}
+
+    self.cachedEllipticNicks[width] = self.cachedEllipticNicks[width]
```
Histalek, line 130: This might lead to inconsistent fonts in cases where some players have a cached ellipticNick for font A and others don't. If then something like the scoreboard wants to get the elliptic nick for those players with font B, some players will retain font A through this caching, correct?

i'm not exactly sure how to nicely prevent this from happening without making this more complex though

Histalek: Additionally we probably need to throw the cache away once the players scale changes

Histalek (changes requested):

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: ah, shit, this makes caching a pain. Might not be worth it to cache it then. Right now the wrapped text in the F1 menu is recreated every frame as well. Which is way more heavy than this

lua/ttt2/extensions/draw.lua
```diff
@@ -678,6 +678,35 @@ function draw.GetTextSize(text, font, scale)
     return w * scale, h * scale
 end
 
+---
+-- Creates a text that is limited to the provided length. Adds a limiting chat (e.g. '...') after the
+-- text if so desired.
```
saibotk, line 683: You should clarify that the limiting char is not counted in to the length.

also typo in "char"

saibotk (changes requested):
Let's get rid of the caching, but otherwise nice!

lua/ttt2/extensions/draw.lua
```diff
@@ -678,6 +678,36 @@ function draw.GetTextSize(text, font, scale)
...
+
+    if limitChar and limitChar ~= "" then
+        width = width - draw.GetTextSize(limitChar, font, scale)
+    end
+
+    -- we use this function here that splits the text in multiple lines
+    local lines = InternalSplitLongWord(text, width, widthText)
+
+    return lines[1] .. (limitChar or "")
+end
```
Histalek, line 709:
```suggestion
function draw.GetLimitedLengthText(text, width, font, limitChar, scale)
    scale = scale or 1.0
    limitChar = limitChar or ""

    local widthText = draw.GetTextSize(text, font, scale)

    if widthText <= width then
        return text
    end

    if limitChar ~= "" then
        width = width - draw.GetTextSize(limitChar, font, scale)
    end

    -- we use this function here that splits the text in multiple lines
    local lines = InternalSplitLongWord(text, width, widthText)

    return lines[1] .. limitChar
end
```
ensuring that `limitChar` is set at the start of the function simplifies 2 expressions later on

i'm fine with either way, i'll let you make the call on this one :)

saibotk (approved):

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added `TTT2CanTakeCredits` hook for overriding whether a player is allowed to take credits from a given corpse. (by @Spanospy)
 - Disabled locational voice during the preparing phase by default
   - Added a ConVar `ttt_locational_voice_prep` to reenable it
+- Added Text / Nickname length limiting and caching (by @TimGoll)
```
Histalek, line 22:
```suggestion
- Added Text / Nickname length limiting (by @TimGoll)
```
Caching is gone :D

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -114,6 +114,18 @@ function plymeta:AnimUpdateGesture()
...
+---
+-- Creates an elliptic nick for a given length.
+-- @param number width The maximum width that should be used to limit the nick
+-- @param[default="DefaultBold"] string font The font ID
+-- @param[default="..."] string limitChar The limiting character(s) that might be appended to the end
+-- @param[default=1.0] number scale The UI scale factor
+-- @return string The length limited nick
+-- @realm client
+function plymeta:NickElliptic(width, font, limitChar, scale)
+    return draw.GetLimitedLengthText(self:Nick(), width, font, limitChar or "...", scale)
```
Histalek, line 126: Do we really want the `limitChar` to be configurable here? i think just using '...' here would be fine and is a bit more in line with the function name

Histalek (changes requested):

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: I don't care. If you say so, I'll change it

Histalek (approved):

PR #1499, opened 2024-03-31
EquipmentEditor: Set clip on buy

TimGoll:
Added `SWEP.SetClipOnBuy` and `SWEP.ClipOnBuy` to set the weapon's clip on buy via the equipment editor. I did not want to just use the latter variable to set a clip on buy to all weapons as most "one time use weapons" instantly remove the weapon on use which makes this slider pointless. It is therefore something swep makers have to integrate into their weapon, which is pretty simple.

I also reordered some elements in the equipment menu because they were out of place. It now looks like this:

image

### DIFF
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -128,6 +128,12 @@ SWEP.silentPickup = false
 -- Can be useful if you have multiple instances, that rely on global variables stored via weapons.GetStored()
 SWEP.HotReloadableKeys = {}
 
+-- Set this to true if the weapon should have a custom clip size on buy that can be set in the equipment editor
+SWEP.SetClipOnBuy = false
+
+-- The defult clip on buy if `SWEP.SetClipOnBuy` is set to true
+SWEP.ClipOnBuy = 1
+
 -- If this weapon should be given to players upon spawning, set a table of the
 -- roles this should happen for here
 --	SWEP.InLoadoutFor = {ROLE_TRAITOR, ROLE_DETECTIVE, ROLE_INNOCENT}
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -129,10 +129,10 @@ SWEP.silentPickup = false
 SWEP.HotReloadableKeys = {}
 
 -- Set this to true if the weapon should have a custom clip size on buy that can be set in the equipment editor
-SWEP.SetClipOnBuy = false
+SWEP.EnableConfigurableClip = false
 
--- The default clip on buy if `SWEP.SetClipOnBuy` is set to true
-SWEP.ClipOnBuy = 1
+-- The default clip on buy if `SWEP.EnableConfigurableClip` is set to true
+SWEP.ConfigurableClip = 1
 
 -- If this weapon should be given to players upon spawning, set a table of the
 -- roles this should happen for here
@@ -1490,10 +1490,10 @@ function SWEP:Initialize()
         self:InitializeCustomModels()
     end
 
-    if self.SetClipOnBuy then
-        self:SetClip1(self.ClipOnBuy or self.Primary.DefaultClip)
+    if self.EnableConfigurableClip then
+        self.Primary.ClipSize = self.ConfigurableClip or self.Primary.DefaultClip
 
-        self.Primary.ClipSize = self.ClipOnBuy or self.Primary.DefaultClip
+        self:SetClip1(self.ConfigurableClip or self.Primary.DefaultClip)
     end
 
     if CLIENT and self:Clip1() == -1 then
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -131,7 +131,7 @@ SWEP.HotReloadableKeys = {}
 -- Set this to true if the weapon should have a custom clip size on buy that can be set in the equipment editor
 SWEP.SetClipOnBuy = false
 
--- The defult clip on buy if `SWEP.SetClipOnBuy` is set to true
+-- The default clip on buy if `SWEP.SetClipOnBuy` is set to true
 SWEP.ClipOnBuy = 1
 
 -- If this weapon should be given to players upon spawning, set a table of the
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -777,16 +777,6 @@ if CLIENT then
 
     ---
     -- @realm client
-    function SWEP:OnRemove()
-        local owner = self:GetOwner()
-
-        if IsValid(owner) and owner == LocalPlayer() and owner:IsTerror() then
-            RunConsoleCommand("lastinv")
-        end
-    end
-
-    ---
-    -- @realm client
     function SWEP:CalcViewModel()
         if not IsFirstTimePredicted() then
             return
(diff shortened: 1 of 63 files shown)


lua/terrortown/lang/en.lua
```diff
@@ -2220,3 +2220,8 @@ L.help_locational_voice_prep = [[By default the proximity chat is disabled in th
 
 Note: Proximity chat is always disabled during the post round phase.]]
 L.help_voice_duck_spectator = "Ducking spectators makes other spectators quieter in comparison to living players. This can be useful if one wants to listen closely to the discussions of the living players."
+
+L.help_equipmenteditor_clip_on_buy = [[The clip size on buy defines the amount of uses the weapon has after it was bought.
+
+Note: This setting is only available for those weapons where its author implemented the handling of it.]]
```
Histalek, line 2226:
```suggestion
L.help_equipmenteditor_clip_on_buy = [[The clip size on buy defines the amount of uses the weapon has when bought in the shop.

Note: This setting is only available for weapons that enable this feature.]]
```

Additionally what does "available" mean in this context? is the option greyed out if `SWEP.SetClipOnBuy` is false?

Histalek (changes requested):

lua/terrortown/lang/en.lua
TimGoll: > is the option greyed out if SWEP.SetClipOnBuy is false?

I could do this. Right now it is just now shown at all.

Histalek: that's fine as well!

Histalek (approved):

saibotk:
Im opposed to the networking part, can you quickly explain why that is necessary?

TimGoll:
> Im opposed to the networking part, can you quickly explain why that is necessary?

It is needed to update the clip variable on the client as well. This was the easiest solution I could find at least. As @ZenBre4ker said the saved keys are networked, but I have to set `wep.Primary.ClipSize` to the value of `wep.ClipOnBuy` if  `wep.SetClipOnBuy` is set to true. I could probably also achieve this with the already networked data and a hook/timer combo. But I did not investigate this. I'm also not at home for the coming two weeks. So if someone wants to look into it, please go ahead, I'm only available via text though

TimGoll:
@saibotk Just to ping you as you did not answer to my comment

saibotk:
But why does the client need this? What happens if you dont do that? 

Seems like something that gmod should/would handle.

TimGoll:
> But why does the client need this? What happens if you dont do that?
> 
> Seems like something that gmod should/would handle.

The Playerinfo HUD element is bugged and shows the wrong max clip size. GMod does not handle this, GMod sadly has no max clip. This is manually handled by TTT

saibotk:
Can this be solved via the weapon entity's network variables? Like we do with the sprint system?
Seems like the more correct place for this instead of another side channel network update.
If not im fine with this. Thanks for explaining!

TimGoll:
> Can this be solved via the weapon entity's network variables? Like we do with the sprint system? Seems like the more correct place for this instead of another side channel network update. If not im fine with this. Thanks for explaining!

Not sure, hm. It is not something that is updated on the fly though. Maybe I can use the synced data from @ZenBre4ker's settings stuff, but I did not investigate this and he did not yet answer

ZenBre4ker:
So, Im wondering why this is something TTT2 should implement?
Isnt that a default setting when you get equipment, that it can have ammo directly?

Otherwise while cliponBuy is synchronized it is just a static setting.
But generally you would only need to check it on the client on the receiving end and add that amount to your clip.
Not really necessary to send it to the client.

TimGoll:
> So, Im wondering why this is something TTT2 should implement?
Isnt that a default setting when you get equipment, that it can have ammo directly?

Yes. But there are multiple weapons that want to have a configurable clip size. Using the `.Primary.Clip` variable doesn't work because it is only updated on weapon initialization when the file is loaded (so only once per map change) and it would also break weapons that do not account for dynamic clip size.

I therefore added the `ClipOnBuy` variable that is only used if the weapon creator enables it. It can be used so that the defi has a definable amount of uses without the hacking that is currently in place.

> Not really necessary to send it to the client additionally.

As I said in a previous comment that does not work. If you do not set the MaxClip on the client manually, it ends up showing stuff like "4 / 1" (four out of one) left, which I consider a bug.

ZenBre4ker:
There should be something in the database you can hook to.

TimGoll:
> There should be something in the database you can hook to.

All I could find is this: https://github.com/TTT-2/TTT2/blob/master/lua/ttt2/libraries/database.lua#L291

But as far as I can see this is only called in the realm where it is stored (the server in this case) and it is also not called when the weapon is initialized and the value from the database is only read, not changed

ZenBre4ker:
Mhh, normally it should be synced. Same way we do it for databaseelement in the ui.

TimGoll:
> Mhh, normally it should be synced. Same way we do it for "database"-element in the ui.

I've never worked with your synced database stuff. So if you think there is a better solution to my custom syncing, I'd really appreciate pointers that show me what to do because I'm not that motivated to start working through all the database and syncing stuff you did, that was quite a lot. Thanks!

ZenBre4ker:
I Thought about this here, should work the same for your case:
https://github.com/TTT-2/TTT2/blob/297be1b71501eef88f96b81602eb5d132785690b/lua/ttt2/libraries/items.lua#L87-L101

TimGoll:
Thank you, I will check it out!

TimGoll:
I simplified the syncing a lot by using the syncing that sven already implemented

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -128,6 +128,12 @@ SWEP.silentPickup = false
 -- Can be useful if you have multiple instances, that rely on global variables stored via weapons.GetStored()
 SWEP.HotReloadableKeys = {}
 
+-- Set this to true if the weapon should have a custom clip size on buy that can be set in the equipment editor
+SWEP.SetClipOnBuy = false
+
+-- The defult clip on buy if `SWEP.SetClipOnBuy` is set to true
```
ZenBre4ker, line 134:
```suggestion
-- The default clip on buy if `SWEP.SetClipOnBuy` is set to true
```

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -1484,6 +1490,10 @@ function SWEP:Initialize()
         self:InitializeCustomModels()
     end
 
+    if self.SetClipOnBuy then
+        self:SetClip1(self.ClipOnBuy or self.Primary.DefaultClip)
+    end
+
```
ZenBre4ker, line 1496: Arent those also called for weapons on the floor? Is that intended?

lua/terrortown/lang/en.lua
```diff
@@ -2222,6 +2222,11 @@ L.help_locational_voice_prep = [[By default the proximity chat is disabled in th
 Note: Proximity chat is always disabled during the post round phase.]]
 L.help_voice_duck_spectator = "Ducking spectators makes other spectators quieter in comparison to living players. This can be useful if one wants to listen closely to the discussions of the living players."
 
+L.help_equipmenteditor_clip_on_buy = [[The clip size on buy defines the amount of uses the weapon has when bought in the shop.
+
+Note: This setting is only available for weapons that enable this feature.]]
```
ZenBre4ker, line 2227: If Initialize is not overriden it also works, right?

saibotk (approved):

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
TimGoll: Yes. It wasn't when I came up with the idea, but I changed my mind. Maybe that means that I should change the variable name. Any good suggestions? Maybe configurableClip

lua/terrortown/lang/en.lua
TimGoll: If it is overridden it breaks, yes. But it also has to be enabled on a per weapon basis because most weapons would "break" if the clip is configurable

TimGoll:
Before merging I will probably rename the variable. I also have to test if the UI is working as intended, I think I forgot to set the MaxClip. But that is an easy fix.

Just so you are aware and don't merge it yet.

TimGoll:
Tested and works

ZenBre4ker (approved):

PR #1502, opened 2024-04-05
Corpse: Corpse velocity can now be properly overridden using TTT2ModifyRagdollVelocity

TW1STaL1CKY:
With this fun little change, a corpse's initial velocity can be properly overridden by returning a vector in a TTT2ModifyRagdollVelocity hook.


TimGoll:
In general we tend to pass these values as parameters to the hook inside a table. That way multiple addons can modify the same value and the change stacks. So if one addon sets the multiplier to 2, the other to 3, the resulting multiplier would be 6. Is this something that you would consider useful?

TW1STaL1CKY:
Ahh I see, I thought updating v was how you modify it but when I tried it wasn't working, hence why I made this PR.
As you noted, I fell into the trap of replacing v with a new vector, which doesn't properly update the value. I just double checked replacing the individual xyz fields and it does update the velocity properly 😄

I'll close this since it isn't needed, the current implementation makes more sense. It would be good if it was noted on the API docs for this hook that this is how you correctly update the velocity

Issue #1491, opened 2024-03-29
Proximity Voice Chat Enhancements


ruby0b:
Planning to implement most of these myself as mentioned on the Discord, just putting a cleaned up list here to keep track / discuss:

**Server-side:**
- [x] Using regular voice chat for everyone during the preparing phase (#1490)
- [ ] Maximum voice chat range cutoff (probably guarded by another ConVar, disabled by default)

**Client-side:**
- [ ] Voice chat toggle keybinding (upstreaming https://github.com/ruby0b/ttt2-voice-chat-toggle)
- [ ] Hiding voice chat panels (with a setting to hide them only for players that you hear through proximity voice chat. Will require clients to receive the information that a voice is heard locationally)

TimGoll:
Nice to see you working on that. Together with my UI rework, the 0.14.0 update will feature quite a lot of voice chat improvements!

TimGoll:
> Voice chat toggle keybinding (upstreaming https://github.com/ruby0b/ttt2-voice-chat-toggle)

What's the general consensus on this? Should this really be a secondary keybind? Or a setting in the client voice setting where you can switch between different modes like "push to talk", "push to mute", "toggle" and "voice activation"?

IMHO we don't need a new keybind for this. It could use the already existing one. We should also make sure that the global voicechat is automatically muted as soon as the team voice chat is engaged.

ruby0b:
I agree that it's probably unnecessary to introduce another keybind, we should just use a dropdown for the different options.
Toggle and voice activation would be the same thing though because as far as I can tell the voice chat always uses some built-in threshold, right?

I found it a bit difficult to wrap my head around how team chat should work when using anything other than push-to-talk, so I made some diagrams.
I guess for toggle it would be like this (G = global voice chat key, T = team voice chat key):
![image](https://github.com/TTT-2/TTT2/assets/106119328/8a29492e-c4be-4af5-88d4-8da97dc6597d)

Push to mute seems a little more confusing since pressing the key is supposed to *mute* you.
But I guess it could work like this, where you switch voice chat by pressing and then *releasing* the key:
![image](https://github.com/TTT-2/TTT2/assets/106119328/143b27d0-453c-42bd-8fda-4ce69cb79f19)

TimGoll:
> as far as I can tell the voice chat always uses some built-in threshold, right?

[...]

PR #1503, opened 2024-04-06
Add locational voice chat range cut-off setting

ruby0b:
#1491

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added `TTT2CanTakeCredits` hook for overriding whether a player is allowed to take credits from a given corpse. (by @Spanospy)
 - Disabled locational voice during the preparing phase by default
   - Added a ConVar `ttt_locational_voice_prep` to reenable it
+- Added `ttt_locational_voice_range` to set a cut-off radius for the locational voice chat range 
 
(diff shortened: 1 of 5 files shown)


ruby0b:
Not sure if using 0 as a special value to disable the feature is idiomatic or if I should add a bool that enables/disables the cut-off being active.

ruby0b:
As for the motivation for this feature: by default, you can basically still hear every player from anywhere on the map, they just get *very* quiet at some point which is probably not what most people want.

EntranceJew:
Does this also account for the "local origin" issue where anyone standing near 0,0,0 will be able to hear conversations at a much higher volume due to the speaking player being in the PAS but not the PVS?

ruby0b:
Sorry, I don't know anything about that issue (didn't even know that was a thing). Would that affect distance calculations?
If not then I guess this would fix that since you're guaranteed not to hear anyone outside the given range.

EntranceJew:
It is more like it is a problem that frequently crops up with distance-cutoff because the issue becomes more apparent when the voice range is relatively small but the map is dense, I'm surprised you haven't had it come up.

saibotk (approved):
Looks very good and i think 0 is fine for disabled 🙌

Thank you so much for improving this.

Histalek (approved):
Looks good to me :)

using 0 to turn this is off is a good choice

PR #1504, opened 2024-04-07
Add convar to prevent players without credits from being able to see credits on a body

nike4613:
The one question left for this is whether to change the default. Currently, this PR defaults to the previous behavior (all players can see credits, except in mode 2), but it seems to me like player expectations would favor not seeing credits by default.

### DIFF
--- a/lua/terrortown/lang/en.lua
+++ b/lua/terrortown/lang/en.lua
@@ -1523,6 +1523,7 @@ L.label_roles_max_roles_pct = "Upper role limit by percentage"
 L.label_roles_max_baseroles = "Upper base role limit"
 L.label_roles_max_baseroles_pct = "Upper base role limit by percentage"
 L.label_detective_hats = "Enable hats for policing roles like the Detective (if player model allows to have them)"
+L.label_inspect_credits_always = "Allow all players to see credits on dead bodies"
 
 L.ttt2_desc_innocent = "An Innocent has no special abilities. They have to find the evil ones among the terrorists and kill them. But they have to be careful not to kill their teammates."
 L.ttt2_desc_traitor = "The Traitor is the enemy of the Innocent. They have an equipment menu with which they are being able to buy special equipment. They have to kill everyone but their teammates."
@@ -1545,6 +1546,10 @@ L.help_detective_hats = [[
 Policing roles such as the Detective may wear hats to show their authority. They lose them on death or if damaged at the head.
 
 Some player models do not support hats by default. This can be changed in 'Administration' -> 'Player Models']]
+L.help_inspect_credits_always = [[
+When players with credits (detectives, traitors, survivalist, etc.) die, their credits can be picked up by other players that can use credits. 
+
+When this option is disabled, only players that can pick up credits can see them on the body.]]
 
 L.label_roles_credits_award_kill = "Credit reward amount for the kill"
 L.label_roles_credits_dead_award = "Enable credits award for certain percentage of dead enemies"
--- a/lua/terrortown/lang/en.lua
+++ b/lua/terrortown/lang/en.lua
@@ -1523,7 +1523,6 @@ L.label_roles_max_roles_pct = "Upper role limit by percentage"
 L.label_roles_max_baseroles = "Upper base role limit"
 L.label_roles_max_baseroles_pct = "Upper base role limit by percentage"
 L.label_detective_hats = "Enable hats for policing roles like the Detective (if player model allows to have them)"
-L.label_inspect_credits_always = "Allow all players to see credits on dead bodies"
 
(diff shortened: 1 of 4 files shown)


Histalek (commented):
@TimGoll could you comment on this? i think you had a reason for showing credits on a body by default

Additionally, wouldn't this be a usecase for the `TTTBodySearchPopulate` hook?

nike4613:
> Additionally, wouldn't this be a usecase for the `TTTBodySearchPopulate` hook?

That hook does not appear to be able to remove things which have already been added to the UI. Credits are added here:
https://github.com/TTT-2/TTT2/blob/d4df150769d45e231fc146502758f9044d637516/gamemodes/terrortown/gamemode/client/cl_search.lua#L292-L301
while the hook is invoked here:
https://github.com/TTT-2/TTT2/blob/d4df150769d45e231fc146502758f9044d637516/gamemodes/terrortown/gamemode/client/cl_search.lua#L309
well after the info items have been added.

TimGoll:
OK, I'd prefer this to be serverside instead of client side, if you want to hide the data, then it should be done in `bodysearch.AssimilateSceneData()`:

https://github.com/TTT-2/TTT2/blob/cb48c43464e4522ed868676b03f3fabce4d1e3e0/lua/ttt2/libraries/bodysearch.lua#L267

That way a detective search would still push the data to everyone.

In my opinion it should be visible to everyone so it is harder for traitors etc to snatch the credits without anyone noticing it.
However this special case here might have to be handled differently then. My implementation seems a bit hacky anyway:

https://github.com/TTT-2/TTT2/blob/cb48c43464e4522ed868676b03f3fabce4d1e3e0/gamemodes/terrortown/gamemode/server/sv_corpse.lua#L293-L294

On the hook: I'm aware that it currently doesn't worky anymore to remove/edit stuff. It is something I did not consider when reworking the UI. I only anticipated addons adding custom data. We should probably introduce a serverside hook that can modify the data, it should be probably right here between the two lines in the middle:

https://github.com/TTT-2/TTT2/blob/cb48c43464e4522ed868676b03f3fabce4d1e3e0/gamemodes/terrortown/gamemode/server/sv_corpse.lua#L296-L301

All in all I like your change here, but I don't like the implementation. This change would be a bit more complex then your implementation, but it should still be doable. If you don't want to do it, you can wait until I'm back at home at the end of the month and I'll take a look at it

nike4613:
> In my opinion it should be visible to everyone so it is harder for traitors etc to snatch the credits without anyone noticing it.

That's actually kinda the point of the change-- that's a whole interesting metagame that should at least be an option. I generally agree that filtering should be done serverside as much as possible, but for my intended behavior, that would require sending different data to each client, which I suspect will get very complex and error-prone very quickly.

If we do add the option for detective search to push the data everywhere, I'd want that to be a separate cvar, *in addition* to the cvar I've already added.

saibotk (approved):

saibotk:
Serverside would be better but since it seems like mode 2 also just does it on the client i think we should change this in another pr.
Otherwise thanks for the comprehensive code and comments, appreciate your work!

lua/terrortown/lang/en.lua
```diff
@@ -1545,6 +1546,10 @@ L.help_detective_hats = [[
 Policing roles such as the Detective may wear hats to show their authority. They lose them on death or if damaged at the head.
 
 Some player models do not support hats by default. This can be changed in 'Administration' -> 'Player Models']]
+L.help_inspect_credits_always = [[
+When players with credits (detectives, traitors, survivalist, etc.) die, their credits can be picked up by other players that can use credits.
```
TimGoll, line 1550: We tend to call them "shopping roles" in code and documentation, maybe you can change the wording a bit here

lua/ttt2/libraries/bodysearch.lua
```diff
@@ -738,7 +747,7 @@ if CLIENT then
             -- special case: mode 2, only shopping roles can see credits
             local client = LocalPlayer()
             if
-                cvInspectConfirmMode:GetInt() == 2
+                (cvInspectConfirmMode:GetInt() == 2 or cvCreditsVisibleToAll:GetInt() == 0)
```
TimGoll, line 750:
```suggestion
                (cvInspectConfirmMode:GetInt() == 2 or not cvCreditsVisibleToAll:GetBool())
```

TimGoll (changes requested):
I guess having it clientside is fine for now. But please put the new language strings at the end of the file, similar to how it is done in other PRs. If you scroll down to the end of the file, you'll understand what to do

TimGoll (approved):
LGTM, that you very much

PR #1511, opened 2024-05-13
UI: Added Combo Cards

TimGoll:
Added Combo-Cards. They act like combo boxes (only one of them can be selected at any time), but they are clickable boxes:

image

```lua
local form = vgui.CreateTTT2Form(parent, "header_maps_select")

-- Create combo boxes to select element from
comboBase = form:MakeIconLayout()

for i = 1, #maps do
    local mapName = maps[i]
    local prefix = map.GetPrefix(mapName)

    form:MakeComboCard({
        icon = map.GetIcon(mapName),
        label = mapName,
        tag = prefix,
        colorTag = util.StringToColor(prefix or ""),
    }, comboBase)
end
```

Similar to other tileable elements in our UI system, this needs an icon layout as well. This is then used as a base for these combo cards. There is no limit to combo cards. All cards that have the same icon layout as a base are tied together.

`comboBase.checked` contains the checked element. It is `nil` if none is checked.

Example where these boxes are used: #1512

**Note:** I previously added different cards. To make it more clear I renamed those. Maybe there is the possibility to merge them in the future, but right now I'm not in favor of that because that would make the code more convoluted.


Histalek (approved):
LGTM :)

PR #1512, opened 2024-05-13
F1 menu: Added Option to Change Maps via the F1 Menu

TimGoll:
This adds the option for admins to change the currently played map while also starting an admin commands menu. This makes use of the combo cards added in #1511.

image

image

image

One thing that was added, is a hook that is called, once a client finished the hot reload. This is needed so that the server knows when the client is ready to resend the data:
```lua
---
-- Called when a player is ready again after they reloaded their game. Can be used to send
-- data that is needed after a reload.
-- @param Player ply The player that finished the reloading
-- @hook
-- @realm server
function GM:TTT2PlayerFinishedReloading(ply)
```

This PR also greatly extended the `map` module which can also then be used by other addons. Details can be checked in the changed files, but most notably are these additions:

This function returns a cached material for a given map. This also includes maps that don't have a material. Since this works by downloading these materials from steam, this system works as long as the map is installed via the workshop and the workshop page has a preview image.
```lua
---
-- Returns the cached icon of a map if there is an icon available.
-- @note This not only uses the icons available on the client through addons, it also
-- searches the workshop and tries to assign the corrent workshop icon if there
-- is no locally available map icon.
-- @param string name The map name
-- @return nil|IMaterial Returns the cached material if available
-- @realm client
function map.GetIcon(name)
```

This function returns a list of all available maps. It already excludes those maps, where the prefix was disabled in the admin settings. This can be used in map mote addons as a map base.
```lua
---
-- Returns a sanitized list of maps installed on the server. Sanitized means that only
-- valid maps with a prefix that is enabled are listed here. This list is automatically
-- synced between the server and all clients.
-- @return table Returns a table with the map names
-- @realm shared
function map.GetList()
```


lua/ttt2/libraries/map.lua
```diff
@@ -504,6 +506,300 @@ function map.GetDataFromSpawnEntity(ent, spawnType)
...
+
+    ---
+    -- Initializes the map list. Searches the file system for available maps, scans those maps
+    -- for their different prefixes and tries to associate a workshop ID with each map.
+    -- @internal
+    -- @realm server
+    function map.InitializeList()
+        local mapsFound = file.Find("maps/*.bsp", "GAME")
+
+        mapsAll = {}
```
Histalek, line 539: You are pulling ALL map files here and then you are painfully trying to match these to an addon below.
This is a lot of work and might even fail i assume.

Why not get all garrysmod maps and cstrike maps here.
Then below while iterating over the addons check if that addon contains a map file and you have the map title and the wsid without any matching.

eg.:
```lua
gmod_maps = file.Find("maps/*.bsp", "garrysmod")
cstrike_maps = file.Find("maps/*.bsp", "cstrike")
```

you can search each mounted addon by itself to see if it contains a map file:
```lua
addon_map = file.Find("maps/*.bsp", addon.title)
```

This would make this a lot cleaner IMO

TimGoll: This is actually a good idea! only thing left: how do we handle the builtin maps (gmod, css etc) or locally installed maps then?

Histalek: > This is actually a good idea! only thing left: how do we handle the builtin maps (gmod, css etc) or locally installed maps then?

handle in what way exactly? we can find the maps via their respective searchpaths [1] with `file.Find` (see above).
Maybe we can get a list of mounted addons first so we don't miss any of those.

As for locally installed maps: if these are present as a .gma file they would still be accessible by the `addon.title`.
For maps in the `garrysmod/addons` folder i'm actually not sure.

We would probably need to iterate over all directories in the `garrysmod/addons` directory and then `file.Find` for each of those?

Pseudocode:
```
local _, directories = file.Find("addons/*", "garrysmod")

foreach directory in directories
	maps[directory] = file.Find("addons/" .. directory .. "/maps/*.bsp", "garrysmod")
```

not the nicest, but useable i think.

[1] https://wiki.facepunch.com/gmod/File_Search_Paths

TimGoll: what abou the following:

1. create a map-name-indexed table of the workshop maps:
  ```
  {
    ["map_name_1"] = "687556776565786578",
    ["map_name_2"] = "76868768767",
    [....]
  }
  ```
2. iterater over all file system maps and assign the workshop ID by the previously generated lookup table
3. if there is no map found, use the old code from this PR to assign an icon

What's your opinion on this? This should be pretty fail safe while also supporting any map that is installed in any way

Histalek: i did not intend to have the icon assignment code changed (besides adapting to the proposed change),
but having a mapname -> wsid lookup table seems reasonable.

TimGoll: Alright, that sounds good to me then. This should make the assignment pretty stable

TimGoll:
I think from my POV this PR is done. Thanks again @Histalek your review was really helpful!

lua/terrortown/lang/en.lua
```diff
@@ -2252,3 +2252,14 @@ When shopping roles die, their credits can be picked up by other players with sh
 
 When this option is disabled, only players that can pick up credits can see them on a body.
 When enabled, all players can see credits on a body.]]
+
+-- 2024-05-13
+L.menu_commands_title = "Admin Commands"
+L.menu_commands_description = "Change maps, spawn bots and edit player roles."
```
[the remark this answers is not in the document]
TimGoll: I'm not sure about that, the other subtitles are a bit more descriptive than that

Histalek (approved):
Nice!

PR #1513, opened 2024-05-13
HUD: Fixed OHI Desync

TimGoll:
Sometimes overhead icons stayed stuck in the world. This is due to the way GMod/source handles player positions. If a player is far enough away, their position is no longer updated (as can be seen with wallhacks for example).

While `:GetPos()` still returns a position that at least represents a rough estimate of their position, the bone position stays at the last updated place. This means that the overhead icon stays floating in random places.

This pullrequest fixes this by introducing a check that uses the fallback position if the player position and the bone position diverge too much.

While I tested this, it would probably be good if @mexikoedi could confirm this as well.

### DIFF
--- a/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
@@ -1235,7 +1235,7 @@ end
 -- @return vector The player height
 -- @realm shared
 function plymeta:GetHeightVector()
-    local matrix, pos
+    local matrix, posHeadBone
     local bone = self:LookupBone("ValveBiped.Bip01_Head1")
 
     local posPlayer = self:GetPos()
@@ -1248,18 +1248,19 @@ function plymeta:GetHeightVector()
     end
 
     if matrix then
-        pos = matrix:GetTranslation()
+        posHeadBone = matrix:GetTranslation()
     end
 
     -- if a player is too far away, their location is only updated sporadically
     -- if the distance between these two positions is too large, the fallback position
     -- should be used
-    if (pos - posPlayer):Length() < 90 then
+    if posHeadBone and (posHeadBone - posPlayer):Length() < 90 then
         -- note: the 8 is the assumed height of the head after the head bone
         -- this might not work for every model
-        pos.z = pos.z + 8 * self:GetModelScale() * self:GetManipulateBoneScale(bone).z
+        posHeadBone.z = posHeadBone.z
+            + 8 * self:GetModelScale() * self:GetManipulateBoneScale(bone).z
 
-        return pos - posPlayer
+        return posHeadBone - posPlayer
     end
 
     -- if the model has no head bone for some reason, use the player
--- a/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
@@ -1235,8 +1235,10 @@ end
 -- @return vector The player height
 -- @realm shared
 function plymeta:GetHeightVector()
-    local matrix
+    local matrix, pos
     local bone = self:LookupBone("ValveBiped.Bip01_Head1")
+
+    local posPlayer = self:GetPos()
 
     -- if the bone is defined, the bone matrix is defined as well;
     -- however on hot reloads this can momentarily break before it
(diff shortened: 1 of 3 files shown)


gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -1246,23 +1248,27 @@ function plymeta:GetHeightVector()
...
 
     if matrix then
-        local pos = matrix:GetTranslation()
+        pos = matrix:GetTranslation()
+    end
 
+    -- if a player is too far away, their location is only updated sporadically
+    -- if the distance between these two positions is too large, the fallback position
+    -- should be used
+    if (pos - posPlayer):Length() < 90 then
```
Histalek, line 1257: pos might be `nil` here

moving this inside the if-clause above should fix this imo. do you agree?

Histalek (changes requested):
While writing my review i'm starting to question if this could really fix this problem. `plymeta:GetHeightVector` implies to only return height (a vector with a non-nil Z value and a nil x and y value). But the problem in the first place is happening on all 3 axis, right?

While this change might still fix issues on the Z-axis, the X- and Y-axis issues can't really be fixed by this.

That is at least what i get from this. Am i missing something?

TimGoll:
> While writing my review i'm starting to question if this could really fix this problem. `plymeta:GetHeightVector` implies to only return height (a vector with a non-nil Z value and a nil x and y value). But the problem in the first place is happening on all 3 axis, right?
> 
> While this change might still fix issues on the Z-axis, the X- and Y-axis issues can't really be fixed by this.
> 
> That is at least what i get from this. Am i missing something?

I thought you were right, but this is a bit more tricky. My intention was to return a height vecvor, yes. But when the positions between the headbone and the player position diverge, x and y offsets are introduced in the height vector.

In general I'd like to keep it that way so that it tracks the position and puts this movement in the x and y offset. So maybe the function name is misleading.

The introduced change catches the divergence though, so my fix should work.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
TimGoll: That's absolutely true, thanks. I fixed it a bit differently as I think that way is the code more readable

TimGoll:
@Histalek I applied your suggestions and also renamed a variable to make it more clear. I'm debating whether I should rename the function. Right now it is called `GetHeightVector()`. Which is only kinda true because it also tracks the X and Y deviation of the head position in relation the the player position. Maybe it should be called `GetHeadPosition` or something similar?

homonovus:
the player's position not updating is due to entity dormancy on the client; you can check this for an early return in rendering using `entity:IsDormant()`

an entity is dormant when it is not in a player's PVS, and when so, the entity will net send updates about its state to players who it is dormant for

for players, an exception is made in GMod, where they are transmitted for a single tick every 10 seconds

Histalek:
> > While writing my review i'm starting to question if this could really fix this problem. `plymeta:GetHeightVector` implies to only return height (a vector with a non-nil Z value and a nil x and y value). But the problem in the first place is happening on all 3 axis, right?
> > While this change might still fix issues on the Z-axis, the X- and Y-axis issues can't really be fixed by this.
> > That is at least what i get from this. Am i missing something?
> 
> I thought you were right, but this is a bit more tricky. My intention was to return a height vecvor, yes. But when the positions between the headbone and the player position diverge, x and y offsets are introduced in the height vector:
> 
> ```lua
> return pos - posPlayer
> ```
> 
> In general I'd like to keep it that way so that it tracks the position and puts this movement in the x and y offset. So maybe the function name is misleading.

definitely misleading imo, but since it already was like this before your changes i don't think we need to address this here.

> 
> The introduced change catches the divergence though, so my fix should work.

Makes sense

Histalek:
> @Histalek I applied your suggestions and also renamed a variable to make it more clear.

Thanks!

> I'm debating whether I should rename the function. Right now it is called `GetHeightVector()`. Which is only kinda true because it also tracks the X and Y deviation of the head position in relation the the player position. Maybe it should be called `GetHeadPosition` or something similar?

`GetHeadPosition` sounds good! although i wouldn't want to rename the function right now as i have no idea who might be using this and they might just rely on this unexpected behaviour.

For now leaving this as is should be the lesser evil imo

Histalek (approved):
LGTM as it is

you could look into the `entity:IsDormant()` that @homonovus mentioned, maybe that could lead to an even nicer if-condition :)

TimGoll:
okay, I will use `entity:IsDorman()`. I wasn't aware that there is a function to check that. Therefore my hacky approach. Thank you @homonovus!

@Histalek about the name: I introduced this function in the last minor release - so I guess it should be fine to rename. Nobody is using it

mexikoedi:
I tested a bit around and with different distances too. I couldn't reproduce this issue anymore. So I think it's fixed or I had luck because it's a bit difficult to reproduce. 
The only thing I noticed was that the OHI is stuck if you go outside the map but I think that's a different topic because other things (player name, ...) get also stuck.

Is a similar fix applicable to this addon: https://steamcommunity.com/sharedfiles/filedetails/?id=1137493106
Because this has the issue which was described in the description.

TimGoll:
Done and tested

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -1246,23 +1248,33 @@ function plymeta:GetHeightVector()
...
+    end
 
+    -- If a player is too far away, their head-bone position is only updated
+    -- sporadically.
+    -- If the head-bone position and the player position are not too far apart
+    -- we assume the head-bone position is accurate and use it. When this is the
+    -- case it is because the entity is dormant and its position is no longer
+    -- updated.
+    -- Otherwise we fall back to the highest corner of the players bounding
+    -- box position.
```
Histalek, line 1261:
```suggestion
    -- If a player is too far away, their head-bone position is only updated
    -- sporadically.
    -- In that case we want to fall back to the highest corner of the players bounding
    -- box position.
```

PR #1521, opened 2024-05-20
UI: Admin Commands

TimGoll:
I added basic admin commands to the UI to restart the round, set the health etc

image

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
@@ -374,6 +374,7 @@ function PANEL:MakeSlider(data)
 
     -- Set default if possible even if the convar could still overwrite it
     right:SetDefaultValue(data.default)
+    right:SetValue(data.initial)
     right:SetConVar(data.convar)
     right:SetServerConVar(data.serverConvar)
     right:SetDatabase(data.database)
@@ -416,7 +417,7 @@ function PANEL:MakeSlider(data)
         left:DockMargin(left:GetIndentationMargin(), 0, 0, 0)
     end
 
-    return left
+    return right, left
 end
 
 ---
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
@@ -465,7 +465,7 @@ function PANEL:MakeButton(data)
         left:DockMargin(left:GetIndentationMargin(), 0, 0, 0)
     end
 
-    return left
+    return right, left
 end
 
 ---
--- a/lua/terrortown/lang/en.lua
+++ b/lua/terrortown/lang/en.lua
@@ -2275,7 +2275,7 @@ L.header_commands_player_add_credits = "Add Equipment Credits"
 L.header_commands_player_set_health = "Set Health"
 L.header_commands_player_set_armor = "Set Armor"
 
-L.label_button_round_restart = "round resart"
+L.label_button_round_restart = "round restart"
 L.label_button_player_slay = "slay player"
 L.label_button_player_teleport = "teleport player"
 L.label_button_player_respawn = "respawn player"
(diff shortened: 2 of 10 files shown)


gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dform_ttt2.lua
```diff
@@ -416,7 +417,7 @@ function PANEL:MakeSlider(data)
         left:DockMargin(left:GetIndentationMargin(), 0, 0, 0)
     end
 
-    return left
+    return right, left
```
TimGoll, line 420: Note: while this looks like a breaking change, I don't think anybody used the return value of that function so far

saibotk (approved):

lua/terrortown/lang/en.lua
```diff
@@ -2263,3 +2263,29 @@ L.header_maps_prefixes = "Enable/Disable Maps by their Prefix"
...
+
+L.header_commands_round_restart = "Round Restart"
+L.header_commands_player_slay = "Slay Player"
+L.header_commands_player_teleport = "Teleport Player to Focused Point"
+L.header_commands_player_respawn = "Respawn Player at Focused Point"
+L.header_commands_player_add_credits = "Add Equipment Credits"
+L.header_commands_player_set_health = "Set Health"
+L.header_commands_player_set_armor = "Set Armor"
+
+L.label_button_round_restart = "round resart"
```
Histalek, line 2278:
```suggestion
L.label_button_round_restart = "round restart"
```

lua/terrortown/lang/en.lua
```diff
@@ -2263,3 +2263,29 @@ L.header_maps_prefixes = "Enable/Disable Maps by their Prefix"
...
+L.label_button_player_add_credits = "add credits"
+L.label_button_player_set_health = "set health"
+L.label_button_player_set_armor = "set armor"
+
+L.label_slider_add_credits = "Set credit amount"
+L.label_slider_set_health = "Set health"
+L.label_slider_set_armor = "Set armor"
+
+L.label_player_select = "Select affected player"
+L.label_execute_command = "Execute command"
```
Histalek, line 2291: is there a specific reason why the button labels start with lowercase and the other labels start uppercase?

lua/terrortown/menus/gamemode/commands/commands.lua
```diff
@@ -0,0 +1,176 @@
...
+    })
+
+    -- SLAY PLAYER --
+
+    local form2 = vgui.CreateTTT2Form(parent, "header_commands_player_slay")
+
+    local playerSlay = form2:MakeComboBox({
+        label = "label_player_select",
+        choices = plyChoices,
+        selectValue = plyChoices[1].value,
```
Histalek, line 40: Do we absolutely need to set a pre-selected value here?

I'd rather have an admin explicitly choose a player for these actions to prevent accidental mishaps.

lua/terrortown/lang/en.lua
TimGoll: In our menus all labels next to a setting always start with an upper case letter. To me it looked best when the inline button is completely lower case (see the image in this PR)

lua/terrortown/menus/gamemode/commands/commands.lua
TimGoll: I absolutely see your point. Hm. I guess I'll change it and if nobody complains then it is fine

Histalek: The buttons `OnClick` needs a nil check then. Or can we disable the buttons outright if there is nothing selected in the dropdown?

TimGoll: I could disable it, yes, Adds a bit of annoying code, but it should work

TimGoll: image

I extended our parenting system to support combo boxes as well. Right now it is pretty basic. No selection counts as false, any selection as true. But it works nice without much additional code

Histalek: neat!

Histalek (approved):
:heart:

PR #1524, opened 2024-05-22
Loadingscreen: Added Screen on Round Restart

TimGoll:
**This PR is not as big as it looks, this is only because I modified the tips strings in all languages!**

This pullrequests adds a loadingscreen that hides the visible and audible lag introduced by the map cleanup on round change. This is done because the lag can hardly be improved since it all takes place in `game.CleanupMap`. It can be disabled if one dislikes it.

This PR also moves the (in my opinion annoying) tips panel to a more useful place by displaying those tips on the new loading screen.

https://github.com/TTT-2/TTT2/assets/13639408/239e0441-0d51-4868-834d-ec121443e86b

Settings panel to disable loading screen and tips:

image

Note: The tips convar is carried over from the original TTT convar.


lua/ttt2/libraries/loadingscreen.lua
```diff
@@ -0,0 +1,195 @@
...
+if SERVER then
+    AddCSLuaFile()
+
+    util.AddNetworkString("TTT2LoadingscreenStart")
+end
+
+loadingscreen = loadingscreen or {}
+
+loadingscreen.isInReloading = false
+loadingscreen.wasInReloading = false
```
saibotk, line 14: Can we maybe rename this to something like `isShown` and `wasShown`, because reloading is rather specific to the round reload use case and this way it is much clearer what those states mean.

gamemodes/terrortown/gamemode/client/cl_tips.lua
saibotk, line 1: This is somewhat breaking, so maybe add this to the changelog even if no one really used this :)

saibotk (changes requested):

saibotk (approved):

TimGoll:
I'm leaving this one open as it is a bit bigger so that there is a chance for someone else to take a look at it

PR #1527, opened 2024-05-22
shifted prep time after cleanup

TimGoll:
Merge after #1524 

Prep time now starts AFTER the map cleanup is done, improving the situation where 4+ seconds of prep are already gone once you are spawned. This still doesn't fix it completely, as the networking still eats up one second or so, but it is way more predictable now, especially when the server is slow.

I also cleaned up the commands a bit to remove code duplication.

TODO: Clean up code once the other PR is merged


TimGoll:
Closing in favor of #1532

PR #1528, opened 2024-05-22
UI: Removed Two Mstack Messages at Round Begin

TimGoll:
Merge after #1524

In an effort to reduce the spam in the mstack at round begin, I removed these two messages. They add no important information to the game and are kinda redundant.

image


saibotk (approved):
very nice

PR #1529, opened 2024-05-22
Fix missing include of cl_voice.lua in pure_skin

Tenrys:
This file is now missing since the rework of the voice HUD


saibotk (approved):
Thank you very much for the PR!
Appreciate it!

Issue #1385, opened 2024-02-08
wallhack effect painting the skybox based on proximity to nearest role color as spectator


EntranceJew:
## Your version of TTT2 (mandatory)

- Latest GitHub version

## Video

![image](https://github.com/TTT-2/TTT2/assets/5711436/24b4ee5f-3e82-45c9-a6c5-f50e1ac3a0c1)
![image](https://github.com/TTT-2/TTT2/assets/5711436/38fd7215-7da6-4cbd-bbf0-af1405710dca)
![image](https://github.com/TTT-2/TTT2/assets/5711436/460d313e-9e5d-4730-99f0-7c323b2e5d3f)
![image](https://github.com/TTT-2/TTT2/assets/5711436/f79aaefe-7e93-46b9-99a8-20c27a6268ce)
![image](https://github.com/TTT-2/TTT2/assets/5711436/9c6f05a5-286d-4b28-9358-5ac860fa665d)

## Describe the bug (mandatory)

- wallhack effect paints the skybox
- proximity to nearer roles cause it to change to that role color
- probably has to do with the surface.Reset function needing to be called
- possibly mcore related but idk

## To reproduce

1. `map ttt_mttresort_v2`
2. Die
3. Freecam into the void
4. Wiggle around until the jet black background becomes the color of a would-be-occluded role.

## Expected behaviour

The skybox doesn't go hog wild in colors.

## Context (please provide as much as you can)

- Collection link of your Server: [here](https://steamcommunity.com/sharedfiles/filedetails/?id=3140327866)

## Additional context

I notified @WardenPotato about it in case they're able to figure it out.

Histalek:
i can't reproduce this with the settings as follows:

- Linux on 64-bit branch
- Launch options `+gmod_mcore_test 1 +mat_queue_mode 2` 
- map: `ttt_minecraft_b5`

maybe this is map dependent, i'll try to test on `ttt_mttresort_v2` next

TimGoll:
I couldn't reproduce it on my personal testmap either

PR #1530, opened 2024-05-23
Marks: Clean up library and improve peformance

WardenPotato:
This pull cleans up the previous changes to the marks library a bit which should also play nicer with other effects.

I also fixed the marks drawing during depth/skybox which resulted in 2-3x the performance impact for no reason.

This likely closes https://github.com/TTT-2/TTT2/issues/1385


TimGoll (commented):
You're doing God's work, awesome! I have to test this if this breaks stuff such as colored ents. But other than that this looks good to me.

Ah, please add an entry to the change log. Thanks

WardenPotato:
Ive done the colored ent tests i did last time to ensure it didnt break, but feel free to test, i'll update the changelog

TimGoll (approved):
Works flawlessly. Amazing, thank you!

image

Issue #1360, opened 2024-02-04
BUG: First Round Problems Mega-Issue


EntranceJew:
There are a slew of problems that occur exclusively on the first round, uncertain as to what the problem may be.

## Door Linkage
A door, during its first round, will behave as expected when paired:
https://github.com/TTT-2/TTT2/assets/5711436/b7b01244-145d-450b-a3aa-e165a2e8af6c

A door, during later rounds, will behave erratically:
https://github.com/TTT-2/TTT2/assets/5711436/b353aa1d-468b-446d-8aed-8aeb722fe788

This may never occur if a door is broken off of its hinges, due to key values not being reset on round restart. Most notably, the stack trace changes to include a concommand even if the round concludes naturally:
```
first round:
Trace:
    1: Line 32    "Trace"        lua/includes/extensions/debug.lua
    2: Line 305    "SetUp"        addons/ttt2/lua/ttt2/libraries/door.lua
    3: Line 891    "nil"        addons/ttt2/gamemodes/terrortown/gamemode/server/sv_main.lua
    4: C function    "CleanUpMap"
    5: Line 895    "CleanUp"        addons/ttt2/gamemodes/terrortown/gamemode/server/sv_main.lua
    6: Line 974    "nil"        addons/ttt2/gamemodes/terrortown/gamemode/server/sv_main.lua

second round:
Trace:
    1: Line 32    "Trace"        lua/includes/extensions/debug.lua
    2: Line 305    "SetUp"        addons/ttt2/lua/ttt2/libraries/door.lua
    3: Line 891    "nil"        addons/ttt2/gamemodes/terrortown/gamemode/server/sv_main.lua
    4: C function    "CleanUpMap"
    5: Line 895    "CleanUp"        addons/ttt2/gamemodes/terrortown/gamemode/server/sv_main.lua
    6: Line 974    "PrepareRound"        addons/ttt2/gamemodes/terrortown/gamemode/server/sv_main.lua
    7: Line 1415    "nil"        addons/ttt2/gamemodes/terrortown/gamemode/server/sv_main.lua
    8: Line 54    "nil"        lua/includes/modules/concommand.lua
```

## Long Load Dragging
If a server has been running for several rounds, a client which has been through several reconnects via mapchange will begin to experience an increasingly long join time to a server, and will frequently miss the first round post-mapchange and become a spectator.
[...]

PR #1532, opened 2024-05-24
Gameloop: Complete Restructure

TimGoll:
This pullrequest is quite large, all over the place, but *mostly* moved code. I try to highlight the changes.

After Gathwar asked for a level reset button (resets the round time, rounds played, karme, score etc) I started investigating this topic. The current gameloop is a mess at best and bug ridden at worst. There are no major bugs, but every now and then someone complains about syncing errors, they are probably all caused by our current game loop. I stumbled upon many bandaid fixes and tried to remove all of them. While doing so I reduced the amount of timers by a lot, reduced the network trafic and made the system more robust.

For some reason every round state change (`ROUND_PREP` etc) triggered a full reset on the client:
image
This might seem like a sane solution, but it actually caused issues because it clashes with `SendFullStateUpdate`. Whoever wrote that part of code was aware of it, but chose to fix it by adding two more syncs triggered via timers:
image
The first sync, called in `roleselection.SelectRoles()` is almost always overwritten while the second one comes through. Since our networking system only syncs differences, this can cause issues with players not knowing about their team mates.

For late joiners I added this instead:
image
It makes sure that the client is receiving the data they need.

Speaking of late joiners. This is some syncing that was done before: 
image
Note here that players are not able to receive data when in `PlayerInitialSpawn`. Therefore the one second timer. This probably works in 99.5% of the cases. But sometimes it fails and the client doesn't know about roles.

I also wrapped many `GlobalXX` variables inside descriptive functions.

Moreover I tried to clean up `sv_main`. I'm not yet done. And the mess with the replicated convars (seriously, client checks in a server file? :D) should be cleaned up as well.

This makes the changes in #1527 redundant and will probably fix some of the issues from #1360

Relevant existing functions still have their alias, so it doesn't break everything:
image


lua/ttt2/libraries/gameloop.lua
```diff
@@ -0,0 +1,540 @@
...
+    function gameloop.End(result)
+        KARMA.RoundEnd()
+
+        -- update the round state that is used to determine how the game behaves
+        SetRoundState(ROUND_POST)
+
+        -- triggers the round state output for every map setting entity
+        ents.TTT.TriggerRoundStateOutputs(ROUND_POST, result)
+
+        PrintResultMessage(result) -- todo what does this do?
```
Histalek, line 218: this just prints the winning team to the server console.

i would just leave it as is (for this PR at least)

Histalek (commented):
already looking good!

just some random comments because i couldn't help myself

TimGoll:
I already removed everything from these synced globals that had something to do with the `gameloop`. But we should remove the remaining ones as well in upcoming PRs as replicated convars have been fixed for years already.

image

Histalek (commented):
whoops. i never send these comments ...

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -844,11 +824,15 @@ function GM:PlayerDeath(victim, infl, attacker)
...
-    -- @realm server
-    -- stylua: ignore
-    if HasteMode() and GetRoundState() == ROUND_ACTIVE and not hook.Run("TTT2ShouldSkipHaste", victim, attacker) then
-        IncRoundEnd(GetConVar("ttt_haste_minutes_per_death"):GetFloat() * 60)
+    if
+        gameloop.IsHasteMode()
+        and gameloop.GetRoundState() == ROUND_ACTIVE
+        ---
+        -- @realm server
+        -- stylua: ignore
```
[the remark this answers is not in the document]
TimGoll: image
 I'm not sure this points to the correct line anymore?

TimGoll:
The last commit fixed the only issue @mexikoedi was experiencing when testing this branch

lua/ttt2/libraries/gameloop.lua
```diff
@@ -0,0 +1,873 @@
...
+        if not keepRoundCount then
+            gameloop.DecreaseRoundsLeft()
+        end
+
+        -- delay the cleanup a bit so that the client starts the loading screen animation before the
+        -- cleanup starts
+        -- note: delaying by a single tick seemed to be ineffective
+        timer.Simple(0.25, function()
+            game.CleanUpMap(false, nil, ents.TTT.FixParentedPostCleanup)
+        end)
```
Histalek, line 311: Maybe this is ping dependent?
It is probably fine to keep this as is for now, but we might need to fine-tune in the future

lua/ttt2/libraries/gameloop.lua
```diff
@@ -0,0 +1,873 @@
...
+
+        -- Remove ULX /me command. (the /me command is the only thing this hook does)
+        hook.Remove("PlayerSay", "ULXMeCheck")
```
Histalek, line 314: Do we know why we do this?
Is this a gameplay consideration or is this an "optimization" simply decreasing the amount of hooks?

Histalek (approved):
LGTM :)

Thanks for tackling this!

lua/ttt2/libraries/gameloop.lua
TimGoll: Maybe it is, I considered that as well. If I don't do that, the first netmessage triggering the loadingscreen sometimes arrives too late. I don't get why exactly, my only idea is that engine-internals use an unsafe net communication that is "faster" or independant of ticks? Not sure

lua/ttt2/libraries/gameloop.lua
TimGoll: I've ggogled for this, because I was puzzled as well. Aparantly ULX adds this hook and it breaks gamemodes such as TTT and DarkRP. Therefore both gamemodes remove this hook again.

Seems really out of place imho, but I didn't want to add new issues, so I just kept it

PR #1534, opened 2024-05-25
Voice: Make team voice chat override global voice chat

ruby0b:
- team voice chat is always push-to-talk
- while using team voice, global voice chat is temporarily disabled
- global voice chat gets restored to its previous state (on/off) afterwards
- removed plymeta:SetSpeakingInVoice and plymeta:IsSpeakingInVoice in favor of VOICE.IsSpeaking
- VOICE.IsSpeaking can now accept a non-client Player to check


gamemodes/terrortown/gamemode/client/cl_voice.lua
```diff
@@ -687,12 +686,18 @@ function VOICE.Tick()
...
 -- @realm client
-function VOICE.IsSpeaking()
-    return LocalPlayer().speaking
+function VOICE.IsSpeaking(ply)
+    ply = ply or LocalPlayer()
+    if ply == LocalPlayer() then
+        return ply.speaking
+    else
+        return ply:IsSpeaking()
+    end
```
TimGoll, line 700:
```suggestion
    if not ply or ply == LocalPlayer() then
        return LocalPlayer().speaking
    else
        return ply:IsSpeaking()
    end
```

Not sure if my implementation is better, just a suggestion

TimGoll (commented):
Looks good to me. I guess you tested the changes with the SetSpeakingInVoice that you removed?

ruby0b:
yes, I did test the SetSpeakingInVoice removal for non-client players, it works fine since it was basically just a manually updated mirror of the built-in Player:IsSpeaking()

TimGoll:
> yes, I did test the SetSpeakingInVoice removal for non-client players, it works fine since it was basically just a manually updated mirror of the built-in Player:IsSpeaking()

It's alweys good to have someone like you at hand, that knows more about a specific topic. Thank you!

TimGoll (approved):

PR #1535, opened 2024-05-27
Weaponbase: Fixed weapon dryfire sound interrupting gunshot sounds

TW1STaL1CKY:
Gunshot sounds would get interrupted by the dryfire sound, sounding quite jank.

This has been fixed by making the dryfire sound use a different channel. Now the last shot in the magazine won't be silenced if you keep holding the trigger!
The parameters in EmitSound match the settings that the "Weapon_Pistol.Empty" soundscript would use (except the channel), so it sounds exactly the same pitch and volume wise.

### DIFF
--- a/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -978,7 +978,7 @@ end
 -- @realm shared
 function SWEP:DryFire(setnext)
     if CLIENT and LocalPlayer() == self:GetOwner() then
-        self:EmitSound("Weapon_Pistol.Empty")
+        self:EmitSound(")weapons/pistol/pistol_empty.wav", 75, 100, 0.7, CHAN_VOICE)
     end
 
     setnext(self, CurTime() + 0.2)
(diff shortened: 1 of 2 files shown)


wgetJane:
`CHAN_ITEM` makes more sense, it's generally what's used for non-gunfire weapon sounds like clicking sounds and such

gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -978,7 +978,7 @@ end
 -- @realm shared
 function SWEP:DryFire(setnext)
     if CLIENT and LocalPlayer() == self:GetOwner() then
-        self:EmitSound("Weapon_Pistol.Empty")
+        self:EmitSound(")weapons/pistol/pistol_empty.wav", 75, 100, 0.7, CHAN_VOICE)
```
TimGoll, line 981:
```suggestion
        self:EmitSound("weapons/pistol/pistol_empty.wav", 75, 100, 0.7, CHAN_VOICE)
```

saibotk: Interesting, let's still switch to `CHAN_ITEM` as @wgetJane suggested :) then this is good to go!

Thanks btw for upstreaming your fixes <3

TW1STaL1CKY:
> `CHAN_ITEM` makes more sense, it's generally what's used for non-gunfire weapon sounds like clicking sounds and such

Sure, done 🙂

TimGoll (approved):

PR #1536, opened 2024-05-30
Voice Chat: Added module for battery drain and bar to UI

TimGoll:
The voice battery drain was previously removed from the UI in the voice rework. It is now added back to the UI. While doing so I moved all convars and functions to a new module called `voicebattery`. That way I was able to remove the global bools and replace them with with replicated convars.

image

I'm not sure that I'm a fan of the voice battery being client only. This seems like a bad idea to me. But I wanted to keep it that way for now, we might want to redo it at some point.


saibotk (approved):
🧼

PR #1539, opened 2024-05-31
Weapons: Minor cleanup and optimizations

TW1STaL1CKY:
I noticed a few things that could be optimised in the weapon base and grenade base so I applied them 😀 Then I thought I'd check some of the other and went on a bit of a spree of applying minor optimisations to those too (sorry if some seem nitpicky). I can't look through all of them now, not that all of them need optimising, but I feel these are the main ones.

I've tested the weapons after making the changes and they still work.
I also applied a previous fix I made to the weapon base to the binoculars where your view could get stuck zoomed in if your binoculars were removed.

By the way, the reason I removed the lines `gren:SetGravity(0.4)` and `gren:SetFriction(0.2)` from the grenade base's `CreateGrenade` function is because they do nothing to physics entities.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Cleanup and performance optimizations for marks library (by @WardenPotato)
 - Updated the Turkish localization file (by @NovaDiablox)
 - The level time now starts with the first preparing phase, meaning that idle on connect doesn't decrease the map time (by @TimGoll)
+- Minor cleanup and optimizations in weapon code (by @TW1STaL1CKY)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed `DynamicCamera` error when a weapon's `CalcView` doesn't return complete values (by @TW1STaL1CKY)
 - Fixed Roundendscreen showing karma changes even if karma is disabled
 - Fixed the player's FOV staying zoomed in if their weapon is removed while scoped in (by @TW1STaL1CKY)
+- Fixed the player's FOV staying zoomed in with the binoculars if they're removed from you (by @TW1STaL1CKY)
 - Fixed weapon unscoping (or generally any time FOV is set back to default) being delayed due to the player's lag (by @TW1STaL1CKY)
 - Fixed overhead icons sometimes being stuck at random places (by @TimGoll)
 - Fixed a null entity error in the ShootBullet function in weapon_tttbase (by @mexikoedi)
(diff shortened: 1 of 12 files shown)


gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
```diff
@@ -351,6 +351,11 @@ if CLIENT then
     -- stylua: ignore
     local cvCrosshairMode = CreateConVar("ttt_crosshair_mode", "0", FCVAR_ARCHIVE)
 
+    ---
+    -- @realm client
+    -- stylua: ignore
+    local cvEnableHUDBoxBlur = GetConVar("ttt2_hud_enable_box_blur")
```
Histalek, line 357:
```suggestion
    local cvEnableHUDBoxBlur = GetConVar("ttt2_hud_enable_box_blur")
```
These doc strings are not needed. Our Doc-tool only looks for `CreateConVar` calls

gamemodes/terrortown/entities/weapons/weapon_zm_improvised.lua
```diff
@@ -264,7 +264,7 @@ function SWEP:SecondaryAttack()
         tr.Hit
         and IsValid(ply)
         and ply:IsPlayer()
-        and (owner:EyePos() - tr.HitPos):Length() < 100
+        and (owner:EyePos() - tr.HitPos):LengthSqr() < 10000 -- 100hu
```
Histalek, line 267: Is there a specific reason why working with the squared length is better here?

Histalek (changes requested):
Nice changes, they also help with readability quite a bit :)

i've only got a small doc suggestion and a question. Otherwise LGTM!

CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Cleanup and performance optimizations for marks library (by @WardenPotato)
 - Updated the Turkish localization file (by @NovaDiablox)
 - The level time now starts with the first preparing phase, meaning that idle on connect doesn't decrease the map time (by @TimGoll)
+- Minor cleanup and optimizations in weapon code (by @TW1STaL1CKY)
```
TimGoll, line 63: you should also add a note to the fixes that you fixed the binoculars

gamemodes/terrortown/entities/weapons/weapon_zm_improvised.lua
TW1STaL1CKY: LengthSqr is more efficient than Length because [Length uses the square root function which is an expensive process](https://wiki.facepunch.com/gmod/Vector:Length). Generally if you only need to calcuate if two vectors within range of each other (ie. you don't need a human readable distance in hammer units), LengthSqr should be used instead

TimGoll: Good to know, never thought about this. We should probably adress this in our whole code base

CHANGELOG.md
TW1STaL1CKY: Added

TimGoll (approved):

Histalek (approved):
Thank you very much :)

PR #1541, opened 2024-06-01
Entity: Player Ragdoll Check

TimGoll:
As discussed in #1539 we now always check if an entity is a true ragdoll. Instead of only checking the class, we now consistently also check if the entity has a player nick to make sure our ragdoll handling is only applied to true ragdolls.


gamemodes/terrortown/gamemode/shared/sh_entity.lua
```diff
@@ -0,0 +1,18 @@
...
+    return
+end
+
+---
+-- Checks if the entity is a ragdoll. It also makes sure it is a true ragdoll
+-- by checking if it has a player nick name because ragdolls can also be used
+-- for props (e.g. a mattress).
+-- @return boolean Returns true if it is a true ragdoll
+-- @realm shared
+function entmeta:IsTrueRagdoll()
```
saibotk, line 16: Wouldn't something like `IsPlayerRagdoll` make more sense? 

"True" can be subjet to interpretation :)

saibotk (changes requested):

gamemodes/terrortown/gamemode/shared/sh_entity.lua
TimGoll: I wasn't happy with my naming either, your suggestion is quite nice, thanks!

saibotk (approved):

PR #1542, opened 2024-06-01
Is(Super)Admin: Consolidation

### LINKED ISSUE #1426: Unify IsAdmin and IsSuperAdmin use
Currently we sometimes use `ply:IsAdmin` and sometimes `ply:IsSuperAdmin`. Most notably, the scoreboard shows the admin icon even for normal admins, while all admin settigs are only visible for superadmins. This leads to confusion.

Besides that, we should probably replace all IsAdmin calls with IsSuperAdmin anyway. There is no point to allow admins to change convars, when the UI is only shown to super admins

TimGoll:
Fixes #1426 

Unifies all calls of `ply:IsAdmin` and `ply:IsSuperAdmin` by making consistent use of our `GM:TTT2AdminCheck` hook. I added a wrapper function for that hook to make is easier to use.


saibotk (approved):
Clean

PR #1544, opened 2024-06-02
Incendiary: Fixed explosion sometimes missing fire

### LINKED ISSUE #1427: Incendiary grenades explode without fire
This seems to be an issue in v0.12.x and v0.13.x. Maybe even earlier. At least from what I heard, it is to confirm if this issue existed before #1318.

When fire grenades explode next to a wall, they always spawn their fire particles behind the wall. This is confusing when there is a room next to it, but outright buggy, if there is nothing behind it because it makes the particles spawn really far out triggering the crazy physics tool that removes them completely.

TimGoll:
Fixes #1427

This pullrequests fixes the issue with incediaries not spawning fire particles (or spawning them in the next room if there is any) when exploding close to a wall. The relevant change for this fix is in the base grenade file:

image

This is not a TTT2 issue, it also [exists in vanilla TTT](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/terrortown/entities/entities/ttt_basegrenade_proj.lua#L54). The issue exists because the trace, that should detect the ground, does not ignore the grenade itself and returns therefore a kind of random HitNormal. Since the grenade is then moved along the normal prior to explosion, this often resulted in the grenade being behing a wall and therefore not spawning any particles.

Adding `self` to the filter would fix this issue as well, but after some quick thought I decided that `MASK_SOLID_BRUSHONLY` would be a fitting mask setting here. [[docs]](https://wiki.facepunch.com/gmod/Enums/MASK).

This issue probably existed with ALL GRENADES that inherited from the weapon base. I wonder why nobody besides me ever reported this. Probably because you do not notice it during normal chaotic gameplay.

Besides this change I also renamed some variables to be more in-line with our style, while also moving the fire grenade decal creation to the server. That way the decal spawns together with the explosion and not a few moments before. It looks cleaner that way.

### DIFF
--- a/gamemodes/terrortown/entities/entities/ttt_basegrenade_proj.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_basegrenade_proj.lua
@@ -56,28 +56,31 @@ end
 ---
 -- @ignore
 function ENT:Think()
-    local etime = self:GetExplodeTime() or 0
-    if etime ~= 0 and etime < CurTime() then
+    local timeExplosion = self:GetExplodeTime() or 0
+
+    if timeExplosion ~= 0 and timeExplosion < CurTime() then
+        local thrower = self:GetThrower()
+
         -- if thrower disconnects before grenade explodes, just don't explode
-        if SERVER and (not IsValid(self:GetThrower())) then
+        if SERVER and (not IsValid(thrower)) then
             self:Remove()
-            etime = 0
+
             return
         end
 
         -- find the ground if it's near and pass it to the explosion
-        local spos = self:GetPos()
+        local pos = self:GetPos()
         local tr = util.TraceLine({
-            start = spos,
-            endpos = spos + Vector(0, 0, -32),
-            mask = MASK_SHOT_HULL,
-            filter = self.thrower,
+            start = pos,
+            endpos = pos + Vector(0, 0, -32),
+            mask = MASK_SOLID_BRUSHONLY,
         })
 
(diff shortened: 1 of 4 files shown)


saibotk (approved):
Thanks 🙏

gamemodes/terrortown/entities/entities/ttt_basegrenade_proj.lua
```diff
@@ -56,28 +56,31 @@ end
...
-        local spos = self:GetPos()
+        local pos = self:GetPos()
         local tr = util.TraceLine({
-            start = spos,
-            endpos = spos + Vector(0, 0, -32),
-            mask = MASK_SHOT_HULL,
-            filter = self.thrower,
+            start = pos,
+            endpos = pos + Vector(0, 0, -32),
+            mask = MASK_SOLID_BRUSHONLY,
```
Histalek, line 76: This removes quite a lot from the mask (which is probably a good thing)

afaik this is essentially only CONTENTS_SOLID and CONTENTS_MOVEABLE [1] which certainly seems fine, but we should extensively test this

[1] https://wiki.facepunch.com/gmod/Enums/CONTENTS

TimGoll: I agree with you here. However at least all players and the entity itself should be excluded here. That would be the alternative. Maybe we should test this on more maps

Issue #1478, opened 2024-03-09
vfire-enabled explosions is not affected by explosion resistance


EntranceJew:
workshop,

![image](https://github.com/TTT-2/TTT2/assets/5711436/ca713f4e-8f59-4fc5-8e4d-384ddbcf9023)

this is more of a concern for [ttt2-ext_addon_settings](https://github.com/TTT-2/ttt2-ext_addon_settings) maybe, but still

EntranceJew:
![image](https://github.com/TTT-2/TTT2/assets/5711436/14c11ccc-1595-44f0-8d5e-b5e9269361df)
not 100% related but these functions also play nicely with vfire related things

TimGoll:
I'm unsure whether that should be added to TTT2 or a new compat addon. Best would be to have this in vFire itself, but I guess they don't care?

TimGoll:
Due to loading order shenanigans it is hard for me to reproduce this issue. I think this issue only exists if a player would die from an explosion, not if they'd only take some damage. If it is the case, I know a fix for that.

PR #1549, opened 2024-06-02
vFire: Improved Integration

TimGoll:
With TTT2 v0.13.0 we started supporting vFire in TTT2. This pullrequest improves the integration by reworking the game effects function to use the vFire particles when vFire is installed. I added a `0.5` modifier to the vFire fire particles as they were quite large and almost overwhelming. Now they are more tuned to TTT2. However I doubled the damage ticks so it more closely matches that of normal TTT.

This PR also adds a hook replacement to the vFire explosion handling reported in #1478. The issue here is that it was applied to all entities, but it shouldn't be applied to players. I couldn't really test this one as I was unable to reproduce the issue due to loading order. See the issue for more details.

Also maybe it is a good idea if @EntranceJew could take a look at this as well because he worked more with vFire than I did.

vFire docs: https://docs.google.com/document/d/1GbSrWMqJYI0f4pP0pYBiG2TsKUCXifQLjCqC0qHA9a0/edit


TimGoll:
I think neodoc hates `hook.Call`?

image

Histalek:
> I think neodoc hates `hook.Call`?
> 
> image

Nope, you were missing the gamemodeTable (2nd argument) [1] and neodoc chokes on that. We really need to revamp Neodoc sometime .-.

[1] https://wiki.facepunch.com/gmod/hook.Call

Histalek (approved):
i don't have the slightest clue about vfire so i can't really attest to the interaction here, but other than that this LGTM

TimGoll:
> i don't have the slightest clue about vfire so i can't really attest to the interaction here, but other than that this LGTM

I leave this open for a day or two so someone else can voice their opinion as well

Racpy:
It would be nice to see an improvement on this, sometimes when I buy explosion resistance I can reproduce when I stand near oil barrel and shoot it, but not every time.

TimGoll:
> It would be nice to see an improvement on this, sometimes when I buy explosion resistance I can reproduce when I stand near oil barrel and shoot it, but not every time.

Should be fixed with this PR. But I'm unable to test this due to loading order.

TimGoll:
There were some comments on discord that this might now fully resolve the issue, however I'm not able to reproduce any problems. Therefore I'll merge it for now and I will reopen the issue if the problem reappears.

PR #1551, opened 2024-06-03
Corpse: On Fire Feedback

TimGoll:
- added "corpse on fire" to targetID when looking at a corpse
- moved the warning message when trying to inspect from chat to mstack
- allow spectators to look at burning corpses
- cleanup of missing `IsPlayerRagdoll` calls


Histalek (approved):
nice!

PR #1554, opened 2024-06-11
Weapons: Shotgun changes

TW1STaL1CKY:
- Added SWEP.DryFireSound to the weapon base to allow the dryfire sound to be easily changed (now used by the shotgun)
- Applied dryfire sound fix to the shotgun
- Allowed the shotgun to attempt reloading on dryfire
- Interrupting the shotgun's reload looks and feels less jank
- Added a third person reload animation to the shotgun

Side note: while making these changes, I only now realised that the dryfire sound only plays for the player holding the weapon. I think this is fine but I thought I would ask, should we instead let the sound be audible to anyone near the player? I can make another PR if so

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Added `admin.IsAdmin(ply)` as a wrapper that automatically calls `GM:TTT2AdminCheck` (by @TimGoll)
   - Made sure this new function is used in our whole codebase for all admin checks
 - Added `ENTITY:IsPlayerRagdoll` to check if a corpse is a real player ragdoll (by @TimGoll)
+- Added the `SWEP.DryFireSound` field to the weapon base to allow the dryfire sound to be easily changed (by @TW1STaL1CKY)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


Racpy:
This was a topic I was discussing with others on the TTT2 discord the other day, so I’m glad to see this update!

personally, I think it's weird that the reload sound is only play on client, but I accept it.

TimGoll:
Regarding the dry fire sound: how is it done for other weapons?

TW1STaL1CKY:
> Regarding the dry fire sound: how is it done for other weapons?

A `SWEP.DryFireSound` field was added to weapon_tttbase and is referenced in `SWEP:DryFire`

Since the shotgun doesn't define its own `SWEP:CanPrimaryAttack` anymore, it'll use the default CanPrimaryAttack (which calls `SWEP:DryFire`) defined in the base. All the shotgun does now is set `SWEP.DryFireSound` to the shotgun empty wav path.
Any other weapon wanting to replace just the sound and nothing else can update that field.

TimGoll (approved):
I took a moment to look through these changes and they look good to me. Especially that it now attempts to reload on dry fire. It always bothered me that shotguns were slightly different. Thanks a lot!

PR #1556, opened 2024-06-13
Weapons: Fixed crowbar attack animation not being visible to the local player

TW1STaL1CKY:
Minor visual fix

Swinging your crowbar in thirdperson would have no animation (everyone else sees it fine). This PR fixes that for both primary and secondary attack


saibotk (approved):

saibotk:
Thank you very much for the fixes 🙏

PR #1558, opened 2024-06-18
fixed spectator voice duck convar typo

mexikoedi:
Fixed the issue that the spectator voice duck amount setting wasn't working as expected because of a typo in the convar name.


Histalek (approved):
Thanks!

PR #1559, opened 2024-06-18
replaced missing body search sounds

mexikoedi:
Replaced the body search sounds because they don't exist in Garry's Mod/CS:S/... .
I tried to use other fitting sounds which can be heard here: https://streamable.com/5rijf8


TimGoll:
Weirdly enough they almost sound the same to the sounds I chose, just a bit shorter and "faster". Sounds good to me!

TimGoll (approved):

PR #1561, opened 2024-06-23
Role Distribution: Add options for "derandomization" for perceptually fairer distributions

nike4613:
Randomness is bursty. Given a fair random roll, it is possible (fairly likely even) that consecutive rolls will be close to each other in value. However, we humans, with our pattern-seeking brains, see that as a pattern, and think "that *must* be wrong".

Rather famously, [a good music shuffle is not truly random](https://keyj.emphy.de/balanced-shuffle/). We can see a similar problem in TTT rounds; over some sessions, some players can consistently get traitor-aligned roles, and others consistently get innocent-aligned roles. Even though it is "fair" in that it is *truly random*, it does not feel good to be that guy who hasn't played traitor for the past two and a half hours, while another player has gotten it multiple times on every map.

This pull request aims to give server admins options to be able to fix this problem.

---

The problem of shuffling a playlist is different enough from the problem of selecting roles, so we take a different approach than is laid out in that post. The broad algorithm is this:

0. Players are initialized with a table containing a weight for all roles, where that weight is `ttt_role_derandomize_min_weight`.
1. When selecting base roles or upgrading to sub-roles, instead of picking a random player from the remaining player list with equal weight, we instead select players according to a per-role weight.
2. After role selection, we increment the weights for each role for each player by 1.
3. Finally, we reset the weights of the selected base- and sub-roles to the value of `ttt_role_derandomize_min_weight`.

The end result of this system is that players who have not gotten a role in several rounds are more likely to be selected for that role. How much each round affects that weighting can effectively be controlled by adjusting the minimum weight, with a higher minimum weight causing each round to have less effect.

The one downside of this system is that it can enable players to metagame, and guess "player A hasn't been traitor for 4 rounds, they're quite likely to be it now", which is why this mode is off by default. A note to this effect is also included in the help text in the UI for server owners. I do not believe that this should be a blocker, however: an appropriate choice of minimum weight can minimize the ability of a player to do that, and for private servers, the players can simply agree to not do that. In those cases (particularly the latter), this change could be a significant improvement over the current role distribution.

As implemented right now, the weights ALWAYS reset on map-change. I'd like to provide options to enable it to persist across map changes, possibly even with an option preventing it from persisting across sessions. I don't know what a good way to do this would be though. Any suggestions?


TimGoll:
This looks really nice. I will look at it depth as soon as I've got time!

lua/ttt2/extensions/math.lua
```diff
@@ -19,3 +19,48 @@ function math.ExponentialDecay(halflife, dt)
...
+    end
+
+    -- first, compute the sum weight
+    local sum = 0
+    for k, v in pairs(tbl) do
+        sum = sum + get_weight(v, k)
+    end
+
+    -- get the random number
+    local rand = math.random(0, sum - 1)
```
saibotk, line 53: What if the sum is < 1 ?

Also this only returns integers, according to the docs, i think you meant to use math.Rand() 

https://wiki.facepunch.com/gmod/math.Rand

lua/ttt2/extensions/math.lua
```diff
@@ -19,3 +19,48 @@ function math.ExponentialDecay(halflife, dt)
...
+
+    -- now do the prefix-sum for the final value
+    sum = 0
+    for k, v in pairs(tbl) do
+        sum = sum + get_weight(v, k)
```
saibotk, line 58: Optional: Just as a sidenote, you could also do this computation while summing up all weights and do get some precomputed prefix sum table. 

I don't think this would make any real difference in performance in our use cases, but i just wanted to let you know!

Bonus effect would be that the weights are ensured to not have changed in between those `get_weight` invocations :)

lua/terrortown/menus/gamemode/administration/roles.lua
```diff
@@ -64,6 +64,47 @@ function CLGAMEMODESUBMENU:Populate(parent)
...
+            },
+            {
+                title = "label_roles_derandomize_mode_sub_only",
+                value = 2, -- ROLE_DERAND_SUB_FLAG
+            },
+            {
+                title = "label_roles_derandomize_mode_base_and_sub",
+                value = bit.bor(1, 2), -- ROLE_DERAND_BASE_FLAG | ROLE_DERAND_SUB_FLAG
+            },
+        },
```
saibotk, line 91: Gameplay related:
Do you have any real use cases to support those rather niche configurations?

We generally try to minimize our configuration values nowadays because they are mostly overwhelming for users and i cannot think of some cases where i would even want to use any other than the `Base + Sub` or `None` configuration.

If you don't, I propose we just reduce this to a simple On/Off toggle :)

lua/terrortown/lang/en.lua
```diff
@@ -2304,3 +2304,36 @@ L.label_button_level_reset = "reset level"
 L.loadingscreen_round_restart_title = "Starting new round"
 L.loadingscreen_round_restart_subtitle = "you're playing on {map}"
 L.loadingscreen_round_restart_subtitle_limits = "you're playing on {map} for another {rounds} round(s) or {time}"
+
+-- 2024-06-23
+L.help_roles_derandomize = [[
+Role derandomization can be used to make role distribution feel more fair over the course of a session.
```
saibotk, line 2310: I think we should maybe try to find some better wording than derandomizing, maybe pseudo-random?
But idk, would also be fine its just somewhat irritating for players.

@TimGoll do you have an idea on what to call this?

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -189,6 +206,28 @@ function roleselection.SaveLayers()
...
+        local roleData = roleList[i]
+
+        if not roleData.isAbstract then
+            roleWeightTable[roleData.index] = minWeight
+        end
+    end
+
+    ply:SetRoleWeightTable(roleWeightTable)
+end
+
```
saibotk, line 230: How about we do not initialize this?
Because it will already use the minWeight when none is found and the table is then being filled inplace when it would increment each entry anyway.

I'd also remove the accessor and directly work on the table, to remove unnecessary functions.
We can then also initialize the table on the player as an empty table.

I think that would just simplify the code.

saibotk (changes requested):
Thank you very very much for the work! 

And for diving deep into the role selection that is some good amount of work!

Love how this is kept simple and how you added comments, followed stylua etc to match the codebase!

Good Job!

I only have some comments we need to discuss but otherwise this is looking very good!

lua/ttt2/extensions/math.lua
nike4613: `sum < 1` suggests either 1) zero or one table elements, which are explicitly checked for, or 2) weights that are `<= 0`, which are mostly nonsense. We need the `- 1` though because we need the upper bound to be exclusive to be correct.

And yes, integers are correct. The algorithm would work with reals, but they are not necessary.

lua/ttt2/extensions/math.lua
nike4613: That would then cause a potentially large extra table allocation for large input tables. I don't think it would necessarily have any meaningful performance improvement for any input size. Would need to measure to be sure, which can be done if there appears to be perf issues in role distribution.

lua/terrortown/menus/gamemode/administration/roles.lua
nike4613: I personally think that `Base only` is the most valuable option. The problem described in the PR description is primarily in the I/T/other distinction, which are base roles. To be honest, I added support for subroles as something of an afterthought, realizing that it would be trivial to do and may be beneficial depending on the server's set of installed roles.

lua/ttt2/extensions/math.lua
saibotk: I think it should handle floats here, even if our special case does not use it, it is not clear from the outside and as a generic math function it should support those cases

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
nike4613: We can't *just* leave it uninitialized; that was actually what my initial version did. I realized though that we still probably want to count rounds before the player gets the role the first time, which means initializing it more eagerly than I originally had it. It might be possible to initialize in the `get_weight` functions when doing role dist, but I worry that it'll miss roles for players, and lead to incorrect counting.

nike4613: This is also related somewhat to the minimum adjustment problem, where the only viable way to adjust the minimum is to fully reset the tables. If/when this table can persist across map changes and server restarts, that'll become an even bigger problem, and we'll probably need a concommand to force-reset anyway.

lua/terrortown/lang/en.lua
```diff
@@ -2304,3 +2304,36 @@ L.label_button_level_reset = "reset level"
...
+
+In essence, when it is enabled, a player's chance of recieving a role increases while they have not been assigned that role. While this can feel more fair, this also enables metagaming, where a player can guess that another will be traitor-aligned based on the fact that they have not been traitor aligned in several rounds. Do not enable this option if this is undesirable.
```
NovaDiablox, line 2312: Typo at "recieving", correct form "receiving".

lua/terrortown/lang/en.lua
TimGoll: In general you call it "weighted random" afaik. So maybe we should lean into that. Something like "Weighted role selection can be used..."

lua/terrortown/menus/gamemode/administration/roles.lua
TimGoll: In general I agree with @saibotk here. However this is handled as a single setting in a dropdown menu. So from my point of view this implementation is alright.

lua/ttt2/extensions/math.lua
```diff
@@ -19,3 +19,48 @@ function math.ExponentialDecay(halflife, dt)
...
     return exp((-0.69314718 / halflife) * dt)
 end
+
+---
+-- Gets the index of an item in the provided table, weighted according to the weights (derived from get_weight).
+-- @param table tbl The array of items to find a weighted item in.
+-- @param function get_weight Called as get_weight(item, index). Must return number.
+-- @return number
+-- @realm shared
+function math.random_weighted(tbl, get_weight)
```
TimGoll, line 29: Please keep in line with (our) lua style which would be `math.RandomWeighted` or `math.GetRandomWeighted`

in general we do not use underscores for functions and variables

TimGoll (changes requested):
Sorry for the very late reply. I was very busy with life and work. Thank you so much for your contribution. Besides some small nitpickings and the naming issue saibotk mentioned, this looks really good to me.

lua/terrortown/lang/en.lua
nike4613: I feel like "weighted" suggests something that is more directly configurable, rather than a mostly automatic. My first thought when seeing "weighted role selection" would be a mode where I (as a server owner) can adjust weights applied to each role to make them more or less likely to appear. While this is *implemented* using weights (and that is how the probabilities are configured), conceptually it is just "reduce the probability of receiving a role according to how recently you've had that role". Hence, "derandomization".

lua/ttt2/extensions/math.lua
nike4613: Hm, I don't know why I did that.

nike4613: On this note, I feel like it would probably be beneficial to include this in linting (if possible). I'm pretty sure `.editorconfig` can describe these rules, but I don't know if any of the Lua tooling available understands that.

TimGoll: @Histalek Is our expert on that matter

lua/terrortown/lang/en.lua
TimGoll: I get what you mean, let me think about it

lua/ttt2/extensions/math.lua
Histalek: this would certainly be nice to have. Sadly i'm not aware of any lua linter checking for function names out of the box

TimGoll (approved):
I think it is fine now. After reading your in-game texts I also think that your name is fitting. It doesn't really matter

saibotk (approved):

TimGoll:
@nike4613 From my POV it is finished. Let me know if you agree with the changes before I merge it

nike4613 (commented):
Is there a reason for removing the maximum?

E: Ah, fixed in a later commit.

nike4613:
> @nike4613 From my POV it is finished. Let me know if you agree with the changes before I merge it

Looks fine to me.

PR #1568, opened 2024-07-04
Voice: fix panel colors and voice chat not working after round start

ruby0b:
`GM:PlayerEndVoice` was resetting players' team voice state even if they were still talking in team voice chat because `GM:PlayerEndVoice` is called client-side whenever a player stops audibly talking (not when they are actually done using voice chat). This was unnecessary anyway so I removed the hook.

There were also multiple issues related to stuff not being updated on round start / team change (panel colors not being updated and your panel being shown but you weren't actually speaking). The simplest solution I found was forcibly re-enabling voice chat on team change.
Unfortunately I discovered that using 0s timers for re-enabling voice chat is inconsistent for whatever reason and wouldn't always call `GM:PlayerStartVoice`.
I didn't fine-tune the timing but using 0.05s (2 or 3 ticks?) seemed to fix this.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_voice.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_voice.lua
@@ -10,7 +10,7 @@ local player = player
 local IsValid = IsValid
 local hook = hook
 
-VOCIE_MODE_GLOBAL = 0
+VOICE_MODE_GLOBAL = 0
 VOICE_MODE_TEAM = 1
 VOICE_MODE_SPEC = 2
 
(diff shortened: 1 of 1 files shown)


ruby0b:
(I don't think a changelog entry would make sense here since this is a fix for unreleased code)

TimGoll (approved):
It took me a while to review this, but I'm fine with it. Thanks a lot! Regarding the weird timer: I think it is ok. In this case it is not network related, at least not lua-network related. Maybe the engine does some weird things here.

We should keep this in mind though in case some issues appear with it.

PR #1570, opened 2024-07-30
Added British English

NovaDiablox:
The changes are mostly "z" to "s" in some words.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
   - Made sure this new function is used in our whole codebase for all admin checks
 - Added `ENTITY:IsPlayerRagdoll` to check if a corpse is a real player ragdoll (by @TimGoll)
 - Added the `SWEP.DryFireSound` field to the weapon base to allow the dryfire sound to be easily changed (by @TW1STaL1CKY)
+- Added British English (by @NovaDiablox)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


saibotk:
Hey 👋 

Sorry for the late reply.

I don't feel like there is much benefit in maintaining another region specific language especially since they do not differ that much. 

So i would tend to not include this. But let's hear someone else's opinion on this.

NovaDiablox:
> Hey 👋
> 
> Sorry for the late reply.
> 
> I don't feel like there is much benefit in maintaining another region specific language especially since they do not differ that much.
> 
> So i would tend to not include this. But let's hear someone else's opinion on this.

To be honest I did it for funzies but it still holds value imo. I would like to hear from others as well.

Histalek:
There is some value to this, but i don't think it is worth the effort

Suppose we add a new translation string to the english language, would someone using the british translation realize that string to be missing? Probably not given that there are only some minor or rare differences e.g.:

- `z` -> `s`
- some `o` -> `ou` (as in `colour` or `favourite`)
- some special distinct words like `flashlight` -> `torch`

To be honest i wouldn't even be sure if our base english file is completely american english. It's probably mushed together anyway

NovaDiablox:
> There is some value to this, but i don't think it is worth the effort
> 
> Suppose we add a new translation string to the english language, would someone using the british translation realize that string to be missing? Probably not given that there are only some minor or rare differences e.g.:
> 
> - `z` -> `s`
> - some `o` -> `ou` (as in `colour` or `favourite`)
> - some special distinct words like `flashlight` -> `torch`
> 
> To be honest i wouldn't even be sure if our base english file is completely american english. It's probably mushed together anyway

Yes I agree.

PR #1574, opened 2024-08-15
Various small fixes found through static analysis

Histalek:
While playing around with the lua language server again i noticed a few things which i have fixed here.

Things marked with `FIXME:` i wasn't exactly sure about and would need some help/input on

### DIFF
--- a/lua/terrortown/menus/gamemode/gameplay/general.lua
+++ b/lua/terrortown/menus/gamemode/gameplay/general.lua
@@ -34,6 +34,5 @@ function CLGAMEMODESUBMENU:Populate(parent)
     form2:MakeCheckBox({
         label = "label_crosshair_ironsight_low_enabled",
         convar = "ttt_ironsights_lowered",
-        master = crossEnb,
     })
 end
--- a/lua/ttt2/libraries/entity_outputs.lua
+++ b/lua/ttt2/libraries/entity_outputs.lua
@@ -52,6 +52,8 @@ function entityOutputs.RegisterHook(hookName)
     entityOutputs.hooks[hookName] = true
 
     hook.Add(hookName .. "_Internal", hookName .. "_name", function()
+        -- The `ACTIVATOR` and `CALLER` globals are only available during execution
+        -- see note at https://wiki.facepunch.com/gmod/ENTITY:TriggerOutput#example
         local activator, caller = ACTIVATOR, CALLER
 
         ---
--- a/lua/ttt2/libraries/gameloop.lua
+++ b/lua/ttt2/libraries/gameloop.lua
@@ -700,7 +700,7 @@ if CLIENT then
     -- @internal
     -- @realm client
     function gameloop.RoundStateChange(oldRoundState, newRoundState)
-        if nnewRoundState == ROUND_PREP then
+        if newRoundState == ROUND_PREP then
             EPOP:Clear()
 
             -- show warning to spec mode players
@@ -754,12 +754,12 @@ if CLIENT then
             -- @realm shared
             -- stylua: ignore
             hook.Run("TTTPrepareRound")
-        elseif oldRoundState == ROUND_PREP and n == ROUND_ACTIVE then
+        elseif oldRoundState == ROUND_PREP and newRoundState == ROUND_ACTIVE then
             ---
             -- @realm shared
             -- stylua: ignore
             hook.Run("TTTBeginRound")
-        elseif oldRoundState == ROUND_ACTIVE and n == ROUND_POST then
+        elseif oldRoundState == ROUND_ACTIVE and newRoundState == ROUND_POST then
             ---
             -- @realm shared
             -- stylua: ignore
--- a/lua/ttt2/libraries/gameloop.lua
+++ b/lua/ttt2/libraries/gameloop.lua
@@ -446,12 +446,12 @@ if SERVER then
             ---
             -- @realm server
             -- stylua: ignore
-            local win = hook.Run("TTT2PreWinChecker", preventWin)
+            local win = hook.Run("TTT2PreWinChecker")
 
             ---
             -- @realm server
             -- stylua: ignore
-            win = win or hook.Run("TTTCheckForWin", preventWin)
+            win = win or hook.Run("TTTCheckForWin")
 
             if win == WIN_NONE then
                 return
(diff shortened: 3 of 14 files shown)


lua/terrortown/menus/gamemode/gameplay/general.lua
```diff
@@ -34,6 +34,6 @@ function CLGAMEMODESUBMENU:Populate(parent)
     form2:MakeCheckBox({
         label = "label_crosshair_ironsight_low_enabled",
         convar = "ttt_ironsights_lowered",
-        master = crossEnb,
+        master = crossEnb, -- FIXME: does this even work? `crossEnb` is a checkbox local to the `appearance/crosshair` settings
```
TimGoll, line 37: no, this doesn't work. This is an artifact from when this was in a different menu. But even then this should not have been tied to the state of the crosshair

lua/ttt2/libraries/entity_outputs.lua
```diff
@@ -52,7 +52,7 @@ function entityOutputs.RegisterHook(hookName)
     entityOutputs.hooks[hookName] = true
 
     hook.Add(hookName .. "_Internal", hookName .. "_name", function()
-        local activator, caller = ACTIVATOR, CALLER
+        local activator, caller = ACTIVATOR, CALLER -- FIXME: where are these defined? we should at least link to where these come from
```
TimGoll, line 55: yes, they are. These are global variables, I forgot where I got that info from. Probably the gmod wiki or the valve dev wiki. It works 100% and it is a wild concept

lua/ttt2/libraries/gameloop.lua
```diff
@@ -446,12 +446,12 @@ if SERVER then
...
             -- @realm server
             -- stylua: ignore
-            local win = hook.Run("TTT2PreWinChecker", preventWin)
+            local win = hook.Run("TTT2PreWinChecker", preventWin) -- FIXME: What do we want to send to the hooks here? the convar value? that would always be true. Do we even want to send anything here?
 
             ---
             -- @realm server
             -- stylua: ignore
-            win = win or hook.Run("TTTCheckForWin", preventWin)
+            win = win or hook.Run("TTTCheckForWin", preventWin) -- FIXME: see above
```
TimGoll, line 454: I'm not sure - isn't prevent win the convar that always prevents a win? What speaks against passing that? It can be true or false. The main TTT2 wincheck, where this convar is used, is also inside this hook

lua/ttt2/libraries/weaponrenderer.lua
```diff
@@ -131,6 +131,8 @@ function weaponrenderer.CreateModel(wep, modelData)
     then
         -- note on the rendergroup: This rendergroup is not listed in the GMod wiki; however all
         -- SWEP construction kit weapons use this rendergroup - and it works
+        -- FIXME: AFAIK this 'rendergroup' evaluates to `nil` so we would be better off simply
+        -- dropping this. Alternatively we could look into setting a proper RENDERGROUP instead
```
TimGoll, line 135: huh, it does work that way - that's all I can say here

lua/terrortown/menus/gamemode/gameplay/general.lua
Histalek: alright i'll remove this then

lua/ttt2/libraries/gameloop.lua
Histalek: jup exactly and this will always be `false` at this point because of:

```lua
        if gameloop.GetRoundState() ~= ROUND_ACTIVE or cvPreventWin:GetBool() then
            return
        end
```
a few lines above.

I'd like to simply remove this, any objections?

lua/ttt2/libraries/weaponrenderer.lua
Histalek: i mean if it works with being `nil` we should simply remove this to prevent further confusion, right?

lua/ttt2/libraries/gameloop.lua
TimGoll: I looked a bit through the code and you are right. I suspect I accidentally added in my gameloop refactor. Maybe I wanted to do something with it, but then scrapped the idea but forgot to add the parameters.

Because it wasn't in the code before my refactor:

https://github.com/TTT-2/TTT2/blob/aa961a58382d77e169dd02eff48dbd66f216e142/gamemodes/terrortown/gamemode/server/sv_main.lua#L793-L817

Also the hook, that is called here, still doesn't accept this value:

https://github.com/TTT-2/TTT2/blob/cc837c42f79254c4c28f1fb1f378bec7b3a5c2b8/gamemodes/terrortown/gamemode/server/sv_main.lua#L848-L854

lua/ttt2/libraries/weaponrenderer.lua
TimGoll: probably yes?

lua/terrortown/menus/gamemode/gameplay/general.lua
Histalek: done in [7ed4347](https://github.com/TTT-2/TTT2/pull/1574/commits/7ed4347fafa536cdf05d130afe3b4dabd48c199e)

lua/ttt2/libraries/weaponrenderer.lua
Histalek: removed in [7ed4347](https://github.com/TTT-2/TTT2/pull/1574/commits/7ed4347fafa536cdf05d130afe3b4dabd48c199e)

lua/ttt2/libraries/gameloop.lua
Histalek: removed in [7ed4347](https://github.com/TTT-2/TTT2/pull/1574/commits/7ed4347fafa536cdf05d130afe3b4dabd48c199e)

lua/ttt2/libraries/entity_outputs.lua
Histalek: done in [7ed4347](https://github.com/TTT-2/TTT2/pull/1574/commits/7ed4347fafa536cdf05d130afe3b4dabd48c199e)

Histalek (commented):
This is finished from my point of view.
Once this is approved i'll add a changelog entry for the `table.FullCopy` fix (the other changes aren't userfacing or noteworthy. The gameloop library wasn't released yet so an entry does not seem necessary to me)

TimGoll (approved):
looks good to me, thank you for investigating such small issues

Issue #1386, opened 2024-02-08
requestrevivalstatus is an expensive netmessage


EntranceJew:
## Your version of TTT2 (mandatory)

- Latest GitHub version

## Video

![image](https://github.com/TTT-2/TTT2/assets/5711436/53996e6d-246a-47f7-b184-ce70b12ba304)
![image](https://github.com/TTT-2/TTT2/assets/5711436/e8ae290f-5dee-4cf4-b097-54ca4d5c5215)

## Describe the bug (mandatory)

requestrevivalstatus is an expensive netmessage

## To reproduce

1. Find a dead person
2. Try to revive them with a defib, or potentially other means
3. Produce 60 netmessages while this occurs.

## Expected behavior

The client doesn't produce so many netmessages.

## Context (please provide as much as you can)

- The code is probably firing in UI code or some trace/tick situation 
- While it isn't long-running or time consuming to process like `ttt2_net_stream`, it does still burden the server.
- It was a highly occurring message in proportion to the amount of times it was utilized during the playsession.

EntranceJew:
![image](https://github.com/TTT-2/TTT2/assets/5711436/fa64f4e9-95d5-4f38-964f-13e1e8366e2e)
can you spot where i was holding a defib

TimGoll:
A band-aid fix would be to limit the defi to send this request only once per second or so. I would still prefer to just have that value shared

Histalek:
i'm somewhat inclined to agree with Tim on making this shared.

IIRC our last discussion about this ended up with the netmessage way because we wanted to prevent other players snooping on revives, but seeing the sheer amount of netmessages players could at least assume that someone is getting revived.

Maybe there is some middle ground to be found.

TimGoll:
> i'm somewhat inclined to agree with Tim on making this shared.
> 
> IIRC our last discussion about this ended up with the netmessage way because we wanted to prevent other players snooping on revives, but seeing the sheer amount of netmessages players could at least assume that someone is getting revived.
> 
> Maybe there is some middle ground to be found.

The cleanest solution would probably be a polling in a good intrerval. But it is way more complex then a shared variable which would reduce the traffic a lot.

Because by polling the client has to request the data. If the server only updates the data when changed, then the traffic is vastly reduced

saibotk:
I think this is not a direct issue for TTT2 but rather the defib repo.

[...]

PR #1577, opened 2024-08-16
Corpse: Improvements and Fixes

TimGoll:
This PR fixes multiple issues with corpses

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_corpse.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_corpse.lua
@@ -475,9 +475,10 @@ realdamageinfo = 0
 -- @param Player ply
 -- @param Player attacker
 -- @param CTakeDamageInfo dmginfo
+-- @param boolean realPlayerCorpse Set to true if this is a real player corpse
 -- @return Entity the CORPSE
 -- @realm server
-function CORPSE.Create(ply, attacker, dmginfo)
+function CORPSE.Create(ply, attacker, dmginfo, realPlayerCorpse)
     if not IsValid(ply) then
         return
     end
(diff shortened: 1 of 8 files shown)


gamemodes/terrortown/gamemode/server/sv_corpse.lua
```diff
@@ -490,6 +491,8 @@ function CORPSE.Create(ply, attacker, dmginfo)
         return
     end
 
+    rag:SetNWBool("real_player_corpse", realPlayerCorpse or false)
```
Histalek, line 494: do we need to network this at all? Only the server uses this if i'm not mistaken

besides that, we have `rag.player_ragdoll = true` a [few lines below](https://github.com/TTT-2/TTT2/pull/1577/files#diff-48172807435d5f9ba26d38fe099bdd5b56057fbb7ac6cbef676d68785c59877bR510)
wouldn't checking that in `FindCorpse` be enough?

Histalek (changes requested):

gamemodes/terrortown/gamemode/server/sv_corpse.lua
TimGoll: TTT2 only needs it on the server, BUt I thought it might be neat for addons such as defis to show why it fails to revive

Histalek: hmm i'm somewhat torn on this

it is indeed neat to provide this to addons, so that they don't have to sync this info themselves
but we would by default send every player what corpses are player corpses ...


i'd say if you want to make use of this in the near future (marker and/or necro defi i assume) i'm fine with it.
Otherwise i don't see a reason to include this "just in case"

TimGoll: I feel like this is the same discussion as this: https://github.com/TTT-2/TTT2/issues/1386

I thought about adding it to the defi, but I'm not sure if I like adding it to the defi or if it would seem to OP

Histalek: you are right :D

Histalek (approved):

PR #1578, opened 2024-08-16
Lang: Fixes and Improvements

### LINKED ISSUE #1565: Localization: missing/broken strings or issues
Hello there!

Long time no see but I didn't forget about you! ;)

Here is some missing or broken strings:
1. This one should use the `item_disguiser_sidebar` line, but for some reason it's not:
Снимок экрана 2024-06-25 024636
2. For some reason last words section in the UI doesn't use quickchat localization: 
image
3. There is no spaces between these lines (for example `search_kill_from_front` and `search_head`).
Снимок экрана 2024-06-25 025017
4. In old_ttt skin the `status_off` line doesn't work.

Issues:
1. These are not presented in commands. They are planned or it's just a false description?
image
2. There is something not okay with this font. I believe it doesn't support cyrillic symbols and there for some are in different font while those which looks similar to latin ones are not. That's why it's shifting and is not lined up. Could you change the font to some that supports cyrillic symbols?
image
3. Is it possible to make this list showing names differently? Like in straight line with the line break only at the end?
image

TimGoll:
Fixes #1565.

Small language and textbox improvements.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
@@ -2169,7 +2169,11 @@ function SKIN:PaintInfoItemTTT2(panel, w, h)
                 continue
             end
 
-            text_translated = text_translated .. DynT(body, params, true) .. ""
+            if i == 1 then
+                text_translated = DynT(body, params, true)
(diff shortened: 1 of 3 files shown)


Histalek (approved):
Nice!

PR #1582, opened 2024-08-17
Bindings: Fixed Undefined Bindings Breaking Gamemode

### LINKED ISSUE #1571: Unbound Keys Can Break the Game
If one uses the spawn editor and has no key bound for "walk", it causes loads of issues.

Relevant parts in code:

https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/entities/weapons/weapon_ttt_spawneditor.lua#L566

https://github.com/TTT-2/TTT2/blob/master/lua/ttt2/extensions/input.lua#L16-L22

TimGoll:
Fixed #1571

Made sure that undefined bindings do not cause errors. While I was at it, I replaced the hardcoded fallbacks with "UNK" (undefined key) so the user is less confused when nothing happens

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed `table.FullCopy(tbl)` behaviour when `tbl` contained a Vector or Angle (by @Histalek)
 - Fixed the bodysearch showing a wrong player icon when searching a fake body (by @TimGoll)
 - Fixed players respawned with `ply:Revive` sometimes spawning on a fake corpse (by @TimGoll)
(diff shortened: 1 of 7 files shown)


NovaDiablox:
Hmm can we use "???" instead. That abbreviation doesn't exist in my language (probably in others too). I can create workaround but it won't be official as a real abbreviation.

TimGoll:
> Hmm can we use "???" instead. That abbreviation doesn't exist in my language (probably in others too). I can create workaround but it won't be official as a real abbreviation.

THat is a brilliant suggestion! I sat here for minutes thinking of a good abbrevation and didn't even think of such a simple solution!

Histalek (approved):

PR #1588, opened 2024-08-20
Remove extra spaces and date swaps.

sbzlzh:


TimGoll (approved):
Thank you. It is always nice to see that other people are caring for clean looking files as well

PR #1595, opened 2024-08-22
Fetch: Removed fetching avatar icons from Steam

mexikoedi:
Avatar icons are no longer fetched from Steam and are being created via [AvatarImage](https://wiki.facepunch.com/gmod/AvatarImage) instead.
The icon should appear much faster than before and therefore the player will always see an icon. (The fetching from Steam took a while and it was possible to see the default icon instead before the real avatar icon was shown.)
Should also solve the issue which some players from China had because Steam is blocked there.
Tested it on v0.14 and it seemed to work great (saving/loading also works).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed undefined Keys breaking the gamemode (by @TimGoll)
 - Fixed markerVision elements being visible to team mates of unknown teams (such as team Innocent) (by @TimGoll)
 - Fixed inverted settings being inverted twice in the equipment editor (by @TimGoll)
+- Fixed missing avatar icons if fetching from Steam fails (by @mexikoedi)
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Updated Russian and English localization files (by @Satton2)
 - Made `ply:IsReviving` a shared player variable (by @TimGoll)
 - Updated and improved the Simplified Chinese localization file (by @sbzlzh and @TheOnly8Z)
+- Avatar icons are not fetched anymore but instead created with `AvatarImage` (fixes missing icons for chinese players) (by @mexikoedi)
 
 ### Fixed
 
@@ -106,7 +107,6 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed undefined Keys breaking the gamemode (by @TimGoll)
 - Fixed markerVision elements being visible to team mates of unknown teams (such as team Innocent) (by @TimGoll)
 - Fixed inverted settings being inverted twice in the equipment editor (by @TimGoll)
-- Fixed missing avatar icons if fetching from Steam fails (by @mexikoedi)
 
 ### Removed
 
@@ -120,6 +120,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Renamed `TTT2ModifyVoiceChatColor(ply, clr)` to `TTT2ModifyVoiceChatMode(ply, mode)`
 - Renamed `ply:GetHeightVector()` to `ply:GetHeadPosition()`
 - Removed the `TIPS` module and replaced it with a new `tips` module
+- Removed `draw.WebImage(url, x, y, width, height, color, angle, cornerorigin)` and `draw.SeamlessWebImage(url, parentwidth, parentheight, xrep, yrep, color)` from the `draw` module
 
 ## [v0.13.2b](https://github.com/TTT-2/TTT2/tree/0.13.2b) (2024-03-10)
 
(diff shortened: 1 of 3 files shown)


saibotk:
Cant we just switch to the native solution entirely?

sbzlzh:
image
image
It just feels like there's a little extra avatar for this voice

TimGoll:
> Cant we just switch to the native solution entirely?

It was me who suggested doing it that way because the AvatarImage doesn't support all resolutions etc. However I'm open to be convinced to vhange the system completely

sbzlzh:
> > 图像 图像感觉就像这个声音有一个额外的形象
> 
> 我不知道你在说什么。但在我看来，它看起来符合预期

image
That's the box that's out of bounds.

TimGoll:
It's intended that way :D

sbzlzh:
> It's intended that way :D

All right

mexikoedi:
> > Cant we just switch to the native solution entirely?
> 
> It was me who suggested doing it that way because the AvatarImage doesn't support all resolutions etc. However I'm open to be convinced to vhange the system completely

I could look into it if it is possible to remove all the fetching and instead use the solution with `AvatarImage` everywhere if this is wanted?
If it is possible then I think it would be even better (more reliable/faster) than to fetch the images from Steam.

TimGoll:
@mexikoedi if you think it is better that way, then I support that approach

sbzlzh:
What would create downloaded_assets, it doesn't look like it does to me

mexikoedi:
> What would create downloaded_assets, it doesn't look like it does to me

The folder is always created if it doesn't exist under "GarrysMod/garrysmod/data".
I haven't changed that.

sbzlzh:
> The folder is always created if it doesn't exist under "GarrysMod/garrysmod/data".
> I haven't changed that.

I mean, I didn't create the file.

sbzlzh:
I know where it is... : )

mexikoedi:
> > The folder is always created if it doesn't exist under "GarrysMod/garrysmod/data".
> > I haven't changed that.
> 
> I mean, I didn't create the file.

Maybe try it with the newest changes.
I tested it again and it worked. 
For example if I don't have the folder and then join a game it will create the folder with three images (32x32, 64x64 and 184x184) which will be then used. If I already have the folder but without the images it also creates the three images.

sbzlzh:
> > > The folder is always created if it doesn't exist under "GarrysMod/garrysmod/data".
> > > I haven't changed that.
> > 
> > 
> > I mean, I didn't create the file.
> 
> Maybe try it with the newest changes. I tested it again and it worked. For example if I don't have the folder and then join a game it will create the folder with three images (32x32, 64x64 and 184x184) which will be then used. If I already have the folder but without the images it also creates the three images.

Yes, indeed.

TimGoll (approved):
This looks good to me. I'm just wondering what happens if a player changes their avatar? The same problem probably existed before as well. But this is something we should think about and find a solution for.

mexikoedi:
> This looks good to me. I'm just wondering what happens if a player changes their avatar? The same problem probably existed before as well. But this is something we should think about and find a solution for.

Yes, this is an issue.
Should I try to solve it in this PR?
Maybe get the already existing file and comparing it with the new one. Something in this direction.

TimGoll:
If nobody complains or notifies me that they want to review it, I'll merge this later today, so I can create a prerelease GMA

TimGoll:
> > This looks good to me. I'm just wondering what happens if a player changes their avatar? The same problem probably existed before as well. But this is something we should think about and find a solution for.
> 
> Yes, this is an issue. Should I try to solve it in this PR? Maybe get the already existing file and comparing it with the new one. Something in this direction.

Not in this PR, this issue already existed before. Feel free to do that in another PR. However, if you compare it with the existing file, you downloaded the new one anyway. So there is no real benefit to comparing instead of just replacing it.

sbzlzh:
> This looks good to me. I'm just wondering what happens if a player changes their avatar? The same problem probably existed before as well. But this is something we should think about and find a solution for.

When you change the avatar, it still shows the previous avatar, but you have to delete the stuff in the downloaded_assets to make it up-to-date, but I didn't find any other problems.

TimGoll:
> > This looks good to me. I'm just wondering what happens if a player changes their avatar? The same problem probably existed before as well. But this is something we should think about and find a solution for.
> 
> When you change the avatar, it still shows the previous avatar, but you have to delete the stuff in the downloaded_assets to make it up-to-date, but I didn't find any other problems.

yes, this seems like an issue to me. But that same issue existed before this PR, so it can be fixed at a later point

mexikoedi:
> > > This looks good to me. I'm just wondering what happens if a player changes their avatar? The same problem probably existed before as well. But this is something we should think about and find a solution for.
> > 
> > Yes, this is an issue. Should I try to solve it in this PR? Maybe get the already existing file and comparing it with the new one. Something in this direction.
> 
> Not in this PR, this issue already existed before. Feel free to do that in another PR. However, if you compare it with the existing file, you downloaded the new one anyway. So there is no real benefit to comparing instead of just replacing it.

Ok, then I will fix that in another PR. Then the solution would be calling  `draw.DropCacheAvatar` in the `CreateAvatarMaterial` at the location where it checks if the file exists. Something like that?

PR #1600, opened 2024-08-26
Fetch: Fixed refreshing of avatar icons

mexikoedi:
Avatar icons are now being refreshed.

It works like this:
If you don't have any icons saved then new ones will be created, cached and used.
Otherwise the already existing icons will be deleted, then they will be created, cached and used.

So if the player changes his avatar icon the already existing ones will be replaced with the new one.

Tested it on v0.14 and it seemed to work.
(The part with the "exists" is not needed anymore because they are always cached or they don't exit.)

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
@@ -238,8 +238,8 @@ end
 net.Receive("TTT_Equipment", UpdateEquipment)
 
 -- Deletes old avatars and caches new ones
-local function CacheAllPlayerAvatars()
-    local plys = player.GetAll()
+local function CacheAllPlayerAvatars(ply)
+    local plys = IsPlayer(ply) and {ply} or player.GetAll()
 
     for i = 1, #plys do
         local plyid64 = plys[i]:SteamID64()
@@ -291,7 +291,7 @@ function GM:SetupMove(ply, mv, cmd)
     end
 
     -- Cache avatars for all players currently on the server
-    CacheAllPlayerAvatars()
+    CacheAllPlayerAvatars(ply)
 end
 
 net.Receive("TTT2NotifyPlayerReadyOnClients", function()
@@ -308,8 +308,8 @@ net.Receive("TTT2NotifyPlayerReadyOnClients", function()
     -- stylua: ignore
     hook.Run("TTT2PlayerReady", ply)
 
-    -- Cache avatars for all players currently on the server (with the new player included)
-    CacheAllPlayerAvatars()
+    -- Cache avatar of the new player
+    CacheAllPlayerAvatars(ply)
 end)
 
 ---
--- a/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
@@ -237,6 +237,16 @@ local function UpdateEquipment()
 end
 net.Receive("TTT_Equipment", UpdateEquipment)
 
+-- Deletes old avatars and caches new ones
+local function CacheAllPlayerAvatars()
+    local plys = player.GetAll()
+
+    for i = 1, #plys do
+        local plyid64 = plys[i]:SteamID64()
+        draw.RefreshAvatars(plyid64)
+    end
+end
+
 ---
 -- SetupMove is called before the engine process movements. This allows us
 -- to override the players movement.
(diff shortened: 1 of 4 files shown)


TimGoll (approved):
LGTM. That way on each connect an avatar is downloaded. This is fine imho. Are you sure that `TTT2PlayerAuthedShared` is also called for players 1 and 2 if I connect after them? I think you also have to use `TTT2PlayerReady` on the client to cache all existing players as well. Alternatively you can use TTT2FinishedLoading which is called in InitPostEntity and in OnReloaded.

mexikoedi:
> LGTM. That way on each connect an avatar is downloaded. This is fine imho. Are you sure that `TTT2PlayerAuthedShared` is also called for players 1 and 2 if I connect after them? I think you also have to use `TTT2PlayerReady` on the client to cache all existing players as well. Alternatively you can use TTT2FinishedLoading which is called in InitPostEntity and in OnReloaded.

Ah yes. I removed the part from `InitPostEntity` because it caused black images but it is needed. Now it works, tested it with a friend.

saibotk:
Could you please explain the need for the timers? 
And if actually needed, add comments to those lines in the code?

mexikoedi:
> Could you please explain the need for the timers? 
> And if actually needed, add comments to those lines in the code?

Sure. I can add code comments.
The timers are needed because otherwise the method is called too early and the result will be black images.
The timers ensure that the method is only called if everything is loaded. (The delay of 0 caused it to be run in the next `GM:Tick`.)

If I should solve it differently then I need to look into it again.
But it seemed to work well with a friend of mine.

saibotk:
Seems rather brittle, can we just move this to another location where the client can draw?
Some other init function maybe? Idk @TimGoll do you have an idea? I just don't want us to add random timers to fix something that can be solved in a deterministic way.

TimGoll:
> Seems rather brittle, can we just move this to another location where the client can draw? Some other init function maybe? Idk @TimGoll do you have an idea? I just don't want us to add random timers to fix something that can be solved in a deterministic way.

I'd use the PlayerReady hook that I introduced because I faced the same problems. At least net messages don't work before the player is ready, so I guess the drawing stuff is similar. Here, check this out @mexikoedi

https://github.com/TTT-2/TTT2/blob/fb43db193ce224f50d9056b9da6a06eb01da807e/gamemodes/terrortown/gamemode/client/cl_player_ext.lua#L250-L276

I'd put it in there after the `hook.Run`. Like I did with the scaling setup.

TimGoll:
#1602 is merged, you can go ahead and hook into it

TimGoll:
Awesome work, thank you!

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -294,6 +307,9 @@ net.Receive("TTT2NotifyPlayerReadyOnClients", function()
     -- @realm shared
     -- stylua: ignore
     hook.Run("TTT2PlayerReady", ply)
+
+    -- Cache avatars for all players currently on the server (with the new player included)
+    CacheAllPlayerAvatars()
```
TimGoll, line 312: Could you add the `ply` as a parameter to that function? There is no point in caching every player every time imho.

The function would then look like this:
```lua
local function CacheAllPlayerAvatars(ply)
    local plys = IsPlayer(ply) and {ply} or player.GetAll()

    for i = 1, #plys do
        local plyid64 = plys[i]:SteamID64()
        draw.RefreshAvatars(plyid64)
    end
end
```

TimGoll (changes requested):

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
mexikoedi: Ok, but be aware that I first tested it with only
```
local plyid64 = ply:SteamID64()
draw.RefreshAvatars(plyid64)
```
here before and it didn't work. Then I used the `CacheAllPlayerAvatars()` function here too to get it to work.

I didn't test your feedback but if you are sure that it works then it's good to go.

TimGoll: Hm, I'm not sure that it works, I thought it would work though. That's a bit weird then

mexikoedi: Nvm, tested it with a friend and it works.

TimGoll (approved):

PR #1602, opened 2024-08-28
PlayerReady: Extended Syncing

TimGoll:
As discussed with @mexikoedi, I made `GM:TTT2PlayerReady` fully shared to be used in his #1600 PR. This might cause issues with third party addons, but at least for our addons, the change is fine. And I'm pretty sure nobody else is using that hook anyway.

To test:
- is `GM:SetupMove` only called for the local player on clients? I think so, but it has to be confirmed
- does this work at all?
- is every player on the client marked as ready, no matter the order in which the players connected?

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
@@ -284,7 +284,9 @@ end
 net.Receive("TTT2NotifyPlayerReadyOnClients", function()
     local ply = net.ReadPlayer()
 
-    if not IsValid(ply) then return end
+    if not IsValid(ply) then
+        return
+    end
 
     ply.isReady = true
 
--- a/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_player_ext.lua
@@ -252,7 +252,13 @@ function GM:SetupMove(ply, mv, cmd)
         return
     end
 
-    ply.isReady = true
+    -- we make the assumption that every player, that already is connected
+    -- is ready. We can't tell for sure, but this is the best we can do here
+    local plys = player.GetAll()
+
+    for i = 1, #plys do
+        plys[i].isReady = true
+    end
 
     net.Start("TTT2SetPlayerReady")
     net.SendToServer()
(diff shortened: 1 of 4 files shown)


mexikoedi:
So I tested a bit around with a friend:
1) Yes.
2) Seems to work, yes.
3) Couldn't test it.

TimGoll:
> So I tested a bit around with a friend:
> 
> 1. Yes.
> 2. Seems to work, yes.
> 3. Couldn't test it.

If 1 and 2 are fine, number 3 works as well. Thank you so much! Then this is ready to review and we can finish your other PR after the merge.

TimGoll:
Any feedback on this PR?

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -252,7 +252,13 @@ function GM:SetupMove(ply, mv, cmd)
...
     end
 
-    ply.isReady = true
+    -- we make the assumption that every player, that already is connected
+    -- is ready. We can't tell for sure, but this is the best we can do here
+    local plys = player.GetAll()
+
+    for i = 1, #plys do
+        plys[i].isReady = true
+    end
```
Histalek, line 261: (no action needed, just me rambling)

I see why this assumption is necessary, but i don't like that this might be wrong for quite some time depending on the circumstances.

Given that i don't know if a more accurate (but more complex) system would give us any benefits i'm fine with it.

gamemodes/terrortown/gamemode/shared/sh_player_ext.lua
```diff
@@ -1067,6 +1067,9 @@ end
 -- Returns whether the player is ready. A player is ready when he is able to look
 -- around and move (first call of @{GM:SetupMove}). A bot player is always considered
 -- as ready.
+-- @note If called on the client, it is only 100% reliable for the local player. While the
+-- restult for other players is most likely still true, it can have some issues in certain
+-- scenarios.
```
Histalek, line 1072: (optional)

I personally would bump this up to a warning as this is an unexpected caveat and might trip people up more than necessary.

Histalek (approved):

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
TimGoll: I thought about this for a while and I'm not sure if there is any good solution for that. Then again, the usecase for IsReady on the client is almost always for the client itself. And for the client the state is reliable.

Moreover it is 100% reliable on the server as well and this is where it matters the most. Therefore I think this issue is neglectable.

PR #1604, opened 2024-09-13
Buttons: Further Fixes

TimGoll:
This pullrequest introduces a further fix to buttons that were introduced by the move of use calls to the client. It fixes the same issue the last PR (#1580) attempted to fix, but this time it improves upon this. Two things are done in this PR:

- added `foundButton:SetSolid(SOLID_BSP)` to make sure all buttons on the map can be actually hit by a trace. This also fixes the issue introduced in the last PR, where some buttons now had a hitbox the player could collide with
- rewrote the syncing to the client as it was a nightmare. Many entites are NULL entities on the client if the player is too far away. This new approach with NWInts fixes this because it makes use of the internal GMod systems.


TimGoll:
It was just confirmed that this PR fixes the issue

saibotk (approved):

PR #1606, opened 2024-09-14
Buttons: Added Cancelable Hook

TimGoll:
Added a cancelable hook that is triggered on button use. This is needed for an addon that I want to create that adds further info to buttons (like a description of the button depending on its state)

```lua
---
-- A hook that is called before a button is pressed. Can be used to cancel the event by returning false.
-- @param Player ply The player that pressed the button
-- @param Entity ent The button entitiy
-- @param boolean oldState The old toggle state of the button before it was pressed
-- @return boolean return false to prevent the player from using that button
-- @hook
-- @realm server
function GM:TTT2OnButtonUse(ply, ent, oldState)
```

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_player.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player.lua
@@ -595,7 +595,7 @@ end)
 ---
 -- A hook that is called before a button is pressed. Can be used to cancel the event by returning false.
 -- @param Player ply The player that pressed the button
--- @param Entity ent The button entitiy
+-- @param Entity ent The button entity
 -- @param boolean oldState The old toggle state of the button before it was pressed
 -- @return boolean return false to prevent the player from using that button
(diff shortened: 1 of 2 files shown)


gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -581,6 +592,16 @@ net.Receive("TTT2PlayerUseEntity", function(len, ply)
     end
 end)
 
+---
+-- A hook that is called before a button is pressed. Can be used to cancel the event by returning false.
+-- @param Player ply The player that pressed the button
+-- @param Entity ent The button entitiy
```
saibotk, line 598:
```suggestion
-- @param Entity ent The button entity
```

saibotk (changes requested):

saibotk (approved):

PR #1608, opened 2024-09-15
Buttons: Added Serverside Fallback

TimGoll:
Since there were still issues with clientside use detection (at least on the map `ttt_mc_jondome` one button refused to work), I added a serverside fallback. If the client doesn't detect any entity on use press, the server uses the engine function to detect any use entity close by. This works.

Only downside is that it triggers uses for entities where the targetID is just slightly out of range. However this is how it has been in 0.13.2 and earlier as well.

Update: In the last commit I implmented `IsUsableEntity` from the sourceSDK in lua. All important links to the original code are in the added code. I had to do this because enties without a `+use` assigned could block the trace. Now it uses the aforementioned serverside fallback if an entity has no use.

image

https://discord.com/channels/442107660955942932/467653559408656384/1285124499401998356

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_entity.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_entity.lua
@@ -2,6 +2,12 @@
 -- @ref https://wiki.facepunch.com/gmod/Entity
 -- @class Entity
 
+-- Caps taken from here: https://github.com/ValveSoftware/source-sdk-2013/blob/55ed12f8d1eb6887d348be03aee5573d44177ffb/mp/src/game/shared/baseentity_shared.h#L21-L38
+FCAP_IMPULSE_USE = 16
+FCAP_CONTINUOUS_USE = 32
+FCAP_ONOFF_USE = 64
+FCAP_DIRECTIONAL_USE = 128
+
 local safeCollisionGroups = {
     [COLLISION_GROUP_WEAPON] = true,
 }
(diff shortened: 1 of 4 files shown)


TimGoll:
image

confirmed to finally resolve all issues with buttons

gamemodes/terrortown/gamemode/server/sv_entity.lua
```diff
@@ -2,6 +2,12 @@
 -- @ref https://wiki.facepunch.com/gmod/Entity
 -- @class Entity
 
+-- Caps taken from here: https://github.com/ValveSoftware/source-sdk-2013/blob/55ed12f8d1eb6887d348be03aee5573d44177ffb/mp/src/game/shared/baseentity_shared.h#L21-L38
+FCAP_IMPULSE_USE = 16
+FCAP_CONTINUOUS_USE = 32
+FCAP_ONOFF_USE = 64
+FCAP_DIRECTIONAL_USE = 128
```
Histalek, line 9: Documenting these is a good idea, but we don't need them accessible via code IMO.

We can simply precompute these into a single value and use that value below instead of the bitwise-or

gamemodes/terrortown/gamemode/server/sv_entity.lua
```diff
@@ -69,3 +75,28 @@ function entmeta:Spawn()
...
+-- @realm server
+function entmeta:IsUsableEntity(requiredCaps)
+    requiredCaps = requiredCaps or 0
+
+    local caps = self:ObjectCaps()
+
+    if
+        bit.band(
+                caps,
+                bit.bor(FCAP_IMPULSE_USE, FCAP_CONTINUOUS_USE, FCAP_ONOFF_USE, FCAP_DIRECTIONAL_USE)
```
Histalek, line 93: replace this with the precomputed value above

gamemodes/terrortown/gamemode/server/sv_entity.lua
Histalek: > You can pass these to the function if you want to get a specific check, so we need them.

Ah fair.

> How would I document them? We never documented enums so far

I meant documenting what fcaps we use in general (e.g. in a full text comment) not as full doc comment. It's like it is now though

Histalek (approved):

PR #1610, opened 2024-09-16
Buttons: Post Button Init Hook

TimGoll:
Added a hook that is called AFTER all buttons on the map are initialized. This happens on hotreload and post map cleanup. I added this hook because I need it in my button name addon so that I don't have to rebuilt the existing list again.

```lua
---
-- Hook that is called after all buttons were initialized on the map. This happens after
-- a map cleanup when a new round starts or when the list is rebuilt after a hotreload.
-- @param table buttonList A table with all buttons found on the map
-- @hook
-- @realm server
function GM:TTT2PostButtonInitialization(buttonList)
```


Histalek (approved):
LGTM

PR #1613, opened 2024-09-17
Plyspawn: Find Spawn on Ground

TimGoll:
While working on the button addon I noticed that `plyspawn.GetSpawnPointsAroundSpawn` doesn't actually make sure that the spawn is on the ground. This change fixes that. This should also slightly improve spawns on maps with too few spawn points.


Histalek (approved):
(optional)
Have you taken a look at https://wiki.facepunch.com/gmod/navmesh.GetGroundHeight ? 
This would save us the manual trace and the 32 units of empty space check also sounds nice.

TimGoll:
> (optional) Have you taken a look at https://wiki.facepunch.com/gmod/navmesh.GetGroundHeight ? This would save us the manual trace and the 32 units of empty space check also sounds nice.

Oh, I will check that out, sounds perfect

PR #1614, opened 2024-09-18
TargetID: Fixed and Improved Vehicle Integration

TimGoll:
Fixed the targetID bug where the player would see their own name in targetID while they were in a vehicle: https://www.youtube.com/shorts/qAhQHJRp7eo

While I was at it, I added targetID and KeyInfo integration to it as well.


Histalek (approved):
LGTM

PR #1621, opened 2024-09-21
Shopeditor: Item Buyability Broken due to Cache State

NickCloudAT:
This pullrequest fixes the issue where items (not weapons) had the wrong state if a custom show was created. This means that they were always marked as selected, even if they weren't. This was due to the caching being done too early, before the items were initialized at all.

The issue didn't exist for weapons because their setup is done in the engine before any lua init code is executed.


ZenBre4ker (approved):

TimGoll:
It is "too", no "to" in front of early :D

TimGoll (approved):

PR #1624, opened 2024-09-22
Radar move from shop UI

ZenBre4ker:
This moves the radar item to a SWEP and therefore eliminates the use of the equipment tab to manually trigger the radar.

Also I cleaned up the weird RADAR class and put it into a library. It now handles only the Target rendering of the Radar and of corpses, bombs and dna samples.

In this draft missing are sounds. Also the current weapon mesh is the radio, I thought to put it in another holding position to distinguish it a bit from the radio.


gamemodes/terrortown/entities/weapons/weapon_ttt_radar.lua
```diff
@@ -0,0 +1,473 @@
...
+SWEP.Primary.Delay = 1.0
+
+SWEP.Secondary.ClipSize = -1
+SWEP.Secondary.DefaultClip = -1
+SWEP.Secondary.Automatic = false
+SWEP.Secondary.Ammo = "none"
+SWEP.Secondary.Delay = 1.0
+
+SWEP.Kind = WEAPON_EQUIP2
+SWEP.CanBuy = { ROLE_TRAITOR, ROLE_DETECTIVE } -- only traitors can buy
```
TimGoll, line 54:
```suggestion
SWEP.CanBuy = { ROLE_TRAITOR, ROLE_DETECTIVE } 
```

gamemodes/terrortown/entities/weapons/weapon_ttt_radar.lua
```diff
@@ -0,0 +1,473 @@
...
+SWEP.LimitedStock = true -- only buyable once
+SWEP.WeaponID = EQUIP_RADAR or 2
```
TimGoll, line 56: I don't think this is needed anymore

TimGoll (commented):
I just looked at the first file so far. One thing I wondered: Some roles were able to modify their radar time (in code) - is this still working after that change?

ZenBre4ker:
No, I removed that part. Seemed kinda silly. But that was at a point where I tried to unpack the radar library, so I could probably easily bring it back.

Also the radar can now be dropped/conquered, so the balance of that item has changed.

TimGoll:
> No, I removed that part. Seemed kinda silly. But that was at a point where I tried to unpack the radar library, so I could probably easily bring it back.

I already removed that from the marker, the necro and supervillain, because they "misused" the radar for other things, such as corpses, crystals etc.

I'm not sure if any third party addons still use this, probably not I'd guess?

ZenBre4ker:
Dont think I'll finish it.

Issue #1411, opened 2024-02-13
Outline library does not behave well with non-screen-sized render targets


nike4613:
## Your version of TTT2 (mandatory)

Leave one of the following:

- Workshop version from Steam

## Describe the bug (mandatory)

When `render.RenderView(...)` is called with a non-screen-sized render target (like multiplier zoom TFA scopes do), the only thing that is visible in the output is a sliver of outline and whatever is inside the outline. (This makes the above scopes useless when there is an outline, because the rest of the world will appear black.)

## To reproduce

Steps to reproduce the behaviour:

1. Install TFA Base + TFA Insurgency: Shared Parts + at least one weapon which can use multiplier sights (such as [the P90](https://steamcommunity.com/sharedfiles/filedetails/?id=2928836309))
2. Spawn in that weapon in TTT, with at least one entity added to outline (my testing was done with [this addon](https://steamcommunity.com/sharedfiles/filedetails/?id=3029677266), using a fellow traitor)
3. Open the TFA customization menu (this gets a bit strange-- you have to bind a key in Sandbox, Q -> Utilities -> Keybinds) and enable the PO 4x24P scope (or any other scope that includes `Nx zoom` in the description; percentage-zoom scopes don't have the issue)
4. ADS, and observe a black world except for the outline

## Expected behaviour

The world should be fully visible through the scope. It'd be nice if the outline renders in the scope, but not necessary.

## Context (please provide as much as you can)

- Collection link of your Server: https://steamcommunity.com/workshop/filedetails/?id=2168680613
- No errors, client or server
- I have several bespoke addons to tweak or fix various parts of the game, none of which are currently available anywhere. (One of them "fixes" this problem, though fairly jankily; it's source is included below)

Without fix:
![image](https://github.com/TTT-2/TTT2/assets/10411129/e66c9328-e34c-43be-b45e-a4bc52ecd2d5)

With fix:
![image](https://github.com/TTT-2/TTT2/assets/10411129/c7bf728d-b5ab-4f5f-9166-d7d37837489c)

Fix addon:

`lua/terrortown/autorun/shared/fix_ttt2_outline.lua`:
```lua
local render = render
local hook = hook
local cam = cam

if SERVER then
  AddCSLuaFile()
  return
end

if not TTT2 then return end

local cvDisableOutlineScoped = CreateClientConVar("cl_tfa_outline_disable_scoped", "0", true, false)

local is_reload = false
[...]

PR #1628, opened 2024-09-22
Rework the outline library and fix the outstanding issues with it

WardenPotato:
Its finally here, the rework of the outline library which started because of https://github.com/TTT-2/TTT2/issues/1411
Ive spent quite some time on this trying over 50 different approaches, running into various setbacks each time.
Some of them got solved: https://github.com/Facepunch/garrysmod-issues/issues/5744
And some didnt: https://github.com/Facepunch/garrysmod-requests/issues/2308

But this is what i consider the best thats possible in gmod today in respect to functionality and performance.

This rework includes(i may have forgotten some, its been a while):
- A switch to using the zbuffer instead of line of sight checking every single entity
- Outline thickness option
- Multiple render types for outlines
- More performant drawing
- Drawing not breaking when scopes are used in weapon addons
- Full documentation
- Fixed outlines being thinner on one side than the other

The comments in the code explain a lot so i hope those answer most questions.

### DIFF
--- a/lua/ttt2/libraries/outline.lua
+++ b/lua/ttt2/libraries/outline.lua
@@ -60,9 +60,6 @@ local ENTS = 1
 local COLOR = 2
 local MODE = 3
 
---FIXME: You can decide if you want this to be a option for users, should be adressed before merge
-local enable_thin_line_workaround = false
-
 local function SetRenderType(render_type)
     if
         render_type ~= OUTLINE_RENDERTYPE_BEFORE_VM
@@ -394,7 +391,7 @@ local function Render()
     -- when its only 1px thick, so if AA is on we add 1px to compensate.
     -- This is done in all cases to still differentiate between 1px and 2px outlines
     local outline_thickness = OutlineThickness
-    if enable_thin_line_workaround and cvMaterialAntialias:GetInt() ~= 0 then
+    if cvMaterialAntialias:GetInt() ~= 0 then
         outline_thickness = outline_thickness + 1
     end
 
--- a/lua/ttt2/libraries/outline.lua
+++ b/lua/ttt2/libraries/outline.lua
@@ -184,7 +184,7 @@ function outline.Add(ents, color, mode, render_type, outline_thickness)
 
     -- Support for passing Entity as first argument
     if not istable(ents) then
-        ents = {ents}
+        ents = { ents }
     end
 
     -- Do not pass empty tables
--- a/lua/ttt2/libraries/outline.lua
+++ b/lua/ttt2/libraries/outline.lua
@@ -217,8 +217,8 @@ function outline.RenderedEntity()
 end
 
 local function RenderModels(render_ents)
-    for j = 1, #render_ents do
-        local ent = render_ents[j]
+    for i = 1, #render_ents do
+        local ent = render_ents[i]
 
         if IsValid(ent) then
             RenderEnt = ent
--- a/lua/ttt2/libraries/outline.lua
+++ b/lua/ttt2/libraries/outline.lua
@@ -441,6 +441,7 @@ end)
 
 ---
 -- A rendering hook that is run in @{GM:PostDrawEffects}, @{GM:PreDrawEffects} and @{GM:PreDrawViewModels} before the outlines are rendered.
+-- @param number render_type [OUTLINE_RENDERTYPE_BEFORE_VM, OUTLINE_RENDERTYPE_BEFORE_EF, OUTLINE_RENDERTYPE_AFTER_EF]
 -- @2D
 -- @hook
 -- @realm client
--- a/lua/ttt2/libraries/outline.lua
+++ b/lua/ttt2/libraries/outline.lua
@@ -25,20 +25,20 @@ OUTLINE_MODE_BOTH = 0
 OUTLINE_MODE_NOTVISIBLE = 1
 OUTLINE_MODE_VISIBLE = 2
 
-OUTLINE_RENDERTYPE_BEFORE_VM = 0	-- Render before drawing the view model
-OUTLINE_RENDERTYPE_BEFORE_EF = 1	-- Render before drawing all effects (after drawing the viewmodel)
-OUTLINE_RENDERTYPE_AFTER_EF = 2		-- Render after drawing all effects
+OUTLINE_RENDERTYPE_BEFORE_VM = 0 -- Render before drawing the view model
+OUTLINE_RENDERTYPE_BEFORE_EF = 1 -- Render before drawing all effects (after drawing the viewmodel)
+OUTLINE_RENDERTYPE_AFTER_EF = 2 -- Render after drawing all effects
 
 local Lists = {
     [OUTLINE_RENDERTYPE_BEFORE_VM] = {},
     [OUTLINE_RENDERTYPE_BEFORE_EF] = {},
-    [OUTLINE_RENDERTYPE_AFTER_EF] = {}
+    [OUTLINE_RENDERTYPE_AFTER_EF] = {},
 }
 
 local ListsSize = {
     [OUTLINE_RENDERTYPE_BEFORE_VM] = {},
     [OUTLINE_RENDERTYPE_BEFORE_EF] = {},
-    [OUTLINE_RENDERTYPE_AFTER_EF] = {}
+    [OUTLINE_RENDERTYPE_AFTER_EF] = {},
 }
 local RenderEnt = NULL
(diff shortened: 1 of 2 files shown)


lua/ttt2/libraries/outline.lua
```diff
@@ -20,37 +25,172 @@ OUTLINE_MODE_BOTH = 0
...
-local storeTexture = render.GetScreenEffectTexture(0)
-local drawTexture = render.GetScreenEffectTexture(1)
+local copy_mat = Material("pp/copy")
 
 local ENTS = 1
 local COLOR = 2
 local MODE = 3
 
+--FIXME: You can decide if you want this to be a option for users, should be adressed before merge
+local enable_thin_line_workaround = false
```
TimGoll, line 64: we do have a performance section in the appearance menu. You could add it there if you like, enabled by default probably

lua/ttt2/libraries/outline.lua
```diff
@@ -20,37 +25,172 @@ OUTLINE_MODE_BOTH = 0
...
-function outline.Add(ents, color, mode)
+function outline.Add(ents, color, mode, render_type, outline_thickness)
+    -- Make the default behaviour comply with older revisions
+    if not render_type then
+        render_type = OUTLINE_RENDERTYPE_AFTER_EF
+    end
+
+    if not outline_thickness then
+        outline_thickness = 1
+    end
```
TimGoll, line 169: is there a reason you didn't do `outline_thickness = outline_thickness or 1`, same for the other checks above

lua/ttt2/libraries/outline.lua
```diff
@@ -20,37 +25,172 @@ OUTLINE_MODE_BOTH = 0
...
+
+    if not SetOutlineThickness(outline_thickness) then
+        return
+    end
+
+    -- Create list if it doesnt exist
+    InitializeCurrentListIfNeeded()
+
+    -- Get List and ListSize for this render type and outline thickness
+    local List, ListSize = GetCurrentList(), GetCurrentListSize()
```
TimGoll, line 184: variables in our lua style are always lower case - or is here a reason for this decision?

lua/ttt2/libraries/outline.lua
```diff
@@ -75,137 +216,237 @@ function outline.RenderedEntity()
...
-        render.SetStencilReferenceValue(i)
-
-        surface.SetDrawColor(List[i][COLOR])
-        surface.DrawRect(0, 0, w, h)
-    end
-
-    cam.End2D()
-    -- end cam modification
+local MATERIAL_WHITE = Material("models/debug/debugwhite")
+local mat_antialias = GetConVar("mat_antialias")
```
TimGoll, line 231: is there a reason you chose two different naming styles? If you want to be in line with our style, it would be `materialDebugWhite` and `materialAntialias`

TimGoll (commented):
I can't say much about the rendering stuff, but it does look really good to me, awesome work!

lua/ttt2/libraries/outline.lua
WardenPotato: the naming i used here mostly reflects how its done by valve in the source engine, in there convars are snakecased similar to how they are in the console because it both communicates that its a convar and what that convar is without having to look it up. Midway this PR ive changed my personal preferences for naming aswell which can be seen

TimGoll: oh, the second one is no material, I'm blind :D

you can keep it that way or rename it to our style which is `cvEnableAntialias`

lua/ttt2/libraries/outline.lua
WardenPotato: because of https://wiki.facepunch.com/gmod/list
though we could find another name

TimGoll: Ah, I see. I'd prefer a different name then. Something like `entsList`, `renderTargetList` or so maybe?

lua/ttt2/libraries/outline.lua
WardenPotato: It really hinges on whether you find the appearance of them being slighty thicker fine, its super subjective but it does look better than the thin line scenarios where it can get cutoff

lua/ttt2/libraries/outline.lua
TimGoll: fair, I don't care that much. I just assumed it had a performance impact and was chosen for that reason

lua/ttt2/libraries/outline.lua
TimGoll: you know what, if the performance is mostly similar, chose whatever you believe looks best and set it as default

lua/ttt2/libraries/outline.lua
WardenPotato: Screenshot_20240922_232117
Screenshot_20240922_232107
Bytecode seems to be the exact same so either will do fine, just a style preference at that point

TimGoll: Keep it as is, that's fine :)

lua/ttt2/libraries/outline.lua
WardenPotato: Ah i remember why i did that, theres also a global Lists and ListSize, i used shadowing locals there to make development easier keeping it portable between different parts of the module, but it does break the no uppercase for non globals rule, can adjust that if you still want it

lua/ttt2/libraries/outline.lua
WardenPotato: no perf difference, it just adjusts the offset by 1 pixel, theres 1 edge case where this looks worse and thats if you have a 1 pixel thick line of geometry on a model being outlined but thats exceptionally rare and even then you'd have to zoom in to notice it, the problem with lines being too thin is a lot more common and noticeable, i'll leave what looks best to you guys in this case because my opinion keeps flipping dependant on which ive been looking at longer so im kinda neutral on it

TimGoll: to me your discription sounds like it should be enabled

TimGoll:
Okay, I tried tackling some of the warnings here and there is one thing I'd like you to change: `List`, `Lists`, `ListSize`, `ListsSize` are way too similar and I have no clue what they are about. Maybe make the name a bit more descriptive.
I don't know what the performance inpact would be, but it would also make sense to put it into the `outline` scope imho.

Once you get that done and the workflow errors are fixed, I can merge this PR

WardenPotato:
Performance impact should be zero because in the worst case i can localise the global anyways, ill get to changing the names then

nike4613:
What still needs to be done on this for it to be mergeable?

WardenPotato:
> What still needs to be done on this for it to be mergeable?

The workflow errors need fixing

There was a suggestion to make the optional thin line fix a toggleable option in the settings menu

and TimGolls comment about variable naming

saibotk:
Jeah we can take over this and fix the rest.
Just a bit slow atm due to vacancies etc. @Histalek and I will try to get this in soon

lua/ttt2/libraries/outline.lua
Histalek: i'm with tim on this one

enabled unconditionally in 4fa8e774e8d40c943d204ce3aa653979cfe3bfe4

lua/ttt2/libraries/outline.lua
Histalek: the inconsistency is already there with how gmod itself handles things ... besides this was fine before so i don't see a reason to change this now

Histalek:
i'm struggling with adding a proper changelog to this, otherwise this should now be good to go IMO

TimGoll:
> i'm struggling with adding a proper changelog to this, otherwise this should now be good to go IMO

@WardenPotato could you summarize your work as a changelog? Otherwise I'd just mention these things:

Change: Outline now support different thicknesses
Fixed: Outline Interactions with certain weapon's scopes
Fixed: Outlines not rendering uniformly

WardenPotato:
> > i'm struggling with adding a proper changelog to this, otherwise this should now be good to go IMO
> 
> @WardenPotato could you summarize your work as a changelog? Otherwise I'd just mention these things:
> 
> Change: Outline now support different thicknesses Fixed: Outline Interactions with certain weapon's scopes Fixed: Outlines not rendering uniformly

whats in the pull description is pretty much what has been done, i dont have more to add to that

TimGoll (approved):

saibotk (approved):

PR #1639, opened 2024-09-28
Loadingscreen: Fixed loading screen alignment and text

### LINKED ISSUE #1630: Loading screen alignment and text issue

The loading screen seems to not fit the whole screen. There is a visible gap at the top and the left side of the screen.
There also seems to be an issue with the text. It looks like there are broken words at the top of the text.

## To reproduce

1. Go into a round
2. Wait until the loading screen appears
3. See described issue

## Expected behaviour

The loading screen should fill the whole screen and the text shouldn't look broken.

- Screenshot:
ttt2_loading_screen

## Additional context

Tested on a 1920x1080p monitor.

mexikoedi:
This PR fixes #1630 .
I tested it and the loading screen fills the whole screen now (the position and the height/width was adjusted) and the text has no artifacts above it anymore.
The issue with the text artifacts was caused because of the scale value.
If the scale is a multiple of 0.2 then the artifacts appear above the text otherwise not.
Maybe the issue lies in `draw.AdvancedText` and how it scales (vector).
I couldn't find the issue there but if someone sees the issue there or if someone knows a better way to fix the text then I can adjust the code.
Screenshot:
ttt2_loading_screen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Fixed
 
 - Fixed missing translation for None role error by removing it (by @mexikoedi)
+- Fixed loading screen alignment and text issue (by @mexikoedi)
 
 ## [v0.14.0b](https://github.com/TTT-2/TTT2/tree/v0.14.0b) (2024-09-20)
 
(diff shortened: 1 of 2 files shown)


TimGoll:
Thank you, but I think the issue is somewhere else. The hook I'm using is probably wrong. After a discussion with @Histalek I came to the conclusion that this hook doesn't provide a real 2D rendering context anymore, hence the artifacts. I will try to fix it with a hook change.

The advanced font module has issues as well. They are a bit more complex to fix. Fixing them would resolve the blurry font issue that you get at scales that are not 1.0, 1.5 and 2.0.

For the future I'd recommend you to discuss such things with us first so you don't spend time fixing a bug that is already resolved in a different way

Issue #1635, opened 2024-09-25
Use Interactions Sometimes Pick-Up Random Weapons Nearby


TimGoll:
TTT2 v0.14.0

When trying to inspect a body whose center point is too far away on the server, the player sometimes picks up weapons that are close by.

TimGoll:
This is probably related to #1634

The pick up of the seemingly random weapon my come from #1608

PR #1641, opened 2024-09-28
Lua ent use fix

### LINKED ISSUE #1634: Use Interactions: Problems for larger objects
TTT2 v0.14.0

Due to the way the server checks the usability of entities, it can happen that the player is out of the 150 units tolerance range, especially for larger entities. Maybe the tolerance range should be determined by getting half the three dimenional size of the bounding box?

TimGoll:
This fixes #1634 and #1635

The issue this PR fixes is that previously the fallback was called for all SENTs (scripted entities) because they lack any engine related use flags. Therefore the clientside trace was completely ignored for this kind of entity.

I added support for our SENTs to `entmeta:IsUsableEntity`. This might look a bit hacky, but let me explain:
- First, the obvious, if an entity has `.CanUseKey` set to true, it has a use functionality attached. This is a good check, because this is a flag that was needed in vanilla TTT as well, so it has good compatibility.
- Second, there are special cases for weapons and ragdolls. This is a bit hacky indeed. However it is handled the same way in the server use handler, so I guess it is fine

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Fixed
 
 - Fixed missing translation for None role error by removing it (by @mexikoedi)
+- Fixed sometimes entity use triggering the wrong or none entity (by @TimGoll)
 
 ## [v0.14.0b](https://github.com/TTT-2/TTT2/tree/v0.14.0b) (2024-09-20)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Fixed
 
 - Fixed missing translation for None role error by removing it (by @mexikoedi)
-- Fixed sometimes entity use triggering the wrong or none entity (by @TimGoll)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Fixed
 
 - Fixed missing translation for None role error by removing it (by @mexikoedi)
+- Fixed sometimes entity use triggering the wrong or none entity (by @TimGoll)
```
Histalek, line 10:
```suggestion
- Fixed sometimes entity use triggering the wrong or no entity (by @TimGoll)
```

gamemodes/terrortown/gamemode/server/sv_entity.lua
```diff
@@ -87,6 +93,17 @@ function entmeta:IsUsableEntity(requiredCaps)
 
     local caps = self:ObjectCaps()
 
+    -- special case: TTT specific lua based use interactions
+    -- when were looking for specfically the lua use, return false it not set
```
Histalek, line 97:
```suggestion
    -- when were looking for specifically the lua use, return false it not set
```

Histalek (approved):
LGTM just some typo nits

PR #1642, opened 2024-09-28
Loadingscreen: Render Hook Change

### LINKED ISSUE #1630: Loading screen alignment and text issue

TimGoll:
Fixes #1630

Fixes the loadingscreen rendering issues that appeared because no true 2D rendering context was provided. This now also means that everything is blurred, even vgui. I think that is fine.

It would be nice if @mexikoedi could confirm that this also fixes the issue on his PC.

image


mexikoedi:
Yep, it works.

Histalek (approved):

PR #1643, opened 2024-09-28
Loadingscreen: Fixed Disabling

### LINKED ISSUE #1622: Bug: Disabled Loadingscreen leads to round not starting
- "v0.14.0b - Github Version"

 [ttt2] addons/ttt2/lua/ttt2/libraries/loadingscreen.lua:73: attempt to perform arithmetic on field 'timeBegin' (a nil value)
  1. End - addons/ttt2/lua/ttt2/libraries/loadingscreen.lua:73
   2. Prepare - addons/ttt2/lua/ttt2/libraries/gameloop.lua:119
    3. unknown - addons/ttt2/gamemodes/terrortown/gamemode/server/sv_main.lua:594

## To reproduce

1. Disable Loadingscreen
2. Restart server
3. Log in
4. Spawn Bots
5. Get Error

## Expected behaviour

No error

Apparently the variable loadingscreen.timeBegin is not available, missing initializer? Not sure if there would be any sideeffects by just making a convar check in loadingscreen.end as well or just initializing it. But the code in .end should not be run, if the loading screen was never shown. Otherwise if the screen was shown, that should be called to reset some variables, even if the convar was changed at some point.

TimGoll:
Fixes #1622

Fixes disabling of the loadingscreen breaking the game. Also I fixed the loading screen displaying one fewer rounds than are actually left because the game counts to 0 and it makes no real sense that in the last round are 0 rounds left.

I also made sure that the prepare time is set to the correct value if the loading screen is disabled.


lua/ttt2/libraries/tips.lua
```diff
@@ -72,7 +72,7 @@ function tips.Initialize()
     tips.Register("tip31")
     tips.Register("tip32")
     tips.Register("tip33")
-    tips.Register("tip34", { helpkey = Key("+gm_showhelp", "F1") })
```
saibotk, line 75: Why is this change in here?

saibotk (changes requested):

lua/ttt2/libraries/tips.lua
TimGoll: Because I noticed that this is unused and I didn't want to open a PR for this single line

saibotk: Okay but this is in general confusing as it is also not mentioned in the pr description.

saibotk (approved):

PR #1660, opened 2024-10-02
ci(neodoc): Bump to v0.2.0

Histalek:
With this we should now never need `-- stylua: ignore` again.

Please let me merge this, the commits need to stay intact for the git-blame ignore to work


TimGoll:
oh, nice. So there is no longer a single 300 character line when calling a hook! :D

TimGoll (approved):
I haven't reviewed every line, but I suppose it will work

PR #1662, opened 2024-10-08
Fix weapon pickup through walls

MrXonte:
Add check for trace start solid to prevent teleporting weapon through walls.

Edgecases can occur where weapons can be picked up through walls (even thick 16 unit walls didn't prevent this). This is because the trace can start solid (most likely if the weapon is too large for the space its resting, in my case a relatively small func_detail shelf) and then the tr.Fraction will return 1.0 through a solid wall, causing weapons the be teleported and picked up.


MrXonte:
Yep! Absolutely no issue without auto pickup since you cant select the weapons through walls, unfortunately people were raiding my traitor room on my current mapping project with auto pickup.
Added myself to the changelog :)

PR #1665, opened 2024-10-13
Addonchecker: Added diagnostics and GMOD version check

NickCloudAT:
This adds a bit more information to the end of the addonchecker for diagnostic purpose.
It also contains a Garry's Mod Version check for the minimum required version which TTT2 can run on.

Print will look like this:

Compatible Version:

image

Not compatible:

image

I sadly don't know which version of GMOD is the current exact minimum on which TTT2 will still run on.

This is a draft, all input welcome :)

(Changelog etc. will follow if accepted and done)


TimGoll:
I can't tell you the exact minimum. But it should be the February 2024 update. The one wher net.ReadPlayer was added

https://wiki.facepunch.com/gmod/net.ReadPlayer

There is a version linked there. Maybe it helps?

NickCloudAT:
> I can't tell you the exact minimum. But it should be the February 2024 update. The one wher net.ReadPlayer was added
> 
> https://wiki.facepunch.com/gmod/net.ReadPlayer
> 
> There is a version linked there. Maybe it helps?

Ah, yeah that helps. That would be version string "20231006" which I could set of course.

TimGoll:
> > I can't tell you the exact minimum. But it should be the February 2024 update. The one wher net.ReadPlayer was added
> > https://wiki.facepunch.com/gmod/net.ReadPlayer
> > There is a version linked there. Maybe it helps?
> 
> Ah, yeah that helps. That would be version string "20231006" which I could set of course.

Hm. That update was released in the beginning of 2024 though. So the year seems wrong? Maybe it was only the dev branch of that year?

NickCloudAT:
> > > I can't tell you the exact minimum. But it should be the February 2024 update. The one wher net.ReadPlayer was added
> > > https://wiki.facepunch.com/gmod/net.ReadPlayer
> > > There is a version linked there. Maybe it helps?
> > 
> > 
> > Ah, yeah that helps. That would be version string "20231006" which I could set of course.
> 
> Hm. That update was released in the beginning of 2024 though. So the year seems wrong? Maybe it was only the dev branch of that year?

Yeah I just noticed.. I would guess it's the date of first implementation, not public release :/
I mean, theoretically we could just set the current GMOD version as the "baseline".. All servers should be up-to-date anyway.

NickCloudAT:
I guess ready for review.

gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
```diff
@@ -6,6 +6,7 @@
 
 ADDON_INCOMPATIBLE = 0
 ADDON_OUTDATED = 1
+TTT2_GMOD_MIN_VERSION = 241009
```
Histalek, line 9:
```suggestion
TTT2_GMOD_MIN_VERSION = 240313
```

For better or worse my dev server setup wasn't up to date and this seems to be the march version, which is able to run the current master branch.

Histalek (approved):
Nice find and cool addition!

It's a pity that we can't get the branch the server is using, but oh well

gamemodes/terrortown/gamemode/server/sv_addonchecker.lua
NickCloudAT: Ahh, very nice thanks :)

I changed it to that version.

PR #1666, opened 2024-10-16
Spectators no longer visible in thermalvision

MrXonte:
A check was added to prevent spectators from being displayed with thermal vision. This is relevant in cases where players are highlighted by TV and then killed, as they are not automatically removed from the entity list. Also prevents SpecDM players from being seen.

### DIFF
--- a/lua/ttt2/libraries/thermalvision.lua
+++ b/lua/ttt2/libraries/thermalvision.lua
@@ -139,7 +139,7 @@ local function RenderHook()
             local entry = thermalvisionList[i]
             local ent = entry.ent
 
-            if not IsValid(ent) or (ent:IsPlayer() and (ent:IsSpec() or ent:IsGhost())) then
+            if not IsValid(ent) or (ent:IsPlayer() and not ent:IsTerror()) then
                 continue
             end
 
--- a/lua/ttt2/libraries/thermalvision.lua
+++ b/lua/ttt2/libraries/thermalvision.lua
@@ -139,7 +139,7 @@ local function RenderHook()
             local entry = thermalvisionList[i]
             local ent = entry.ent
 
-            if not IsValid(ent) then
+            if not IsValid(ent) or (ent:IsPlayer() and not ent:Alive()) then
                 continue
             end
 
--- a/lua/ttt2/libraries/thermalvision.lua
+++ b/lua/ttt2/libraries/thermalvision.lua
@@ -138,8 +138,8 @@ local function RenderHook()
         for i = 1, #thermalvisionList do
             local entry = thermalvisionList[i]
             local ent = entry.ent
-
-            if not IsValid(ent) or (ent:IsPlayer() and not ent:Alive()) then
+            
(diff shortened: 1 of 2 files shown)


lua/ttt2/libraries/thermalvision.lua
```diff
@@ -139,7 +139,7 @@ local function RenderHook()
             local entry = thermalvisionList[i]
             local ent = entry.ent
 
-            if not IsValid(ent) then
+            if not IsValid(ent) or (ent:IsPlayer() and (ent:IsSpec() or ent:IsGhost())) then
```
TimGoll, line 142: I'm not that big of a fan to include SpecDM related checks inside TTT2. Especially since this code breaks if specDM is not installed.

@mexikoedi do you have an idea how to resolve this? could this be done in SpecDM? Do we have to add a filter interface in TTT2 maybe?

TimGoll (changes requested):

lua/ttt2/libraries/thermalvision.lua
MrXonte: So its probably best to move the SpecDM check out of here, but the general check if the entity is a player and dead (aka a spectator) still makes sense right? (which does seem to not be used directly in TTT2 at all except for CleanUpMap)
Or if the functionality should be moved out into addons anyway removing the whole library since its not used?

TimGoll: It is a bit more complicated, I get the issue now. We'd need some way to prevent ents from being rendered while they are already added. Something like a hook. Although I'm not sure if a hook is a good idea inside a render hook.

Not sure what a clean solution would be here. Maybe hide it if the player is team spec? I'd have to look a bit more into that and I'm not sure how SpecDM handles that. The same issue also exists for markerVision etc. We probably need a solution that works everywhere.

TimGoll: > but the general check if the entity is a player and dead (aka a spectator) still makes sense right?

yes, that's what I'm thinking as well. Not sure how IsGhost is implemented, but if it is another team such as TEAM_TERROR and TEAM_SPEC a `not ply:IsTerror()` check should suffice.

MrXonte: Updated it to use IsTerror!

TimGoll: Did you test it, if it still works with SpecDM the way you intended to?

MrXonte: thermalAlive
Yep works perfect! No thermal visible spectator after killing an alive player, and no visible specDM player (standing right next to alive player)

mexikoedi: Great that it works with `IsTerror()`.

TimGoll:
Looks good to me now. We probably have to do the same for the outline and marks module.

May I ask which weapon/addon you are using that has thermal vision?

MrXonte:
I was using a custom weapon that I haven't published to the workshop yet, waiting for the thermalvision library update to be public. Basicly just a sniper that adds all alive players to thermalvision when scoping in

PR #1668, opened 2024-10-16
Hook: Added hook "TTT2PlayDeathScream" for deathscreams

NickCloudAT:
This hook allows to change the sound of the deathscream or cancel it outright.

Maybe I'm mistaken but I thought the table approach would be the most appropriate one here. Please correct me if not.

Changelog will follow when this is done.

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_player.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player.lua
@@ -682,6 +682,8 @@ local function PlayDeathSound(victim, isSilent)
         sound = deathsounds[math.random(deathsounds_count)],
     }
 
+    ---
+    -- @realm server
     if hook.Run("TTT2PlayDeathScream", tbl, isSilent) == false then return end
 
     sound.Play(tbl["sound"], victim:GetShootPos(), 90, 100)
--- a/gamemodes/terrortown/gamemode/server/sv_player.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player.lua
@@ -684,7 +684,9 @@ local function PlayDeathSound(victim, isSilent)
 
     ---
     -- @realm server
-    if hook.Run("TTT2PlayDeathScream", tbl, isSilent) == false then return end
+    if hook.Run("TTT2PlayDeathScream", tbl, isSilent) == false then
+        return
+    end
 
     sound.Play(tbl["sound"], victim:GetShootPos(), 90, 100)
 end
@@ -796,9 +798,12 @@ function GM:DoPlayerDeath(ply, attacker, dmginfo)
     local killwep = util.WeaponFromDamage(dmginfo)
 
     ---@cast killwep -nil
-        PlayDeathSound(ply, ply.was_headshot
-        or dmginfo:IsDamageType(DMG_SLASH)
-        or (IsValid(killwep) and killwep.IsSilent))
+        PlayDeathSound(
+            ply,
+            ply.was_headshot
+                or dmginfo:IsDamageType(DMG_SLASH)
+                or (IsValid(killwep) and killwep.IsSilent)
+        )
 
     credits.HandleKillCreditsAward(ply, attacker)
 end
--- a/gamemodes/terrortown/gamemode/server/sv_player.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_player.lua
@@ -680,10 +680,9 @@ local function PlayDeathSound(victim, isSilent)
     local tbl = {
         victim = victim,
         sound = deathsounds[math.random(deathsounds_count)],
-        silent = isSilent,
     }
 
-    if not hook.Run("TTT2PlayDeathScream", tbl) or tbl["silent"] then return end
+    if not hook.Run("TTT2PlayDeathScream", tbl, isSilent) then return end
 
     sound.Play(tbl["sound"], victim:GetShootPos(), 90, 100)
 end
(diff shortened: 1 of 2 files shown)


Aeconomy7:
Woohoo, awesome to see this making it in!
- Sc00by

NickCloudAT:
(any feedback appreciated)

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -672,12 +672,20 @@ local deathsounds = {
...
     end
 
-    sound.Play(deathsounds[math.random(deathsounds_count)], victim:GetShootPos(), 90, 100)
+    local tbl = {
+        victim = victim,
+        sound = deathsounds[math.random(deathsounds_count)],
+        silent = isSilent,
+    }
+
+    if not hook.Run("TTT2PlayDeathScream", tbl) or tbl["silent"] then return end
```
Histalek, line 686: pulling in `isSilent` into the function doesn't make sense in my opinion.
With this there are two ways to prevent the death scream which complicates things.

IF we go with this hook implementation (which i'm fine with) we should either handle the isSilent part as we did before (old lines 792ff.) and prevent the call to `PlayDeathSound` outright. Or we should handle the isSilent part in the gamemodehook (new lines 1662ff.) you've introduced

NickCloudAT: I understand.
Though I'm unsure how I would get the isSilent into the hook otherwise?
For me it makes perfect sense to pull isSilent into the function.. But maybe I'm just dumb.

I understand that I should use isSilent in the gm hook I introduced, so if isSilent is true > cancel it outright if it ever goes to the gm hook and is not canceled earlier.. but I still need to pass isSilent to the hook then.

I guess I could take out the isSilent from the table and make it a direct parameter in the hook..

Histalek: You are right, without `isSilent` we won't be able to handle this in the gamemodehook

In that case i'd rather use the old `if..then` approach

That way we (as the gamemode) still determine first if a death is silent, if it is not we allow addons to hook into this and either change the deathsound or silence the death (by returning false in the hook)
so no `isSilent` would be necessary

NickCloudAT: Well.. I would agree with you but find it a bit more complicated that way..

For example: If I would check isSilent in the gamemode and not even call the Scream Hook if it is, other addons would need to use a Death Hook to play the sound if they want to play it even if it should be silent.. Actually they would need to then hook both the Scream and the Death hook and check in the deathhook if it was silent themselves.. Which makes this all way more complicated.

If I just pass isSilent to the hook too, I would just say
`if isSilent then return false end`
in the GM hook and remove the check from the PlayDeathSound function.

Histalek: Ah i see. I was giving feedback based on what you wrote in the PR description

> This hook allows to change the sound of the deathscream or cancel it outright.

If the hook should allow addons to "unsilence" a playerdeath then that's a different situation

NickCloudAT: > Ah i see. I was giving feedback based on what you wrote in the PR description
> 
> > This hook allows to change the sound of the deathscream or cancel it outright.
> 
> If the hook should allow addons to "unsilence" a playerdeath then that's a different situation

Welp.. Maybe my description was a bit unclear, sorry for that :')

But yes, I think it would be more beneficial if the hook did both. It would at least make it more versatile.

NickCloudAT:
Latest commit changed to how isSilent should be handled instead, as per suggestion.

gamemodes/terrortown/gamemode/server/sv_player.lua
```diff
@@ -672,12 +672,19 @@ local deathsounds = {
...
         return
     end
 
-    sound.Play(deathsounds[math.random(deathsounds_count)], victim:GetShootPos(), 90, 100)
+    local tbl = {
+        victim = victim,
+        sound = deathsounds[math.random(deathsounds_count)],
+    }
+
+    if not hook.Run("TTT2PlayDeathScream", tbl, isSilent) then return end
```
TimGoll, line 685: I personally would check for an explicit `false` here, so a normal `if not xyz then return end` would still work. Or is this against the purpose of this PR?

NickCloudAT: After thinking about this:
Normally I would agree, but in this case I don't think it's necessary. The Gamemode Hook would always return either false or true.

But if you think it fits better into TTT2 general code, I will change it.

NickCloudAT: Changed this as I see that it fits the overall TTT2 style more.

RFR

TimGoll (approved):

NickCloudAT:
Anything else for this to get merged? Just so it does not stale :P

TimGoll:
> Anything else for this to get merged? Just so it does not stale :P

image

NickCloudAT:
> > Anything else for this to get merged? Just so it does not stale :P
> 
> image

dangit, ofc I missed that. Should be resolved now.

TimGoll:
> dangit, ofc I missed that. Should be resolved now.

Style is still complaining :D

NickCloudAT:
> > dangit, ofc I missed that. Should be resolved now.
> 
> Style is still complaining :D

I might just be retarded.. But I actually don't get why it's complaining o.O?

NickCloudAT:
Ok.. ffs.. Took me 10 years now to know how the freak this style check works.. Well, better now than never.

TimGoll:
> Ok.. ffs.. Took me 10 years now to know how the freak this style check works.. Well, better now than never.

FYI, if you install stylua in VSCode it autodects the config file in the repo and does everything for you everytime you save the file

NickCloudAT:
> > Ok.. ffs.. Took me 10 years now to know how the freak this style check works.. Well, better now than never.
> 
> FYI, if you install stylua in VSCode it autodects the config file in the repo and does everything for you everytime you save the file

Ohhhhhhh wait actually? Damn, will do that rn.

TimGoll:
> Ohhhhhhh wait actually? Damn, will do that rn.

Yes! I wasn't a fan of it initially, but Histalek was right, it is really nice. If you have any issues with it, ask him, he knows how to fix it :D

PR #1669, opened 2024-10-20
fix(map): Streamline behaviour for the ttt2_enable_map_prefix_ convars

### LINKED ISSUE #1632: Server cvar message constantly showing up in chat

The following message shows up in the chat each map on the first round: "Server cvar 'ttt2_enable_map_prefix_gm' changed to 0".

## To reproduce

1. Go into a round 
2. Start the round (maybe change the map a couple of times)
3. See described issue

## Expected behaviour

This message shouldn't appear each time in the chat.

Histalek:
Updated changes after a few iterations are now the following:

* Unconditionally create gm, ttt and ttt2 prefixes

We can reasonably assume `gm` maps to be present on the server, the same
goes for `ttt` and `ttt2` maps.

Creating convars dynamically if we find more map prefixes is still
done. And only prefixes for installed maps will be shown anyway.

* Disable all map prefixes but `ttt` and `ttt2` by default

We will assume people only want to run maps intended for ttt(2).

Given that the prefix selection can be modified on the same admin
settings page that is (currently) the only one using these convars this
seems fine.

* Drop FCVAR_NOTIFY for all `map_prefix_enabled` convars

Given that these are purely used in an admin context there is no need to
announce changes to all users on the server.

The last changes especially fixes the original Issue https://github.com/TTT-2/TTT2/issues/1632

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Updated targetID to use `Vehicle:GetDriver` instead of the `ttt_driver` NWEntity (by @Histalek)
 - Updated Russian and English localization files (by @Satton2)
 - Updated old TTT HUD to show name of spectated player (by @somefnfplayerlol)
+- Changes to the enabled map prefixes will not be announced to players anymore (by @Histalek)
+- By default only `ttt` and `ttt2` map prefixes are enabled (by @Histalek)
 
 ## [v0.14.0b](https://github.com/TTT-2/TTT2/tree/v0.14.0b) (2024-09-20)
 
--- a/lua/ttt2/libraries/map.lua
+++ b/lua/ttt2/libraries/map.lua
@@ -502,19 +502,32 @@ local mapsPrefixes = {}
 local mapsWSIDs = {}
 
 if SERVER then
-    -- by default cs, de and test maps should be hidden
+    -- by default cs, de, gm and test maps should be hidden
+    -- while ttt and ttt2 maps should be shown
 
     ---
     -- @realm server
-    CreateConVar("ttt2_enable_map_prefix_cs", "0", { FCVAR_ARCHIVE, FCVAR_NOTIFY })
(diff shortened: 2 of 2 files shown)


TimGoll:
This only resolves the issue for the prefixes you added. If one adds a map with another prefix, the issue still presists

Histalek:
> This only resolves the issue for the prefixes you added. If one adds a map with another prefix, the issue still persists

i'm aware of that, if we can find a fix for all prefixes that would be nice!
This would only fix the common occurrences of the issue

if we would flip all other prefixes to off by default this issue would only occur for servers that enable any other prefixes and i would guess that not many servers would want to play on non-ttt prefixed maps

TimGoll:
> > This only resolves the issue for the prefixes you added. If one adds a map with another prefix, the issue still persists
> 
> i'm aware of that, if we can find a fix for all prefixes that would be nice! This would only fix the common occurrences of the issue
> 
> if we would flip all other prefixes to off by default this issue would only occur for servers that enable any other prefixes and i would guess that not many servers would want to play on non-ttt prefixed maps

If you are aware, then it is fine with me. I already tried for an hour or so, sadly without luck so far.

But mexikoedi and I discussed if we should maybe remove the convar change announcement to chat for all convars. That would resolve this issue as well. And announcing convar changes to chat doesn't see polished to me. We could probably make sure that it is still logged to the console or so.

TimGoll (approved):

Histalek:
Reading over this again i'm somewhat inclined to only default `ttt` and `ttt2` maps to enabled and disable the rest

Reason being that this would be the most likely setup people are running, correct?

> But mexikoedi and I discussed if we should maybe remove the convar change announcement to chat for all convars. That would resolve this issue as well. And announcing convar changes to chat doesn't see polished to me. We could probably make sure that it is still logged to the console or so.

I don't think removing this for all convars would be correct, but we could just drop `FCVAR_NOTIFY` for the dynamically created convars as a workaround

TimGoll (approved):
I like these changes, thank you. In my opinion this can be merged

Histalek:
cleaned up the commit and wrote changelog entries

PR #1683, opened 2024-11-27
fix(random_shop): Do not put minimum player limited equip in random shop

### LINKED ISSUE #1679: Items that cant be bought due to minimum player amount being set below current playercount show up in Random Shop.

Workshop version from Steam (v0.14.0b)

Items with a minimum player number set show up in random shops, taking away options.

## To reproduce

1. Set items to a minimum player number
2. Enable random shop
3. play with less players than that number
4. Have "wasted" shop slots

## Expected behaviour

Disabled items (which I count items that cant be used due to player count) should not be in the random shop.

clean TTT2

Histalek:
This fixes #1679, very well aware that other limitations (team/global limited) are still not respected and might still show up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed spectating player still being visible through thermalvision after killing that player (by @MrXonte)
 - Fixed Magneto-stick not using C_Hands (by @SvveetMavis)
 - Fixed console error when dropping ammo for weapons with no AmmoEnt (by @MrXonte)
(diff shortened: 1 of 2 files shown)


saibotk (approved):

Histalek:
FYI i put this into draft again as this only works correctly for non-Team randomshops.
Team wide randomshops still contain minimum player limited items.

I really need to get a dedicated `IsItemAvailable()` or smth check going and use it for these kinds of cases ...

Histalek:
i will probably need to refactor quite a bit to properly fix this.
I'll close this attempt for now

PR #1692, opened 2024-12-06
TargetID: Make `ttt_idenfity_body_woconfirm` replicated; move and clean up additional ConVars surrounding corpses/ragdolls

Wryyyong:
In #1684, `targetid.HUDDrawTargetIDRagdolls` was updated to check if the `ttt_idenfity_body_woconfirm` ConVar was enabled to determine what hint gets displayed.

Problem is, this ConVar only exists on the server, and the TargetID library is entirely clientside. This resulted in Lua errors being thrown on every frame the function was called and managed to reach that check.

~~This PR adds a GlobalBool and ChangeCallback for `ttt_idenfity_body_woconfirm`, and updates the check to use it instead of a nonexistent ConVar object. It also adds a ChangeCallback for `ttt2_confirm_team`, and changes its initial `SetGlobal` call to grab its name from its ConVar object to be consistent with the other GlobalVars in `SyncGlobals`.~~

This PR adds an additional table to the `CORPSE` namespace for the purpose of storing ConVar objects related to corpses and bodysearching. The following `CreateConVar` calls have been moved from `sv_main`:
- `ttt_idenfity_body_woconfirm` (to `sh_corpse`, with `FCVAR_REPLICATED` added)
- `ttt2_confirm_team` (to `sv_corpse`)
- `ttt2_confirm_killlist` (to `sv_corpse`)

I've also moved the `CreateConVar` calls for `ttt_announce_body_found` and `ttt_ragdoll_collide` into the new table for consistency, and so scripts for other parts of the gamemode can access the ConVar objects directly without having to store their own separate copies (or calling `GetConVar` repeatedly). I've also already adjusted all existing checks for these ConVars across the gamemode accordingly.

The changelog entry is worded the way it is because I just can't think of any other way to describe the results of the changes without going into a level of detail that'd be unnecessary for an end-user, but still useful information for addon developers.

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_main.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_main.lua
@@ -132,24 +132,6 @@ local map_switch_delay = CreateConVar(
 
 ---
 -- @realm server
-local confirm_team = CreateConVar(
-    "ttt2_confirm_team",
-    "0",
-    { FCVAR_NOTIFY, FCVAR_ARCHIVE },
-    "Show team of confirmed player"
-)
-
----
--- @realm server
-CreateConVar(
-    "ttt2_confirm_killlist",
-    "1",
-    { FCVAR_NOTIFY, FCVAR_ARCHIVE },
-    "Confirm players in kill list"
-)
-
----
--- @realm server
 CreateConVar(
     "ttt_enforce_playermodel",
     "1",
@@ -496,8 +478,6 @@ function GM:SyncGlobals()
         )
     end
 
-    SetGlobalBool(confirm_team:GetName(), confirm_team:GetBool())
-
     ---
     -- @realm server
     hook.Run("TTT2SyncGlobals")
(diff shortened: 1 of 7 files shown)


gamemodes/terrortown/gamemode/server/sv_main.lua
```diff
@@ -132,7 +132,7 @@ local map_switch_delay = CreateConVar(
 
 ---
 -- @realm server
-CreateConVar(
+local identify_body_woconfirm = CreateConVar(
```
TimGoll, line 135: best would be to move this line to a shared file and add the replicated convar flag. Global bools were added in the past when replication was broken, we're slowly trying to remove this from the codebase in recent updates.

Wryyyong: Mmm, while we're there, should I also remove the one for `ttt2_confirm_team` and move its `CreateConVar` call to `sv_corpse`? If GlobalVars are gonna be cleaned up and replaced with replicated ConVars, I figure we might as well move it to where it belongs.

There's also a reference to `ttt2_confirm_team` in `/lua/autorun/gs_crazyphysics.lua` that uses its `GetGlobal`, should we just store a copy of the ConVar object in that script as well?

Histalek (commented):
First of all thanks for catching my mistake! That's what i get for only testing locally i suppose

As Tim somewhat started the review i'll leave it to him unless he tells me otherwise :)

Wryyyong:
I've updated the title and description of the PR to more accurately describe its new goal(s). I'll also quickly update the changelog entry as well, for good measure.

TimGoll (approved):
that looks really good - thank you!

PR #1694, opened 2024-12-09
Ammo Drop from reserve now tries to drop a full box, instead of, at most, a full clip

MrXonte:
Ammo dropping from the reserve was a bit weird. You currently drop `Min(Clipsize, Reserve)`, which is fine for default weapons where clip size = box ammo, but for any custom weapon with less ammo, things get increasingly weird.
The worst-case scenario involves weapons with super-low ammo. Consider the often-used pocket rifle, which has only a clip size of 1; here, you must drop ammo 10 times to drop the same as a regular 357 box.
My change makes it so that instead, you now drop `Min(Box Size, Reserve)`, which feels much better and has no downsides I can see. Dropping from the clip remains unchanged as it should.


TimGoll (approved):
that is a good change, well done

PR #1698, opened 2024-12-25
Fix: Prevent spectators, covert, and long range corpse searches from making sound

TW1STaL1CKY:
The new corpse searching sounds play when anything searches them in any manner. The main example being spectators - if a spectator examined a body, the body would make a sound seemingly out of nowhere for players who are alive.
Covert searches and long range searches also cause the search sound to play which seems wrong.

The change in this PR fixes that. The sound will now only play if the player is alive (using IsTerror), and if the search wasn't covert or long range.


saibotk (approved):
Very nice change thank you very much! 🙏

saibotk:
Can you quickly merge the if statement into a single line to make stylua happy?

PR #1702, opened 2025-01-01
Improve admin UI around roles

nike4613:
All role-related logic is moved into the "Roles" admin tab, and the search bar is now sticky-scrolling above searchable submenus. Role layering description improved, and algorithm overview of role distribution added.

I added the algorithm overview as HTML in the language file; is there a better way to have large translatable strings?


TimGoll:
> I added the algorithm overview as HTML in the language file; is there a better way to have large translatable strings?

Not really yet. I wanted to add a base-text system for that, that supports headings, text and bullet points. But so far, I did not have the time to do any of that. But I'm not a fan of using HTML in the UI, because the rendered text looks completely out of place.

Speaking of it, could you show screenshots of what you did? In general your changes sound like a solid idea! I think we already talked about this on discord.

For the future: Stuff like this could easily be separated into multiple smaller PRs. But I struggle with that as well, so I get why you did it like this.

nike4613:
> Stuff like this could easily be separated into multiple smaller PRs

Yeah, the multiple changelog entries tipped me off there, but by that point it was already done. I can pull the overview page and description changes out easily enough if you want.

> But I'm not a fan of using HTML in the UI, because the rendered text looks completely out of place.

Nor am I, but the overview doc needs nested lists. I'm sure I could do it in vgui, but I think I'd rather spend the time doing HTML or Markdown to vgui automatically. 

> Speaking of it, could you show screenshots of what you did? 

Here's the video I recorded the other day of the UI changes without the overview page: https://files.catbox.moe/zn3kqz.mkv (MKV w/ AV1; if browser/media player doesn't work, use mpv)

I can take more screenshots later, not super sure when. Only things new from that video are the overview page (looks like the changelog pages) and text changes in the role layering page.

nike4613:
Screenshots:

image
image
image
image

TimGoll:
Ok, I watched the video and took a look at the screenshots. First of, it's awesome what you did here.

However, to me it feels a bit "janky"(?). I absolutely understand why you did it the way you did it, but it feels weird to have entries above the search bar. That looks wrong to me on first glance. It looks a bit out of place.

I'm not sure if this needs some color to differentiate it from the rest. Or if it should be below the search bar and the "pinning" should be shown in a different way.

Side note, this is how winXP did the pinning:
image

And win10 had the search bar at the bottom:
image

Not sure if these windows solutions point into a better direction, but I think we should open a discussion about the design here. Because your implementation seems a bit unintuitive to me. Albeit be it very impressive from a technical point of view.

TimGoll:
mockup

Here's a quick mockup of the ideas I had in mind. There's absolutely room for improvement here, I just wanted to show something so we have a base for discussions

TimGoll:
On a technical side I'm wondering how you did it that some submenus are "pinned", and some are normal menus. Personally I'd have added a new flag to it, something like:

```lua
CLGAMEMODESUBMENU.pinned = true
```
or
```lua
function CLGAMEMODESUBMENU:IsPinned()
    return true
end
```

Histalek:
First of all: amazing work! Especially the terrifically in-depth description for the rolelayering, I appreciate it!

> > Stuff like this could easily be separated into multiple smaller PRs
> 
> Yeah, the multiple changelog entries tipped me off there, but by that point it was already done. I can pull the overview page and description changes out easily enough if you want.

It might be worth to split off the overview page for now. I'm quite worried about how feasible it is to translate this properly
But i also don't want your in-depth description to go to waste or for my concern to block your other changes

I'd also appreciate it if some translators could chime in if my concern is warranted in the first place

nike4613:
> @TimGoll: ...

I have 2 arguments for why the current presentation is the correct choice:

1. By consistency. All other menus (having all items being searchable) have, and have always had, the search bar at the *top*, *above* the items that are searched. It follows then that a search bar searches items listed below it.

2. By literary convention. Consider a heading:
   > Some Text
   > # Heading
   > More Text

   The underline under the heading connects the heading to the text *after* it, on the other side of the line. The search bar UI already has a highlight line underneath, so this connection is maintained. 

Windows 10 is an interesting comparison, but the search bar is *on the taskbar*, not *at the bottom*. The taskbar can be placed on any edge of the screen. Plus, the menu is *invoked* by first selecting the search bar, making it a somewhat different situation to our UI. I will also reference several other applications for placing the search bar above the searched content:

- Visual Studio
- VSCode
- Rider
- PowerToys Run
- Explorer
- KDE Plasma (basically everything)
- Thunderbird
- Every search engine ever
- ...

I strongly suspect that, one way or another, placing the search bar *above* the content it searches is the way to go. There might be something worth doing to highlight more that the items above it *aren't* searched, though I also think that there's not much point, as even in the worst case, the user tries to search and immediately finds that the items above are not filtered.

> On a technical side I'm wondering how you did it that some submenus are "pinned", and some are normal menus. 

It's implemented as a new function on `CLGAMEMODEMENU` returning a list of submenus which are 'non-searched'. The logic here is 2-fold:

1. This method is compatible with the previous approach of `GetVisibleSubmenus()` + `HasSearchbar()`, and the defaults are such that other menus continue to behave exactly as before with no changes.
2. This method is compatible with the existing approach to searching, just with some slightly different setup in the submenu list. (I personally don't like this search approach, and would personally do it differently, but no sense in changing what works IMO)

It would be very possible to *also* have an `IsPinned` option on the submenu itself, and adjust the default logic to continue to preserve behavior, if that's desirable. For the purposes of the roles menu, there's not *much* point, because we treat the auto-detected submenus fundamentally differently from the role list anyway, so we already have the searchable and non-searchable menus split out from each other.

NickCloudAT:
If nobody minds my opinion on this:

I like how nike made this. I think putting the searchbar at the bottom is a bad design solution in this case, but that's also just my taste.

But: The searchable content should always be the last thing in the overall list. Also, above a searchfield, I would like to see a "split" line, pretty much like this:

image

nike4613:
What about something like this?

image

I think I prefer the line on the bottom, but I don't hate this.

TimGoll:
To me it has the same problem still. Having a search bar that is neither at the top or bottom looks wrong to me.

saibotk:
I think both top and below pinned is okayish (with a separator).
Personally id also prefer it at the top because thats just less disruptive and id expect a search there.
Also dont we atm always show these searches at the top by default in other menues? Then id definitely say lets keep it this way for consistency with other Menues to provide a predictable UI.

And yes its not worth discussing too much about it.

Id say @TimGoll you can just go ahead and select the solution you see for TTT2 as you have he greater vision for our UI and this just needs a decision from someone.

Thanks @nike4613 for your efforts and your patience with this discussion! 🫶

nike4613:
If possible, I'd like to work out what needs to be done here to be mergeable soon. 

Re: the UI design: I still think that my original design is the best overall. Every single person I have asked about this has given a different answer to what's clearest and how it could be improved. 

One interesting idea (that unfortunately I don't think is particularly feasible, nor that good of an idea) is to add another nested sub-menu for the search bar and roles. 

I also asked Claude about it because I figured it can't possibly give more conflicting suggestions than I've already gotten, and it made a few points:
- My design keeps the searchable objects close to the search bar, and also at the head of their visual scope, which is typically something UI designers try to do
- It suggested changing the filler text for the search bar to something like "Search roles..." (I will probably do this)
- It suggested adding some extra visual scoping (i.e. a different background color or border) around the search bar and its items. I'm not sure how feasible this is, but it might help a little.

TimGoll:
@nike4613 I think we won't come to an agreement here, but that is ok. It seems like no one here is "on my side", so I think your implementation of it is as least as good as my suggestion.

And since I've seen your newest PR, I got a lot more understanding for your vision and respect for your skill.

That being said, there are two things:
- I wonder if a graphical devider *above* the search bar will improve upon the thing I'm most worried about. Something like the double line I did, but above the search bar, while keeping the rest as you did it:
  image
- The way you define the "pinned" pages is something I have to look into. But I have the time for that later today

nike4613:
The only problem I have with a separator is this: what color is it? If it's the same color as the background of the search bar, then it looks strange; if it's a much lighter color, it would stand out too much; if it's the highlight color, then isn't that basically just the last image that I posted? We could also have the highlight both above *and* below, and that would probably work, but personally I'm not sure if that is a meaningful improvement over just below.

TimGoll:
> The only problem I have with a separator is this: what color is it? If it's the same color as the background of the search bar, then it looks strange; if it's a much lighter color, it would stand out too much; if it's the highlight color, then isn't that basically just the last image that I posted? We could also have the highlight both above _and_ below, and that would probably work, but personally I'm not sure if that is a meaningful improvement over just below.

seperator

I think it looks alright? Maybe I'm biased though, what is your opinion on that? (ignore the pin symbols)

The color is the same as the horizontal line below a form heading

TimGoll:
Okay, I gave it a try in-game. First of all, please revert the blue line back to the bottom. That looks definitely better that way. I think I'd still prefer a visual devider in the way I suggested it. But if you are of different opinion, it is fine for me without it. I can give it a try by myself in a different PR.

Regarding the long text in HTML. I hate it :D But we do not have a better system for now. So it is fine!

I will take a look at your code now

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -177,3 +188,26 @@ end
...
+    submenuList:Dock(FILL)
+    submenuList:SetPadding(padding)
+    submenuList:SetBasemenuClass(self, contentPanel)
+    if self.searchBarPlaceholder then
+        submenuList:SetSearchBarPlaceholder(self.searchBarPlaceholder)
+    end
+
+    -- REFRESH SIZE OF SUBMENULIST FOR CORRECT SUBMENU DEPENDENT SIZE
+    submenuList:InvalidateLayout(true)
+end
```
TimGoll, line 213: I'm not sure I like it that GUI code is moved from the gui file to the class. The way the classes are designed, it was intended that they only contain data, no UI code. I'm not sure how complicated that change would be though. But if it is feasable, I'd like this part to be moved to the file where it was before.

The whole menu structure right now is defined in cl_help.lua. Down to the position of the search bar. Now this one nav panel is moved to a different file for no apparant reason. This makes mainting more tedious in my opinion.

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -10,6 +10,7 @@ CLGAMEMODEMENU.priority = 0
 CLGAMEMODEMENU.icon = nil
 CLGAMEMODEMENU.title = ""
 CLGAMEMODEMENU.description = ""
+CLGAMEMODEMENU.searchBarPlaceholder = nil
```
TimGoll, line 13: It took me a few minutes to figure out what this does. This has to be renamed for sure. May I suggest something like `CLGAMEMODEMENU.searchBarPlaceholderText` would probably suffice.

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -177,3 +188,26 @@ end
...
+-- @hook
+-- @realm client
+function CLGAMEMODEMENU:FillNavPanel(navPanel, contentPanel, padding)
+    -- MAKE SEPARATE SUBMENULIST ON THE NAVAREA WITH A CONTENT AREA
+    local submenuList = vgui.Create("DSubmenuListTTT2", navPanel)
+    submenuList:Dock(FILL)
+    submenuList:SetPadding(padding)
+    submenuList:SetBasemenuClass(self, contentPanel)
+    if self.searchBarPlaceholder then
+        submenuList:SetSearchBarPlaceholder(self.searchBarPlaceholder)
```
TimGoll, line 208: Same here, append a `Text` to the function name

lua/terrortown/menus/gamemode/roles.lua
```diff
@@ -36,13 +37,18 @@ function CLGAMEMODEMENU:InitializeVirtualMenus()
...
+        virtualSubmenu.title = roleData.name
+        virtualSubmenu.icon = roleData.iconMaterial
+        virtualSubmenu.roleData = roleData
+        virtualSubmenu.iconBadge = roleData.builtin and builtinIcon
+        virtualSubmenu.iconBadgeSize = 8
+        virtualSubmenu.basemenu = self
+        -- make sure that the virtual submenu will be shown
+        function virtualSubmenu:ShouldShow()
+            return true
+        end
```
TimGoll, line 50: can you explain why this had to be added? Because it wasn't necessary before.

lua/terrortown/menus/gamemode/roles.lua
```diff
@@ -62,6 +68,24 @@ function CLGAMEMODEMENU:HasSearchbar()
...
+
+        if not submenu:ShouldShow() then
+            continue
+        end
+
+        visibleSubmenus[#visibleSubmenus + 1] = submenu
+    end
+
+    return visibleSubmenus
+end
```
TimGoll, line 87: This is how you differentiate between "pinned" submenus and searchable submenus? I think I still don't get how you do it. My assumption here is that virtual submenus are generated *after* this function is called?

If this is true, then I dislike it for two reasons:
- a developer calling this function at a later point in time would receive a different output then the one you designed it for
- this system seems pretty fixed it is use; it only works when virtaul submenus are in use. Why don't add a new attribute? Either `CLGAMEMODESUBMENU.isSearchable = true` or `CLGAMEMODESUBMENU.static = false` to submenus that are sarchable and not "fixed" (static being a different words for the menus that are fixed to the top)

Moreover, it looks like the content of this function is identical to `CLGAMEMODEMENU:GetVisibleSubmenus()` that already exists which seems to confirm my point, while also raising the question: why do you need this at all?

I'd like to also have @saibotk's input on this point.

lua/terrortown/menus/gamemode/roles/roles.lua
```diff
@@ -0,0 +1,178 @@
+--- @ignore
+
+CLGAMEMODESUBMENU.base = "base_gamemodesubmenu"
+
+CLGAMEMODESUBMENU.priority = 97
+CLGAMEMODESUBMENU.title = "submenu_administration_roles_general_title"
```
TimGoll, line 6: this title identifier should be edited to match the new file location. Same holds true for the other submenus.

lua/terrortown/menus/gamemode/roles/base_roles.lua
```diff
@@ -167,3 +167,8 @@ function CLGAMEMODESUBMENU:Populate(parent)
 
     hook.Run("TTT2OnRoleAddToSettingsMenu", self.roleData, parent)
 end
+
+-- Override this here, because we don't want auto-discovery to show this menu
+function CLGAMEMODESUBMENU:ShouldShow()
+    return false
+end
```
TimGoll, line 174: I'm not 100% sure what you're doing here and why. What is "auto-discovery"? What is the reason why this had to be changed from how it was before?

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dsubmenulist_ttt2.lua
```diff
@@ -4,13 +4,8 @@
...
 local PANEL = {}
 
----
--- @accessor function
--- @realm client
-AccessorFunc(PANEL, "searchFunction", "SearchFunction")
-
 -- Define sizes
-local heightNavHeader = 10
+local heightNavHeader = 0 --10
```
TimGoll, line 8: if you keep it a 0, then you can remove this variable completely imho

TimGoll (changes requested):
Like I already said: This is awesome and you're doing good work!

But, I also have a few things that I'd like to be adressed. Most of them are minor. But I don't like the way you are differentiating between searchable and non-searchable submenus. To me this is counter-intuitive

lua/terrortown/menus/gamemode/roles/base_roles.lua
nike4613: Since the roles menu now contains normal, non-generated submenus, the *base* for the submenu would get picked up and shown *as a non-generated submenu*, which is undesirable. We set this here so that it doesn't do that. The virtual submenu logic is otherwise kept exactly.

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
nike4613: This is mostly an artifact of a different approach that wanted to let submenus define structurally *very* different UIs. That design didn't work out for what I wanted here, but I left the extension point mostly because I didn't see much reason to remove it again.

lua/terrortown/menus/gamemode/roles.lua
nike4613: See my reply here: https://github.com/TTT-2/TTT2/pull/1702#discussion_r1915704387 The virtual submenus need to be marked as visible once created to behave correctly.

lua/terrortown/menus/gamemode/roles.lua
nike4613: Menu creation looks like this:

1. The menu class is instantiated.
2. Submenus are discovered, and added to the menu's `submenus` field. (This field is normally read by `GetVisibleSubmenus()`, the result of *that* is always "searchable" for compatibility.)
3. If the menu wants to define non-searchable submenus, it can override `GetVisibleNonSearchedSubmenus()`, which will be called first. In the case of roles, we override it to provide the same behavior as `GetVisibleSubmenus()` does by default, because we override `GetVisibleSubmenus()` to return the virtual role submenus.

I chose this approach because it seemed to match what was already being done conceptually, and required *no changes whatsoever* to any other menu. Adding a new attribute could be made to work for sure, but my worry becomes that it doesn't mesh well with the dynamic submenu creation which is used for (I think?) all of our searchable menus right now (which **bypass** the normal `submenus` field). It would not be a bad idea to pull out the core body of `GetVisibleNonSearchedSubmenus()` into a reusable helper or something though.

TimGoll: Maybe I'm missing something here, but to me this looks like a hacky solution to a simple problem. Especially with what you said here: https://github.com/TTT-2/TTT2/pull/1702#discussion_r1915706908

Just the fact that you had to explain to me twice what it does and how it works, even though I've seen your code and I wrote all the remaining code in the UI shows to me, that it is not a straight forward solution.

I'm also still wondering what `GetVisibleNonSearchedSubmenus()` would return during runtime. I'm pretty sure it is different to the return value at menu creation. And this is a big no-go for me.

I'm not seeing the benefit of your implementation here. To me it seems obvious to add a flag to submenus that defines if they are pinned or not. The same flag can be correlated with the searchability as well. In my opinion that would make it obvious to any future dev on how to add custom searchable submenus.

What happens if I want to add a menu with searchable submenus, that doesn't use virtual menus? With your implementation this wouldn't work.

Or what if a menu action triggers a hook at runtime that has to rebuilt the UI (there are some places in the UI that do that). Wouldn't it break if the `GetVisibleNonSearchedSubmenus()` returns something different than it did in the beginning?

TimGoll: Not different places, different times. The only thing you seem to use to differentiate is if `ShouldShow` is true. At a later point in time this should show is true for all menus: virtual and real submenus

nike4613: Virtual menus don't end up in the `submenus` list. They get their own list, which is returned from `GetSubmenus()` earlier in this file. This implementation is *exactly* the normal implementation of `GetVisibleSubmenus()`, only bypassing `GetSubmenus()` and hitting the underlying list directly.

TimGoll: Oh, now I understand. I completely forgot that we did it this way. I think I get now what you did here. I have to look at our submenu code again in detail to give more feedback. It has been a while.

saibotk: I also think a flag would be cleaner and simpler. If necessary i can also dive deeper into this code to give a more refined opinion. But id let @TimGoll decide if this is fine for now because it matches the previous work and we refactor this in a follow up.

TimGoll: For the sake of keeping things moving, I think this is fine then. I don't have the time right now to review and rework this in depth. So it's ok from my POV, @saibotk if you still agree, you can merge the PR then

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
TimGoll: thanks for reverting that change!

TimGoll (approved):
Thanks a lot for taking your time to add this to the code! I might have sounded a bit anti, but this is not because I dislike what you did here. I'm focusing on some small details.

@saibotk if you're fine with the most recent changes, please go ahead and merge this

PR #1705, opened 2025-01-03
Various small chores

Histalek:
i've pooled these changes together as all of them are extremely small and all don't warrant a changelog entry imo


TimGoll:
Always great to see that you find so many minor bugs that nobody ever notived. I love this, great work

Issue #1706, opened 2025-01-03
Visibility of the prop spec icon


mexikoedi:
## Your version of TTT2 (mandatory)

- TTT2 v0.14.0b Workshop version from Steam

## Describe the bug (mandatory)

The prop spec icon on the bottom of the screen is visible even though the feature is deactivated.

## To reproduce

Steps to reproduce the behaviour:

1. Deactivate the prop spec feature
2. Join the round
3. Look at the icons at the bottom
4. See the described issue

## Expected behaviour

The icon shouldn't be visible if the prop spec feature is disabled.

## Context (please provide as much as you can)

- Screenshot:
[Prop spec icon](https://cdn.discordapp.com/attachments/493712038216990720/1324634777386745877/image.png?ex=6778ddc5&is=67778c45&hm=d3c599c3ddc221c59f49949de2d16213e687d521e1e9c37319d149399aaf25be&)

## Additional context

Someone reported this on Discord in the suggestions channel.
He wrote: "i think if the the icon telling you how to use prop spec was disabled when the server has prop spec off that would be neat"

PR #1709, opened 2025-01-04
UI: Hide Propspec bind if disabled

NickCloudAT:
This PR would "fix" https://github.com/TTT-2/TTT2/issues/1706

I'm just not sure: Is using globals here a good idea? I though about using a replicated cvar, but that would require at least a cl_propspec.lua file to be added.

Changelog will follow when finished.

### DIFF
--- a/gamemodes/terrortown/gamemode/server/sv_propspec.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_propspec.lua
@@ -9,9 +9,7 @@ local timer = timer
 
 PROPSPEC = {}
 
----
--- @realm server
-local cvPropspecToggle = CreateConVar("ttt_spec_prop_control", "1", { FCVAR_NOTIFY, FCVAR_ARCHIVE })
+local cvPropspecToggle = GetConVar("ttt_spec_prop_control")
 
 ---
 -- @realm server
@@ -30,14 +28,6 @@ local cvPropspecMax = CreateConVar("ttt_spec_prop_maxbonus", "16", { FCVAR_NOTIF
 -- @realm server
 local cvPropspecDashMulitplier =
     CreateConVar("ttt_spec_prop_dash", "2", { FCVAR_NOTIFY, FCVAR_ARCHIVE })
-
-hook.Add("TTT2SyncGlobals", "AddPropspecGlobals", function()
-    SetGlobalBool(cvPropspecToggle:GetName(), cvPropspecToggle:GetBool())
-end)
-
-cvars.AddChangeCallback(cvPropspecToggle:GetName(), function(cv, old, new)
-    SetGlobalBool(cvPropspecToggle:GetName(), tobool(tonumber(new)))
-end)
 
 ---
 -- Forces a @{Player} to spectate an @{Entity}
--- a/gamemodes/terrortown/gamemode/server/sv_propspec.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_propspec.lua
@@ -31,6 +31,14 @@ local cvPropspecMax = CreateConVar("ttt_spec_prop_maxbonus", "16", { FCVAR_NOTIF
 local cvPropspecDashMulitplier =
     CreateConVar("ttt_spec_prop_dash", "2", { FCVAR_NOTIFY, FCVAR_ARCHIVE })
 
+hook.Add("TTT2SyncGlobals", "AddPropspecGlobals", function()
+    SetGlobalBool(cvPropspecToggle:GetName(), cvPropspecToggle:GetBool())
+end)
+
+cvars.AddChangeCallback(cvPropspecToggle:GetName(), function(cv, old, new)
+    SetGlobalBool(cvPropspecToggle:GetName(), tobool(tonumber(new)))
+end)
+
 ---
 -- Forces a @{Player} to spectate an @{Entity}
 -- @param Player ply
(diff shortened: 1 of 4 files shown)


TimGoll:
I see the problem here. I'd prefer not to use global bools. Maybe we can create a shared convar file where stuff like this could be moved to? hm.

NickCloudAT:
> I see the problem here. I'd prefer not to use global bools. Maybe we can create a shared convar file where stuff like this could be moved to? hm.

I just saw that here is a file called "sh_cvar_handler.lua".. Which would be exactly what we need, or am I wrong?
It's just not really used as of now.

TimGoll:
> > I see the problem here. I'd prefer not to use global bools. Maybe we can create a shared convar file where stuff like this could be moved to? hm.
> 
> I just saw that here is a file called "sh_cvar_handler.lua".. Which would be exactly what we need, or am I wrong? It's just not really used as of now.

Exactly. Not that you reminded me, I created this file with the purpose of moving ALL convars in there. But then ditched the idea because I thought it would be better to keep them in their modules.

But imho this file is fine for your use case

NickCloudAT:
> > > I see the problem here. I'd prefer not to use global bools. Maybe we can create a shared convar file where stuff like this could be moved to? hm.
> > 
> > 
> > I just saw that here is a file called "sh_cvar_handler.lua".. Which would be exactly what we need, or am I wrong? It's just not really used as of now.
> 
> Exactly. Not that you reminded me, I created this file with the purpose of moving ALL convars in there. But then ditched the idea because I thought it would be better to keep them in their modules.
> 
> But imho this file is fine for your use case

I changed it to what I thought about when mentioning the replicated cvar.
Only "problem" i see now is, that the keyhelp.lua seems to be loaded before the sh_cvar_handler file. So I cant cache the convar directly..

gamemodes/terrortown/gamemode/server/sv_propspec.lua
```diff
@@ -9,9 +9,7 @@ local timer = timer
 
 PROPSPEC = {}
 
----
--- @realm server
-local cvPropspecToggle = CreateConVar("ttt_spec_prop_control", "1", { FCVAR_NOTIFY, FCVAR_ARCHIVE })
+local cvPropspecToggle = GetConVar("ttt_spec_prop_control")
```
TimGoll, line 12: just to be sure: did you test that the other file is loaded first?

TimGoll:
> I changed it to what I thought about when mentioning the replicated cvar. Only "problem" i see now is, that the keyhelp.lua seems to be loaded before the sh_cvar_handler file. So I cant cache the convar directly..

That's fine, a convar lookup is not that expensive. I think it's good as it is now

gamemodes/terrortown/gamemode/server/sv_propspec.lua
NickCloudAT: Yes, this works as expected.

It is loaded first:
cl main:
image

sv main:
image

NickCloudAT:
> > I changed it to what I thought about when mentioning the replicated cvar. Only "problem" i see now is, that the keyhelp.lua seems to be loaded before the sh_cvar_handler file. So I cant cache the convar directly..
> 
> That's fine, a convar lookup is not that expensive. I think it's good as it is now

The GMOD wiki on this is also not really helpful for me.. Either I understand something wrong, or the information on there is very convoluted.. https://wiki.facepunch.com/gmod/Global.GetConVar
It mentiones that it caches? But also that it does not?

TimGoll:
> > > I changed it to what I thought about when mentioning the replicated cvar. Only "problem" i see now is, that the keyhelp.lua seems to be loaded before the sh_cvar_handler file. So I cant cache the convar directly..
> > 
> > 
> > That's fine, a convar lookup is not that expensive. I think it's good as it is now
> 
> The GMOD wiki on this is also not really helpful for me.. Either I understand something wrong, or the information on there is very convoluted.. https://wiki.facepunch.com/gmod/Global.GetConVar It mentiones that it caches? But also that it does not?

Ah, interesting. It is cached, but it a bad way.

you can do it like this then:

```lua
-- top of the file
local cvPropspecToggle

-- in function
cvPropspecToggle = cvPropspecToggle  or GetConVar("ttt_spec_prop_control")
```

This is not perfect, but should be faster than the internal caching.

NickCloudAT:
> > > > I changed it to what I thought about when mentioning the replicated cvar. Only "problem" i see now is, that the keyhelp.lua seems to be loaded before the sh_cvar_handler file. So I cant cache the convar directly..
> > > 
> > > 
> > > That's fine, a convar lookup is not that expensive. I think it's good as it is now
> > 
> > 
> > The GMOD wiki on this is also not really helpful for me.. Either I understand something wrong, or the information on there is very convoluted.. https://wiki.facepunch.com/gmod/Global.GetConVar It mentiones that it caches? But also that it does not?
> 
> Ah, interesting. It is cached, but it a bad way.
> 
> you can do it like this then:
> 
> ```lua
> -- top of the file
> local cvPropspecToggle
> 
> -- in function
> cvPropspecToggle = cvPropspecToggle  or GetConVar("ttt_spec_prop_control")
> ```
> 
> This is not perfect, but should be faster than the internal caching.

Hm, yeah you are right. I added this, guess that's the best way in this case.

TimGoll (approved):

TimGoll:
awesome work, thank you!

Issue #1707, opened 2025-01-03
Make map time optional


mexikoedi:
## Is your feature request related to a problem? Please describe.
Currently there is a map time and a number of rounds.

## Describe the solution you'd like
The map time should be optional so it only goes by rounds if the map time is deactivated.

## Additional context
Someone reported this on Discord in the suggestions channel.
He wrote: "it would also be neat if there was a way to turn off map time limits entirely so it only goes by round"

PR #1710, opened 2025-01-05
Gameloop: Split round and time

NickCloudAT:
This PR will tackle https://github.com/TTT-2/TTT2/issues/1707

Not finished yet, still have to do all LANG files and some testing + cleanup if neccessarry.

Would love some input on this though!

Only "real" problem I have so far is the settings in the F1 menu: I can not "disable" the round time or the round count setting because the "master" is never null but always a value.. Should not really be confusing though.


NickCloudAT:
First tests seem to be working as expected on my end.

TimGoll:
yeah, we should probably extend the parenting system to dropdown menus as well. It would be a bit more complex through

NickCloudAT:
> yeah, we should probably extend the parenting system to dropdown menus as well. It would be a bit more complex through

Yeah.. I looked at the vgui stuff cause I wanted to know how that works.. Then I said "nope" and went on with just this PR..
I mean, nothing that can't be changed later.. Also, like said, I don't think it's actually confusing to not grey out those two sliders.. for now

NickCloudAT:
Not really able to give a full overview of it in this video, but you at least see the settings and the change in the Scoreboard:


https://github.com/user-attachments/assets/4e582a0f-b324-438a-9f3d-1a0a53288edc

Everything else I was able to test was successful.. A MapVote always triggered at the right moment (only testet with PAM though).

TimGoll:
Looks good to me, I'd replace "mapvote" with "map change", since the vote is only there with map vote addons. Also just replace "TTT(2)" with TTT2 imho.

NickCloudAT:
> Looks good to me, I'd replace "mapvote" with "map change", since the vote is only there with map vote addons. Also just replace "TTT(2)" with TTT2 imho.

Yeahh after I wrote it I was pretty sure somebody will tell me to change that :P
I will do so.

NickCloudAT:
Should be good now..
Oh well, I just noticed with his PR that here are many new DE language lines that need to be translated.. Well well

Also: Will add a changelog now as I forgot that...

TimGoll (approved):
But make sure to add empty lines at file endings:

image

NickCloudAT:
> But make sure to add empty lines at file endings:
> 
> image

ah, yeah thats not done automatically.. Well, fixed.

TimGoll (approved):

PR #1712, opened 2025-01-06
Markervision fix: Fixed registry table being able to have duplicates

TW1STaL1CKY:
I discovered that when new marker vision objects are made with `markerVision.Add`, it doesn't clear out existing versions of the same object (matching ent and indentifier). This leads to the `markerVision.registry` table being filled up with objects that do not clear themselves, and makes `markerVision.Get` potentially return obsolete markers (mainly observed serverside).

This PR resolves this possible bloat issue in the registry. This also fixes an issue where marker vision objects can incorrectly change to being visible to team or visible to role (clientside), when it's supposed to be visible to everyone.


saibotk (approved):

TimGoll:
That's a smart change! While it was intentional the way I did it, I do like your change. We will merge it as soon as you add yourself to the changelog.

TW1STaL1CKY:
Sorry, there we go 😄

PR #1714, opened 2025-01-12
Role distribution inspection

nike4613:
This PR adds an admin UI to inspect how and why role distribution selects certain roles, and why it distributes those roles to players.

The goal here is to assist admins in understanding the role distribution process, so they can more effectively tweak the parameters involved, and better see their effects.

This also includes a graphic showing player weights during role distribution (if derandomization is enabled) which can assist debugging that as well as help intuition when tweaking it.

Some of the graphics are a touch jank, and in particular I wish there was a good way to get an icon in the label of a collapsible form, but just role names probably works fine anyway.

This PR is based on (and soft depends-on) #1702. That dependency can be removed if necessary.

Screenshots:
Screenshot_20250111_221601
Screenshot_20250111_222215
Screenshot_20250111_222252
Screenshot_20250111_222312
Screenshot_20250111_222324
Screenshot_20250111_222328
Screenshot_20250111_221616
Screenshot_20250111_221629
Screenshot_20250111_221656
Screenshot_20250111_221731
Screenshot_20250111_221747


TimGoll:
WTF, I'm speechless. This is so impressive, wow. Thank you! I'll take a look at it as soon as I've got the time!

TimGoll:
image

Is there a reson why this is turned off by default? And it also seems to glitch a bit showing some text here

TimGoll:
I was unable to test this due to the following error:

```
[ttt2] addons/ttt2/lua/terrortown/menus/gamemode/roles/roleinspect.lua:221: attempt to get length of local 'subroleSelectBaseroleOrder' (a nil value)
  1. populateFn - addons/ttt2/lua/terrortown/menus/gamemode/roles/roleinspect.lua:221
   2. unknown - addons/ttt2/lua/terrortown/menus/gamemode/roles/roleinspect.lua:528
    3. callback - addons/ttt2/gamemodes/terrortown/gamemode/shared/sh_roleinspect.lua:244
     4. func - addons/ttt2/lua/ttt2/extensions/net.lua:206
      5. unknown - lua/includes/extensions/net.lua:38

```

TimGoll:
Some high level feedback points on the UI:

> image

I think you should make it more clear that this is about the last round / role selection process and not about the future / some kind of avarage. Also maybe we could make it more clear which role is "diabled" here, because right now one has to look really close to spot the difference. Maybe a colored outline or a small icon as an indicator would help. Or maybe the removal of the drop shadow from the role icon would already suffice.

> image

I think I don't understand what these bar graphs tell me. But they do look good!

This is all I could find on a higher level. All in all this looks pretty good and is really impressive. A code review will follow soon

nike4613:
> I was unable to test this due to the following error:
> 
> ```
> [ttt2] addons/ttt2/lua/terrortown/menus/gamemode/roles/roleinspect.lua:221: attempt to get length of local 'subroleSelectBaseroleOrder' (a nil value)
>   1. populateFn - addons/ttt2/lua/terrortown/menus/gamemode/roles/roleinspect.lua:221
>    2. unknown - addons/ttt2/lua/terrortown/menus/gamemode/roles/roleinspect.lua:528
>     3. callback - addons/ttt2/gamemodes/terrortown/gamemode/shared/sh_roleinspect.lua:244
>      4. func - addons/ttt2/lua/ttt2/extensions/net.lua:206
>       5. unknown - lua/includes/extensions/net.lua:38
> ```

Aha, I think this is because you don't have any subroles. Easy fix.

> Also maybe we could make it more clear which role is "diabled" here, because right now one has to look really close to spot the difference. Maybe a colored outline or a small icon as an indicator would help. Or maybe the removal of the drop shadow from the role icon would already suffice.

Something like https://github.com/TTT-2/TTT2/pull/1715 would probably be sufficient, though I think it would be worth keeping the grey-out as well.

> I think I don't understand what these bar graphs tell me.

They're the weights assigned to each player during selection. I'd thought I mentioned that in the help text, but maybe it's not clear enough.

TimGoll:
> They're the weights assigned to each player during selection. I'd thought I mentioned that in the help text, but maybe it's not clear enough.

It wasn't shown on the screenshots and I was unable to see it in-game. So maybe it is fine and I just did not see it

TimGoll:
Now that the base code for this was merged, could you please clean up the merge conflicts? That would make the review easier for me. Thank you!

Also I'm working right now on #1715. Maybe this PR should wait on this so you can use the new status feature for the role icons?

nike4613:
> Now that the base code for this was merged, could you please clean up the merge conflicts? That would make the review easier for me. Thank you!

Done.

> Also I'm working right now on #1715. Maybe this PR should wait on this so you can use the new status feature for the role icons?

Maybe. It could also be a decent idea to use the `DPiPPanel` I've defined here as the basis for that graphic instead, though I don't know how you'd want to handle doing that if so.

TimGoll:
#1715 is merged. Could you merge it to your PR and give the UI a try? Maybe the icon preview also looks decent for your usecase. Then we can discuss if we should add another mode to the role-image

nike4613:
> #1715 is merged. Could you merge it to your PR and give the UI a try? Maybe the icon preview also looks decent for your usecase. Then we can discuss if we should add another mode to the role-image
> 
> **Edit:** Ah, you use your own panel `DPiPPanel`? That is fine as well (although it would be probably best to minimize the amount of panels that have to be maintained)

The `DPiPPanel` mostly serves a different purpose: It's for displaying the baserole and subrole icons over each other, as well as he final selected role over player avatar. I'm not sure there's another decent way to do that without a dedicated panel for it. My suggestion above was that we could use the `DPiPPanel` along with some existing panel to display a little red X in the corner to indicate disabled or ignored.

nike4613:
Hmm, something might be wrong...

image

nike4613:
It looks like it's because I don't associate a convar with the role images, so `panel:GetValue()` in paint returns something falsey, resulting in it always being grayed out

TimGoll:
Ahh, well, how do you set the value? A `SetValue` should already work I think

nike4613:
I've been using `:SetEnabled()`. I can switch to `SetValue`, but that's still only half; the corner icon not scaling well with size is still an issue (albeit one relatively easily solved). I also think that about half of the usages here shouldn't have that extra icon or border at all.

nike4613:
> > And it looks like the X icon doesn't respect overall size image
> 
> The icon x currently has a hardcoded size, yes. I can make this scaleable. But I'm more confused where the bright speck in the top and left comes from? I only use a fill in lua, I don't understand where that line is coming from.

That's from another, smaller, icon! That's the point of the `DPiPPanel` Refer back to the screenshots in OP.

TimGoll:
> I've been using `:SetEnabled()`. I can switch to `SetValue`, but that's still only half; the corner icon not scaling well with size is still an issue (albeit one relatively easily solved). I also think that about half of the usages here shouldn't have that extra icon or border at all.

I think for your usecase, there should be probably be no icon, only the outline. I'd suggest making the icon toggleable. I can do this quick in a new PR if you like

nike4613:
After a quick fix to use `:SetValue` and scale the corner icon correctly:
image

I'm not completely convinced that this is the right look for this bit of the UI. I'm not sure it "not selected" as opposed to "not enabled". Same for in pre-selection honestly.
image

I also notice that in the role layering tab, layered roles always appear as disabled and don't do anything on right-click.

TimGoll:
I made them toggleable (hidden by default) in #1724

TimGoll:
Regarding you long comment: The icons look completely different for you and for me. That is strange. Your outlines are thicker in role layering and the icon is way too small?

> I also notice that in the role layering tab, layered roles always appear as disabled and don't do anything on right-click.

You're right and I see why

nike4613:
> Regarding you long comment: The icons look completely different for you and for me. That is strange. Your outlines are thicker in role layering and the icon is way too small?

This is the inspection menu; I size the icons at half what the role layering menu does so that more can fit in one row. I personally find them uncomfortably large in role layering anyway, but here it's more important that a bunch are visible.

TimGoll:
Interesting that they are too large in your opinion. I thought the icons in your design were too small :D

nike4613:
Much more reasonable, I think.

image

TimGoll:
> Much more reasonable, I think.
> 
> image
> 
> 

Looks very clean, I like it. I also think the fact that the layering UI has an outline while this hasn't works well because this UI only shows a state, while the other is there to edit it

lua/terrortown/entities/roles/ttt_role_base/init.lua
```diff
@@ -24,26 +24,42 @@ function ROLE:RemoveRoleLoadout(ply, isRoleChange) end
 -- Checks whether a role is able to get selected (and maybe assigned to a @{Player}) if the round starts
 -- @param boolean avoidHook should the @{hook.TTT2RoleNotSelectable} hook be ignored?
 -- @return boolean
+-- @return ROLEINSPECT_REASON
 -- @realm server
 function ROLE:IsSelectable(avoidHook)
```
TimGoll, line 29: keep in mind, that roles can overwrite that function and probably already do?

TimGoll:
Regarding my first comment, is there a reason it is disabled by default?

image

nike4613:
> Regarding my first comment, is there a reason it is disabled by default?
> 
> image

It records quite a lot of information, including lots of cloned tables, and persists them for an arbitrary amount of time. For large player counts, this could hypothetically become an issue, and I dislike silently changing behavior, even if it should be mostly transparent like this.

TimGoll:
> It records quite a lot of information, including lots of cloned tables, and persists them for an arbitrary amount of time. For large player counts, this could hypothetically become an issue, and I dislike silently changing behavior, even if it should be mostly transparent like this.

Ok, makes sense. If you think that it is too heavy, then it is probably the best. Personally I chose to enable things by default and give users the option to disable it. So they see that it is there and can disable it if they so chose

gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
```diff
@@ -1345,15 +1346,30 @@ function SKIN:PaintTooltipTTT2(panel, w, h)
...
+            drawSimpleText(
+                text,
+                panel:GetFont(),
+                0.5 * w,
+                0.5 * (h + sizeArrow),
+                utilGetDefaultColor(colors.background),
+                TEXT_ALIGN_CENTER,
+                TEXT_ALIGN_CENTER
+            )
+        end
```
TimGoll, line 1372: Hm. I see why you did it this way and this probably also works, but I'd recommend using our function `draw.GetWrappedText()` [(reference)](https://api-docs.ttt2.neoxult.de/module/draw/none/client/draw.GetWrappedText) for this.

Here's a usage sample: https://github.com/TTT-2/TTT2/blob/cb34ac03f11993486f33ccdbb007a89648814ee2/gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua#L1438-L1452

The benefit here is that it supports single and multiline, while also adding word wrap to limit the width.

gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
```diff
@@ -1893,7 +1909,10 @@ function SKIN:PaintSearchbar(panel, w, h)
     )
 
     -- Draw small blue bar on the bottom
-    drawBox(leftPad, h - sizes.border - bottomPad, w - widthPad, sizes.border, colorBar)
+    --drawBox(leftPad, h - sizes.border - bottomPad, w - widthPad, sizes.border, colorBar)
+
+    -- Draw small blue bar across the top
+    drawBox(leftPad, topPad, w - widthPad, sizes.border, colorBar)
```
TimGoll, line 1915: your reverted change got merged back into here

gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
```diff
@@ -2293,5 +2312,87 @@ function SKIN:PaintWeaponPreviewTTT2(panel, w, h)
...
+                panel:GetFont(),
+                item.x + item.w + padding,
+                item.y,
+                valueOutsideColor,
+                TEXT_ALIGN_LEFT,
+                TEXT_ALIGN_TOP
+            )
+        end
+    end
+end
```
TimGoll, line 2395: make sure to remove dead code

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dpippanel_ttt2.lua
```diff
@@ -0,0 +1,243 @@
...
+}
+local axisSizeTbl = {
+    [axisX] = "w",
+    [axisY] = "h",
+}
+
+---
+-- @ignore
+function PANEL:PerformLayout()
+    --self:InvalidateChildren() -- Make sure we invoke our childrens' layout functions
```
TimGoll, line 111: dead code

gamemodes/terrortown/gamemode/shared/sh_roleinspect.lua
```diff
@@ -0,0 +1,256 @@
+---
```
TimGoll, line 1: I think I'd prefer it if this file was located in `lua/ttt2/libraries/`

gamemodes/terrortown/gamemode/shared/sh_roleinspect.lua
```diff
@@ -0,0 +1,256 @@
...
+    end)
+end
+
+---
+-- Gets the decision table recorded by roleinspect.
+-- @note When called on the server, the callback is provided with the actual decision table reference.
+-- @note When called on a client which does not have admin permissions, the callback is still called, but with an empty table.
+-- @param function callback Called with the decision table as the only argument.
+-- @realm shared
+roleinspect.GetDecisions = getDecisionsImpl
```
TimGoll, line 256: I'm not sure if I like this way of implementing it. I think it would be more readable if you followed the normal approach used in our codebase. Meaning: A shared function body with checks for client and server inside.

lua/terrortown/menus/gamemode/roles/roleinspect.lua
```diff
@@ -0,0 +1,559 @@
...
+    [ROLEINSPECT_STAGE_FINAL] = PopulateFinalStage,
+}
+
+local function PopulateUnhandledRoleStage(stage, form, stageData)
+    form:MakeHelp({
+        label = "help_" .. roleinspect.GetStageFullName(stage),
+    })
+
+    -- TODO: read decisions to try to create a crude approximation of the data in
+    -- the case of new stages
```
TimGoll, line 505: is this still todo?

TimGoll (changes requested):
Not an in-depth review. I might take another look at your derma code. But overall this looks pretty good to me. You have a few blobs of dead code in your code, which should be removed. Also either tackle the ToDos, or remove them

gamemodes/terrortown/gamemode/shared/sh_roleinspect.lua
nike4613: I agree; I would prefer that as well. That would mean making what are currently server-only or client-only variables present on both, if unused. This isn't necessarily a bad thing, but that's the reason I did it this way.

nike4613: I originally did that, but the doc tool didn't like it.

lua/terrortown/menus/gamemode/roles/roleinspect.lua
nike4613: Yes. This function is where it would be implemented. I don't think its very important though.

gamemodes/terrortown/gamemode/shared/sh_roleinspect.lua
TimGoll: Ahh, then the question is directed at @histalek, our style guru :D

Histalek: i would need to take a more in-depth look at this, which will have to wait until the weekend

to be honest i'm more surprised to see that this **doesn't** break our doc tool lol

TimGoll: Probably only works because there are no parameters in this function

Histalek: > You could also implement it in both scopes individually
>> I originally did that, but the doc tool didn't like it.

This is expected because our doc tool uses the function name as an identifier

> Probably only works because there are no parameters in this function

Not quite, it is missing the `function` keyword and is therefore ignored by neodoc. The doc comments are recognized but the function itself is not -> the doc comments get discarded because there is no function to relate them to.

So the bad news: this function doesn't exist as far as the doc tool is concerned and therefore we won't have proper docs for it

The good news: the doc comments won't "spill over" into the next function if we would ever add another function below this.

---

i guess the way to workaround this would be to make this an actual shared function and handle the `if CLIENT` and `if SERVER` inside of it. that is how we do this in other cases, given it might be a bit scuffed for this case the following just might work

```lua
function roleinspect.GetDecisions(callback)
    if CLIENT then
        ---@diagnostic disable-next-line:undefined-global
        recvCallbacks[#recvCallbacks + 1] = callback

        net.Start("TTT2SyncRoleInspectInfo")
        net.SendToServer()
    else
        callback(roleinspect.decisions)
    end
end
```

Please let me know if this doesn't work!

TimGoll: This was my initial suggestion as well, their answer was:

> I agree; I would prefer that as well. That would mean making what are currently server-only or client-only variables present on both, if unused. This isn't necessarily a bad thing, but that's the reason I did it this way.

I share your opinion though that this would be the cleanest soultion

lua/terrortown/menus/gamemode/roles/roleinspect.lua
TimGoll: I think you should either implement it, or remove it. Feel free to create an issue with a feature request that points to that line. TODOs in code tend to get lost and never be addressed

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dplayergraph_ttt2.lua
```diff
@@ -0,0 +1,274 @@
...
+        if self.button then
+            self.button:SetPos(padding + tw, y)
+            local bw, bh = self.button:GetSize()
+            w = math.max(w, tw + padding + bw + padding)
+            titleHeight = math.max(titleHeight, bh)
+        end
+
+        y = y + titleHeight + padding
+        --sepY = y
+        --y = y + 1 + padding -- separator
```
TimGoll, line 209: dead code

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dplayergraph_ttt2.lua
```diff
@@ -0,0 +1,274 @@
...
+    elseif self.button then
+        self.button:SetPos(padding, y)
+        local bw, bh = self.button:GetSize()
+        w = math.max(w, bw + padding)
+        y = y + bh + padding
+        --sepY = y
+        --y = y + 1 + padding -- separator
```
TimGoll, line 216: dead code

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
```diff
@@ -212,3 +212,23 @@ end
...
+function CLGAMEMODEMENU:FillNavPanel(navPanel, contentPanel, padding)
+    -- MAKE SEPARATE SUBMENULIST ON THE NAVAREA WITH A CONTENT AREA
+    local submenuList = vgui.Create("DSubmenuListTTT2", navPanel)
+    submenuList:Dock(FILL)
+    submenuList:SetPadding(padding)
+    submenuList:SetBasemenuClass(self, contentPanel)
+
+    -- REFRESH SIZE OF SUBMENULIST FOR CORRECT SUBMENU DEPENDENT SIZE
+    submenuList:InvalidateLayout(true)
+end
```
TimGoll, line 234: Is this still in use? I think this is a leftover from a merge conflict, isn't it? See this comment: https://github.com/TTT-2/TTT2/pull/1702#discussion_r1915373028

I can't find the place where you call this from

lua/terrortown/menus/gamemode/roles/roleinspect.lua
```diff
@@ -0,0 +1,559 @@
+--- @ignore
+
+local table = table
+--local TryT = LANG.TryTranslation
+--local ParT = LANG.GetParamTranslation
```
TimGoll, line 5: dead code

lua/terrortown/menus/gamemode/roles/roleinspect.lua
```diff
@@ -0,0 +1,559 @@
...
+        --local subroles = assignment.subroles
+        --local players = assignment.players
+
+        local baseroleForm = vgui.CreateTTT2Form(
+            form,
+            DynT("header_inspect_upgrade_order", { name = baseroleData.name }, true)
+        )
+        baseroleForm:SetExpanded(false) -- default to being collapsed
+
+        -- TODO: display available subroles
```
TimGoll, line 383: todo?

lua/terrortown/menus/gamemode/roles/roleinspect.lua
```diff
@@ -0,0 +1,559 @@
...
+        params = {},
+    })
+
+    -- go through the assignment order to display selection info
+    for i, assignment in pairs(stageData.extra.upgradeOrder) do
+        -- assignment has baserole, subroles, players
+        local baserole = assignment.baserole
+        local baseroleData = roles.GetByIndex(baserole)
+        --local subroles = assignment.subroles
+        --local players = assignment.players
```
TimGoll, line 375: dead code

lua/terrortown/menus/gamemode/roles/roleinspect.lua
```diff
@@ -0,0 +1,559 @@
...
+            -- empty table, put in an appropriate message
+
+            local labelHolder = vgui.Create("DPanelTTT2", parent)
+            labelHolder:Dock(TOP)
+            labelHolder:DockMargin(20, 20, 20, 20)
+            local labelNoContent = vgui.Create("DLabelTTT2", labelHolder)
+            labelNoContent:SetText("label_roleinspect_no_data")
+            labelNoContent:SetFont("DermaTTT2Title")
+            labelNoContent:Dock(FILL)
+            --labelNoContent:FitContents()
```
TimGoll, line 539: dead code

lua/terrortown/menus/gamemode/base_gamemodemenu.lua
nike4613: Argh, this is why I dislike squashing. Fixing.

TimGoll: Understandable. But without squashing we'd need all contributors to have clean commits, because we don't want stuff like this in the history:

image
(No worries, my commits look the same)

Histalek for example follows clean commit strategies, but this is (in my opinion) too advanced for the average contributor. Moreover, this project already has more than 2000 commits with squashed PRs. Without it, we would be way north of 10k. Which isn't a bad thing per se, but probably makes tracking down problems in the history even harder.

nike4613: I personally dislike rewriting git history in general. and would rather a messy but accurate commit history over a clean but less-than-helpful history.

TimGoll: Take a look at our first 500 commits and then tell me if you stand by your opinion :D

lua/ttt2/libraries/roleinspect.lua
```diff
@@ -257,4 +245,15 @@ end
 -- @note When called on a client which does not have admin permissions, the callback is still called, but with an empty table.
 -- @param function callback Called with the decision table as the only argument.
 -- @realm shared
-roleinspect.GetDecisions = getDecisionsImpl
+function roleinspect.GetDecisions(callback)
+    if CLIENT then
+        recvCallbacks[#recvCallbacks + 1] = callback
+
+        net.start("TTT2SyncRoleInspectInfo")
```
TimGoll, line 252: net.Start with capital S

TimGoll (approved):
From my POV this PR is good as it is now. Since this is a rather big PR, I'd like a quick ok from @Histalek or @saibotk here as well. You don't have to review this in depth, but please give it a quick scroll through

gamemodes/terrortown/gamemode/server/sv_roleselection.lua
```diff
@@ -840,6 +1121,14 @@ local function SelectForcedRoles(plys, selectableRoles)
             roleselection.finalRoles[ply] = subrole
             curCount = curCount + 1
 
+            roleselect.ReportDecision(
```
Histalek, line 1124:
```suggestion
            roleinspect.ReportDecision(
```

Histalek (changes requested):
the function typo needs fixing, otherwise this seems good to go :)

Amazing work!

lua/ttt2/libraries/roleinspect.lua
```diff
@@ -0,0 +1,259 @@
...
+        end
+
+        -- otherwise, send back the full decision table
+        net.SendStream("TTT2SyncRoleInspectInfo", roleinspect.decisions, { ply })
+    end)
+end
+
+local recvCallbacks
+if CLIENT then
+    recvCallbacks = {}
```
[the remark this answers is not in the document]
nike4613: Yep. The `if` is a local scope, meaning either `GetDecisions` needs to be defined in it, or `recvCallbacks` needs to be defined outside it.

Histalek (approved):

PR #1715, opened 2025-01-14
Rolelayering: Show Role Enable Status

### LINKED ISSUE #1700: Improve Role Layering Menu
## Is your feature request related to a problem? Please describe.

There are currently at least 2 pain points with the role layering menu:
* each baserole that has less than 2 subroles installed doesn't have its subrole-layering section displayed leading to confusion when switching between different servers/collections
* disabled subroles still show up the same as enabled ones do, which makes it harder to understand what the rolelayering will do
* (does the order of subroles in a layer even have any effect? being allowed to manually order items in a layer gives the impression that the order matters)

## Describe the solution you'd like

* Show a notice why subrolelayering is not possible for each installed Baserole instead of hiding the whole section with less than 2 subroles installed
* visually distinct roleicons for disable subroles, maybe grayscale the color (subroles are sorted into their baseclasses section here, so the color isn't as important here anyways)
* (if the order inside a layer has no effect: disable manual ordering inside of a layer and autosort by roleid or rolename)

TimGoll:
Fixes #1700 by showing the enable status of a role.

Looks something like this:
image

Currently I'm struggling with the fact that the role state cvar is server only. Altough sven already implemented a nice system for that, it requires a lot of code duplication to get it working in this UI element as well.

This is how it works for checkboxes: https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcheckboxlabel_ttt2.lua#L90

Best case would be to use inheritance here, but our vgui inheritance is pretty broken.

I'm open for ideas on this!

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
@@ -1721,11 +1721,70 @@ end
 -- @param number h
 -- @realm client
 function SKIN:PaintRoleImageTTT2(panel, w, h)
+    local widthBorder = 2
+    local widthBorder2 = widthBorder * 2
+    local padding = 3
+    local sizeMode = 18
+    local sizeIconMode = sizeMode - 2 * padding
+    local posIconModeX = w - sizeMode + padding
+    local posIconModeY = h - sizeMode + padding
+
     local colorBackground = panel:GetColor()
     local colorIcon = utilGetDefaultColor(colorBackground)
 
-    drawRoundedBox(sizes.cornerRadius, 0, 0, w, h, colorBackground)
-    drawFilteredShadowedTexture(0, 0, w, h, panel:GetMaterial(), colorIcon.a, colorIcon)
+    local colorBorder = colorCardAdded
+    local iconBorder = materialCardAdded
+
+    if not panel:GetEnabled() then
+        colorBorder = colorCardInheritRemoved
+        iconBorder = materialCardRemoved
+    end
+
+    local colorBorderIcon = utilGetDefaultColor(colorBorder)
+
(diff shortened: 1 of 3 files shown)


TimGoll:
image

Slowly making progress with the inheritance. But it is a bit painful

TimGoll:
Got it working. Will finish this PR once #1722 is merged

TimGoll:
@nike4613 Maybe you want to take a quick look on this as well

nike4613:
Is the visual appearance still the same as in the OP? If so, I might suggest only showing an extra icon when it's disabled, and dimming the role icon in that case as well.

TimGoll:
> Is the visual appearance still the same as in the OP? If so, I might suggest only showing an extra icon when it's disabled, and dimming the role icon in that case as well.

I'm confused, the role icon is dimmed in the image in the OP?

(yes, I updated the screenshot)

saibotk (approved):

nike4613:
I personally would prefer keeping the coloration (though with dimming/darkening/whatever) as opposed to completely removing it here. I quite like the way that the dimmed icons look over in #1714 even without the extra overlay clarifying that they're disabled.

TimGoll:
> I personally would prefer keeping the coloration (though with dimming/darkening/whatever) as opposed to completely removing it here. I quite like the way that the dimmed icons look over in #1714 even without the extra overlay clarifying that they're disabled.

I tried that at first, but I didn't like it because it wasn't that clear on first glace. For your usecase I see the reason why you'd like that though. You use the role-image element to show if a role was selected or not - am I right?

We should probably extend the role image then a bit. Since a gray version with a red border probably doesn't make too much sense for your use-case? I'm not sure.

I'll merge this PR for now, but I'm open to add more to this for your UI. Personally, I think the design language I chose here is better for this specific use case, but not for yours.

PR #1717, opened 2025-01-15
Fixed error in new Ammo drop

MrXonte:
now references the already created ammo box instead of trying to get data from via weapons ammo entity reference which can return nil.


TimGoll (approved):
LGTM, thank you

saibotk (approved):
Very nice 👍 Thanks for submitting the fix

Issue #1402, opened 2024-02-09
Fonts: Rework


TimGoll:
This is an issue with multiple things that should be changed for fonts. This can happen over multiple PRs

- Switch from `Tahoma` to a different font that is distributed with TTT2. Currently `OpenSans` seems like a good canidate
- Move all font creation to a single file, that way we might be able to remove some duplicates
- Re-/Create fonts in `TTT2FinishedLoading` so that our advanced font system is hotreload safe

Idea:
Don't pregenerate advanced fonts for 1.0, 1.5 and 2.0 scale but generate the fitting font on scale change. So if you set the scale to 1.3, all fonts are regenerated to use that scale. We might have to support the old system still, because some addons might use it

EntranceJew:
the main thing is that we generate a font starting from the global scale, so that the initial font selected is not stretched, like on 1.7:
![gmod_vN624aaBSH](https://github.com/TTT-2/TTT2/assets/5711436/f8ed4cdb-3fdb-40d5-b661-36c30d065178)
vs 2.0 (applied same-session):
![gmod_CP4fU1NVzu](https://github.com/TTT-2/TTT2/assets/5711436/19b55410-712f-4b81-9f8d-df1137e99c00)
vs 2.0 (applied after a map restart):
![gmod_UrerlAPqGn](https://github.com/TTT-2/TTT2/assets/5711436/02f99c32-a493-49fd-9d8b-f8c7ef344136)

and for the "breakpoint" font sizes to correspond with a multiple of the initial font size, e.g.
`1.7, 2.55, 3.4, 4.25` and we use those scales when the requested font scale is `1, 1.5, 2.0, 2.5` so that they're crisp as well

internally, i believe nearly all the font scale calls are using one of the preset breakpoints

TimGoll:
> internally, i believe nearly all the font scale calls are using one of the preset breakpoints

internally we only use the 1.0 scale, or the scale defined with the slider. We never increase a font by using a scale factor. So if we were to recreate fonts on scale change, we only would need a single font

EntranceJew:
[TTT2_OS_SC_SB.zip](https://github.com/TTT-2/TTT2/files/14221221/TTT2_OS_SC_SB.zip)
also here's the font i edited the name to if we wanted to use `Open Sans SemiCondensed SemiBold` across the board, i renamed it to `TTT2_OS_SC_SB` and that's what the `CreateFont` declaration should call it as well

PR #1719, opened 2025-01-18
Implement UI scaling in Derma UIs

nike4613:
Currently, only the HUD is scaled. This is rather annoying for those with high DPI displays, as much of the text is too small to be readable. Instead, we want to present as much of our UI according to the user's specified UI scale as possible, which this PR does.

This unfortunately means threading `* scale` through a bunch of layout and skin code, and it also means that non-default skins will not work until updated (This is particularly onerous with the role debugger and embedded legacy VGUI menus). I tried, at first, implementing a fully-general system that would automatically attempt to scale things which are not hidpi-aware, but that NOT ONLY had terrible performance, but ALSO didn't work.

In the process of doing this, I also discovered that `draw.AdvancedText` was implemented fundamentally incorrectly, and have fixed it.

Some limitations:
- For some reason, I *cannot* convince `Panel:DrawTextEntryText()` to use a different font set dynamically or be scaled. Is `Panel:SetFont()` frame-delayed?
- Body search result items have their description currently hard-coded to `DermaDefault`, which doesn't have scaled variants and so looks bad (though it is scaled! I spent 3 whole days on this!)

Currently unchanged:
- Scoreboard
- Buy menu

I intend to continue this and apply scaling to the scoreboard as well, but I don't think the buy menu is reasonably possible without giving it a whole UI overhaul.

Before:
Screenshot_20250118_024453
Screenshot_20250118_024459
Screenshot_20250118_024507
Screenshot_20250118_024548

After: (1.5x UI scale)
Screenshot_20250118_025243
Screenshot_20250118_025245
Screenshot_20250118_025250
Screenshot_20250118_025410

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_help.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_help.lua
@@ -224,13 +224,21 @@ function HELPSCRN:ShowMainMenu()
     local maxHeight = rows * heightMainMenuButton * scale
         + (rows - 1) * padding
         + ((#menusAdmin == 0) and 0 or heightAdminSeperator * scale)
-    local heightScroll = height * scale - vskin.GetHeaderHeight() - vskin.GetBorderSize() - 2 * padding
+    local heightScroll = height * scale
+        - vskin.GetHeaderHeight()
+        - vskin.GetBorderSize()
+        - 2 * padding
 
     local scrollSize = (heightScroll < maxHeight and 20 or 0) * scale
 
     local widthMainMenuButton = (width - 4 * padding - scrollSize) / 3
 
-    AddMenuButtons(menusNormal, dsettings, widthMainMenuButton * scale, heightMainMenuButton * scale)
+    AddMenuButtons(
+        menusNormal,
+        dsettings,
+        widthMainMenuButton * scale,
+        heightMainMenuButton * scale
+    )
 
     -- only show admin section if player is admin and
     -- there are menues to be shown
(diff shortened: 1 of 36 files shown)


TimGoll:
Wow, you're really doing the heavy lifting here. I'm impressed. Another point on my todo list I thought I'd never get around to.

> This is particularly onerous with the role debugger 

This addon is broken anyway since a while. This is the reason I started adding lots of features of that addon to TTT2. The actual role debugging is missing though. The dev of that addon said they wanted to contribute to TTT2 and add that stuff, but I guess they didn't have the time so far.

A few things I want to add:

- font scaling is currently very basic. I spent some time in summer 2024 digging deeper and I have an idea how to vastly imrove the performance and visual fifelity of fonts. Didn't have the time and motivation yet though
- Right now, for vgui all sizes are recalculated and all strings are translated every frame; this could be cached which would improve the performance quite a lot

nike4613:
There's quite a lot of VGUI which makes this painful. Honestly, I'm tempted to try to make another layer above VGUI which enables UI code to not have to care about scaling or translation caching, and have all of that handled automatically. That would require a LOT more work, but may wind up being more worthwhile long-term.

TimGoll:
> There's quite a lot of VGUI which makes this painful. Honestly, I'm tempted to try to make another layer above VGUI which enables UI code to not have to care about scaling or translation caching, and have all of that handled automatically. That would require a LOT more work, but may wind up being more worthwhile long-term.

I tried this at one point with the `drawsc` library. And while this improved a lot of things for the HUD code, in the end it was way too basic to make a different.

VGUI right now is aweful, mostly due to me. VGUI is built on top of inheritance, but I hacked quite a lot of features in it which breaks the inheritance. Moreover vgui by itself is designed to be cached and only updated when changes happen, but at the time I wasn't aware, so my code completely ignores it.

If one would want to exted it, it probably needs a partial rewrite.

TimGoll:
Ah, and here's the issue about the font stuff. Some of my investigation results are in there, but I guess most of them are just in my head and on paper: https://github.com/TTT-2/TTT2/issues/1402

TimGoll:
IN #1752 I'm currently rebuilding how the color caching works. Once this is done I can show you how I did this and you can use the same system to rescale alle UIs.

PR #1722, opened 2025-01-19
vgui: Reworked inheritance

TimGoll:
This PR adds no new code or new functionality, it only moves code around in files so that everything that inherits from buttons can use (server) convars. Therefore the code from the ttt2 checkboxlabel was moved to button. This caused this change to track the underlying button in those labels:

image

Moreover I added our tooltip code from the panel to the label as well, as both of these are a starting point for the inheritance that can't be merged.

I'd appreaciate a quick merge so that #1715 can be finished.

Also note, that there are MANY things that could be further improved by improving the inheritance. This only tackles a small part of that.

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcheckboxlabel_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcheckboxlabel_ttt2.lua
@@ -10,11 +10,6 @@ local PANEL = {}
 AccessorFunc(PANEL, "m_iIndent", "Indent")
 
 ---
--- @accessor bool
--- @realm client
-AccessorFunc(PANEL, "ignoreCallbackEnabledVar", "IgnoreCallbackEnabledVar", FORCE_BOOL)
-
----
 -- @ignore
 function PANEL:Init()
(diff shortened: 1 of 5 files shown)


TimGoll:
image

this is done since our custom tooltip doesn't ignore empty setters

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dcheckboxlabel_ttt2.lua
```diff
@@ -64,125 +63,7 @@ function PANEL:Paint(w, h)
...
-
-    if self.inverted then
-        val = not val
-    end
-
-    self.Button:SetValue(val)
-end
-
----
+-- overwrites the default function
```
ZenBre4ker, line 66: Shouldnt you call the parent function here and only add code for the reset?

ZenBre4ker (approved):
So checkbox does not really inherit, but roleimage does, right?

TimGoll:
> So checkbox does not really inherit, but roleimage does, right?

Checkbox is still a gmod based element that therefore can't inherit from our code. The checkboxlabel, which includes a checkbox, inherits from our button same as role image

PR #1724, opened 2025-01-19
vgui: Toggleable Role Indicator

TimGoll:
Added option to show and hide these indicators for `DRoleImageTTT2`.

image

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
@@ -1777,6 +1777,10 @@ function SKIN:PaintRoleImageTTT2(panel, w, h)
         )
     end
 
+    if not panel:IndicatorEnabled() then
+        return
+    end
+
     drawRoundedBoxEx(
         sizes.cornerRadius,
         w - sizeMode,
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/droleimage_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/droleimage_ttt2.lua
@@ -9,6 +9,7 @@ function PANEL:Init()
     self.data = {
         color = COLOR_WHITE,
         icon = nil,
+        indicatorState = false,
     }
 end
 
@@ -47,6 +48,20 @@ function PANEL:DoRightClick()
 
     self:SetValue(newValue)
     self:ValueChanged(newValue)
+end
+
+---
+-- param boolean state
+-- @realm client
+function PANEL:EnableIndicator(state)
+    self.data.indicatorState = state
+end
+
+---
+-- @return boolean
+-- @realm client
+function PANEL:IndicatorEnabled()
+    return self.data.indicatorState or false
 end
 
 ---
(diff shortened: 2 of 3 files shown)


nike4613:
I think if the icon goes, so should the outline. That will be the most useful state for #1714.

TimGoll:
> I think if the icon goes, so should the outline. That will be the most useful state for #1714.

Can do that

TimGoll:
image

done

gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
```diff
@@ -1723,6 +1723,7 @@ end
 function SKIN:PaintRoleImageTTT2(panel, w, h)
     local widthBorder = 2
     local widthBorder2 = widthBorder * 2
+    local sizeIconRole = w - widthBorder2
     local padding = 3
     local sizeMode = 18
```
nike4613, line 1728: Can we make sure this scales with the size of the image as painted?

nike4613 (changes requested):

gamemodes/terrortown/gamemode/client/cl_vskin/default_skin.lua
TimGoll: I'm no longer on my PC for today. So yes, we can do this, but I can't do it today.

Also keep in mind that these outlines are hard coded for all UI elements so far

nike4613: The outlines are mostly fine, it's the child icon that poses a problem. See https://github.com/TTT-2/TTT2/pull/1714#issuecomment-2600934166

TimGoll: Ah, understood. I will probably have some time in the coming days. Otherwise we merge this as is and you modify it to your liking

nike4613:
```suggestion
    local sizeMode = 18 * w / 64
```

nike4613 (approved):

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/droleimage_ttt2.lua
```diff
@@ -49,6 +50,20 @@ function PANEL:DoRightClick()
...
+function PANEL:EnableIndicator(state)
+    self.data.indicatorState = state
+end
+
+---
+-- @return boolean
+-- @realm client
+function PANEL:IndicatorEnabled()
+    return self.data.indicatorState or false
+end
```
saibotk, line 65: I think indicator is pretty generic and let's stick to the vgui naming scheme for such states.
It seems like the majority uses `SetXY` namings e.g. https://wiki.facepunch.com/gmod/DLabel:SetDark
Let's maybe use `SetIsActiveIndicator` and `GetIsActiveIndicator`

saibotk (changes requested):
only a minor nit

gamemodes/terrortown/gamemode/client/cl_vskin/vgui/droleimage_ttt2.lua
TimGoll: Done

saibotk (approved):

PR #1726, opened 2025-01-20
HUD: Added Team name next to Role name

milkwxter:
Hello!

I have encountered a problem on my friends server where if we have many roles installed, some players who join will not know what every single role is, and they will become confused, especially with what team they are supposed to play on.

This new change will allow players to see what team they are on, immediately next to their role name. I know there is a team ICON next to the time remaining in the match on the top of the screen, but our players either never even noticed it, or did not know what it meant.

I have included a toggle in the F1 menu, under HUD Switcher so that players who feel like they don't need to see the team can turn it off and enjoy the game their way.

If you guys have any changes to be made I would love to hear it, I think this change will help new player retention!

Here are some previews of how it will look, including the F1 menu and an example with a long role name!
f1_menu
short_name
long_name


TimGoll:
> but our players either never even noticed it, or did not know what it meant.

I've noticed this as well. Quite a few YouTubers that played TTT2 for hours, did never notice any of the icons. Peter from PietSmiet was once confused what the red knife next to the corpse of a dead traitor means. I was taken aback, because this icon is shown as a team and role icon, ALL THE TIME.
The same holds true with the key bind info. They are always on the bottom right corner. Even with explanation when you hold tab. But still, I've seen them wondering how to interact with an item, while it is shown there.

I'm conflicted if we should make it more prominent like you did, or just find a way to teach players to use the UI.

---

Regarding your implementation: You have to support translations

lua/terrortown/lang/en.lua
```diff
@@ -2403,6 +2403,10 @@ L.searchbar_roles_placeholder = "Search roles..."
 L.label_menu_search_no_items = "No items matched your search."
 
 L.submenu_roles_overview_title = "Roles Overview (READ ME)"
+
+-- 2025-01-20
+L.label_hud_show_team_name = "Should you see your team name next to your role name?"
```
TimGoll, line 2408: please rename this to something like "Show team name next to role name" or "Enable showing team name next to role name"

TimGoll: Also, this should be at the end of the file

milkwxter: For sure, I will do this now!

gamemodes/terrortown/gamemode/shared/hud_elements/tttinfopanel/pure_skin_playerinfo.lua
```diff
@@ -62,6 +62,16 @@ if CLIENT then
                 slf:SaveData()
             end,
         }
+		
+		savingKeys.showTeamName = {
+            typ = "bool",
+            desc = "label_hud_show_team_name",
+            default = true,
```
TimGoll, line 69: The default should be false. I'm fine with this being an option, but the team shouldn't be shown there by default

milkwxter:
Regarding the translations, will Google Translate do? Or is there a better resource to use, since I only speak English?

TimGoll:
> Regarding the translations, will Google Translate do? Or is there a better resource to use, since I only speak English?

Sorry, I was unclear. The team name should be translated. That is already a feature in TTT2. just use `LANG.TryTranslation(team)`, where `team` is the identifier such as "traitors" or "innocents"

milkwxter:
Awesome! I got the translation working I will show it below!

image

TimGoll:
This is better, but the default team is probably not the best. You should probably get the team of the player (e.g. if you're a sidekick, your team won't be the default team). `L[client:GetTeam()]` should work here - I think

TimGoll:
Just a hint: we use spaces to indent.

And if you install stylua in VSCode it autoformats the file for you

milkwxter:
Haha I think I got it now thank you! I am used to tabs and I use notepad++ so it was a little confusing figuring out what was wrong

TimGoll:
That's weird. This setting should be disabled by default

PR #1732, opened 2025-01-21
Fix the ordering of param declarations

SmartMario1:
Switch 1 line to make the param ordering of GM:TTT2GiveFoundCredits(ply, rag) in the docs follow the ordering of the actual function.


saibotk (approved):
Thank you very much!

PR #1733, opened 2025-01-21
Added Win Score Event

MrXonte:
Added win score to finish event with parameter for amount of score in role base and LANG entries

I felt like it was weird that no score was gained for winning, especially for roles that cant gain score as quickly in other ways. This adds an event for that where score amount can also be configured for every role.


lua/terrortown/events/finish.lua
```diff
@@ -122,6 +122,7 @@ function EVENT:CalculateScore()
             score_timelimit = wintype == WIN_TIMELIMIT and math.ceil(
                 otherAlivePlayers * roleData.score.timelimitMultiplier
             ) or 0,
+            score_win = wintype == team and roleData.score.winMultiplier or 0,
```
TimGoll, line 125: you can expect this value to be set since you set it in the base role

TimGoll (approved):
I personally do like this. TTT handled winning differently. Each surviving player from a team gave a bonus. That way you were awarded for winning and keeping your team mates alive.

But since TTT2 puts a bigger focus on different teams, I think this change is valid. In general I'm all for adding more events that give score points (like a single point for the first person to call a dete to a corpse) to emphasis beneficial actions.

Before merging I think that @Histalek and @saibotk should give their opinion on this as well

TimGoll:
image

I'd recommend making feature branched for each PR you do, that way you make sure that old commits are no longer listed in the changelog. Alternatively you could sync your fork before adding new commits.

lua/terrortown/entities/roles/ttt_role_base/shared.lua
```diff
@@ -40,9 +40,12 @@ ROLE.score = {
     -- round ended with nobody winning, usually a negative number.
     timelimitMultiplier = 0,
 
-    -- The amount of points gained by killing yourself. Should be a
+    -- The amount of score points gained by killing yourself. Should be a
     -- negative number for most roles.
     suicideMultiplier = -1,
+
+    -- The amount of score points gained when your team wins.
+    winMultiplier = 1,
```
TimGoll, line 48: I think you should increase this value, even for default TTT roles. It should probably be in the region of 2 to 4 points

MrXonte: changed it to 3, which i use as the default on my own server 👍 

as for the old commit, that's the first time I've had this issue, I did sync the fork before making the changes, no idea why this time there was an issue.

saibotk (approved):

lua/terrortown/events/finish.lua
MrXonte: updated my code

PR #1734, opened 2025-01-21
C4: Reworked UI

TimGoll:
Currently done:

- basic UI element
- press&hold buttons to replace the old weird 2 stage button to destroy the entity

### DIFF
--- a/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua
@@ -7,6 +7,74 @@ local starttime = C4_MINIMUM_TIME
 local T = LANG.GetTranslation
 local PT = LANG.GetParamTranslation
 
+---
+-- Calculates and caches the dimensions of the bodysearch UI.
+-- @realm client
+function ENT:CalculateMenuSizes()
+    self.sizes = {}
+
+    self.sizes.width = 600
+    self.sizes.height = 500
+    self.sizes.padding = 10
+
+    self.sizes.heightButton = 45
+    self.sizes.widthButton = 160
+    self.sizes.widthButtonClose = 100
+    self.sizes.heightBottomButtonPanel = self.sizes.heightButton + self.sizes.padding + 1
+
+    self.sizes.widthMainArea = self.sizes.width - 2 * self.sizes.padding
+    self.sizes.heightMainArea = self.sizes.height
+        - self.sizes.heightBottomButtonPanel
+        - 3 * self.sizes.padding
+        - vskin.GetHeaderHeight()
+        - vskin.GetBorderSize()
+end
+
+function ENT:ShowMenu()
+    self:CalculateMenuSizes()
+
+    local frame = self.menuFrame
+
+    -- IF MENU ELEMENT DOES NOT ALREADY EXIST, CREATE IT
+    if IsValid(frame) then
+        frame:ClearFrame(nil, nil, "c4_arm")
+    else
+        frame = vguihandler.GenerateFrame(self.sizes.width, self.sizes.height, "c4_arm")
+    end
+
+    frame:SetPadding(self.sizes.padding, self.sizes.padding, self.sizes.padding, self.sizes.padding)
+
+    -- BUTTONS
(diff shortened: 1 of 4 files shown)


TimGoll:
image

some progress with the UI. While it is far from done, my general idea is already there:

1. A text box on top that explains some things that can be ignored once you know how it works
2. The time slider is replaced by a preset of 5(?) predefined times. Makes it easier to select and define differences
3. the area below will show the effects of the selected time on the explosion in a visible pleasing manner; right now my idea was to have the following effects:
  - volume of C4 beeper
  - amount of wires that lead to instant explosion on defusal attempt
  - amount of fire spawned around C4
  - C4 killradius
 These values should be balanced in such a way that no C4 is inherintly better than the other, only situational differences

MrXonte:
2) While i like removing the clonky slider, it did allow you to set the timer close to important round progress points like roundtime and overtime end. And depending on your server config there can be huge differences in max/min time available.
Would it maybe make sense to instead have a Time with a -30s and +30s button next to it? This is a granular enough increment i think and you can still clearly show the effects from the increase 🤔 

3) Great Idea! Its probably a huge pain to do, but it would be amazing if the kill radius would be shown visually, like this:
image
So showing a person as a size reference, and the a half circle for the radius? (bigger radius = smaller reference person)

4) What exactly do you mean here? I assume something like more time -> range++, defuse difficulty++, offset by beep volume++ and the increased time to find it? or did you have a more complex interaction in mind?

TimGoll:
2. While we can make the times dynamic, your +/- 30s remove the possibility to define a handful scenarios. Allowing for dynamic time changes defeats this purpose
3. I honestly wanted to make this waaay simpler. Something like "Radius: 20m" in a neatly rendered way
4. yeah, something along those lines, I have not yet figured this out

Crashington:
2. Hmm, personally, I am not sure about the complete removal of the slider as an option. I see where you're coming from but personally greatly appreciate the exact control over the timer.
3. I like it just being displayed as a readable distance in irl units. Is an F1 menu toggle to switch between meters and feet needed :D?
4. A longer C4 should be quieter, but should it maybe scale higher? I can see arguements for basically all implementations of volume scaling/levels i can think of.

I think personally I would make it so there are only two or three discrete levels. the normal beeps, which are only audible within the same room. Then the beeps withing the last minute or so, where they should be audible in a few rooms over. and finally the the last 30 seconds or so, where it gets as loud as it gets currently. I think putting it in discrete stages like this makes it easier  for (trained) innocents to understand where they are at, and how to react, while it should at the same time make it a bit more tense when players immediatly notice "oh we are already in the endgame"

I am not sure how to scale this to the fuse time though. if this should all be based on percents of the chosen of time, or if shorter bombs should immediatly start in the later levels or what

TimGoll:
2. I personally hated the slider because nobody really bothers to set an exact time anyway. And probably nobody noticed the correlation between the defuse wires and the time. Having a fixed amount of presets makes that clear. Also I don't get why you have to set it to 2:06 instead of 2 minutes - is there really a difference? Because I'm always trying to achieve to things: simplify the options and explain more.
3. Currently we only support metric, and I'm not really interested in supporting anything else

Crashington:
> 2. I personally hated the slider because nobody really bothers to set an exact time anyway 

awww, i definitly do that haha. but i get what you mean. I've got to say, I did not entirely get the problem with having discrete 30 second or so steps to choose from. It feels like a great compromise to me. these can still be discretely defined, it would be more a matter of displaying them so that there arent like 10 different options to choose from but you instead cycle through them by pressing the "time up" or "time down" buttons

TimGoll:
There is no real problem, maybe your solution is even the best. My point was: Nobody wants to fine tune their C4, nobody has the time. I want to provide 3 to 5 discrete time stamps that have their own strategic benefits. When having a +-30s button, you have a slider, but worse :D It is even slower than a slider, while providing less granularity

TimGoll:
After discussions that we had in different PRs I decided to change gears on that. First of, I will finish https://github.com/TTT-2/TTT2/pull/1752 first. After that, I will go forward with my initial idea, however same as with armor I will add a "classic mode" that is exactly as it is now (with a nicer UI though)

Issue #1730, opened 2025-01-20
C4 distance indicator not correct


Crashington:
- Workshop version from Steam

## Describe the bug (mandatory)

The indicator that appears when looking at C4 (the indicator that team members can see through walls) does not accurately reflect the kill, damage and no hurt zones of C4. players will survive standing within the "kill distance" taking anywhere from close to lethal (but not lethal) or no damage at all.

## To reproduce

Steps to reproduce the behaviour:

1. Plant a C4
2. Stand near where "Kill Distance" changes to "Damage Distance"
3. Wait for the C4 to explode and check the damage dealt

## Expected behaviour

As long as the inidcator says "Kill distance" in red, it should actually kill the player.

Crashington:
maybe it could be changed from "kill distance" to "sure kill distance" or something, and get the inner bounds of that range variation?

TimGoll:
Maybe. When I do the C4 rework, I will look into that as well

MrXonte:
from a quick look at the warning, there is no LoS check in there
Within LoS we do util.BlastDamage in a 1000 unit radius (although i cant find any reference how util.BlastDamage calculates its damage)
Regardless of LoS we also do SphereDamage in a 750 unit radius

Unless i made a grave mistake somewhere, that means we are lethal within LoS for muuuuch further than out of LoS, while the check only checks hard distances. From the indicator code we have the following states based on distance:

- >1000 (>outerRadius): Safe
- 1000 to 750 (outerRadius to innerRadius): Damage Zone. Only true if we are in LoS
- <750 (<innerRadius): Kill Zone. Now if i did my math right, that's actually not the case. The calculation for SphereDamage uses absolute numbers (instead of a linear or exponential scaling with a surekill radius) and somehow that means we have a space of about 150 units where we do no/negative damge:
![Image](https://github.com/user-attachments/assets/e8cc9526-3c69-43b8-874b-59fe7ca465c6)

So we either need to include a LoS check, or have uniform damage distances for LoS and no LoS.

MrXonte:
Also, how _should_ C4 deal its damage? Currently its has a reasonable model of more damage range in LoS, but at least in my experience that really doesn't matter. I think I've never seen a C4 sitting in the open, they are always in a vent or something like this where LoS damage is meaningless.

[...]

Issue #1729, opened 2025-01-20
Buff C4


Crashington:
## Is your feature request related to a problem? Please describe.
I think C4 is consistently underwhelming and underpowered. Let alone the many maps that are simply too big to effectively use C4 (which isnt the item#s fault), but even on small maps, C4 only gets even a single kill *extremely* rarely, even when not noticed.
 
## Describe the solution you'd like
I would like for C4 to scale its range (and damage) to the length of the timer. 
This way, a long timer puts much more pressure on the innocents to actually try and defuse it, instead of just walking out of the (static) range.

Ideally, I would want some convars to customize this via the F1 menu, like being able to change much the lenght of the timer scales explosion distance, explosion damage and beeping volume.

## Describe alternatives you've considered
Alternatively the item could just be buffed in the raw numbers.

TimGoll:
I think it already works that way? If it doesn't, then it definitely should!

Crashington:
I also though this is how it worked before opening the issue! But nothing ingame says it works this way, the range indicator does not change, and anecdotally, if it changed, it definitly doesnt change enough (hence the want for f1 convars to adjust it),

Glad you agree in principle :D

TimGoll:
at some point in 2025 I will rework the UI. While doing so, I'll rework this as well

MrXonte:
as I mentioned in reply to #1730 , the non LoS damage seems to completely disregard radius when it comes to damage. Radius only affects the cutoff for the damage calculation, not the damage itself. So even if you had a radius of 999999999 units, you wouldn't do any damage beyond ~600 units as the scaling then turns damage negative. The hardkill radius of up to 490 units is hardcoded, as is the falloff to 0 over the next sqrt(12500) units.

PR #1739, opened 2025-01-23
C4 Rework

MrXonte:
Now scales C4 damage with radius and gives consistent damage through walls regardless of line of sight.
Relates to #1729 and #1730 although the calculation for proper Killzone is not yet included, but the safe and killzones are now 100% correct.


MrXonte:
Also, whats wrong with the style check? Cant figure out what the problem here is
image

TimGoll:
> im not sure what im doing wrong for the total history to show up, but I sycnched my fork before PR #1738 and it showed up, now I made a branch and its still all there

If you make a new branch AFTER the commits were added, the new branch won't change anything. 

I barely work with forks, so my knowledge is spotty. But you probably not only need to sync to the host, but also rebase it, so your commit history is removed in favor of ours (we squash PRs, so your commits are not in our repository)

TimGoll:
> Also, whats wrong with the style check? Cant figure out what the problem here is image

@Histalek looks like stylua crashed?

lua/ttt2/libraries/game_effects.lua
```diff
@@ -148,6 +148,73 @@ function gameEffects.RadiusDamage(dmginfo, pos, radius, inflictor)
...
+        innerRadius = outerRadius
+    end
+
+    -- pre-declare to avoid realloc
+    local d = 0.0
+    local dFraction = 0.0
+    local diff = nil
+    local dmg = 0
+    local radiDiff = (outerRadius - innerRadius)
+    for _, ply in pairs(player.GetAll()) do
```
TimGoll, line 188: pairs is really slow, ipairs is better, but best is:

```lua
local plys = player.GetAll()

for i = 1, #plys do
    local ply = plys[i]
end
```

Histalek:
> > Also, whats wrong with the style check? Cant figure out what the problem here is image
> 
> @Histalek looks like stylua crashed?
> 
> @MrXonte the crash might have happened because you removed the empty line at file end (which always should be there)
> 
> image

that's not a crash and only stylua reporting a style issue (but that style issue is about non-printable characters, which makes the output you see kinda useless)

and yes you're spot on with the missing newline at the end of the file

lua/ttt2/libraries/game_effects.lua
MrXonte: done

lua/ttt2/libraries/game_effects.lua
```diff
@@ -148,6 +148,75 @@ function gameEffects.RadiusDamage(dmginfo, pos, radius, inflictor)
...
+    -- pre-declare to avoid realloc
+    local d = 0.0
+    local dFraction = 0.0
+    local diff = nil
+    local dmg = 0
+    local radiDiff = (outerRadius - innerRadius)
+    local plys = player.GetAll()
+    for i = 1, #plys do
+        local ply = plys[i]
+        if IsValid(ply) and ply:Team() == TEAM_TERROR then
```
TimGoll, line 191:
```suggestion
    for i = 1, #plys do
        local ply = plys[i]

        if not IsValid(ply) or not ply:IsTerror() then
            continue
        end
```

make sure to remove the end statement at the end as well

lua/ttt2/libraries/game_effects.lua
```diff
@@ -148,6 +148,75 @@ function gameEffects.RadiusDamage(dmginfo, pos, radius, inflictor)
...
+            diff = origin - ply:GetPos()
```
TimGoll, line 192:
```suggestion
            diff = origin - (ply:GetPos() + ply:OBBCenter())
```

While this doesn't _really_ matter, this takes the center of the player. I used this for all distance calculations in the code base and it doesn't cost much.

lua/ttt2/libraries/game_effects.lua
```diff
@@ -148,6 +148,75 @@ function gameEffects.RadiusDamage(dmginfo, pos, radius, inflictor)
     end
 end
 
+-- Creates explosion damage in a sphere through walls. Very useful for making explosives that aren't line of sight based.
+-- @param Player dmgowner The player that causes this explosion.
```
TimGoll, line 152: this should be `attacker` according to [the wiki](https://wiki.facepunch.com/gmod/CTakeDamageInfo)

lua/ttt2/libraries/game_effects.lua
```diff
@@ -148,6 +148,75 @@ function gameEffects.RadiusDamage(dmginfo, pos, radius, inflictor)
...
+-- @param Entity source The entity that causes this explosion.
```
TimGoll, line 153: this should be `inflictor` according to [the wiki](https://wiki.facepunch.com/gmod/CTakeDamageInfo)

lua/ttt2/libraries/game_effects.lua
```diff
@@ -148,6 +148,75 @@ function gameEffects.RadiusDamage(dmginfo, pos, radius, inflictor)
...
+-- @param number damage The maximum damage done by this explosion.
+-- @param Vector origin The center of the explosion.
+-- @param number outerRadius The outer border for the explosion damage and its falloff.
+-- @param number innerRadius The inner border for the explosion damage where falloff starts.
+-- @param boolean exponentialFalloff If the damage falloff should be FALSE: Linear, TRUE: exponential.
```
TimGoll, line 158: personally, I'd settle for linear or exponential and then only support one of these two

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
```diff
@@ -244,11 +195,15 @@ function ENT:Explode(tr)
             r_outer = r_outer / 2.5
         end
 
-        -- damage through walls
-        self:SphereDamage(dmgowner, pos, r_inner)
-
-        -- explosion damage
-        util.BlastDamage(self, dmgowner, pos, r_outer, self:GetDmg())
```
TimGoll, line 251: the removal of this line causes these two questions for me:

- does the C4 now cause less damage when in the deadly zone? does this have to be compensated?
- does this remove the fire particles and explosion animation that was spawned around C4?

TimGoll (changes requested):

gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
MrXonte: the damage in LOS will be lower, but technically we did double damage in LOS if we are in the deadly zone which was weird to begin with I think. It should still absolutely be deadly to any normal player.

Those should come from the effect that I left in, not from the blast/spheredamage.

lua/ttt2/libraries/game_effects.lua
MrXonte: Should i also use this for SetDamagePosition then?

lua/ttt2/libraries/game_effects.lua
TimGoll: sounds good to me!

lua/ttt2/libraries/game_effects.lua
TimGoll: In theory, yes. However the C4 is so small I did not consider this at all. But yes, please do!

TimGoll:
Just tested it in-game and it looks good to me, I think. I noticed though that the fall-of after leaving the kill zone is pretty steep. I was probably 25 units away from the kill zone and received 5 damage. Is this as expected?

TimGoll:
image

Also, I'm not sure if this was different before, but C4 is not listed as inflictor. Do you know how it was before?

MrXonte:
> Just tested it in-game and it looks good to me, I think. I noticed though that the fall-of after leaving the kill zone is pretty steep. I was probably 25 units away from the kill zone and received 5 damage. Is this as expected?

that's waaaayyy too steep. Ill double check. Should be like this: 
image

TimGoll:
Maybe my judgement was off. I walked for about a second as soon as the icon turned orange

MrXonte:
> Maybe my judgement was off. I walked for about a second as soon as the icon turned orange

default walk speed is 220u/s, so if you walked for exactly 1s you should be at 750+220 = 970 Units, which fits 5 damage

TimGoll:
If you change something, I'm available for another 30 minutes or so

MrXonte:
> If you change something, I'm available for another 30 minutes or so

if its ok with you, I'd set the Outer range to 1500? Heres a comparison:
image
image

or we set it to linear

TimGoll:
yeah, that looks good and it compensates the old LOS kill I think

TimGoll:
> no idea how it was before, but im doing it the same way the old spheredamage did. It used self for the inflictor, and im parsing self as the inflictor into the new spheredamage, so it should be the same as before

Yes I looked at the code as well. I'm confused. I guess it did not work before because you pass `self`, which is the c4 ent, but not the c4 weapon.

could you try to pass `ents.Create("weapon_ttt_c4")` instead of `self`?

MrXonte:
> > no idea how it was before, but im doing it the same way the old spheredamage did. It used self for the inflictor, and im parsing self as the inflictor into the new spheredamage, so it should be the same as before
> 
> Yes I looked at the code as well. I'm confused. I guess it did not work before either because you pass `self`, which is the c4 ent, but not the c4 weapon.
> 
> could you try to pass `ents.Create("weapon_ttt_c4")` instead of `self`?

it should also be possible to parse the entity instead of the weapon, but I'm gonna check later

TimGoll:
> it should also be possible to parse the entity instead of the weapon

Actually I don't think so. Looking at the code here makes me believe this never worked:
image

TimGoll:
If my suggestion fixes it, could you please test it in an old TTT2 version (before your PR) to confirm it was broken there as well? If it was, could you then please add this to the changelog? Thank you!

TimGoll:
Also, I don't know if this should be part of this PR, but if you changed the kill/damage radius, you should probably also change the UI to this?

MrXonte:
> Also, I don't know if this should be part of this PR, but if you changed the kill/damage radius, you should probably also change the UI (marker vision) to this?

I've included it now since i do think it should be part of the PR. if we are gonna change how it works, we should also make sure the display is (finally) correct

MrXonte:
> > it should also be possible to parse the entity instead of the weapon
> 
> Actually I don't think so. Looking at the code here makes me believe this never worked: image

hm this complicates things, but i guess that just means we should finally expand this class to allow for entities to be displayed here? A well defined entity should include all the relevant data we want to display here since we just do a lot of back and forth to print out the entity printname/id anyways?

TimGoll:
> hm this complicates things, but i guess that just means we should finally expand this class to allow for entities to be displayed here? A well defined entity should include all the relevant data we want to display here since we just do a lot of back and forth to print out the entity printname/id anyways?

I have to look into that. In theory that could be an easy change, but I'm not sure if this has any unwanted side effects.

I know I spent some time getting the boomerang to work because it is also a generic entity, not a weapon. I solved this, but I forgot how

MrXonte:
> > hm this complicates things, but i guess that just means we should finally expand this class to allow for entities to be displayed here? A well defined entity should include all the relevant data we want to display here since we just do a lot of back and forth to print out the entity printname/id anyways?
> 
> I have to look into that. In theory that could be an easy change, but I'm not sure if this has any unwanted side effects.
> 
> I know I spent some time getting the boomerang to work because it is also a generic entity, not a weapon. I solved this, but I forgot how

if your entity has a .ScoreName then this is taken instead if its not a weapon, at least for the kill event.
image
In the corpse its worse, since it only allows for a weapon or nothing at all, but if you parsed an entity here it might also work (or break horribly)
image

TimGoll:
We have to make TTT3 that breaks compatibility with all that shit, that looks awful :D

I'm not sure what I changed for bodysearch. But I did something to support weapons such as the boomerang

MrXonte:
> If my suggestion fixes it, could you please test it in an old TTT2 version (before your PR) to confirm it was broken there as well? If it was, could you then please add this to the changelog? Thank you!

as the placer of the C4, I get the C4 note
image
but definitely no indication that C4 was the cause when it kills someone else
image

TimGoll:
In case you're waiting on me. I'm fine with this as soon as you test the `ents.Create("weapon_ttt_c4")` suggestion

TimGoll:
this change actually fixed it:

image

For me this PR is done. If you want to investigate registering entites as inflictors, feel free to go ahead in a further PR

MrXonte:
> In case you're waiting on me. I'm fine with this as soon as you test the `ents.Create("weapon_ttt_c4")` suggestion

i actually just didn't have time yet, but great that it seems to work now! :D (also Eieruhr C4 xD)

TimGoll (approved):

TimGoll:
> (also Eieruhr C4 xD)

:D That's for the PietSmiet easteregg pack

FYI, I've got some plans for the rework of the C4 UI I'm currently working on. This will trigger some reworks of the damage again. So don't be confused when I change the parameters again.

MrXonte:
> > (also Eieruhr C4 xD)
> 
> :D That's for the PietSmiet easteregg pack
> 
> FYI, I've got some plans for the rework of the C4 UI I'm currently working on. This will trigger some reworks of the damage again. So don't be confused when I change the parameters again.

aahhhh i knew i knew it from somewhere xD

the MV UI elements or the use UI? 🤔 Also I think scaling damage with the fuze time would be great. Its a bit weird to only make it harder to defuse since you can just go away from it much more easily if the fuze is longer.

TimGoll:
See #1734, the PR is about the UI and some light rebalancing

If you want to discuss my ideas, it is best to post them there

PR #1752, opened 2025-01-30
VGui: Rework

TimGoll:
This PR has _many_ things yet to do, but I wanted to open a draft so it is open to discussion. There are a few things I want to tackle in this PR. It will probably be rather large, but without any meaningful new features. It is mostly renamed, reordered and cleaned up.

### 1. Broken Inheritance

The first issue I want to address in this PR is the broken vgui inheritance in TTT2. There are quite a few places in the codebase with code duplications and also many unnecessary vgui elements.

After some investigation I learned that both `PANEL` and `LABEL` are defined in the GMod codebase, while every other element with a D prefix is defined in lua. We could use those elements already existing, but there are so many modifications by now, that is is useful to stick to our copied and modied files.
However, in our code base `TTT2:DLabel` and `TTT2:DPanel` have many duplications because they both inherit from their respective engine defined base. In the engine `LABEL` inherits from `PANEL`. We can make `TTT2:DPanel` inherit from TTT2:DLabel`. This drastically cleans up our code and has only a single side effect: the two elements are mostly similar (with some changed default values).

### 2. Repetitive Code

The issue that triggered this rework is the following scenario. This is how it looks if you want to create a button:

```lua
local buttonReport = vgui.Create("DButtonTTT2", buttonArea)
buttonReport:SetText("search_call")
buttonReport:SetSize(self.sizes.widthButton, self.sizes.heightButton)
buttonReport:SetPos(0, self.sizes.padding + 1)
buttonReport:SetIcon(roles.DETECTIVE.iconMaterial, true, 16)
buttonReport.DoClick = function(btn)
    bodysearch.ClientReportsCorpse(data.rag)
end
```

When using _method chaining_ this can be cleaned up a lot:

```lua
local buttonReport = vgui.Create("DButtonTTT2", buttonArea)
    :SetText("search_call")
    :SetSize(self.sizes.widthButton, self.sizes.heightButton)
    :SetPos(0, self.sizes.padding + 1)
    :SetIcon(roles.DETECTIVE.iconMaterial, true, 16)
    :On("Click", function(btn)
        bodysearch.ClientReportsCorpse(data.rag)
    end)
```

This makes it way more clean and less cluttered.

### 3. Events and Hook Overwrites

It is typical that our custom elements have something like this in their initialize function:

```lua
local oldClick = self.DoClick

self.DoClick = function(slf)
    sound.ConditionalPlay(soundClick, SOUND_TYPE_BUTTONS)

    oldClick(slf)
end
```

While this does work, it is rather tedious and prone to errors. This has to be done since there is no real inheritance in vgui elements that could use `BaseClasses`, at least not as far as I could tell. Additionally one might want to attach a function to an event after a element was already created. The developer here has to make sure that they don't overwrite an already existing hook.
[The way the sound was added to buttons](https://github.com/TTT-2/TTT2/blob/master/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dbutton_ttt2.lua#L38-L58) is a good example of the issue at hand.

The solution is to introduce a simple system, that mimics hooks. Elements can be registered with `PANEL:On(name, function)`. That supports multiple events on a single event. So we can handle our button sounds internally in a clean way, while also exposing the event to the developer.

Similarly `PANEL:After(delay, function)` was introduced to provide a clean wrapper for timers.

Adding sound to a button would look like this:

```lua
local buttonReport = vgui.Create("DButtonTTT2", buttonArea)
    :On("Click", function(slf)
        sound.ConditionalPlay(soundClick, SOUND_TYPE_BUTTONS)
    end)
```

### 4. Other Ideas

While researching this, I also found this library which seems quite nice: https://github.com/Threebow/better-derma-grid

### DIFF
--- a/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dlabel_ttt2.lua
+++ b/gamemodes/terrortown/gamemode/client/cl_vskin/vgui/dlabel_ttt2.lua
@@ -1,48 +1,38 @@
 ---
 -- @class PANEL
--- @section DLabelTTT2
+-- @section TTT2:DLabel
 
 local PANEL = {}
 
 ---
--- @accessor Color
--- @realm client
-AccessorFunc(PANEL, "m_colText", "TextColor")
-
----
--- @accessor Color
--- @realm client
-AccessorFunc(PANEL, "m_colTextStyle", "TextStyleColor")
-
----
 -- @accessor string
 -- @realm client
-AccessorFunc(PANEL, "m_FontName", "Font")
+AccessorFunc(PANEL, "m_FontName", "Font", true)
 
 ---
 -- @accessor boolean
 -- @realm client
-AccessorFunc(PANEL, "m_bDoubleClicking", "DoubleClickingEnabled", FORCE_BOOL)
+AccessorFunc(PANEL, "m_bDoubleClicking", "DoubleClickingEnabled", FORCE_BOOL, true)
 
 ---
 -- @accessor boolean
 -- @realm client
-AccessorFunc(PANEL, "m_bAutoStretchVertical", "AutoStretchVertical", FORCE_BOOL)
+AccessorFunc(PANEL, "m_bAutoStretchVertical", "AutoStretchVertical", FORCE_BOOL, true)
 
 ---
 -- @accessor boolean
 -- @realm client
-AccessorFunc(PANEL, "m_bIsMenuComponent", "IsMenu", FORCE_BOOL)
+AccessorFunc(PANEL, "m_bIsMenuComponent", "IsMenu", FORCE_BOOL) -- ??? needed in TTT2:DComboBox?
 
 ---
 -- @accessor boolean
 -- @realm client
-AccessorFunc(PANEL, "m_bBackground", "PaintBackground", FORCE_BOOL)
+AccessorFunc(PANEL, "m_bBackground", "PaintBackground", FORCE_BOOL) -- ??? needed in TTT2:DButton?
 
 ---
 -- @accessor boolean
 -- @realm client
-AccessorFunc(PANEL, "m_bIsToggle", "IsToggle", FORCE_BOOL)
+AccessorFunc(PANEL, "m_bIsToggle", "IsToggle", FORCE_BOOL, true)
 
 ---
 -- @accessor boolean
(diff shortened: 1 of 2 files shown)


nike4613:
I've been on-and-off working on a framework to replace VGUI as much as possible myself. It has a **lot** of work to go (it can't even show anything yet!), but I have a vision for how I want it to be used:

```lua
{
    ty = "box",
    margin = 5,
    fill = true,
    {
        ty = "box",
        align = TOP,
        height = 20,
        fill = true,
        {
            ty = "text",
            translate = true,
            text = "some_text_key",
            align = { CENTER, CENTER },
        },
    },
    {
        vgui = "DPiPPanelTTT2",
        set = { Player = LocalPlayer() },
        init = function(pnl) end,
    },
}
```

Part of the goal is also to remove the pixel-sizing that is currently pervasive, provide automatic UI scaling, more aggressively cache data, etc.

TimGoll:
Uh, this is also really neat! However, this won't stop my project. In fact, they don't interfere with each other. If you ever come around doing your idea, this rework will probably make it simpler!

nike4613:
Something in the PR description read as asking for ideas or some such, though I can't see it now. I figured it'd be worth mentioning here that I've been working on that as a result.

TimGoll:
> Something in the PR description read as asking for ideas or some such, though I can't see it now. I figured it'd be worth mentioning here that I've been working on that as a result.

Yes, I asked for feedback. However, this PR is ambitious as it is - I won't start implementing a new UI framework for now :D

TimGoll:
Also, I want to mention something. The changes in this PR bring minimal issues with compatibility. Hooks, functions etc mostly stayed the same, I only introduced a new layer of interacting with them on top. While cleaning up the backend. _If_ any addon actually uses our vgui, it should work without changes, maybe minimal changes

TimGoll:
stylua sometimes handles these snippets in a bad way, not sure if there is anything that could be done about this. Most of the time it is as it is in the second example, but sometimes it is broken.

image

For vgui elements created with `:Add` this issue also exists. I worked around this by doing this:

image

Histalek:
> stylua sometimes handles these snippets in a bad way, not sure if there is anything that could be done about this. Most of the time it is as it is in the second example, but sometimes it is broken.

you mean the `.Create(` being pushed to the next line?

I'm not sure why this happens, but method chaining isn't really common in lua as far as i'm aware. So stylua is probably just not tested for this.

TimGoll:
> you mean the `.Create(` being pushed to the next line?

Yes. Especially for the spacer example. Without my doubled variable, it would the `:Add` to the second line as well which makes it hard to read imho.

Yeah, method chaining is mainly used in UI applications afaik. I've seen it being heavily used in the KiCad UI source code.

TimGoll:
I'm now at a point where i probably rebuilt 80% of the UI code. With some further refactoring I can probably get rid of many code duplications in the rendering. Right now panels, labels and buttons are working in their basic form. They all support translations, colors, outlines, content alignment and more. All of this while caching language, colors and sizes.

If one ignores the special UI elements (probably around 15 or so) all the others can be moved to this. E.g a menu button, a submenu button, ... are all basically the same. My new structure now supports all of the very specific features directly at the core, so adding them is easy now.

The only feature missing now is a multiline description in the same panel. While this sounds easy enough, the automatic perform layout will probably be tricky.

TimGoll:
My goal for this PR is to get it working for all normal panels. There is probably lots of stuff that can be improved in the special panels we created, but this can be done at a later point.

TimGoll:
I'm amazed that I only needed 3h of debbuging to get it working! Here are some updates:

image
**Buttons:** Work with icons, and without. Always center text alignment was used. Border radius and outline were changed. Colors were set and button number 7 is disabled. Text length limiting works.

image
**Panels:** (Basically buttons without hover and click.) Different text alignments were used, icons and text respect that alignment. Colors, border etc are supported here as well.

image
**Stretchable Panels:** This was the most complex, but it works. It automatically sizes itself to the content.

1. horizontal stretch enabled, text align top left
2. horizontal stretch enabled, text align center left
3-5. horizontal stretch disabled, vertical stretch enabled, with description, different text alignments
6. horizontal stretch enabled, vertical stretch enabled, with description, bottom right text alignment (do note, that only the title text influences the hor stretch, the description width is ignored)
7-9. horizontal stretch enabled, fixed height, different text alignment and icon size change

Do also note that everything here is cached and way more performant than before. It still supports vskin changes and language changes.

The next step is to make sure all basic TTT2 elements actually use the new system.

TimGoll:
image

The F1 menu is currently broken due to all the changes, but this will be fixed.

TimGoll:
image
**[Stretch with vertical alignment](https://github.com/TTT-2/ttt2-testing/blob/main/lua/autorun/client/cl_vgui_panels_stretch_v.lua):** Panels now also support vertical alignment by setting `PANEL:SetVerticalAlignment(true)`. All the other parameters are the same. And of course this works the same for buttons ans stretch panels.

Note: That the first three elements look broken is intended. The vertical height is too narrow to fit the content, auto stretch is disabled.

TimGoll:
I also removed the background bleeding for outlines with rounded corners:

old:
image

new:
image

zoomed in (old vs new):
image

TimGoll:
I have to restructure the core again. I thought this PR was on a good path, but when I got to elements such a DScrollBar and DTextEntry (which are basically the same element twice, one can be removed), I noticed that they inherit from source defined elements.

This has one major issue: Resizing, coloring etc has to be redefined again for those elements. And there are a few more of these.

So the third rewrite looks as follows: There are now core-modules. These are basically just files with code that implement a special feature. Like Recoloring for example. A panel can then define a list of stuff it implements, like "backgroundColor" or "outline".

This way the same code can be shared over multiple elements. Moreover, this prevents the base elements from having 2k lines of code.

Overall I like this change. But it will take a while to break stuff up so it is modular.

TimGoll:
image

I'm almost there with custom text boxes, that support all TTT2 textbox features. However, I'm currently unable to type anything. I can select the text, but not type it. If someone has an idea, I'm all for it

TimGoll:
https://github.com/user-attachments/assets/94e870d5-1da4-466f-8ff3-7660d2a863ad

Ha, I got it working now! The textbox works and has now console-like autocomplete. It always shows the common base of all possible entries. Hitting tab adds this common base to the string. Doesn't show anything when there is no common base.

If there are multiple options, nothing will be shown. However, this data is accessible and could be printed in a next panel if one wants to implement some kind of rcon console or so.

See this example code: https://github.com/TTT-2/ttt2-testing/blob/main/lua/autorun/client/cl_vgui_textentry.lua#L19-L28

TimGoll:
image

It is working now!

TimGoll:
image

ever so slowly getting there with the F1 menu

TimGoll:
Still a lot to do, but an end is in sight!

TimGoll:
Note so self: This fix should probably be included in the search bar UI element: https://github.com/TTT-2/TTT2/pull/1801

hubertoschusch:
Could you maybe also fix the dbinder? I think a setconvars is missing on the PANEL:makebinder method, when trying to set a key the convars does not get updated

TimGoll:
> Could you maybe also fix the dbinder? I think a setconvars is missing on the PANEL:makebinder method. When trying to set a key the convars does not get updated automatically

Ah, that was never needed in core TTT2, hence I did not think of this. You want a bind stored in a convar? You are aware of TTT2's bind system, aren't you?

hubertoschusch:
Ah, I guess I wasn't aware, my bad

TimGoll:
@hubertoschusch There could still be a usecase for your request, feel free to explain what you want to do

TimGoll:
> I just wanted to add a keybind configuration option to an addon submenu in addons menu, and thought it would work like sliders, checkbox etc., but I guess I didn't see there is already a whole extra menu for that

if you register a key bind, it will be automatically to the binds menu. You should probably use that to not confuse the users.

Here's an example that also adds the keyinfo on the bottom right: https://github.com/TTT-2/ttt2-heroes/blob/0d31e137cac4302737fe65a7c4000a7fc7d209d0/lua/terrortown/autorun/client/cl_crystal.lua#L55-L69

PR #1753, opened 2025-01-30
M16 Ironsights Allignment and Zoom Tweak

SvveetMavis:
This aligns the iron sights position of the m16 to fit better and lessens the zoom on the iron sights to look at a lot better, the little default zoom I feel is a little much.

Before 
Desktop Screenshot 2025 01 30 - 05 30 32 86

After 
Desktop Screenshot 2025 01 30 - 05 31 08 57


SvveetMavis:
Tweaked a bit further, looks better I think

Desktop Screenshot 2025 01 30 - 05 49 23 57

SvveetMavis:
you can see a bit more of the side where the m16 isnt modeled, but you could already see it when moving around with the default iron sights so I think its okay, idk how easy of a fix just like applying the texture of the other side of the viewmodel there would be

TimGoll (approved):
LGTM, thank you.

Regarding the model error: I'm aware and I noticed this previously as well. Would love to fix this, but I have no clue how.

PR #1754, opened 2025-01-31
feat(radio): Update radio description to reflect current interaction

### LINKED ISSUE #1708: Adjust the text of the radio in the shop

The shop description of the radio wasn't updated when the radio got its big update.
So the shop description can be misleading.

## To reproduce

1. Go into a round
2. Open the shop
3. Read the description of the radio
4. See the described issue

## Expected behaviour

The description of the radio in the shop should be updated so that the player can't misunderstand it.

## Additional context

Someone reported this on Discord in the help1 channel.
He wrote: "the description of the radio in the shop should probably be changed to reflect that I think then, because it still says you use it in the menu"

Histalek:
This is a followup to #1437 where this was initially missed.

I've also removed the duplicated `SWEP.EquipMenuData`.

Fixes #1708


TimGoll (approved):

TimGoll:
However, you should remove it in all languages. Otherwise the translators won't know that the text changed

Histalek:
> However, you should remove it in all languages. Otherwise the translators won't know that the text changed

right, i was wondering how the language tool would handle this but removing these makes a lot more sense

PR #1761, opened 2025-02-04
Localization: Russian and English

Satton2:
- Made the way "base roles" and "subroles" spelled consistent.
- Added some minor changes to make text a bit more clearly understandable (including more consistent usage of "submenu" and "distribution" terms).
- Other minor changes.
- Localized new/updated text into Russian (for some reason it says in-game that it's 99.9% while I don't see anything left: bug/actually don't see a thing?)


saibotk (approved):
Looks good thank you very much, as always 🫶

This might be a bug? Idk if so we can track that in a new issue.

TimGoll:
> Localized new/updated text into Russian (for some reason it says in-game that it's 99.9% while I don't see anything left: bug/actually don't see a thing?)

this counter also includes installed addons. Might be that you have one addon that adds one localized line that is not translated into russian

Satton2:
> > Localized new/updated text into Russian (for some reason it says in-game that it's 99.9% while I don't see anything left: bug/actually don't see a thing?)
> 
> this counter also includes installed addons. Might be that you have one addon that adds one localized line that is not translated into russian

No, sadly. There was only one subrole (to test that inspection stuff) and admin mod. They both were deleted/disabled when I found that percentage issue. I will test a bit more latter and maybe will create an Issue.

PR #1766, opened 2025-02-12
Incendiary grenade lights bodies on fire

milkwxter:
I thought it would be neat if incendiary grenades lit bodies on fire.

Here are some notes:
- It does NOT destroy the bodies
- It does light them on fire for a few seconds so you can't search them

What do you guys think?

I mostly made this because I have a role in development where if his body is lit on fire, he revives. But I thought it would be worth adding to the main mod as it gives the incendiary grenade some extra power since the damage isn't too great.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 All notable changes to TTT2 will be documented here. Inspired by [keep a changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## Unreleased
+- Added convar option to make incendiary grenades light bodies on fire (by @milkwxter)
 
 ### Changed
 - Updated Russian and English localization files (by @Satton2)
--- a/gamemodes/terrortown/entities/entities/ttt_flame.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_flame.lua
@@ -152,6 +152,13 @@ if SERVER then
 
                 gameEffects.RadiusDamage(dmg, self:GetPos(), self.hurt_radius, self)
 
+                -- set nearby bodies on fire so people cant search them
+                for _, nearbyEnt in pairs(ents.FindInSphere(self:GetPos(), self.hurt_radius)) do
+                    if nearbyEnt:GetClass() == "prop_ragdoll" then
+                        nearbyEnt:Ignite(5)
+                    end
+                end
+
                 self.next_hurt = CurTime() + self.hurt_interval
             end
 
--- a/gamemodes/terrortown/entities/entities/ttt_flame.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_flame.lua
@@ -153,9 +153,11 @@ if SERVER then
                 gameEffects.RadiusDamage(dmg, self:GetPos(), self.hurt_radius, self)
 
                 -- set nearby bodies on fire so people cant search them
-                for _, nearbyEnt in pairs(ents.FindInSphere(self:GetPos(), self.hurt_radius)) do
-                    if nearbyEnt:GetClass() == "prop_ragdoll" then
-                        nearbyEnt:Ignite(5)
+                if GetConVar("ttt2_incendiary_grenade_ignite_bodies"):GetBool() then
+                    for _, nearbyEnt in pairs(ents.FindInSphere(self:GetPos(), self.hurt_radius)) do
+                        if nearbyEnt:GetClass() == "prop_ragdoll" then
+                            nearbyEnt:Ignite(5)
+                        end
                     end
                 end
 
--- a/gamemodes/terrortown/entities/entities/ttt_flame.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_flame.lua
@@ -155,9 +155,8 @@ if SERVER then
                 -- set nearby bodies on fire so people cant search them
                 if GetConVar("ttt2_incendiary_grenade_ignite_bodies"):GetBool() then
                     for _, nearbyEnt in pairs(ents.FindInSphere(self:GetPos(), self.hurt_radius)) do
-                        if nearbyEnt:GetClass() == "prop_ragdoll" then
-                            nearbyEnt:Ignite(5)
-                        end
+                        if nearbyEnt:GetClass() ~= "prop_ragdoll" then continue end
+                        nearbyEnt:Ignite(5)
                     end
                 end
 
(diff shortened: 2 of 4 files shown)


nike4613:
This should be controlled by an off-by-default convar, with associated admin UI. It's a significant enough change (a *very* low bar) to meaningfully affect strategy, and so should be opt-in. If we had a way to detect that an installation was customized, I'd say it should be on for fresh installs, but we don't have that.

milkwxter:
Sounds good I will add your ideas!

TimGoll:
> This should be controlled by an off-by-default convar, with associated admin UI. It's a significant enough change (a _very_ low bar) to meaningfully affect strategy, and so should be opt-in. If we had a way to detect that an installation was customized, I'd say it should be on for fresh installs, but we don't have that.

We had many gameplay changes in the past that we enabled by default, with a setting to disable. Personally I have nothing against this being enabled by default. But I'm not alone on deciding that.

TimGoll:
Just a note: We added vFire support quite a while ago. The way you implemented it, it won't work when vFire is installed. Although it could be that vFire alreay sets corpses on fire. I'm not sure

milkwxter:
> Just a note: We added vFire support quite a while ago. The way you implemented it, it won't work when vFire is installed. Although it could be that vFire alreay sets corpses on fire. I'm not sure

Yes vFire does light bodies on fire! I was testing my role with that installed and it worked great.

nike4613:
> > This should be controlled by an off-by-default convar, with associated admin UI. It's a significant enough change (a _very_ low bar) to meaningfully affect strategy, and so should be opt-in. If we had a way to detect that an installation was customized, I'd say it should be on for fresh installs, but we don't have that.
> 
> We had many gameplay changes in the past that we enabled by default, with a setting to disable. Personally I have nothing against this being enabled by default. But I'm not alone on deciding that.

And it is precisely those changes that make people ask "why is this different now?"

Especially because the primary distribution channel is the Steam Workshop, which involves **automatic, involuntary** updates, behavioral compatibility should be our number one priority. Your server should only work differently when you tell it to.

You may note that back when I implemented role derandomization, despite how unintrusive it is, and how it leads to objectively better meta-gamefeel, **I was still hesitant to enable it by default**. *That* is the attitude we need to apply.

I don't think this change will make a significant difference in 90% of cases (most people wouldn't approach the body because of the fire anyway); that's why I'd advocate for it to be on in new installations. But it does make a difference, and will necessarily lead to strategy changes.

Histalek:
> > This should be controlled by an off-by-default convar, with associated admin UI. It's a significant enough change (a _very_ low bar) to meaningfully affect strategy, and so should be opt-in. If we had a way to detect that an installation was customized, I'd say it should be on for fresh installs, but we don't have that.
> 
> We had many gameplay changes in the past that we enabled by default, with a setting to disable. Personally I have nothing against this being enabled by default. But I'm not alone on deciding that.

We did that for additional features and not for altered features only if possible. Enabling altered features by default defies expectations, which is obviously bad

So in this case i agree this feature should default to `off`

milkwxter:
I agree with TimGoll that if implemented, it should be default behavior, I think since there is already vFire support and vFire incendiary grenades already do this, it isn't the biggest change to include in the base mod. (Since vFire doesn't tell you incendiary grenades light bodies on fire now)

I think it will feel very natural to include, as the body will be clearly on fire and already has target ID support, where it tells you something like "Ouch! This body is on fire, please wait a bit for it to cool down!"

But if it stays far away from admins view in the equipment settings menu then they probably won't know it exists. What do you guys think? (I will update the PR if you guys still think it's a good idea)

MrXonte:
I'd also argue that the case of someone checking a body inside a fire is currently almost nonexistent. Like, who doesn't just wait a few seconds? The binoculars might be an edge case, but I have no input here since I have not seen anyone use it ever.
But to avoid something like the C4 buff issue, we could do a poll on Discord to check if our bubble and the general public are aligned here?

milkwxter:
> You bring up a good point with the binoculars, I need to test this

Yeah if a body is on fire, you can't scan it with binoculars, it says "Ouch! This body is on fire!" still.

nike4613:
> The reason I favor it being enabled is the following: Nobody will enable it, which defeats the purpose of adding it

This is an argument which is too broad to be useful; it can apply to any new feature, *even ones which are obviously bad*.

The general form of the argument goes like this:
1. Some new feature X is being proposed. Without doing lots of playtesting, it is neither obviously good nor obviously bad. It may enable some functionality which previously didn't exist, but that functionality is not a clear improvement in play experience over the status quo.
2. People are resistant to change. They will not change their setup unless they see compelling reason to do so. 
3. Once feature X is rolled out, if feature X is disabled-by-default, people will never turn it on and try it.
4. Therefore, X should be enabled-by-default.

The problem with this argument is step 3 -> 4. Fundamentally, our goal is not *get people to try feature X*, but rather *provide people with a balanced, engaging, and above all **fun** game experience*. If the status quo provides that, and we have no concrete evidence that feature X provides a *more* balanced, engaging, and fun game experience, then enabling feature X by default runs the risk of actually making it worse. The conclusion from step 3 should be "Oh well, their loss," and we should encourage playtesting and feedback or whatever it takes to get evidence that it meaningfully improves the experience.

MrXonte:
>  Fundamentally, our goal is not _get people to try feature X_, but rather _provide people with a balanced, engaging, and above all **fun** game experience_. If the status quo provides that, and we have no concrete evidence that feature X provides a _more_ balanced, engaging, and fun game experience, then enabling feature X by default runs the risk of actually making it worse. The conclusion from step 3 should be "Oh well, their loss," and we should encourage playtesting and feedback or whatever it takes to get evidence that it meaningfully improves the experience.

I fully agree, but which means are there to encourage people to playtest or adopt a feature? The vast majority of people don't read the patch notes in detail (if at all) or search for new options. We would need a way to make people aware of the changes and then force them to decide if they want to try it out or not. That would create the necessary awareness, but I have no idea if that's feasible to implement in a way that's not obnoxious.

By forcing a new default (with a fallback option to the old system or similar enough to old) people will be able/forced to provide feedback and return to default if it bothers them, or leave it as is if it doesn't. But I think this needs to be a discussion of the core team internally with a fixed final decision, otherwise this discussion will come up for every change? Regardless, fallback to the old or old equivalent should be a requirement.

TimGoll:
It seems like most people here on GitHub don't share my opinion. So I guess the decision is made then

milkwxter:
I will close this for now, thank you guys for the input. I don't think this idea is ready yet!

nike4613:
> I will close this for now, thank you guys for the input. I don't think this idea is ready yet!

I think the feature and implementation are fine, as long as its off-by-default.

Issue #1751, opened 2025-01-29
Ammo Mangement F1 Menu & Better Ammo Management


MrXonte:
## Is your feature request related to a problem? Please describe.
The original setup of TTT ammo entities quickly falls apart when using addons. Some of this was addressed (like dropping wep.clipsize ammo, which caused huge issues with low clipsize guns), but the underlying problem is that the ammo is matched to the original guns where one ammo entity is exactly one clip, and the max reserve is exactly two clips, and the guns ClipMax is two ammo entities. This can cause a lot of weird behavior with guns that don't stick to the original values or larger modifications that add/remove ammo types.
For addon creators, this is also an issue since you need to directly modify core TTT files if you want to modify ammo, which can cause conflicts if you want to make "drop-in" solutions that just downloads from the workshop.

## Describe the solution you'd like
Dedicated F1 submenu for managing ammo entities (aka anything that inherits from the base ammo entity) and some reworks/additional options for ammo.
Some settings for each ammo type I would like are:
- [Bool] Enable/Disable Ammo Spawns
- [Bool] Replace with Random Spawn (requires disabled ammo spawn)
- [Integer] Box Capacity: Default amount of ammo in an ammo entity and the maximum dropable at once.
- [Integer] Static Reserve Amount: Maximum size of reserve ammo.
- [Bool] Dynamic Reserve: An option to enable dynamic reserve. This system makes your total reserve equal to the ClipMax (which is unused in TTT/2 except for some bad interactions during Equip that can delete ammo from the round and which is equal to the static ammo reserve anyway) of all your weapons, summed. That means your reserve increases accordingly if you have multiple weapons of the same ammo type. Disables Static Reserve Amount.

This should also not break any default behaviour or existing addons (that don't directly modify these files) while giving great customization options.

## Describe alternatives you've considered
If you want to be able to modify ammo, there really isn't a good alternative that comes to mind since there is no system in place.

TimGoll:
will consider this and come back to the ideas once we plan the weapon/ammo rework. Maybe I should start planning things so other people (like you) can implement subtasks

MrXonte:
That would be great! And there's a non-zero chance I might have some things ready anyway when it comes to weapons since I'm using a hugely modified weapon base. Ill try to make a PR for any non-breaking improvements to the weapon base moving forward, and for any base gamemode stuff I do PRs now anyway 👍 

[...]

PR #1773, opened 2025-02-20
Better Ammo Boxes

MrXonte:
Ammo boxes no longer despawn after the ammo has been picked from them, and they have less than 25% capacity left. It was always possible to create boxes with less than 25% capacity since boxes check your clip size when dropping ammo but check their max size for despawning. Aka, if an SMG with a 10-round clip drops 3 ammo, that's valid, even though that's less than 25% of the 30-round SMG ammo box capacity. This should be more consistent now as there is no lower limit.

Ammo boxes are now visually scaled with ammo stored from 50-100% model scale.


wgetJane:
i assume this just removes the weird behaviour ttt does where it "rounds down" ammo?

> Removing Weapon Ammo Storage

while i think the way ammo gets stored in dropped guns is silly as well, i think it's going to be very difficult to change it in a way that keeps ttt2 compatible with most of existing weapon addons: everything from custom t weps, custom floor weps, to completely different weapon bases, idk how you're supposed to really handle that

at most i think it would really only make sense for the ttt weapon base and only ones that use smg, shotgun, rifle, deagle, and pistol ammo

MrXonte:
> i assume this just removes the weird behaviour ttt does where it "rounds down" ammo?

yep exactly. This weird <25% rounding that has been an inconsistent mess ever since you can drop from reserve instead of only from clip.

> > Removing Weapon Ammo Storage
> 
> while i think the way ammo gets stored in dropped guns is silly as well, i think it's going to be very difficult to change it in a way that keeps ttt2 compatible with most of existing weapon addons: everything from custom t weps, custom floor weps, to completely different weapon bases, idk how you're supposed to really handle that
> 
> at most i think it would really only make sense for the ttt weapon base and only ones that use smg, shotgun, rifle, deagle, and pistol ammo

My plan was to make verify that the ammo type exists and then drop them as boxes, that would also allow for this to work for third party ammo types.
To not break any addons, the original code would need to remain for unknown ammo types, still storing the ammo in the gun. I would change the weird behaviour there though, since currently you store all your reserve ammo in the gun, but when you pick it up again you delete any stored ammo above your clipMax. For default weapons clipmax is equal to the ammo box max so no issue, but if you have a gun with less or more than that you can give yourself ammo above max or delete stored ammo.

wgetJane:
not sure how you'd make easy to use without adding a sort of ammo inventory ui for players

maybe i guess have a radial menu that pops up when you hold the current ammo drop bind?

it could easily display all your reserve ammo and tou mouseover and release the key to drop some amount from it (and of course an option at the top or bottom of the radial menu to drop your current gun's clip)

it could internally work as just calling a concommand with the ammo type as the arg so it'd work with potential custom ammo types

also how do you handle picking up ammo for guns you don't own? i assume you'd be able to manually pick up ammo?

ammo boxes also would need to drop on death as well

wgetJane:
oh that's much simpler than i imagined you were gonna implement it

imo it'll feel a little odd to drop a bunch of ammo boxes when trying to drop a weapon, maybe keep it to 1 ammo box at most even if it'll have a ton of ammo inside

you'd also need to delay their drop or offset their initial pos so they dont immediately collide with each other in the air

MrXonte:
> oh that's much simpler than i imagined you were gonna implement it

yea for now its just a small little change. Bigger changes to all ammo management are on my todo in the future.

> imo it'll feel a little odd to drop a bunch of ammo boxes when trying to drop a weapon, maybe keep it to 1 ammo box at most even if it'll have a ton of ammo inside

given the current rules you would at most drop 2 boxes since that's the max reserve which should be ok i think? Tons of boxes would be super weird, I agree. 

> you'd also need to delay their drop or offset their initial pos so they dont immediately collide with each other in the air

true! Probably calling the default ammo drop function with a small delay.

MrXonte:
So ive been busy and also tackled #1751 . Here just a short list of what ive done:
Added ammo target ID, inkluding showing how much is left in a box (toggleable per client)
Added ammo box size scaling with ammo left (toggleable per client)
image
image

Changed reserve ammo to be a global parameter that has nothing todo with an entities AmmoMax. Theoretically now allows for other ammo entities for the same ammo type without breaking. Supports old custom ammo that uses AmmoMax out of the box.
Added F1 menu ammo management tab. Allows settings reserve max, box ammo, spawnable (if not, replace with random ammo spawn)
image

MrXonte:
@TimGoll @saibotk can you review?

saibotk:
Sorry, we are quite busy with work! 
You can rest assured that im quite diligent in tracking my open Todos & mentions, i'll come around to this whenever my time allows for it. 

@mexikoedi would you mind checking this out and help us with testing this independently? Id love to have you try this out as one of the most active people :)

mexikoedi:
I tested it and it works without errors.
The spare ammo and the ammo in the boxes works as it's described here.
The F1 menu settings also do what they should (for example I limited the spare ammo and the ammo in the boxes for the smg type).

But I also found some stuff which could be adjusted.

1) I would remove the TargetID for the ammo boxes because it conflicts with the [[TTT2] Weapon Statistics [FEATURE]](https://github.com/TTT-2/ttt2-weaponstats) addon. This one shows the the names of the ammo boxes for spectators and also for the players. If I use the addon while testing the changes of this pull request the TargetID doesn't work for the spectators anymore. Maybe remove it or you could add the ammo count to the addon instead (if needed).

2) I would remove the model scaling of the ammo boxes. It looks weird and also causes issues like ammo boxes flying above the ground. See:
small ammo boxes fly above ground

3) I would adjust the text in the F1 menu. Is the type (for example `item_ammo_pistol_ttt`) really needed if the title says `9MM AMMO`? And the description of the `AMMO SETTINGS` section could be written a bit simpler. It's a bit difficult to read (the second sentence). See:
f1 menu text

MrXonte:
> 1. I would remove the TargetID for the ammo boxes because it conflicts with the [[TTT2] Weapon Statistics [FEATURE]](https://github.com/TTT-2/ttt2-weaponstats) addon. This one shows the the names of the ammo boxes for spectators and also for the players. If I use the addon while testing the changes of this pull request the TargetID doesn't work for the spectators anymore. Maybe remove it or you could add the ammo count to the addon instead (if needed).

Sure, no problem!

> 2. I would remove the model scaling of the ammo boxes. It looks weird and also causes issues like ammo boxes flying above the ground. See:

Yeah, that's because I'm only scaling the visuals on the client side, not touching the server-side physics. It could be fixed by turning it into a server-side convar instead, but that would lose client-side control over this feature. Which do you think is the best approach?

> 3. I would adjust the text in the F1 menu. Is the type (for example `item_ammo_pistol_ttt`) really needed if the title says `9MM AMMO`? And the description of the `AMMO SETTINGS` section could be written a bit simpler. It's a bit difficult to read (the second sentence). See:
>

It allows you to have multiple ammo entities for the same ammo type. So you could have a big and a small 9mm box if you wanted to. Very niche feature, but I thought it would be nice to have. Can be removed though.

mexikoedi:
> > 1. I would remove the TargetID for the ammo boxes because it conflicts with the [[TTT2] Weapon Statistics [FEATURE]](https://github.com/TTT-2/ttt2-weaponstats) addon. This one shows the the names of the ammo boxes for spectators and also for the players. If I use the addon while testing the changes of this pull request the TargetID doesn't work for the spectators anymore. Maybe remove it or you could add the ammo count to the addon instead (if needed).
> 
> Sure, no problem!
> 
> > 2. I would remove the model scaling of the ammo boxes. It looks weird and also causes issues like ammo boxes flying above the ground. See:
> 
> Yeah, that's because I'm only scaling the visuals on the client side, not touching the server-side physics. It could be fixed by turning it into a server-side convar instead, but that would lose client-side control over this feature. Which do you think is the best approach?
> 
> > 3. I would adjust the text in the F1 menu. Is the type (for example `item_ammo_pistol_ttt`) really needed if the title says `9MM AMMO`? And the description of the `AMMO SETTINGS` section could be written a bit simpler. It's a bit difficult to read (the second sentence). See:
> 
> It allows you to have multiple ammo entities for the same ammo type. So you could have a big and a small 9mm box if you wanted to. Very niche feature, but I thought it would be nice to have. Can be removed though.

2. I think it's ok to turn it server-side.
But note that I didn't test it with addons like that one: [JersonGaming's Model/Real RNG Ammo Boxes](https://steamcommunity.com/sharedfiles/filedetails/?id=1741790902) addon which is used in many collections (it's also in the [[OFFICIAL] TTT2 Basics - Graphics](https://steamcommunity.com/sharedfiles/filedetails/?id=3152311816) collection).
The addon will most likely still work but it would be great if you could test it thoroughly so it doesn't conflict with addons like that one.

3. Ah ok. That's up to you. You can remove that or adjust the text.

saibotk:
I think we should remove the model scaling part for now and can then iterate on that in a follow up PR instead. I like the other changes though. And thx @mexikoedi for helping us out here

PR #1778, opened 2025-02-24
Added incompatible addon to the addonchecker

NickCloudAT:
The addon (https://steamcommunity.com/sharedfiles/filedetails/?id=2795122660) is no longer supported too and seems to cause some issues with TTT2. For example: How corpses are created is overwritten and causes them to act like they are fake.


saibotk (approved):

saibotk:
Thx <3

Issue #1776, opened 2025-02-22
Last words: Does not respect convar + last words being client side


vertiKarl:
## Disclaimer

If you are missing the following information your issue will be closed!

- master branch with commit "7001db9a"
- "Last words are getting printed when dying to fall damage even when having the convar: `ttt_lastwords_chatprint` set to 0."

Also please make sure to read the [FAQ](https://docs.ttt2.neoxult.de/troubleshooting) and [existing issues](https://github.com/TTT-2/TTT2/issues?q=is%3Aissue) first!

## Your version of TTT2 (mandatory)

Leave one of the following:

- Latest GitHub version

## Describe the bug (mandatory)

Last words are getting printed when dying to fall damage even when having the convar: `ttt_lastwords_chatprint` set to 0.

This is due to this line:

https://github.com/TTT-2/TTT2/blob/7001db9a4c2c02312e87d0f6062b9ee0bca10c1a/gamemodes/terrortown/gamemode/server/sv_gamemsg.lua#L484

I do not think this is intended even though it seems like this has not changed since the very first commit in 2018:

https://github.com/TTT-2/TTT2/blame/908b0ff4febd926ee89cbb032868de0e706efb77/gamemodes/terrortown/gamemode/gamemsg.lua#L419C65-L419C65

## To reproduce

Steps to reproduce the behaviour:

1. Start the game
2. type something in chat (without submitting)
3. die from fall damage
4. last words in chat (even when disabled via convar)

## Expected behaviour

Do not send last words when the convar is disabled, no matter the reason for death.

## Considerations / Exploits
I would have just opened a PR for this, but I saw an open issue in the Shinigami repository which stated that the Shinigami does not stop sending last words. For which I considered writing a fix but then noticed that there are no hooks to overwrite the last words behavior, so I wanted to propose a rework for last words to be overwritable by other addons and also being handled on the server side.

Because as it seems, currently the clients broadcast the last words peer-to-peer? So it's also exploitable as clients are not forced to respect convars (I assume?) and clients **always** send the message but only display when the convar is set.

NickCloudAT:
Just to jump in; Yes, this is nothing TTT2 ever touched. This is still the same in default TTT.

TimGoll:
I wonder if this was intentional in original TTT? Because why else would there be the explicit fall damage check? That being said, it doesn't make much sense to me.

As a sidenote, @vertiKarl This is an awesome issue, thank you!

TimGoll:
[...]

PR #1780, opened 2025-03-01
LastWords: Fixed convar not being respected and added a new hook

vertiKarl:
- Added a new hook `TTT2AvoidLastWords` which is getting checked when the server tries to send out the last words message to other clients. The parameters are the player that is trying to send the message and the chat message they were trying to send. To block said message, just add this hook to your addon and return false based on conditions you define. 
- Fixed (#1776) not respecting ConVar if player dies by fall damage.

I have tested the following:
- Dying by fall damage when ConVar is disabled and having a chat message written but not sent. (works: is getting blocked)
- Adding the hook to the Shinigami and:
  - Dying as Shinigami when not reborn (works: sends message)
  - Dying as Shinigami when reborn (works: blocks message)

I will also make a PR for the Shinigami once this PR gets merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Added
 
 - Added option to select preferred unit of length for distance displays (by @wgetJane)
+- Added a new hook `TTT2AvoidLastWords` to allow blocking last word messages on a per-message basis (by @vertiKarl)
 
 ### Changed
+
 - Updated Russian and English localization files (by @Satton2)
 - Updated the list of troublesome addons used by the addonchecker
 
@@ -17,6 +19,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Fixed
 
 - Fixed new outlines' `OUTLINE_MODE_VISIBLE` and `OUTLINE_MODE_BOTH`
+- Fixed ConVar `ttt_lastwords_chatprint` not being respected if player dies by fall damage (by @vertiKarl)
 
 ## [v0.14.1b](https://github.com/TTT-2/TTT2/tree/v0.14.1b) (2025-02-01)
 
@@ -58,7 +61,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 - Fixed `markerVision`'s registry table being able to contain duplicate obsolete entries, thus fixing potential syncing issues with markers (by @TW1STaL1CKY)
 - Fixed issue in new Ammo dropping that could cause an error when dropping for modified weapon bases. (by @MrXonte)
 - Fixed C4 not showing the correct inflictor when the player is killed (by @TimGoll)
-- Fixed M16 Ironsight misalignment (by @SvveetMavis) 
+- Fixed M16 Ironsight misalignment (by @SvveetMavis)
 - Fixed outline interactions with certain weapon scopes (by @WardenPotato)
 - Fixed outlines not rendering uniformly (by @WardenPotato)
 
--- a/gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
+++ b/gamemodes/terrortown/gamemode/server/sv_gamemsg.lua
@@ -481,7 +481,7 @@ local function deathrec(ply, cmd, args)
         return
     end
 
-    if ttt_lastwords:GetBool() or ply.death_type == KILL_FALL then
+    if ttt_lastwords:GetBool() and hook.Run("TTT2AvoidLastWords", ply, words) ~= false then
         LastWordsMsg(ply, words)
     end
 
@@ -582,3 +582,12 @@ function GM:TTT2AvoidTeamChat(sender, team, msg) end
 -- @hook
 -- @realm server
 function GM:TTT2AvoidGeneralChat(sender, msg) end
+
+---
+-- Cancelable hook to block a the last words chat message.
+-- @param Player sender The player that sends the message.
+-- @param string words The message that is about to be sent
+-- @return nil|boolean Return false to block message
+-- @hook
+-- @realm server
+function GM:TTT2AvoidLastWords(sender, words) end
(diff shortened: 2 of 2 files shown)


saibotk (approved):
Thanks 🙏

saibotk:
@vertiKarl can you quickly resolve the merge issues? Then ill merge this in

wgetJane:
note: there's a hook in ttt called TTTLastWordMsg which ttt2 removes for some reason

https://github.com/Facepunch/garrysmod/blob/998cab5c145cf339bbb02d00b82997832c16dc07/garrysmod/gamemodes/terrortown/gamemode/gamemsg.lua#L284

also btw last words being sent on a fall death is a default ttt behaviour, it's not a mistake

wgetJane:
> Doesnt make sense that you can turn of last words but can still send a last word message when you die through fall damage. Its probably an oversight from base TTT which is now getting fixed in TTT2.

this is what the original condition is:

```lua
if ttt_lastwords:GetBool() or ply.death_type == KILL_FALL then
```

im not sure how this could be interpreted as an oversight, the intention here is clearly for fall deaths to trigger last words

ttt_lastwords is also an undocumented cvar, it does not appear on the ttt settings menu and it does not appear on the official ttt website

idc if ttt2 wants to change this little thing, but it's silly to keep acting like every obscure thing you didnt know about base ttt is a stupid mistake, you guys need to give bku far more credit than this

TimGoll:
Not sure what the state of this PR is right now, especially with the changes introduced in https://github.com/TTT-2/TTT2/pull/1784

G4PLS:
> Not sure what the state of this PR is right now, especially with the changes introduced in #1784

Do I read it correctly that the behaviour got changed so that you can now return nil or false in `TTTLastWordsMsg` and the message wont be sent?

So this PR would still be valid as the ConVar didnt get respected on falldamage. In my understanding setting ttt_lastwords to false should block the last words completely, otherwise people could abuse it with custom roles.

Basically you have a role that is not allowed to write could still say which people are Traitors by abusing the falldamage.
Ofc this should get handled by the role aswell so the change #1784 is good so that roles can block the last message but the ConVar should also just block lastwords completely.

TimGoll:
Honestly, and I don't mean that in a negative way, I don't really care. If you think this is a useful addition, then I'm open to merge this if no one objects from our team. I'm also open to @wgetJane's input here

TimGoll:
@G4PLS I talked to wget in the linked PR and the only thing to consider here is if the hook is redundant now. What's your opinion on this?

vertiKarl:
Well yeah the hook added in #1784 should be used instead of the newly added one in this PR.
The only relevant change is the removal of `ply.death_type == KILL_FALL` in sv_gamemsg.

as leaving this in allows muted roles (like the shinigami) to reveal who the traitors are by typing their name in the chat and then dying by fall damage.
In base TTT this was never an issue as you basically didn't have any "confirmed" roles or roles with extra information.
The shinigami (and all other addons working similar) could manually override this behavior with the newly added hook in #1784.

But from a server admin pov, ignoring the convar setting should never happen in the first place. If I set `ttt_lastwords_chatprint` to 0, I expect no last words to be printed to chat. ever. No matter if this was intended in TTT.

TimGoll:
You could add a fourth parameter to the hook, the kill-type. Then you can check it in the shinigame code. Would this work for you? That way you wouldn't break the default TTT behaviour and wget is happy as well

vertiKarl:
The kill type isn't relevant for the shinigami (and it also seems to be attached to the player table).
It's possible to overwrite this from a addon point of view with the newly added hook now anyway.

My goal is only to allow server admins to disable last words properly, without addons needing to be updated **when playing with last words disabled**.
So for example as long as the shinigami hasn't implemented this new hook, it is still possible to send messages when dying by fall damage even though the convar is disabled!
This goes for any addon that gives the player valuable information that they shouldn't be able to share.
I don't even know why it would ever be useful for it to ignore the convar at all. But if you consider this unimportant and don't want this behavior to change, feel free to close this and I'll just maintain it for my server myself.

TimGoll:
Hm. Maybe a dropdown menu then. "disable all", "classic" and "enable all"? Not sure. For me personally an enable/disable would suffice. But I don't want to keep discussing with people that appreciate these nuances.

G4PLS:
> @G4PLS I talked to wget in the linked PR and the only thing to consider here is if the hook is redundant now. What's your opinion on this?

I think the hook introduced with this PR is not needed anymore because of #1781, but as @vertiKarl said the main point wasnt the hook but to actually respect the convar.

I also think that a dropdown is overkill here. 
I see it like that: TTT2 is its own thing, I understand backwards compatibility is a thing but at some point you just have to do stuff.
So for me the fix introduced with this PR is absolutely fine and could imho just be merged but the hook added with this could be removed.


But at this point this is way to much discussion over a trivial improvement so I wont give more opinions on this as my point should stand.

saibotk:
Yeah lets just remove the hook and also remove this weird fall death case.

wgetJane:
(moved this comment from the other pr since it's more relevant here)

-----

last words on fall deaths being removed is a change i disagree with but tbh it's very minor and is pretty rare, i just didn't like the reasoning used for its removal and wanted to set the record straight that it's not a mistake/oversight with ttt:

image

the last words on fall deaths is pretty obviously supposed to be like someone screaming into their radio as they fall, since falling to your death isn't an instant death

though i would prefer that it looked like a normal message (https://github.com/TTT-2/TTT2/issues/1782) so that it doesnt confirm a death

there's archives of the ancient ttt forums where bku has patch notes, explains changes, and answers feedback, ttt was a very extensively playtested game and a lot of the little things added or cut from the game weren't just done completely arbitrarily

in fact, if you're not sure about a certain obscure mechanic in ttt, you can just ask bku himself (through his reddit account or whatever) and he answers pretty quickly and surprisingly remembers a lot

basically: i personally think this should be distinguished as a change/removal and not as a fix

saibotk:
Now with the new hook in, i think we should close this PR for now.
The shinigami can use the new method to prevent the last words and i think based on @wgetJane's comment, that we should just keep in the fall damage case.

I agree that we are a standalone thing and can do things differently, but in this case i think it is a valid game mechanic.

As this PR otherwise will start to rot, i'll go ahead and close it.
If any of you feel like we should discuss the default behavior more, please go ahead and open up and issue :)

I still want to thank everyone here for their participation and elaborate comments / thoughts! We really appreciate it!

Issue #1728, opened 2025-01-20
HL2 Small and Large Healthkits create errors


Crashington:
- Workshop version from Steam

## Description

Picking up small and large health kits (the hl2 item) causes errors. Depending on the installed assets it might break the UI entirely, preventing it from rendering. To be clear, the healing part still works.

## To reproduce

1. take damage
2. walk into a healthkit

## Expected behaviour

TTT2 shouldnt have an issue with such a simple, basegame interaction.

## Context (please provide as much as you can)

I tested this with only TTT2 and a single map enabled.

[TTT2 (Base) - v0.14.0b] lua/ttt2/extensions/draw.lua:259: bad argument #1 to 'SetMaterial' (IMaterial expected, got nil)

TimGoll:
Uhm, where do I find these health kits?

Crashington:
These are item_healthkit and item_healthvial and may be spawned by any arbitrary means. In my case, one of my maps spawns them from a vending machine.

TimGoll:
Hmmm. Any non-ttt entity can break the game. I will keep it in mind and will take a look at it

Crashington:
thank you!

vertiKarl:
This happens reliably on this [map](https://steamcommunity.com/sharedfiles/filedetails/?id=2245449608) when picking up these items. Picking them up breaks the whole HUD and spams the console with errors like the ones I attached in this file [console.txt](https://github.com/user-attachments/files/19042944/console.txt).

Also worth noting, you don't even have to take damage, picking them up suffices.

![Image](https://github.com/user-attachments/assets/ec0e1001-23cb-40d1-b825-e7760a50c563)
![Image](https://github.com/user-attachments/assets/ecb76027-a0ae-4d4a-834f-71f791ef33a4)

NickCloudAT:
Not sure if this is just me, but I'm unable to reproduce this because the items you show don't even work for me.

On the map that was mentioned; No items on this location, also not after several restarts:

![Image](https://github.com/user-attachments/assets/f829d40c-f3cf-40f6-bafe-a5b95942d05a)

Spawning the items myself with ent_create just creates them, but I cannot pick them up or anything else.

vertiKarl:
> Not sure if this is just me, but I'm unable to reproduce this because the items you show don't even work for me.

[...]

PR #1781, opened 2025-03-02
Fixed Hud error on pickup

NickCloudAT:
This fixes Hud errors when picking up an item or weapon with no viable icon by falling back to the item_special icon

Fixes https://github.com/TTT-2/TTT2/issues/1728

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Fixed
 
 - Fixed new outlines' `OUTLINE_MODE_VISIBLE` and `OUTLINE_MODE_BOTH`
+- Fixed Hud errors when picking up items or weapons with no viable icon (by @NickCloudAT)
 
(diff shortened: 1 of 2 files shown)


NickCloudAT:
Before I forget: I did not find a "fall back" icon that would be suited for this case and the item_special is already pre-defined at the top of the file.. So I decided to just use that as a fallback.

Crashington:
awesome, very happy this seemed to have been an easy fix! hope it hoes live soon

TimGoll (approved):
Seems like a valid solution to me. I wonder if the same thing has to be done for the old TTT hud?

NickCloudAT:
> Seems like a valid solution to me. I wonder if the same thing has to be done for the old TTT hud?

Ah, huh.. Yeah I completely forgot about the old hud (definitely not because it's ugly and I hate it and force disabled it on my end).
Will test later and get back, so maybe wait for my reply.

NickCloudAT:
Okay, no fix for old Hud needed.. It has no icons:
image

TimGoll:
Thanks for checking it!

Issue #1782, opened 2025-03-05
Add setting to make last words look like a normal message.


Dolphin2Point1:
## Is your feature request related to a problem? Please describe.
Some players find it frustrating when someone gives them a KOS as last words. A normal KOS is fine because it can easily be faked, but when players are given the context that the person saying the KOS died, it adds extra weight and all but validates the KOS.
I also just believe that telling you that someone died at that exact moment is just too much information for many lobbies. If you make it unclear whether they actually died or just sent an unfinished message, that would encourage use of gamesense and good communication for innos, and allow more tricks by traitors, thus making the game more fun.

## Describe the solution you'd like
A setting that makes last words look like an ordinary message, as if the player had just typed and sent it in chat.

## Describe alternatives you've considered
Some of this problem stems from quickchat triggering last words, which I think is very strange behavior. In my mind, quickchat sends messages instantly so it shouldn't trigger quickchat, however a toggle for that deserves it's own separate issue.
You could also just disable last words, but last words makes the game more fair for people who can't type a message out as quickly, and I also just think last words can be pretty funny, so I don't think they should be outright disabled. This may make them a bit less funny, because having the context that they're tied to a player's last moments changes the way players interpret it, but I believe it's still a setting worth adding.

## Additional context
No additional context is required to understand this feature request.

wgetJane:
made this pr for ttt specifically to make it possible to change lastwords to appear like normal messages: https://github.com/Facepunch/garrysmod/pull/2227
(which should be brought over to ttt2 along with a ton of other changes from upstream that ttt2 needs to catch up with)

would be neat if lastwords looking normal was a built-in ttt2 option as well

TimGoll:
Agreed on the part with the upstream changes that have to be ported over. I've seen that there are a few valid ones

PR #1784, opened 2025-03-07
Let TTTLastWordsMsg hook override default behaviour

wgetJane:
from https://github.com/Facepunch/garrysmod/pull/2227

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 ### Added
 
 - Added option to select preferred unit of length for distance displays (by @wgetJane)
-- Added `GM:TTTLastWordsMsg` hook from base TTT (Facepunch/garrysmod#2227, by @wgetJane)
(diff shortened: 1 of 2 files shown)


TimGoll (approved):

wgetJane:
@TimGoll btw i cant comment in #1780 for some reason but imo it doesn't conflict with this one, but i guess maybe the TTT2AvoidLastWords hook is a little redundant

wgetJane:
last words on fall deaths being removed is a change i disagree with but it's very minor tbh, i just didn't like the reasoning used for its removal and wanted to set the record straight that it's not a mistake/oversight with ttt:

image

the last words on fall deaths is pretty obviously supposed to be like someone screaming into their radio as they fall, since falling to your death isn't an instant death

though i would prefer that it looked like a normal message (https://github.com/TTT-2/TTT2/issues/1782) so that it doesnt confirm a death

there's archives of the ancient ttt forums where bku has patch notes, explains changes, and answers feedback, ttt was a very extensively playtested game and a lot of the little things added or cut from the game weren't just done completely arbitrarily

in fact, if you're not sure about a certain obscure mechanic in ttt, you can just ask bku himself (through his reddit account or whatever) and he answers pretty quickly and surprisingly remembers a lot

PR #1787, opened 2025-03-07
Fix C4/Radio sounds not playing outside of PAS

wgetJane:
from https://github.com/Facepunch/garrysmod/pull/2203


TimGoll (approved):
Thanks, this is an important fix that is brought over!

PR #1788, opened 2025-03-07
Allow workshop addons to pack rearm scripts

wgetJane:
from https://github.com/Facepunch/garrysmod/pull/2139


TimGoll (approved):
Nice, this improves support for classic TTT rearm script which is always very welcome

Issue #1786, opened 2025-03-07
Support for new TTT entity ttt_filter_role


wgetJane:
https://github.com/Facepunch/garrysmod/pull/2225

this needs to be using the same logic as ttt_logic_role in ttt2

PR #1789, opened 2025-03-07
Support for new TTT entity ttt_filter_role

wgetJane:
closes https://github.com/TTT-2/TTT2/issues/1786

new ttt_filter_role entity is from https://github.com/Facepunch/garrysmod/pull/2225


TimGoll (approved):
Thank you!

PR #1793, opened 2025-03-11
fix lua errors after autorefresh

wgetJane:
editing these files will cause errors from trying to use nil upvalues


TimGoll (approved):
LGTM. We should in general consider integrating crazyphysics fully into TTT2. Right now it works as an addon inside TTT2 and the convars are not even shown in the UI

PR #1794, opened 2025-03-11
Sprinting tweaks, 2 new sprint settings

wgetJane:
`ttt2_sprint_stamina_cooldown 0.8` : time in seconds before stamina begins regenerating after sprinting
`ttt2_sprint_forwards_only 1` : require forward key to be held to be able to sprint, prevents sprinting backwards or side to side, still allows sprinting diagonally with W+A or W+D

not sure if these should be enabled by default, but feels closer to how sprint/stamina in other games work


TimGoll:
I think sprinting was mostly done by @sai, so it is up to him to review. On first glance this looks solid though

TimGoll:
Speaking of sprint: I think you made a rework of the sprint animation years ago. This is something that I always wanted to bring into TTT2 as well, but never had the time to do

wgetJane:
> Speaking of sprint: I think you made a rework of the sprint animation years ago. This is something that I always wanted to bring into TTT2 as well, but never had the time to do

yeah i abandoned it after both ttt2 and crttt rewrote their sprint implementation to fix movement prediction

imo in retrospect those changes in the addon might've been too harsh (i still personally prefer sprint disabled)

saibotk:
Ill review this later or tomorrow!
Thanks for the PR though :)

TimGoll:
> > Speaking of sprint: I think you made a rework of the sprint animation years ago. This is something that I always wanted to bring into TTT2 as well, but never had the time to do
> 
> yeah i abandoned it after both ttt2 and crttt rewrote their sprint implementation to fix movement prediction
> 
> imo in retrospect those changes in the addon might've been too harsh (i still personally prefer sprint disabled)

were they? I liked the short video where I saw the animation. I think TTT2 (and most of GMod) lacks quite a lot of polish. This is why I started adding sounds to stuff for example. To make it feel less static. And these animations fit into this idea.

I personally hate playing without sprinting :D

PR #1792, opened 2025-03-10
Fix up issues related to +use, doors, buttons, and such

wgetJane:
fixes #1775
fixes #1691
fixes #1633
fixes #1626

PR #1799, opened 2025-03-20
Fix health station continuous use not working

wgetJane:
fixes a bug re-introduced by #1792

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


TimGoll:
Interesting that this already fixes it

TimGoll:
This change doesn't interfere with stuff such as C4 and marker vision?

wgetJane:
> This change doesn't interfere with stuff such as C4 and marker vision?

nope, checking for `self.ClientUse` was a mistake, `self.CanUseKey` is enough

TimGoll (approved):

wgetJane:
(mistake that i made, to be specific :trollface:)

Issue #1800, opened 2025-03-23
Buy menu breaks after using the search bar


OrdinaryOreo:
## Your version of TTT2 (mandatory)

TTT2 v0.14.3b

## Video

https://github.com/user-attachments/assets/384db630-7e6a-4e7e-9a59-77bcb630f06f

## Describe the bug (mandatory)

No buy menu items gets displayed and this stays after the round ends.

## To reproduce

By starting typing with certain special characters in the search bar. I barely did any testing and only know that `[` and `%` causes this behavior but probably many more do as well.

PR #1801, opened 2025-03-23
Fixed shop search

NickCloudAT:
This fixes the shop search breaking when using certain special characters because of luaPatterns

Fixes https://github.com/TTT-2/TTT2/issues/1800


saibotk (approved):
Thx !

PR #1819, opened 2025-06-19
Fixed guns not applying themselves as their damage inflictor for FireBullets

TW1STaL1CKY:
I've noticed that guns in TTT2 don't set their Inflictor as themselves for FireBullets ([which is a new-ish addition](https://wiki.facepunch.com/gmod/Structures/Bullet#Inflictor)), so I've fixed that.

I'd like to note it seems `util.WeaponFromDamage` was partially used as a workaround for FireBullets not setting an Inflictor, do people think it's still needed?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Fixed
+
+- Fixed guns not applying themselves as their damage inflictor (by @TW1STaL1CKY)
+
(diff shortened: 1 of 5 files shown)


wgetJane:
> I'd like to note it seems `util.WeaponFromDamage` was partially used as a workaround for FireBullets not setting an Inflictor, do people think it's still needed?

whatever you're planning for this function is definitely not a good idea

a lot of addons, like damagelogs, rely on this function

saibotk (approved):
Nice addition!

And yes we should keep the util function for now. Thx!

PR #1820, opened 2025-06-19
Slight code cleanup in weapon code

TW1STaL1CKY:
Noticed things could be tidied a bit in these two files

(I don't think this warrants an entry in the changelog)


saibotk (approved):
Thanks!

PR #1821, opened 2025-06-19
Input: Refactor +use handling

### LINKED ISSUE #1816: Health station continuous use doesn't work correctly

- Latest GitHub version

If you look at the health station and press the use key then you only get one health point and it stops working.
But if you press the use key near the health station (without first looking at it) then it works perfectly.

## To reproduce

1. Go into a round
2. Spawn a health station
3. Use the use key near it and then use the use key while looking at it
4. See the described issue

## Expected behaviour

If you press the use key while looking at the health station then it should work.
It should still work if you look at something else while holding the use key.
But it shouldn't work if you just press the use key near the health station without looking at it prior.

## Additional context

There was a fix recently which seems to have no effect because it's an issue in the TTT2 v0.14.3b Workshop version from Steam and also an issue in the latest GitHub version: https://github.com/TTT-2/TTT2/pull/1799

saibotk:
This will get a proper description & commits later.

Fixes #1816

### DIFF
--- a/gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_base_placeable.lua
@@ -14,8 +14,6 @@ ENT.isDestructible = true
 
 ENT.pickupWeaponClass = nil
 
-ENT.CanUseKey = true
-
 local soundDeny = Sound("HL2Player.UseDeny")
 
 ---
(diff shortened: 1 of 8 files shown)


wgetJane:
skimmed through the changes, are you sure these won't reintroduce any of the bugs fixed by #1792?

saibotk:
Pretty sure, as all the changes that i removed were mainly introduced to workaround the broken use system.
But will check doors etc. on the weekend

saibotk:
Okay @TimGoll , i finished this one up and fixed any remaining bugs i could find when testing this locally.

Also added a description and changelog!

The change is actually rather simple now, but understanding the code parts can be hard, if you have any questions ask me and i'll happily help out!

PR #1825, opened 2025-06-27
Markervision: Added UpdateTransmitState handling

NickCloudAT:
This will now handle UpdateTransmitState for markervision entities.

Please give me all feedback you have.. Right now, this is the easiest solution I was able to come up with.

(will add changelog later)


TimGoll (approved):
To me it looks like a solid implementation

TimGoll:
Good question. I'd say it is a fix

NickCloudAT:
That should be it then

PR #1826, opened 2025-07-05
Fixed not being able to use entities through water and clip brushes

TW1STaL1CKY:
Currently, trying to use entities and pick up weapons through water and clip brushes doesn't work, even though your TargetId HUD says you can.
This was happening because the clientside trace was using MASK_ALL. I've now changed to MASK_SOLID + CONTENTS_DEBRIS, which is roughly the same mask that the TargetId HUD uses.

I'm aware #1821 exists which I assume would fix this problem when it's eventually merged in, but I thought I would patch this issue specifially while that PR is still in the works.


saibotk (approved):
Nice 👍 yes thats still good to have fixed here until i can work on the use PR again

PR #1828, opened 2025-07-31
TargetId: Let GM:TTTModifyTargetedEntity modify the target distance

TW1STaL1CKY:
You can modify the target entity, but the distance value would remain the same since it couldn't be changed. This can make targetids to appear and disappear unexpectedly if the target entity was changed.

I've made `GM:TTTModifyTargetedEntity` read a second return value for distance so devs can correct the distance or do as they please with it. I've checked and it doesn't let you use entities from far away if you force target them with distance set to 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


TimGoll (approved):
Makes sense, thank you! :)

TW1STaL1CKY:
Fixed the merge conflict

PR #1831, opened 2025-08-12
Implement TT2 Guide -> Roles

ROODAY:
Implemented a very basic roles page that lists the available roles, and when a role is clicked it will display some basic info about the role

image


TimGoll:
First of: I really appreciate what you did here. However, I have to say that it does not fit our design language so far. I'm a bit confused why your scrollbar looks different. And you should put the content in forms instead of bare strings

ROODAY:
@TimGoll sure, I can use forms like how admin Roles page works, I guess these would be in form help text and just not have any of the config toggles/inputs?

As for the scrollbar looking different, I'm not sure why it's like that. I'm using a DScrollPanel instead of a DSubmenuList as the latter required a bunch of fields/data to be set that seems to only be on pages derived from `base_gamemodemenu.lua`, but this guide->roles page is derived from `base_gamemodesubmenu.lua`. I initially tried just replicating the admin roles page but ran into a ton of errors/stuff just not rendering and not knowing what was going on, so it felt easier to just rebuild from scratch with the components that DSubmenuList uses internally.

Sidenote, would you know how to remove the margin around the content area in this gamemode submenu? I don't think I explicitly set any margin here (and in other submenus I see there is a margin so it looks like something set at a default level)

PR #1835, opened 2025-08-21
Markervision: Fixed null entity error

### LINKED ISSUE #1834: Marker Visions cause error on removal

TTT2 (Base) - v0.14.5b (Steam)

Every time a marker vision object is removed (markerVision.Remove() is called), the following error occurs:

```
[TTT2 (Base) - v0.14.5b] lua/ttt2/libraries/marker_vision.lua:108: Tried to use a NULL entity!
  1. AddEFlags - [C]:-1
   2. Remove - lua/ttt2/libraries/marker_vision.lua:108
    3. func - lua/ttt2/libraries/marker_vision.lua:240
     4. unknown - lua/includes/extensions/net.lua:34
```

## To reproduce

1. Place a Traitor item that creates a marker vision in 3D space (e.g. Radio)
2. Destroy the item, pick it up or end/restart the round
3. Error occurs

## Context

My workshop collection: https://steamcommunity.com/sharedfiles/filedetails/?id=2788998348

## Additional context

Maybe only an additional NULL check is needed.

mexikoedi:
This PR fixes #1834 .
The null entity error was introduced in #1825 .

I tested it and I couldn't reproduce the error from the above issue again.

Error reason:
markerVision.Remove is called twice.
The first time ent is valid but the second time it's a null entity and that caused the error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to TTT2 will be documented here. Inspired by [keep a changel
 
 ## Unreleased
 
+### Fixed
+
+- Fixed markerVision null entity error (by @mexikoedi)
+
 ## [v0.14.5b](https://github.com/TTT-2/TTT2/tree/v0.14.5b) (2025-08-18)
 
 ### Changed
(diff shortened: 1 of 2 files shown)


mexikoedi:
> Shouldn't line 108 be RemoveEFlag? It doesn't make sense to AddEFlag in the .Add method and .Add it again in the .Remove

Makes sense, I adjusted/tested it and it still works fine without any errors.

TimGoll (approved):
Thank you!

PR #1840, opened 2025-11-09
Contributing: Adjusted outdated information

mexikoedi:
This PR removed outdated information from the CONTRIBUTING file and replaced it with newer information.

I also checked some other places (README, DESIGNGUIDLINES, RELEASE_PROCESS and the Steam Workshop page) for dead urls/outdated information/... and found only one issue:
- At the [Steam Workshop page of TTT2 Base ](https://steamcommunity.com/sharedfiles/filedetails/?id=1357204556) under "ADDONS" there is an url for an example for the addonchecker with the name "looks like this" which is not available anymore. This should be replaced with for example this url: https://docs.ttt2.neoxult.de/troubleshooting/#addon-checker


saibotk (approved):
Thank you! This is also a very important contribution <3

PR #1842, opened 2025-11-14
Fixed missing ttt_filter_role in fgd and added ttt_filter_subrole entity

MrXonte:
Fixed missing ttt_filter_role in ttt2.fgd

Added ttt_filter_subrole entity
Added ttt_filter_subrole  in ttt2.fgd
This, as the name implies, filters by subrole. Created it mainly for edgecases like the Defective role or map specific roles


TimGoll (approved):
I'm fine with this, however I do not know how these changes affect the game. Probably not at all? This is only relevant for new maps, isn't it?

PR #1854, opened 2026-01-17
Consolidate playermodel management logic

nike4613:
This PR consolidates the playermodel and player color management logic into one place, accessed via extensible hooks.

As far as I can tell, based on the testing I have done with bots, by default this has exactly the same behavior as without the PR. The PR just inserts more extensibility points. As I'm working on the addon I need this for, I might inject more hooks.

I am not a fan of the exact set of hook here either, but I'm reasonably certain they need to be structured this way to not break compatibility with other addons. In a perfect world, I'd be able to completely redo it to have fewer moving pieces, but alas.


nike4613:
> Just be aware that this still has to work with pointshop 2 and enhanced player model selection. This is a reason why the current approach is so convoluted.

I believe it does still; the actual flow should be unchanged from what it was, only with some organizational changes (i.e. adding hooks inbetween, and consolidating all logic into one place). I've only tested so far with EPMS, but as long as pointshop works in a similar way, there shouldn't be a problem I think.

PR #1858, opened 2026-02-23
Fixed thirdperson animations from third-party addons

### LINKED ISSUE #1690: Issue with thirdperson animation from third-party addon

- Workshop version from Steam

I noticed an issue with the following addon: [Custom Taunt](https://steamcommunity.com/sharedfiles/filedetails/?id=1310909460)
This addon allows you to use custom taunts (dances) which will be visible for everyone and yourself.
It worked prior to TTT2 v0.13.0 and stopped working in TTT2 v0.13.0. (Thirdperson doesn't work. Only firstperson is visible.)
Then it started to work as it should in TTT2 v0.13.1 and TTT2 v0.13.2 before being broken again in TTT2 v0.14.0 (Thirdperson works, but the animation is only visible for everyone else but not yourself.).

I think it has something to do with the following PR or something related to it:
- https://github.com/TTT-2/TTT2/pull/1439

Maybe this is a deeper issue which could also affect similar addons which also do animations in thirdperson.

## To reproduce

1. Join a round with the addon installed
2. Press F6 (or !dance in chat)
3. Select a taunt (dance)
4. See the described issue

## Expected behaviour

The animation shouldn't be just visible for everyone else but also for yourself too.

## Additional context

This issue was tested with only TTT2 and the mentioned addon installed.

mexikoedi:
This PR fixes #1690 .

Third-party addons that create thirdperson animations like [Custom Taunt](https://steamcommunity.com/sharedfiles/filedetails/?id=1310909460) can use gesture slots that conflict with TTT2 gesture handling (for example: GESTURE_SLOT_CUSTOM).
Because of this, the local player’s third-person animation could be visually overridden and not appear correctly.

There was also a second issue with the TTT2 voice chat gesture (ACT_GMOD_IN_CHAT) which interfered with active third-party thirdperson animations.

This fix adds clientside external gesture detection and suppresses the TTT2 ACT_GMOD_IN_CHAT gesture while an external gesture is active.
To make this more robust for all kinds of addons, both AddVCDSequenceToGestureSlot and AnimRestartGesture are tracked, while TTT2 internal gestures are excluded from that detection.

I tested this with the addon mentioned above and can confirm the issue is resolved (thirdperson animation is now visible for the local player).
The voice chat ACT_GMOD_IN_CHAT behavior also works as expected, and I did not encounter errors during testing.

Edit:
Description doesn't fit the code changes anymore, see comments below.


gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -7,6 +8,94 @@ if not plymeta then
...
+    -- @realm client
+    -- @see https://wiki.facepunch.com/gmod/Player:AnimRestartGesture
+    function plymeta:AnimRestartGesture(slot, activity, autokill)
+        -- some addons use AnimRestartGesture -> map activity to sequence to track duration
+        if
+            (self._ttt2InternalGestureWriteDepth or 0) <= 0
+            and isnumber(activity)
+            and activity >= 0
+        then
+            local sequence = self:SelectWeightedSequence(activity)
```
saibotk: I could not find how this is handled internally within GMOD, but this can be wrong, when the sequence here is different from the one that GMOD chooses when playing ACT gestures that hold multiple sequences. 

This is because you select one sequence at random here, you would need to pass this exact sequence to the `AnimRestartGesture` call, as far as i can see.

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -7,6 +8,94 @@ if not plymeta then
     return
 end
 
+-- use a dedicated slot if available to reduce collisions with third-party gesture addons
+local ttt2GestureSlot = GESTURE_SLOT_VCD or GESTURE_SLOT_CUSTOM
```
saibotk: I cannot really understand the reasoning here. What is the intention of using the VCD slot? 
GMOD and sandbox gamemode both default to the GESTUR_SLOT_CUSTOM.

Besides, those two slot variables are always defined, so i'd assume this or statement to not be effective at all.

saibotk (changes requested):
I looked over the changes and got some concerns and questions we need to discuss and i would prefer not to merge those in their current state.

- Not to shame you, but if this code was written with AI, I'd prefer if you'd disclose so (if not, I'm sorry its just hard to tell in todays world)
- Can you elaborate on the exact problem that was fixed here? The issue is rather old and lacks some information on how this behaves in the current version and what you expect it to do
- Can you elaborate how the Custom Taunts addon uses these functions or GMOD functions to handle animations, so we can understand why and what we need to adjust here?
- I'd prefer not to introduce additional public methods, especially since GMOD also has functions like `IsPlayingTaunt` already. This will be very confusing and we need to check if we handle all cases correctly

Let's first collect more information on this and then we can discuss a potential solution together. This way you don't waste time iterating too much on a potential solution.

Thanks for your understanding!

TimGoll:
> @TimGoll what do you think about just not playing the hand to head animation in general when another animation plays? I think this should be fine, but want a second opinion on this.

Good question, I have to think about this. Not sure honestly. While I think TTT2 would improve a lot by better/more animations and sounds (like the corpse search sound I added), I never actually looked at such details

mexikoedi:
I adjusted the code.
It's now much simpler/cleaner.
`IsPlayingTaunt` didn't work and `GetLayerCycle` is not needed.
Now there is only one new function which has a couple of checks to make sure that custom gestures come through if the voice chat gesture (the only problematic gesture) is active.
I tested it with a player and it worked without any errors (I and he could see the custom gestures.).

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
```diff
@@ -32,11 +32,41 @@ local function MakeSimpleRunner(act)
...
+    end
+
+    -- if currentSeq is invalid, the layer is not active, so no collision can occur
+    local currentSeq = ply:GetLayerSequence(slot)
+    if not currentSeq or currentSeq < 0 then
+        return false
+    end
+
+    -- if currentSeq is different from voiceSeq, then another gesture has taken over
+    return currentSeq ~= voiceSeq
```
saibotk: The `voiceSeq` depends on a random selection at least in theory, dunno if there are multiple of such animations. 

So I'd propose to use https://wiki.facepunch.com/gmod/Entity:GetSequenceActivity on the current sequence, to check if the activity id is equal to `ACT_GMOD_IN_CHAT`, i think that might be a little more safe to check if this is a sequence that is associated with the activity.

saibotk (changes requested):
Just a small suggestion.

Otherwise clean!

Thanks for your effort and the simplification!

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
mexikoedi: I tested it with `GetSequenceActivity` but it was unreliable.
It didn't work sometimes because `act` was invalid (`-1`).
That's why I left `voiceSeq` with the `SelectWeightedSequence` and the `currentSeq ~= voiceSeq` check as a fallback instead of removing it.
Then I tested it and it worked for all the custom taunts from the addon.

Is that alright like that?

saibotk: Ah yes -1 is fine, custom sequences often do not have an activity associated with them.

What i meant here was:
- Remove the fallback
- Remove the voiceSeq
- Just do `return act ~= ACT_GMOD_IN_CHAT`, because we only need to know if the current sequence is associated with the chat activity all other results indicate that another animation is currently playing.

saibotk (changes requested):

gamemodes/terrortown/gamemode/client/cl_player_ext.lua
mexikoedi: Ah, ok.
That was my mistake.
I tried this like that but I assumed that `act` should be valid (not `-1`) all the time so I added a check for that scenario.

Tested it again and it works great.

saibotk (approved):
Thanks 🙏

PR #1873, opened 2026-05-03
Fix DTooltipTTT2 assigning random globals

nike4613:
I can only assume this was just a silly mistake whenever this was last touched. I don't imagine it really affects anything though.


saibotk (approved):
Oh thanks! Good catch!

PR #1875, opened 2026-05-19
Add momentary_rot_button to TargetID

MrXonte:
Added momentary_rot_button to TargetID so that this third type of button also displays the usual interact text when hovering over.


saibotk (approved):
Thanks!

Issue #1841, opened 2025-11-13
TTT2 radio UI now obsolete due to Main TTT branch updates on November 13th


7script:
## Disclaimer

If you are missing the following information your issue will be closed!

- "Your version of TTT2"
- "Describe the bug"

Also please make sure to read the [FAQ](https://docs.ttt2.neoxult.de/troubleshooting) and [existing issues](https://github.com/TTT-2/TTT2/issues?q=is%3Aissue) first!

## Your version of TTT2 (mandatory)

Leave one of the following:

- Workshop version from Steam - **https://steamcommunity.com/sharedfiles/filedetails/?id=1357204556**
- Latest GitHub version
- Non-latest GitHub version (please provide what you can to specify)

## Video

If it's a very strange problem, it's better to add a video to see what happens.  
This step can replace the following steps: "Describe the bug", "To reproduce"

**Pictures:**

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/d64225c3-84a4-4c60-991d-3809019d74d9" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/81f5bcb3-7cba-4ad9-8f1f-40ccb48ca363" />

## Describe the bug (mandatory)

**In vanilla TTT, the menu has options to play the glock, silenced pistol, health station and teleport noises. The TTT2 version is missing these - as well as any custom radio sounds introduced via the framework created in the same update.**
An error is **not** a description.

## To reproduce

Steps to reproduce the behaviour:

**1. Disable all addons and open vanilla TTT
2. Become traitor (either naturally or through console commands
3. Purchase and deploy radio
4. Open radio menu
5. Observe screen
6. Close game and enable TTT2 and relaunch
7. Become traitor, purchase, and deploy radio
8. Open radio menu
9. Observe lacking buttons**

## Expected behaviour

**1.Enable TTT2 and launch game
2. Become traitor, purchase, and deploy radio
3. Open radio menu
4. Observe correct buttons**

## Context (please provide as much as you can)

- Collection link of your Server: [e.g. https://steamcommunity.com/sharedfiles/filedetails/?id=1888516630] **-None**
- **All** Server errors in the server console **- None**
- Client errors in the developer console **- None**
- Other addons installed in your servers `addons` folder **- None**
- Screenshots **- Done**

## Additional context

[...]

PR #1876, opened 2026-05-21
Allow insertion, deletion & query to Traitor Radio sound table

Guy-L:
Currently, there's no clean way for addons to add new sounds to the Radio weapon's sound list. This limits opportunities for addons that have unique sound cues that may be worth faking as Traitor. This PR adds a few API methods to interact with the radio sound list, which addons can use at initialization time.

### DIFF
--- a/gamemodes/terrortown/entities/entities/ttt_radio.lua
+++ b/gamemodes/terrortown/entities/entities/ttt_radio.lua
@@ -119,97 +119,170 @@ function ENT:AddSound(snd)
     end
 end
 
-local simplesounds = {
-    scream = {
-        Sound("vo/npc/male01/pain07.wav"),
-        Sound("vo/npc/male01/pain08.wav"),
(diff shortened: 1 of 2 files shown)


Guy-L:
Seems there was a base TTT update to address this specific issue that hasn't been added to TTT2 yet: https://github.com/TTT-2/TTT2/issues/1841

Leaving this up in case this is PR is an acceptable implementation

saibotk:
Hey, thanks for your work on this! 

Though, we would prefer an implementation with the same API as the original TTT code here, so we can stay compatible with TTT addons. It seems that in https://github.com/Facepunch/garrysmod/pull/2354, they added `TRADIO.AddNewSound()` for this.

So if you want to go forward with this and get this in, i'd recommend you implement the same changes from the base TTT PR in TTT2 :) 

Thanks!

PR #1878, opened 2026-05-29
Fix Markervision PVS Edgecase

MrXonte:
Fixes an edgecase of markervision with engine-native entities not getting relayed properly when out of PVS. I encountered specifically when using markervision on corpses while playin the vulture. Bodies often did not show up on MV when the corpse that should get markervisioned was too far away (out of PVS).


MrXonte:
@TimGoll @saibotk can you review?

saibotk (approved):
Nice! Thanks and thanks for the comment!

